PR #23, opened 2015-06-23
Atribute based XSS fixed on components and applications views

CalldiDoctor:
It was possible to perform an atribute based cross site scripting attack. 
The code was already protecting XSS attacks on HTML context, so it was not possible to exploit an XSS by setting the component or application name with the following payload:
```javascript
<script>alert(1)</script>
```
However, the code was also reflected on several attributes of the Edit button, and on some option tags. One of the possible payloads for exploiting this new XSS is:
```javascript
"onmouseover="alert(1)
```
Now, XSS cannot be performed on these fields:
* Application name
* Vendor
* Component name
* Component version
* License
* Language


stevespringett:
Thanks for catching this.

PR #26, opened 2015-11-19
Spring4 hibernate4 migration

wjase:
Bumped versions of spring and hibernate. 
Refactored transaction behaviour.
Started work on converting xml config to java config.
Added dao/config test

PR #27, opened 2015-11-28
Revert "Enable HEAD request and add Last-Modified header for nist cache"

stevespringett:
Reverts stevespringett/dependency-track#24


stevespringett:
Will apply functionality when #26 is fully committed and functional.

PR #28, opened 2015-11-28
Revert "enable HEAD method and add Last-Modified header"

stevespringett:
Reverts stevespringett/dependency-track#25


stevespringett:
Will apply functionality when #26 is fully committed and functional.

PR #30, opened 2016-01-03
enable HEAD method and add Last-Modified header

bezda:
required for dependency-check-maven plugin to be able to use custom cve cache - cveUrl properties


stevespringett:
Thanks @bezda

PR #32, opened 2016-03-03
Added maven enforcer plugin and fixed warnings

wjase:
This resolves a number of duplicate classes and multiple concurrent versions of dependencies which go undetected without the enforcer plugin.


stevespringett:
Thanks for the PR. Feel free to add yourself as a developer in the pom.

PR #62, opened 2017-07-28
Feature/fix issues

MichaelGissingNC:
1. make it work when the app does not run at /
1. fix a JS error


MichaelGissingNC:
@stevespringett Is this PR acceptable for you?

stevespringett:
Thanks for the PR. I've verified the changes fix the issue reported. Also noticed the logout functionality doesn't work correctly in a non-root context, but I'll work on a fix for that. Again, thanks for the PR.

PR #80, opened 2017-11-29
add optional to change the attribute that all queries use

m1a0yu3:
fit the https://github.com/stevespringett/Alpine/pull/9


stevespringett:
Thanks again for the PR 😄

PR #148, opened 2018-05-10
Config the LDAP security authentication level to use

ldmara:
Cofig the LDAP security authentication level used.


stevespringett:
Thanks again for the PR

PR #157, opened 2018-05-29
Fix GAV parsing so it can handle scan reports from Gradle (fixes #156)

### LINKED ISSUE #156: GAV identifier output from Gradle is not in parenthesis
So scan imports will fail with the message:

Error while processing scan result
null

A gradle scan report will contain names like "de.schlichtherle.truezip:truezip-swing:7.7.7", not "(de.schlichtherle.truezip:truezip-swing:7.7.7)". I guess the latter is what would appear in a Maven generated scan?!

Anyway, I have an incoming fix for it.

jskovjyskebankdk:
The code looks like Maven scans would result in GAVs on the form
(G:A:V), while they look like "G:A:V" in the Gradle report.

Strip the parenthesis if present to handle both cases.


stevespringett:
Perfect. Thanks for the PR.

PR #359, opened 2019-06-06
Add a view to list affected projects / dependencies by specific vulnerability with analysis comment. 

mathuriga:
This PR modify the existing view with some additional metadata of affected products/dependencies by the specific vulnerability. Refer below for more information. 

**Current Behaviour:** 

When we search for specific vulnerability, it shows the page which contains the details of that particular vulnerability with affected products. 
vulnerability

**Proposed Improvement:** 

Currently above feature list down only affected project and version. If we want to get the metadata of affected components of a particular project and analysis metadata we have to navigate to that specific project and search for that vulnerability which increases users navigation effort. In order to avoid this, We can have following a complete view of particular vulnerability and how it affects projects dependencies.

**Solution:**

When we search particular vulnerability in the vaulnerability page, list down the projects, it's components which are bound to a specific vulnerability. In DT point of view, it considers a project as affected when the vulnerability is not suppressed for that project. It would be beneficial if we provide the two view which lists down only affected projects (Exclude Suppression) and which lists down affected projects as well as suppressed projects to get an idea why it was suppressed (Include Suppression). The following metadata will give a complete view of particular vulnerability across all the projects. 

01. Project
02. version
03. Component
04. Comment - Last audit comment (This excludes the state change comments)
05. Analysis State
06. Suppressed

**View 01:** Exclude Suppression - Lists down only affected projects which is default behavior and existing behavior

excludeSuppression

**View 02:** Include Suppression - Lists down all projects that are affected and suppressed. This will give a complete view of particular vulnerability. 

includeSuppression

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
@@ -205,7 +205,6 @@ public class VulnerabilityResource extends AlpineResource {
                         for (Component component : vulnerability.getComponents()) {
                             DependencyAnalysisResponse dependencyAnalysisResponse;
                             Analysis analysis = qm.getAnalysis(project, component, vulnerability);
-                            System.out.println(analysis.getProject().getName());
                             // to hold the last mitigation comment of a dependency
                             String lastComment;
                             if (analysis != null) {
@@ -214,7 +213,6 @@ public class VulnerabilityResource extends AlpineResource {
                                 List<String> comments = new ArrayList<>();
                                 for (AnalysisComment analysisComment : analysis.getAnalysisComments()) {
                                     String comment = analysisComment.getComment();
-                                    System.out.println("current comment:" + comment);
                                     // if a comment contains any analysis state transition , ignore that comment.
                                     if (comment.contains("EXPLOITABLE") || comment
                                             .contains("FALSE_POSITIVE") || comment
@@ -231,7 +229,6 @@ public class VulnerabilityResource extends AlpineResource {
                                 } else {
                                     lastComment = null;
                                 }
-                                System.out.println(lastComment);
                                 dependencyAnalysisResponse = new DependencyAnalysisResponse(
                                         project.getName(), project.getVersion(), project.getUuid(), component.getName(),
                                         lastComment, analysis.getAnalysisState(), analysis.isSuppressed());
--- a/src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
@@ -178,13 +178,14 @@ public class VulnerabilityResource extends AlpineResource {
     }
 
     @GET
-    @Path("/source/{source}/vuln/{vuln}/suppressed/{suppressed}/analysis")
+    @Path("/source/{source}/vuln/{vuln}/includeSuppressed/{includeSuppressed}/analysis")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(
-            value = "Returns a list of all analysis metadata of dependencies based on given vulnerability ",
+            value = "Returns a list of all analysis metadata of affected dependency based on given vulnerability ",
             response = DependencyAnalysisResponse.class,
             responseContainer = "List",
-            responseHeaders = @ResponseHeader(name = TOTAL_COUNT_HEADER, response = Long.class, description = "The total number of projects")
+            responseHeaders = @ResponseHeader(name = TOTAL_COUNT_HEADER, response = Long.class,
+                    description = "The total number of projects")
     )
     @ApiResponses(value = {
             @ApiResponse(code = 401, message = "Unauthorized"),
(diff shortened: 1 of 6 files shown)


mathuriga:
Hi @stevespringett ,

Appreciate if you could provide a suggestion on this PR regarding the changes needs to be done or introduced improvement?

Further, Please note that currently, I'm looking at the test case failure.

stevespringett:
@mathuriga big thanks for this PR. Due to the changes and how close it was to the 3.5 release date, I couldn't include it. But this is a really useful view. Also thanks for pinging. It dropped off my radar after the 3.5 release.

I'm going to review it now and merge if everything is ok. The only thing I don't like is the comments, but I also don't have a suggestion on how to make it more useful. Perhaps a way to expand the comments to reveal them all? Not entirely sure.

mathuriga:
@stevespringett,

Thanks for the prompt update. On behalf of @WSO2, I have evaluated currently using Dependency Track to monitor third-party dependencies. During our evaluation, we came up with the following use cases. 

01. During the product release time, teams have to analysis the reported vulnerabilities and update the analysis findings in the audit tab. The CVE analysis state depends on the usages of the particular dependencies. If the CVE analysis state comes under not a threat or false positive, we give a reason why this CVE falls under not a threat or false positive as a comment in audit tap. In this case, this comment column lists the last comment (Excludes the comments related to audit changes). 

02. During the analysis time, the analyzer should able to get a complete view of particular CVE and how effects across the products in a single view rather than spending time to navigating through all products view. 

03. If any queries come from users on how particular CVE affects WSO2 products, we would able to get a complete view with the comment at once rather than navigating to each and every product.

IMHO, We can consider this use case as a generic way.  Especially this will be useful for folks who handle the support tickets and dependency auditor. Further, If anyone wants to remove the comment view, he/she can unselect the "Comment" column.

If you have any suggestions let me know.

mathuriga:
Hi @stevespringett ,

Ping. 😊

mathuriga:
Hi @stevespringett

mathuriga:
Hi @stevespringett ,

Any feedback or improvements for this PR.. 🙂

msymons:
This addition to functionality really would be useful.  For instance, the display of affected component can result in the display JQuery or Primefaces (or other packages) for CVE-2019-11358.

It would be useful if the component version were also displayed, as seeing that in a list of multiple projects would let one immediately identify "low hanging fruit"...  the projects that will likely be easiest to update because the vulnerable component version is already "almost up to date".

stevespringett:
Apologies for the delay. 

I think the only thing I would like changed is instead of having tabs with two tables for affected projects, I'd rather have a single table with a checkbox above the table for an option to include suppressed findings.

This way, only one request would be made by default (instead of two), and if the user wanted to also include suppressed, they would have the option to do that. And only at that time would another request be made.

Make sense. Agree? Disagree?

If agree, then there's a few things that would need to happen in order for this change to work.
1) the table would have to have a `data-toolbar` defined
2) the toolbar would include the checkbox option
3) a listener would be added to:
  a) update the url of the table on state change
  b) add/remove the suppressed column on state change
  c) refresh the contents silently

If you're comfortable doing this, then go ahead. If not, then I'll merge it in and make the changes real quick.

src/main/webapp/vulnerability/index.jsp
```diff
@@ -33,17 +33,64 @@
...
-                   data-show-refresh="true" data-show-columns="true" data-search="true"
-                   data-response-handler="formatProjectsTable" data-height="100%">
-                <thead>
-                <tr>
-                    <th data-align="left" data-field="projecthref">Name</th>
-                    <th data-align="left" data-field="version">Version</th>
-                </tr>
-                </thead>
-            </table>
+            <div class="panel with-nav-tabs panel-default tight">
```
stevespringett: Lets make this into a single table again, but with the additional Component, Comment, and Analysis State fields with a table toolbar containing a checkbox to optionally include suppressed. When suppressed is included, the Suppressed field would be made visible in the table, the URL would be updated to specify `?includeSuppressed=true` and the table would silently refresh.

stevespringett: If you need my help here. Not a problem. Just let me know.

msymons:
@mathuriga,  as commented previously, I believe that this enhancement will be really useful for Dependency-Track.  Is there any chance that the requested changes could be done in time for 3.6 release?  I do notice that the last travis build was a couple of months ago.  I seem to recall that a travis change was implemented since then that should (all other things being equal) address the build fail problem.

stevespringett:
FYI. I will be releasing v3.6 in the next few days.

mathuriga:
@, Sorry for the delayed response. Thanks for the suggested changes. I will do the possible requested change within couple of days and update you.

stevespringett:
I plan on releasing v3.6 at approx 05:00 UTC Sunday. If you think you can have the changes made by then, great. If you don't think you can have the changes made by then, please let me know ASAP.

All work targeted for v3.6 has been completed. Documentation is ready for publication. The only remaining items that I need to do is perform upgrade testing one last time and write a release announcement. 

If you're not able to get these changes done by then, please let me know as soon as possible. I might be able to squeeze it in (if I know early enough). As a last resort, it might have to be moved to v3.7.

src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
```diff
@@ -173,27 +178,74 @@ public Response getVulnerabilityByVulnId(@PathParam("source") String source,
     }
 
     @GET
-    @Path("/source/{source}/vuln/{vuln}/projects")
+    @Path("/source/{source}/vuln/{vuln}/includeSuppressed/{includeSuppressed}/analysis")
```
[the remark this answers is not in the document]
mathuriga: I have resolved this suggestion by following way:

Leave the "/source/{source}/vuln/{vuln}/projects" as it is (Inorder to avoid the breaking existing API).
Introduce new API '/source/{source}/vuln/{vuln}/dependencyAnalysis' with optional parameter to achieve above use case.

**Reason to introduce new API.**
   "/source/{source}/vuln/{vuln}/projects" API returns Project model. Since Project model does not have analysis metadata, I have introduced above mentioned new API which returns DependencyAnalysisResponse.

If you have any concerns about this, please let me know. :)

src/main/webapp/vulnerability/index.jsp
mathuriga: Ack.. Due to time constraint, I couldn't able to finish this on time. If I suppose to do, it will take bit time since I have to look at a single-table structure. Therefore IMHO, It would be great if you help on this.

mathuriga:
> I plan on releasing v3.6 at approx 05:00 UTC Sunday. If you think you can have the changes made by then, great. If you don't think you can have the changes made by then, please let me know ASAP.
> 
> All work targeted for v3.6 has been completed. Documentation is ready for publication. The only remaining items that I need to do is perform upgrade testing one last time and write a release announcement.
> 
> If you're not able to get these changes done by then, please let me know as soon as possible. I might be able to squeeze it in (if I know early enough). As a last resort, it might have to be moved to v3.7.

Thank you for your patience. I have modified the request changes in API Part. For Table View suggestion, I need help on this as mentioned above.

Thanks in advance.

stevespringett:
Thanks for the updates. I'll assist the UI table elements. However, the new API to retrieve affected projects has some issues. For example, if there are multiple versions of a component in a single project (common with .js) , there will be duplicate affected projects. Each one could potentially have a different audit decision.

Screen Shot 2019-09-28 at 9 20 09 PM

I have a good idea on how this could potentially be fixed and will work on getting it done in the coming weeks.

mathuriga:
@stevespringett ,

Thanks for the update. May I know the status of this PR? Will you able to merge this? 😊

Thanks.

stevespringett:
The comment from September 28 will need to be resolved. I have not had a chance to look into this yet.

https://github.com/DependencyTrack/dependency-track/pull/359#issuecomment-536241872

Because of the observations noted in this comment, the code hasn't been merged yet. I think this is the only thing preventing the code from being merged.

nscuro:
Closing due to inactivity. Please open a new PR if you still want to introduce the proposed functionality.

PR #360, opened 2019-06-06
Fetch JDBC drivers over HTTPS

angrylogic:
This change updates the Dockerfile so that the JDBC drivers are fetched over a secure channel via HTTPS.  This prevents a MITM attack against these dependencies which might result in a compromised docker container.


stevespringett:
Thanks for the PR

Issue #354, opened 2019-05-29
Dependency Check Support


jaskaransinghdr6j:
Hey Steve,

We've been using Dependency Track 3.4.0 for a while now, and have recently been getting inconsitent errors while importing Dependency check reports.

The API posting the scans is successful, the project is created in Dependency Track (PROJECT table entry present), but the Project has "No Data Available" and no dependencies are associated with the project as well. The SCAN table also does not create an entry for the project and it's associated uploaded scan.

Current Setup:
Dependency Track : 3.4.0
Dependency Chck : 5.0.0M2

I have not noticed the same in Dependency Check v4, hence would like to know if Dependency Check v5 reports will be supported in 3.4.1 or 3.4.5 or in future so we can plan accordingly. We have a service based Dependency Check implementation which offloads scanning to our servers instead of Jenkins/Dev machines, due to large volume, and scans are then posted to Dependency Track for aggregation and notifications. 

Keep up the good work!

stevespringett:
Importing Dependency-Check scans is not recommended. It can be done, but the result will not be great. See: https://docs.dependencytrack.org/best-practices/

Dependency-Check 5 introduces a new XML schema which Dependency-Track does not support. Last I checked, the timestamp in the XML report was either in a different format than before, or was missing, and the XSD only specified it as being a string, so its hard to know what the specific issue is. But the v2 XSD is incompatible with the 1.x XSD's which Dependency-Track v1 - v4 used. 

In Dependency-Track v3.5, I will likely put in a check for the Dependency-Check XML version and abort with warning (log) if schema v2 or higher is uploaded.

At the moment, importing ODC5 is a very low priority since it's not a best practice and there is a backlog of risk identification and reduction enhancements that the community is waiting for.

PR #375, opened 2019-06-18
Parse latest date format from dependency-check XML reports

angrylogic:
At some point [Dependency Check](https://github.com/jeremylong/DependencyCheck/blob/55d8f6659746e71b00f1f480d683b9660a5679d7/core/src/main/java/org/owasp/dependencycheck/reporting/ReportGenerator.java#L204) moved to using [DateTimeFormatter.ISO_INSTANT](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_INSTANT) to generate the scan date inside the XML report.  This format reports dates in UTC without a timezone and causes Dependency Track to fail to parse these reports (and thus throw a NPE when one is ingested via the API).

This patch updates the date parsing code so that both the old and new formats are tried.  It should allow Dependency Track to parse reports from both versions.

### DIFF
--- a/src/main/java/org/dependencytrack/parser/dependencycheck/model/DateAdapter.java
+++ b/src/main/java/org/dependencytrack/parser/dependencycheck/model/DateAdapter.java
@@ -20,7 +20,9 @@ package org.dependencytrack.parser.dependencycheck.model;
 
 import javax.xml.bind.annotation.adapters.XmlAdapter;
 import java.text.SimpleDateFormat;
+import java.text.ParseException;
 import java.util.Date;
+import java.util.TimeZone;
 
 /**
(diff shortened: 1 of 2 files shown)


stevespringett:
Thanks for the PR. I don't know when/if DC 5 will be supported as an import format though. 

Refer to #354

It's not a best practice so it really depends on the community to prioritize support for it over other things.

angrylogic:
Thanks @stevespringett, I've removed the vim swp file that snuck in.  

If it's an intentional decision not to support this format going forward I can withdraw this PR.

stevespringett:
Thanks @angrylogic

PR #400, opened 2019-07-24
Fix "Remove Dependency" functionality in project view

nscuro:
This PR fixes an issue where removing a Dependency from a Project was not possible via web UI. The UI was submitting `null`s as Component UUIDs to the backend, which responded with HTTP 500. 

Turns out that simply one word was missing. `selections` contains objects of type `Dependency`, not `Component`.


stevespringett:
Thanks for finding and correcting the issue. The fix will be included in v3.6.0

PR #420, opened 2019-08-22
New Dockerfile based on Alpine linux and compatible with Kubernetes and OpenShift

mterron:
Make it compatible with running as an arbitrary user as long as it's a member of GID 1000. This is useful in Kubernetes and OpenShift environments where you don't have control over the UID your container runs under.

### DIFF
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -17,7 +17,6 @@ COPY --chown=1000 ./target/dependency-track-embedded.war "$APP_DIR"/
 # Create a user and assign home directory to a $DATA_DIR
 # Download optional JDBC drivers to the external library directory
 # Ensure UID 1000 & GID 1000 own all the needed directories
-# Ensure GID 1000 have access to the database drivers
 # Install JRE
 RUN mkdir -p -m 770 "$DATA_DIR" "$EXTLIB_DIR" \
     && adduser -D -h "$DATA_DIR" -u 1000 "$USERNAME" \
@@ -25,7 +24,6 @@ RUN mkdir -p -m 770 "$DATA_DIR" "$EXTLIB_DIR" \
     && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-connector-java-5.1.47.jar \
     && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar \
 	&& chown -R "$USERNAME":"$USERNAME" "$EXTLIB_DIR" \
-    && chmod -R 440 "$EXTLIB_DIR"/* \
 	&& apk --no-cache add openjdk8-jre-base
 
 # Specify the user to run as (in numeric format for compatibility with Kubernetes/OpenShift's SCC)
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -24,7 +24,7 @@ RUN mkdir -p -m 770 "$DATA_DIR" "$EXTLIB_DIR" \
 	&& wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/org/postgresql/postgresql/42.2.5/postgresql-42.2.5.jar \
     && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-connector-java-5.1.47.jar \
     && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar \
-	&& chown -R "$USERNAME":"$USERNAME" "$EXTLIBDIR" \
+	&& chown -R "$USERNAME":"$USERNAME" "$EXTLIB_DIR" \
     && chmod -R 440 "$EXTLIB_DIR"/* \
 	&& apk --no-cache add openjdk8-jre-base
 
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8u201-jre-alpine
+FROM alpine:3.10
 LABEL maintainer="steve.springett@owasp.org"
 LABEL vendor="OWASP"
 
@@ -8,8 +8,9 @@ ARG DATA_DIR=/data
 ARG EXTLIB_DIR=/extlib
 ARG USER=dtrack
 
+RUN apk --no-cache add openjdk8-jre-base \
 # Create the application directory where Dependency-Track will be installed and the external library directory
-RUN mkdir -p "$APP_DIR" "$DATA_DIR" "$EXTLIB_DIR"
+	&& mkdir -p "$APP_DIR" "$DATA_DIR" "$EXTLIB_DIR"
 
 # Copy the compiled WAR to the application directory created above
 COPY ./target/dependency-track-embedded.war "$APP_DIR"
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -3,40 +3,47 @@ LABEL maintainer="steve.springett@owasp.org"
 LABEL vendor="OWASP"
 
 # Arguments that can be passed at build time
-ARG APP_DIR=/opt/owasp/dependency-track
-ARG DATA_DIR=/data
-ARG EXTLIB_DIR=/extlib
-ARG USER=dtrack
-
-RUN apk --no-cache add openjdk8-jre-base \
-# Create the application directory where Dependency-Track will be installed and the external library directory
-	&& mkdir -p "$APP_DIR" "$DATA_DIR" "$EXTLIB_DIR"
+# Directory names must end with / to avoid errors when ADDing and COPYing
+ARG APP_DIR=/opt/owasp/dependency-track/
+ARG DATA_DIR=/data/
+ARG EXTLIB_DIR=/extlib/
+ARG USERNAME=dtrack
 
 # Copy the compiled WAR to the application directory created above
-COPY ./target/dependency-track-embedded.war "$APP_DIR"
+# Automatically creates the $APP_DIR directory
+COPY --chown=1000 ./target/dependency-track-embedded.war "$APP_DIR"/
+
+# Create the directory where Dependency-Track will store its data ($DATA_DIR) and the external library directory ($EXTLIB_DIR)
+# Create a user and assign home directory to a $DATA_DIR
 # Download optional JDBC drivers to the external library directory
-ADD https://repo1.maven.org/maven2/org/postgresql/postgresql/42.2.5/postgresql-42.2.5.jar "$EXTLIB_DIR"
-ADD https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-connector-java-5.1.47.jar "$EXTLIB_DIR"
-ADD https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar "$EXTLIB_DIR"
+# Ensure UID 1000 & GID 1000 own all the needed directories
+# Ensure GID 1000 have access to the database drivers
+# Install JRE
+RUN mkdir -p -m 770 "$DATA_DIR" "$EXTLIB_DIR" \
+	&& apk --no-cache add openjdk8-jre-base \
+	&& wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/org/postgresql/postgresql/42.2.5/postgresql-42.2.5.jar \
+    && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-connector-java-5.1.47.jar \
+    && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar \
+	&& chown -R "$USERNAME":"$USERNAME" "$EXTLIBDIR" \
+    && chmod -R 440 "$EXTLIB_DIR"/* \
+    && adduser -D -h "$DATA_DIR" -u 1000 "$USERNAME"
 
-# Create a user and assign home directory to a non-standard path
-RUN adduser -h "$DATA_DIR" -D "$USER" \
-# Ensure "$USER" have access to all the needed directories
-    && chown -R "$USER":"$USER" "$APP_DIR" "$EXTLIB_DIR" \
-    && chmod -R 770 "$APP_DIR" "$DATA_DIR" "$EXTLIB_DIR"
-
-# Specify the user to run commands as
+# Specify the user to run as (in numeric format for compatibility with Kubernetes/OpenShift's SCC)
 USER 1000
+
+# Specify the container working directory
 WORKDIR ${APP_DIR}
 
 # Launch Dependency-Track
 CMD java $JAVA_OPTIONS -DdependencyTrack.logging.level=$LOGGING_LEVEL -jar dependency-track-embedded.war
 
+# Specify which port Dependency-Track listens on
 EXPOSE 8080
 
-# The default logging level
+# Dependency-Track's default logging level
 ENV LOGGING_LEVEL=INFO
+
 # Environment variables that can be passed at runtime
 ENV JAVA_OPTIONS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Xmx4G"
-
+ 
 #VOLUME "$DATA_DIR"
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -46,3 +46,5 @@ ENV LOGGING_LEVEL=INFO
 ENV JAVA_OPTIONS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Xmx4G"
  
 #VOLUME "$DATA_DIR"
+
+HEALTHCHECK --interval=5m --timeout=3s CMD wget -q -O/dev/null http://127.0.0.1:8080/api/version
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -43,7 +43,7 @@ EXPOSE 8080
 ENV LOGGING_LEVEL=INFO
 
 # Environment variables that can be passed at runtime
-ENV JAVA_OPTIONS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Xmx4G"
+ENV JAVA_OPTIONS="-XX:MaxRAMPercentage=80.0"
  
 #VOLUME "$DATA_DIR"
 
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -20,13 +20,13 @@ ADD https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-conne
 ADD https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar "$EXTLIB_DIR"
 
 # Create a user and assign home directory to a non-standard path
-RUN adduser -h "$DATA_DIR" -s bash -D "$USER" \
+RUN adduser -h "$DATA_DIR" -D "$USER" \
 # Ensure "$USER" have access to all the needed directories
     && chown -R "$USER":"$USER" "$APP_DIR" "$EXTLIB_DIR" \
     && chmod -R 770 "$APP_DIR" "$DATA_DIR" "$EXTLIB_DIR"
 
 # Specify the user to run commands as
-USER "$USER"
+USER 1000
 WORKDIR ${APP_DIR}
 
 # Launch Dependency-Track
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -20,13 +20,13 @@ COPY --chown=1000 ./target/dependency-track-embedded.war "$APP_DIR"/
 # Ensure GID 1000 have access to the database drivers
 # Install JRE
 RUN mkdir -p -m 770 "$DATA_DIR" "$EXTLIB_DIR" \
-	&& apk --no-cache add openjdk8-jre-base \
+    && adduser -D -h "$DATA_DIR" -u 1000 "$USERNAME" \
 	&& wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/org/postgresql/postgresql/42.2.5/postgresql-42.2.5.jar \
     && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-connector-java-5.1.47.jar \
     && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar \
 	&& chown -R "$USERNAME":"$USERNAME" "$EXTLIBDIR" \
     && chmod -R 440 "$EXTLIB_DIR"/* \
-    && adduser -D -h "$DATA_DIR" -u 1000 "$USERNAME"
+	&& apk --no-cache add openjdk8-jre-base
 
 # Specify the user to run as (in numeric format for compatibility with Kubernetes/OpenShift's SCC)
 USER 1000
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -5,39 +5,37 @@ LABEL vendor="OWASP"
 # Arguments that can be passed at build time
 ARG APP_DIR=/opt/owasp/dependency-track
 ARG DATA_DIR=/data
+ARG EXTLIB_DIR=/extlib
 ARG USER=dtrack
 
-# The default logging level
-ENV LOGGING_LEVEL=INFO
-
-# Environment variables that can be passed at runtime
-ENV JAVA_OPTIONS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Xmx4G"
-
-# Create the application directory where Dependency-Track will be installed
-RUN mkdir -p ${APP_DIR}
-
-# Create the external library directory
-RUN mkdir -p /extlib
-
-# Create a user and assign home directory to a non-standard path
-RUN adduser -h ${DATA_DIR} -s bash -D ${USER}
+# Create the application directory where Dependency-Track will be installed and the external library directory
+RUN mkdir -p -m 770 "$APP_DIR" "$DATA_DIR" "$EXTLIB_DIR"
 
 # Copy the compiled WAR to the application directory created above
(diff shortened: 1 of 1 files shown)


mterron:
This seems to be the reason why Travis fails: https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

mterron:
Hi Steve, this are the minimum set of changes to achieve a smooth running image (in a secured Kubernetes cluster) and rebasing on Alpine + Dockerfile best parctices. Happy to discuss line by line.

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,42 @@
-FROM openjdk:8u201-jre-alpine
+FROM alpine:3.10
```
mterron: Rebase on Alpine as per our discussion in Gitter

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,42 @@
...
 LABEL maintainer="steve.springett@owasp.org"
 LABEL vendor="OWASP"
 
 # Arguments that can be passed at build time
 ARG APP_DIR=/opt/owasp/dependency-track
 ARG DATA_DIR=/data
+ARG EXTLIB_DIR=/extlib
```
mterron: extlib was the only hardcoded directory so thought it'd be good to parametrise it too.

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,42 @@
...
 
-# The default logging level
-ENV LOGGING_LEVEL=INFO
-
-# Environment variables that can be passed at runtime
-ENV JAVA_OPTIONS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Xmx4G"
-
-# Create the application directory where Dependency-Track will be installed
-RUN mkdir -p ${APP_DIR}
-
```
mterron: Consolidate multiple RUN statements.

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,42 @@
...
-# Create the external library directory
-RUN mkdir -p /extlib
-
-# Create a user and assign home directory to a non-standard path
-RUN adduser -h ${DATA_DIR} -s bash -D ${USER}
```
mterron: `bash` is not present on the base Alpine image

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,42 @@
...
+RUN apk --no-cache add openjdk8-jre-base \
```
mterron: Install JRE

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,42 @@
...
-USER ${USER}
+# Create a user and assign home directory to a non-standard path
+RUN adduser -h "$DATA_DIR" -D "$USER" \
+# Ensure "$USER" have access to all the needed directories
+    && chown -R "$USER":"$USER" "$APP_DIR" "$EXTLIB_DIR" \
+    && chmod -R 770 "$APP_DIR" "$DATA_DIR" "$EXTLIB_DIR"
 
-EXPOSE 8080
+# Specify the user to run commands as
+USER 1000
```
mterron: OpenShift requires the user to be an UID

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,42 @@
...
+WORKDIR ${APP_DIR}
 
 # Launch Dependency-Track
-WORKDIR ${APP_DIR}
 CMD java $JAVA_OPTIONS -DdependencyTrack.logging.level=$LOGGING_LEVEL -jar dependency-track-embedded.war
+
+EXPOSE 8080
```
mterron: Move to the bottom as to not invalidate cached layer if these values need to change.

mterron:
Codacy "error" should be a "warning". I know what `mkdir -p -m 770 $DIR" does and it is exactly what I want in this case.

mterron:
@stevespringett , `dockle` and `trivy` are scanning a different image from the one that is built by Travis.

See https://travis-ci.org/DependencyTrack/dependency-track/builds/577642216#L4550 pointing to a non existent ADD instruction and https://travis-ci.org/DependencyTrack/dependency-track/builds/577642216#L4566 mentioning Alpine 3.9.3 when the Dockerfile points to alpine:3.10

mterron:
> Overall, there were a lot of changes, many of which do not appear to be related to using native alpine or K8s. I would recommend sticking with only the minimum amount of changes required in order to meet these requirements while still addressing the comments.

I guess it is fair to say that this is a completely new (and hopefully improved) Dockerfile and not a modification of the existing one.

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,48 @@
...
-
-# Create the external library directory
-RUN mkdir -p /extlib
-
-# Create a user and assign home directory to a non-standard path
-RUN adduser -h ${DATA_DIR} -s bash -D ${USER}
+# Directory names must end with / to avoid errors when ADDing and COPYing
+ARG APP_DIR=/opt/owasp/dependency-track/
+ARG DATA_DIR=/data/
+ARG EXTLIB_DIR=/extlib/
```
mterron: Parametrise location of the external library directory in line with the rest.

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,48 @@
...
 # Download optional JDBC drivers to the external library directory
-ADD https://repo1.maven.org/maven2/org/postgresql/postgresql/42.2.5/postgresql-42.2.5.jar /extlib
-ADD https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-connector-java-5.1.47.jar /extlib
-ADD https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar /extlib
-RUN chown -f -R ${USER}:${USER} /extlib
+# Ensure UID 1000 & GID 1000 own all the needed directories
+# Install JRE
+# hadolint ignore=SC2174
+RUN mkdir -p -m 770 "$DATA_DIR" "$EXTLIB_DIR" \
+    && adduser -D -h "$DATA_DIR" -u 1000 "$USERNAME" \
```
mterron: `bash` is not present in the default Alpine image

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,48 @@
...
+# Create the directory where Dependency-Track will store its data ($DATA_DIR) and the external library directory ($EXTLIB_DIR)
+# Create a user and assign home directory to a $DATA_DIR
 # Download optional JDBC drivers to the external library directory
-ADD https://repo1.maven.org/maven2/org/postgresql/postgresql/42.2.5/postgresql-42.2.5.jar /extlib
-ADD https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-connector-java-5.1.47.jar /extlib
-ADD https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar /extlib
-RUN chown -f -R ${USER}:${USER} /extlib
+# Ensure UID 1000 & GID 1000 own all the needed directories
+# Install JRE
+# hadolint ignore=SC2174
```
mterron: Ignore bogus error from Codacy Dockerfile linter.

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,48 @@
...
+RUN mkdir -p -m 770 "$DATA_DIR" "$EXTLIB_DIR" \
```
mterron: Consolidate multiple RUN statements as per Docker best practices.

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,48 @@
...
+    && adduser -D -h "$DATA_DIR" -u 1000 "$USERNAME" \
+	&& wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/org/postgresql/postgresql/42.2.5/postgresql-42.2.5.jar \
+    && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/mysql/mysql-connector-java/5.1.47/mysql-connector-java-5.1.47.jar \
+    && wget -P "$EXTLIB_DIR" https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/7.1.3.jre8-preview/mssql-jdbc-7.1.3.jre8-preview.jar \
+	&& chown -R "$USERNAME":"$USERNAME" "$EXTLIB_DIR" \
+	&& apk --no-cache add openjdk8-jre-base
+
+# Specify the user to run as (in numeric format for compatibility with Kubernetes/OpenShift's SCC)
+USER 1000
+
```
mterron: Move this block to the end as to not invalidate cached layer if these values need to change.

src/main/docker/Dockerfile
```diff
@@ -1,43 +1,48 @@
...
-# Launch Dependency-Track
-WORKDIR ${APP_DIR}
-CMD java $JAVA_OPTIONS -DdependencyTrack.logging.level=$LOGGING_LEVEL -jar dependency-track-embedded.war
+# Dependency-Track's default logging level
+ENV LOGGING_LEVEL=INFO
+
+# Environment variables that can be passed at runtime
+ENV JAVA_OPTIONS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Xmx4G"
+ 
+#VOLUME "$DATA_DIR"
```
mterron: VOLUME declared on the Dockerfile makes extending the image more difficult. See https://github.com/moby/moby/issues/3639

stevespringett:
I've decided to not use Alpine directly and not use the generic OpenJDK distribution. This is due to the changes OpenJDK has in version 11 onwards. AdoptOpenJDK will be used as it will maintain LTS releases directly and also has an Alpine base. The Alpine base is not perfect. I've submitted a change which should be implemented soon.

The three things from your contributions that I have not accepted are the volume statement, changes to java_options, and quoted variables. Other than that, I think everything should meet the K8/OS requirements.

The new proposed Dockerfile is here: https://github.com/DependencyTrack/dependency-track/blob/master/src/main/docker/Dockerfile.next

The original Dockerfile is still checked in. Once Dockerfile.next is finalized, it'll be merged into Dockerfile.

mterron:
> I've decided to not use Alpine directly and not use the generic OpenJDK distribution. This is due to the changes OpenJDK has in version 11 onwards. AdoptOpenJDK will be used as it will maintain LTS releases directly and also has an Alpine base. The Alpine base is not perfect. I've submitted a change which should be implemented soon.

AdoptOpenJDK does not officially support Alpine linux (see: https://adoptopenjdk.net/supported_platforms.html) what you're using in `Dockerfile.new` is an unofficial release. What was missing from the official Alpine image that steered you in this direction? 

> The three things from your contributions that I have not accepted are the volume statement, changes to java_options, and quoted variables. Other than that, I think everything should meet the K8/OS requirements.

Why do you need the anonymous VOLUME declared on the Dockerfile? It is generally a bad idea as per the link I provided where Docker employees themselves argue that it was a bad decision to have it available in the first place.

Java8 from u131 onwards has backported the docker support changes from 9 and 10 so `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` is not needed anymore as that is the default configuration. The configuration I suggested would adapt the heap size to the memory assigned to the container instead of capping it at 4g as it currently does and also it'll be forward compatible with Java 11 option syntax.

As for quoted variables, it is safer and have no downsides so I see no reason not to do it. As soon as someone tries to set `$DATA_DIR` to `/my data directory/` or anything with spaces  the Dockerfile will break.

Cheers

stevespringett:
> What was missing from the official Alpine image that steered you in this direction?

Java 11 LTS was the deciding factor for that. Oracle maintains OpenJDK LTS releases for only 6 months. After 6 months, you have to use Oracle, a JDK distro that maintains LTS releases (such as AdoptOpenJdk), or rely on individual OpenJDK distributors (such as RedHat) to maintain OpenJDK. It's unclear who maintains Alpine OpenJDK releases. Having a dedicated project maintain OpenJDK LTS releases was the deciding factor. The requirement for glibc applies to any Java release running on Alpine, not just AdoptOpenJdk.

I will be experimenting with removing `XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` this week. I have not verified the behavior works, but once verified they will be removed.

The question of Xmx4G is a matter of adoption vs optimization. Dependency-Track will not get adopted if a users first experience is their inability to execute it. I think the Dockerfile comments and perhaps the documentation should include the percentage.

I will need to experiment without having a volume in the Dockerfile. If that will impact the complexity of execution or any doc updates. It was left in there for the time being because, at this point, I don't know what some of those answers are.

mterron:
> > What was missing from the official Alpine image that steered you in this direction?
> 
> Java 11 LTS was the deciding factor for that. Oracle maintains OpenJDK LTS releases for only 6 months. After 6 months, you have to use Oracle, a JDK distro that maintains LTS releases (such as AdoptOpenJdk), or rely on individual OpenJDK distributors (such as RedHat) to maintain OpenJDK. It's unclear who maintains Alpine OpenJDK releases. Having a dedicated project maintain OpenJDK LTS releases was the deciding factor. The requirement for glibc applies to any Java release running on Alpine, not just AdoptOpenJdk.

That is definitely not the case Steve. Alpine native openjdk packages (as well as Azul's container we discussed on Gitter) are both linked against musl libc, the native libc on Alpine linux. 
There's no glibc requirement for Java, it is just what AdoptOpenJDK chooses to support and the reason I mentioned that the base image you are using on the new Dockerfile is unsupported by AdoptOpenJDK.

```
 # docker run -it --rm alpine
/ # apk add openjdk8-jre-base file
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/community/x86_64/APKINDEX.tar.gz
...
/ # file $(which java)
/usr/bin/java: symbolic link to ../lib/jvm/default-jvm/jre/bin/java
/ # file /usr/lib/jvm/default-jvm/jre/bin/java
/usr/lib/jvm/default-jvm/jre/bin/java: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, stripped
```
The Alpine linux team maintains and supports the openjdk builds on Alpine for openjdk 8, 9, 10 and 11. 
Azul systems maintain, support and certify their own "distro" of OpenJDK too without using glibc.


> I will be experimenting with removing `XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` this week. I have not verified the behavior works, but once verified they will be removed.
> 
> The question of Xmx4G is a matter of adoption vs optimization. Dependency-Track will not get adopted if a users first experience is their inability to execute it. I think the Dockerfile comments and perhaps the documentation should include the percentage.
>

The main problem with -Xmx is that if the user assigns more than 4g of memory to Dependency Track's container, it won't be able to use it and if he assigns less than 4g and DTrack tries to use it, it'll be killed by the OOMKiller in Docker for excessive memory consumption.
Using the percentage adapts to whatever the user defines as the available resources. Probably we should also add -XX:InitialRAMPercentage set to some sensible value.

> I will need to experiment without having a volume in the Dockerfile. If that will impact the complexity of execution or any doc updates. It was left in there for the time being because, at this point, I don't know what some of those answers are.
I think a documentation change to remind users to create a named volume and attach it to /data is a good idea.
Another thing would be to assign memory limits in the example `docker run` and also in the `docker-compose.yml` file. Happy to send a PR.

stevespringett:
> The Alpine linux team maintains and supports the openjdk builds on Alpine for openjdk 8, 9, 10 and 11.

Actually, the IcedTeam team does, not Alpine. There have been multiple occasions where JDK releases for Alpine have been delayed for months due to delays with IcedTea. As of today 8u212 is the latest for Alpine 3.10. 8u222 is what AdoptOpenJdk provides which was released in July and fixes 17 security flaws. 

One interesting reason why AdoptOpenJdk was chosen is due to initial seed times. It turns out that glibc is actually a faster than musl. In repeated tests with v3.6 Docker snapshots (on Alpine 3.10), AdoptOpenJdk would seed in about 13 minutes whereas with OpenJdk would seed in about 22 minutes. This wasn't all that noticeable in Dependency-Track v3.5, as NVD mirroring only accounted for about 130K records. However, in v3.6, there are over 2M records created during NVD seeding. Upon further investigation, there seems to be some benchmarks that back up this finding. https://medium.com/rocket-travel-engineering/alpine-vs-debian-images-for-java-jvm-builds-b8f8e1cc58a8

I've been doing a lot of testing with AdoptOpenJdk on Alpine and it seems to work fine. However, if it runs into trouble, I may consider using a Debian slim image instead of Alpine as I think glibc (at least for the time being) provides the performance necessary.

---
I'm still experimenting with some of the other leftover paramaters, i.e. volume, java_options, etc. I'd like to finalize the dockerfile this weekend if possible, as I'm aiming to release v3.6 next week

mterron:
That's really interesting, I never noticed any statistically significant performance difference between musl and glibc.
I'll run some benchmarks next week.

mterron:
Hi Steve, I tried to reproduce your results but failed.

Tried 4 different configurations of dependency-track:3.5.1: 
1.  std docker image
2. image based on azul/zulu-openjdk:8-jre
3. image based on alpine:3.10 (openjdk8)
4. RedHat UBI8 (glibc based) openjdk8

For all 3 non-std images I copied the same binary and used the same JAVA_OPTS

On my hardware, all 4 off them run [NistMirrorTask] in ~23 minutes +- 10s 

I'll try again on 3.6.0

stevespringett:
Two tests were conducted - one informal with results posted above, and one more formal with results posted below.

Previously, I was testing on my development macOS machine. The deltas on that machine were much more exaggerated. Testing was not dedicated and other processes running. This might be more true to real-world results.

The following test was conduced in a more formal environment. All tests were conduced on a dedicated Docker physical host with a single container running.

### Unconstrained resources

In this configuration, the Docker host did not constrain the amount of RAM available to the container. Usage was around 13GB in all test cases.

| OS | libc | Java Version | NVD Parse Time | Tag |
| :- | :--- | :----------- | -------------: | --- |
| Alpine 3.10 | musl | 8.0.212 (IcedTea) | 535085ms | alpine:3.10 |
| Alpine 3.10 | glibc | 8.0.222 (AdoptOpenJDK) | 486059ms | adoptopenjdk/openjdk8:alpine-jre |
| Ubuntu 18.04 | glibc | 8.0.222 (AdoptOpenJDK) | 491864ms | adoptopenjdk:8-jre-hotspot-bionic |


* Fastest: Alpine 3.10 (glibc) with OpenJDK (AdoptOpenJDK) 8.0.222
* Slowest: Alpine 3.10 (musl) with OpenJDK (IcedTea) 8.0.212
* Delta: __9.6%__


### 8GB RAM Constraint

Container runtime RAM limited via `--memory=8192m`.

| OS | libc | Java Version | NVD Parse Time | Tag |
| :- | :--- | :----------- | -------------: | --- |
| Alpine 3.10 | musl | 8.0.212 (IcedTea) | 557643ms | alpine:3.10 |
| Alpine 3.10 | glibc | 8.0.222 (AdoptOpenJDK) | 507000ms | adoptopenjdk/openjdk8:alpine-jre |
| Ubuntu 18.04 | glibc | 8.0.222 (AdoptOpenJDK) | 518060ms | adoptopenjdk:8-jre-hotspot-bionic |


* Fastest: Alpine 3.10 (glibc) with OpenJDK (AdoptOpenJDK) 8.0.222
* Slowest: Alpine 3.10 (musl) with OpenJDK (IcedTea) 8.0.212
* Delta: __9.5%__


### 4.5GB RAM Constraint

Container runtime RAM limited via `--memory=4608m`.

| OS | libc | Java Version | NVD Parse Time | Tag |
| :- | :--- | :----------- | -------------: | --- |
| Alpine 3.10 | musl | 8.0.212 (IcedTea) | 894447ms | alpine:3.10 |
| Alpine 3.10 | glibc | 8.0.222 (AdoptOpenJDK) | 695166ms | adoptopenjdk/openjdk8:alpine-jre |
| Ubuntu 18.04 | glibc | 8.0.222 (AdoptOpenJDK) | 695676ms | adoptopenjdk:8-jre-hotspot-bionic |


* Fastest: Alpine 3.10 (glibc) with OpenJDK (AdoptOpenJDK) 8.0.222
* Slowest: Alpine 3.10 (musl) with OpenJDK (IcedTea) 8.0.212
* Delta: __25%__


### Java 11 Comparison

| Constraint | OS | libc | Java Version | NVD Parse Time | Tag |
| :- | :- | :--- | :----------- | -------------: | --- |
| None | Ubuntu 18.04 | glibc | 11.0.4 (AdoptOpenJDK) | 557923ms | adoptopenjdk:11-jre-hotspot-bionic |
| 8GB RAM | Ubuntu 18.04 | glibc | 11.0.4 (AdoptOpenJDK) | 576257ms | adoptopenjdk:11-jre-hotspot-bionic |
| 4.5GB RAM | Ubuntu 18.04 | glibc | 11.0.4 (AdoptOpenJDK) | 635250ms | adoptopenjdk:11-jre-hotspot-bionic |

* Java 11 was the 9% faster than Alpine 3.10 (glibc) in the 4.5GB RAM constraint test. Java 11 in this test was 33.8% faster than Alpine 3.10 (musl). However, Java 11 was slower in the other two tests.

stevespringett:
Using the same environment and testing conditions above, additional testing was performed against the Azul Systems OpenJDK distribution.

### Azul Systems Java 8 Comparison

| Constraint | OS | libc | Java Version | NVD Parse Time | Tag |
| :- | :- | :--- | :----------- | -------------: | --- |
| None | Alpine | musl | 8.0.222 (Azul Systems) | 492794ms | azul/zulu-openjdk-alpine:8-jre |
| 8GB RAM | Alpine | musl | 8.0.222 (Azul Systems) | 531405ms | azul/zulu-openjdk-alpine:8-jre |
| 4.5GB RAM | Alpine | musl | 8.0.222 (Azul Systems) |  | azul/zulu-openjdk-alpine:8-jre |

* The unconstrained test showed that Azul Systems OpenJDK had similar performance to the glibc variants, although a bit slower than both. It was 8.2% faster than Alpine OpenJDK (IcedTea) and 1.3% slower than AdoptOpenJDK on Alpine.
* The 8MB constraint test showed that Azul Systems OpenJDK was 4.8% faster than Alpine OpenJDK (IcedTea) and 4.7% slower than AdoptOpenJDK on Alpine.
* The 4.5MB constraint test showed that Azul Systems OpenJDK was 46.7% faster than Alpine OpenJDK (IcedTea) and 22.3% faster than AdoptOpenJDK on Alpine. 

Observation: The Azul Systems OpenJDK distro appears to be optimized for small container usage and outperforms all others under heavy resource constraints. This advantage disappears when the container can provide adequate RAM and is slightly slower than AdoptOpenJDK on Alpine.

mterron:
I ran the test again on `snapshot` and again couldn't reproduce. I'd say it is hardware dependent, this are the results on my hardware sorted from fastest to slowest. This was with 8Gb of ram assigned and the default 90% config.

Image | Time | %
------ | ----- | --
Azul | 1494177ms | 0%
Red Hat UBI8 | 1528754ms | +2.31%
Alpine | 1583205ms | +5.96%
Official (OpenJDK) | 1682638ms | +12.61%

```
owasp/dependency-track:snapshot
[Config] --------------------------------------------------------------------------------
[Config] OS Name:      Linux
[Config] OS Version:   4.9.184-linuxkit
[Config] OS Arch:      amd64
[Config] CPU Cores:    2
[Config] Java Vendor:  AdoptOpenJDK
[Config] Java Version: 8.0.222
[Config] Java Home:    /opt/java/openjdk
[Config] Java Temp:    /tmp
[Config] User:         dtrack
[Config] User Home:    /data/
[Config] --------------------------------------------------------------------------------
[Config] Initializing Configuration
[Config] System property alpine.application.properties not specified
[Config] Loading application.properties from classpath
[Config] --------------------------------------------------------------------------------
[Config] Application:  Dependency-Track
[Config] Version:      3.6.0-SNAPSHOT
[Config] Built-on:     2019-09-24T15:52:27Z
[Config] --------------------------------------------------------------------------------
[Config] Framework:    Alpine
[Config] Version :     1.6.0-SNAPSHOT
[Config] Built-on:     2019-09-24T15:12:41Z
[Config] --------------------------------------------------------------------------------
[NistMirrorTask] Time spent (d/l):   32617ms
[NistMirrorTask] Time spend (parse): 1682638ms
[NistMirrorTask] Time spent (total): 1942381ms

azul/zulu-openjdk-alpine:8-jre
[Config] --------------------------------------------------------------------------------
[Config] OS Name:      Linux
[Config] OS Version:   4.9.184-linuxkit
[Config] OS Arch:      amd64
[Config] CPU Cores:    2
[Config] Java Vendor:  Azul Systems, Inc.
[Config] Java Version: 8.0.222
[Config] Java Home:    /usr/lib/jvm/zulu-8
[Config] Java Temp:    /tmp
[Config] User:         dtrack
[Config] User Home:    /data/
[Config] --------------------------------------------------------------------------------
[Config] Initializing Configuration
[Config] System property alpine.application.properties not specified
[Config] Loading application.properties from classpath
[Config] --------------------------------------------------------------------------------
[Config] Application:  Dependency-Track
[Config] Version:      3.6.0-SNAPSHOT
[Config] Built-on:     2019-09-24T15:52:27Z
[Config] --------------------------------------------------------------------------------
[Config] Framework:    Alpine
[Config] Version :     1.6.0-SNAPSHOT
[Config] Built-on:     2019-09-24T15:12:41Z
[Config] --------------------------------------------------------------------------------
[NistMirrorTask] Time spent (d/l):   34451ms
[NistMirrorTask] Time spend (parse): 1494177ms
[NistMirrorTask] Time spent (total): 1795346ms

alpine:3.10
[Config] --------------------------------------------------------------------------------
[Config] OS Name:      Linux
[Config] OS Version:   4.9.184-linuxkit
[Config] OS Arch:      amd64
[Config] CPU Cores:    2
[Config] Java Vendor:  IcedTea
[Config] Java Version: 8.0.212
[Config] Java Home:    /usr/lib/jvm/java-1.8-openjdk/jre
[Config] Java Temp:    /tmp
[Config] User:         dtrack
[Config] User Home:    /data/
[Config] --------------------------------------------------------------------------------
[Config] Initializing Configuration
[Config] System property alpine.application.properties not specified
[Config] Loading application.properties from classpath
[Config] --------------------------------------------------------------------------------
[Config] Application:  Dependency-Track
[Config] Version:      3.6.0-SNAPSHOT
[Config] Built-on:     2019-09-24T15:52:27Z
[Config] --------------------------------------------------------------------------------
[Config] Framework:    Alpine
[Config] Version :     1.6.0-SNAPSHOT
[Config] Built-on:     2019-09-24T15:12:41Z
[Config] --------------------------------------------------------------------------------
[NistMirrorTask] Time spent (d/l):   33976ms
[NistMirrorTask] Time spend (parse): 1583205ms
[NistMirrorTask] Time spent (total): 1789089ms

registry.access.redhat.com/ubi8/ubi-minimal
[Config] --------------------------------------------------------------------------------
[Config] OS Name:      Linux
[Config] OS Version:   4.9.184-linuxkit
[Config] OS Arch:      amd64
[Config] CPU Cores:    2
[Config] Java Vendor:  Oracle Corporation
[Config] Java Version: 8.0.222
[Config] Java Home:    /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.222.b10-0.el8_0.x86_64/jre
[Config] Java Temp:    /tmp
[Config] User:         dtrack
[Config] User Home:    /data/
[Config] --------------------------------------------------------------------------------
[Config] Initializing Configuration
[Config] System property alpine.application.properties not specified
[Config] Loading application.properties from classpath
[Config] --------------------------------------------------------------------------------
[Config] Application:  Dependency-Track
[Config] Version:      3.6.0-SNAPSHOT
[Config] Built-on:     2019-09-24T15:52:27Z
[Config] --------------------------------------------------------------------------------
[Config] Framework:    Alpine
[Config] Version :     1.6.0-SNAPSHOT
[Config] Built-on:     2019-09-24T15:12:41Z
[Config] --------------------------------------------------------------------------------
[NistMirrorTask] Time spent (d/l):   34126ms
[NistMirrorTask] Time spend (parse): 1528754ms
[NistMirrorTask] Time spent (total): 1739216ms
```

Issue #303, opened 2019-03-27
OSS Index: Allow exclusion of components to scan


nscuro:
### Issue Type:
  - [ ] defect report
  - [x] enhancement request
  
### Current Behavior:

When using the OSS Index scanner, Dependency-Track will send the packageURLs of all components to OSS Index. This of course includes those of internal components. The namespace of a packageURL may contain the company's name, making it fairly easy to find out who's making the requests and what their application landscape may look like.

Someone with access to OSS Index's request logs may be able to find out how often a given project is built, how many vulnerable components it has and how quickly vulnerable components get patched. The main issue here is that we simply cannot know what Sonatype does with the data being sent to it.

Although I'm explicitly mentioning Sonatype's OSS Index here, I'm sure this also affects other scanners that work similarly.

### Expected Behavior:

It should be possible to exclude information about specific components from being sent to external services like OSS Index.

It should be considered that new projects (so potentially new namespaces) will be added dynamically (e.g. through the Jenkins plugin), which is why I'm certain that a Regex or "namespace contains" type of exclusion list would be optimal. E.g.:

**Exclude namespaces:**
* `^com\.acme.*`
* `.*mycompanyname.*`

### Environment:
  
  - Dependency-Track Version: 3.4.0

stevespringett:
The same exclusions should likely apply to outd

nscuro:
Good point. Yes, they should.

msymons:
This idea of implementing the idea of "teaching" Dependency-Track about internal components is pretty much the same enhancement that I had been thinking for a while "I really need to log this", although @nscuro identified a use case (preventing exposure of internal information) that I had not considered.  Nice one!

1) The enhancement should reduce load and speed up overall analysis time.  I have checked a couple of my projects and see figures such as "196 total dependencies in project...  42 of which are internal".

2) This enhancement would also (hopefully) address a problem I have seen where (very occasionally) internal components are incorrectly identified as having threats.  At the risk of leaking a bit of my own internal information, DT v3.4.0 using OSS Index as the only enabled analyser is telling me that....

`pkg:maven/uk.co.card.post/iQSim-SmartCard-Client@1.0.16?type=jar`

..is vulnerable to:

* [CVE-2011-5036](https://nvd.nist.gov/vuln/detail/CVE-2011-5036)
* [CVE-2012-6109](https://nvd.nist.gov/vuln/detail/CVE-2012-6109)
[...]

Issue #164, opened 2018-06-08
Make repositories configurable


stevespringett:
Repositories should be configurable. Create REST endpoint and UI metaphors to facilitate configuration of repositories by repo type and order.

stevespringett:
Finally, Sonatype Nexus Repo now supports npm dist-tag

https://issues.sonatype.org/browse/NEXUS-9862

Support for this was introduced in Nexus Repo v3.19.0. Anything prior to that and Dependency-Track will not be able to query for updated versions of npm packages.

Will need to find out what the status and limitations for jFrog Artifactory are as well.

stevespringett:
Need to fix skipping of meta analyzers introduced in #512 with internal component pattern. 

> One thing that may need to change is when #164 is implemented. Once users are able to define their own repositories (nexus, artifactory, github package repo, etc) there will be a need to identify if internal components are out of date or not. If we skip the meta analyzers entirely, we won't be able to analyze them for being out-of-date.

> I was thinking of adding a flag to the repos that would indicate if the repo was internal or not. If the repo is internal, then the meta analyzers should execute regardless. So in this scenario, if the regex specifies org.acme for example, then any component with org.acme would not be analyzed with NPMAudit, OSS Index, or VulnDB, nor would they be meta analyzed using external repos, but would be meta analyzed against internal repos.

PR #512, opened 2019-11-14
Identification of internal Components

nscuro:
This PR addresses issue #303 where I explained how submitting PURLs or CPEs to external systems or services like Sonatype's OSSIndex might not be desirable for a company.

### What has been changed?

Administrators can now specify regular expressions for groups and names of components. The screenshot below shows a fictional scenario where components with groups that start with `us` or `com.github` should be considered internal.

2019-11-14 08_35_30-Dependency-Track - Administration

Providing a regex for Component names is necessary for ecosystems that simply don't support namespaces / groups. Python is one of them. Here, companies may choose to use common pre- or suffixes for their packages.

A Component will be considered internal when at least one of the regexes matches.

Components are checked for being internal in the following cases:

- a new BOM is uploaded / parsed
- a new Component is created manually
- an existing Component is updated manually
- the task `InternalComponentIdentificationTask` is executed

The latter can be triggered with a button as shown in the screenshot above. It will scan the entire portfolio and mark all Components as internal that match the given regular expressions:

```
09:14:39.029 INFO [InternalComponentIdentificationTask] Starting internal component identification task
09:14:39.045 INFO [InternalComponentIdentificationTask] Component pkg:maven/com.github.virtuald/curvesapi@1.04?type=jar was identified to be internal
09:14:39.060 INFO [InternalComponentIdentificationTask] Component pkg:maven/us.springett/cvss-calculator@1.2.1?type=jar was identified to be internal
09:14:39.060 INFO [InternalComponentIdentificationTask] Component pkg:maven/com.github.spullara.mustache.java/compiler@0.7.9?type=jar was identified to be internal
09:14:39.060 INFO [InternalComponentIdentificationTask] Component pkg:maven/us.springett/vulndb-data-mirror@1.0.0?type=jar was identified to be internal
09:14:39.387 INFO [InternalComponentIdentificationTask] Component pkg:maven/com.github.cliftonlabs/json-simple@2.3.1?type=jar was identified to be internal
09:14:39.402 INFO [InternalComponentIdentificationTask] Component pkg:maven/us.springett/cpe-parser@2.0.1?type=jar was identified to be internal
09:14:39.402 INFO [InternalComponentIdentificationTask] Component pkg:maven/us.springett/alpine@1.6.3-20191026.064014-1?type=jar was identified to be internal
09:14:39.402 INFO [InternalComponentIdentificationTask] Component pkg:maven/com.github.andrewoma.dexx/collection@0.7?type=jar was identified to be internal
09:14:39.402 INFO [InternalComponentIdentificationTask] Internal component identification task completed
```

New columns have been added to the "Dependencies" tab of the Projects view:

2019-11-14 09_08_32-Dependency-Track - Project Details

as well as the Components view:

2019-11-14 08_35_12-Dependency-Track - Components

Internal components will be skipped for:

* `NPMAuditAnalysisTask`
* `OssIndexAnalysisTask`
* `VulnDbAnalysisTask`
* `GemMetaAnalyzer`
* `HexMetaAnalyzer`
* `MavenMetaAnalyzer`
* `NpmMetaAnalyzer`
* `NugetMetaAnalyzer`
* `PypiMetaAnalyzer`

### Caveats:

- Manually changing individual components to be internal is not possible. This is because there's no sensible way to keep this change persistent over time, while also allowing for retroactive internal component identification throughout the entire portfolio.
- As it's implemented now, MetaAnalyzers are bluntly skipped for internal components. This might not be desirable when #164 has landed. Checking for updates of internal components in internal repositories is a great feature that we shouldn't miss out on due to this PR.
- There's currently no straightforward way to easily test the regexes. `InternalComponentIdentificationTask` runs asynchronously, so we don't get any feedback in the UI when errors occur. There's also a noticable lack of [online regex testers](https://regexr.com/) that support the Java syntax.

### Points for discussion

- **Why Regular Expressions?**
  - The bigger the company, the more chaotic are the naming schemes. There might even be cases where every team has their own scheme. Can we identify internal groups by their prefix? Suffix? We simply don't now in advance.
  - Letting users specify regexes that are evaluated by the application is generally considered bad due to the possibility of DOSing. I'm aware of that risk, but the related setting is only available to admins. Additionally, this attack vector [might get eliminated when upgrading to Java 11](https://stackoverflow.com/a/53049175).


stevespringett:
Regular expressions I think are fine for this feature. In fact, Sonatype Nexus IQ does the same thing - just a single regex field. It can be challenging for some orgs to come up with a regex that handles all cases, but I think it's fine for minimal viable.

One thing that may need to change is when #164 is implemented. Once users are able to define their own repositories (nexus, artifactory, github package repo, etc) there will be a need to identify if internal components are out of date or not. If we skip the meta analyzers entirely, we won't be able to analyze them for being out-of-date.

I was thinking of adding a flag to the repos that would indicate if the repo was internal or not. If the repo is internal, then the meta analyzers should execute regardless. So in this scenario, if the regex specifies org.acme for example, then any component with org.acme would not be analyzed with NPMAudit, OSS Index, or VulnDB, nor would they be meta analyzed using external repos, but would be meta analyzed against internal repos.

PR #550, opened 2019-12-19
Enable Hex in IMetaAnalyzer.build

voltone:


stevespringett:
Ah, sorry about that. I didn't have a bom containing hex components to test with. Thanks for the PR.

PR #562, opened 2020-01-07
Update the CPE field on components from BOM

angrylogic:
When an existing component is detected whilst processing a BOM most fields are updated to reflect the latest value from the BOM.  The `Cpe` field appears to be missing from this list.  This makes it difficult to update the `Cpe` and thus match against NVD vulnerabilities.

This patch adds the `Cpe` field to the list updated from the BOM.


stevespringett:
Thanks for the PR

PR #564, opened 2020-01-08
Fix building of Maven GAV URL in MavenMetaAnalyzer

nscuro:
This PR fixes an issue where Dependeny-Track couldn't check Scala components for their latest version using `MavenMetaAnalyzer`. 

The issue was that the analyzer would replace all dots (`.`) in a components name with slashes (`/`). The names of Scala components are always suffixed with `_<COMPILER_VERSION>`, e.g. `_2.12`.

For `pkg:maven/com.typesafe.akka/akka-actor_2.13@2.5.23?type=jar`, the analyzer would produce the Maven GAV URL `http://central.maven.org/maven2/com/typesafe/akka/akka-actor_2/13/maven-metadata.xml`.

I removed the replacement of dots with slashes from the component name. As far as I understand, only the groupId / namespace is split into "folders" inside maven repos.


nscuro:
@stevespringett It would be great if we could backport this to 3.7.X as well.

stevespringett:
Indeed a defect. Thanks for the PR.

PR #641, opened 2020-03-30
Updated postgresql JDBC driver

svenlange:
Updated postgresql JDBC driver from 42.2.5 to 42.2.11


svenlange:
This should be documented in the release notes as users have to adjust the application.properties. If they dont, dependency track will fail to start as it doesnt find the old jar

PR #652, opened 2020-04-07
Delete the trailing curly bracket

roycyt:


stevespringett:
Thanks for catching this and for submitting a PR.

PR #659, opened 2020-04-16
Revise the email template for BOM consumed and processed

roycyt:
When a BOM is consumed or processed, it is good to tell the email receiver the affected project.


stevespringett:
Thanks for the PR.

PR #696, opened 2020-06-06
bom endpoint: grammar fix in api operation notes

Collinux:
"If a UUID is not specified, than the projectName and ..." --> "If a UUID is not specified then the projectName and ..."


stevespringett:
Thanks for the PR.

Issue #721, opened 2020-07-08
Mail notifications fails if multiple addresses are configured


jensborrmann:
### Current Behavior:
Configuring a mail notification with more than one recipient 
* does not result in a UI error
* results in no mails being sent at all

### Expected Behavior:
One of the following two things should be acchieved (with a preference for the first one, maybe both is the optimum):
* Allow lists of email addresses (suggestion: comma separated)
* Provide user feedback on invalid input

### Environment:

* Dependency-Track Version: 3.8.0

### Additional Details:
https://github.com/DependencyTrack/dependency-track/blob/466929ae66bce18cbb91512c32a3bcdf208b0451/src/main/java/org/dependencytrack/notification/publisher/SendMailPublisher.java#L76 
could be changed.

pmckeown:
This is merged now - can be added to Milestone 4.0 and closed?

stevespringett:
Closing. Fixed in v4.0 - to be released soon.

PR #722, opened 2020-07-08
Allow comma separated list of email addresses as destination

jensborrmann:
Suggested fix for #721


stevespringett:
Thanks for the PR.

PR #792, opened 2020-10-10
Defectdojo integration point

alitheg:
Adds DefectDojo integration point. Documentation updates still required (can do these before merging anything!)


src/test/java/org/dependencytrack/integrations/defectdojo/DefectDojoClientTest.java
```diff
@@ -0,0 +1,118 @@
...
+                                )
+                )
+                .respond(
+                        response()
+                                .withStatusCode(201)
+                                .withHeader(HttpHeaders.CONTENT_TYPE, "application/json")
+                );
+        DefectDojoUploader uploader = new DefectDojoUploader();
+        DefectDojoClient client = new DefectDojoClient(uploader, new URL("https://localhost/defectdojo"));
+        client.uploadDependencyTrackFindings(token, engagementId, new NullInputStream(0));
```
pmckeown, line 90: There is no assertion done in this or the following test.  Since there is no response stubbing to verify a value was returned, perhaps you should look to use the Verification API with MockServer to prove that the expected request was made? https://5-2.mock-server.com/mock_server/verification.html

pmckeown (commented):
Looks good to me - only a couple of minor pieces of feedback.

pmckeown:
Just noticed that the build is failing with 2 failing tests:
```
Error:    DefaultObjectGeneratorTest.testContextInitialized:35->testLoadDefaultConfigProperties:80 expected:<33> but was:<37>
Error:    DefaultObjectGeneratorTest.testLoadDefaultConfigProperties:80 expected:<33> but was:<37>
```
Looks like you just need to change the epxectations in them now that you have added 4 new properties.

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -51,6 +51,10 @@
     FORTIFY_SSC_URL("integrations", "fortify.ssc.url", null, PropertyType.URL, "Base URL to Fortify SSC"),
     FORTIFY_SSC_USERNAME("integrations", "fortify.ssc.username", null, PropertyType.STRING, "The username to authenticate to Fortify SSC with"),
     FORTIFY_SSC_PASSWORD("integrations", "fortify.ssc.password", null, PropertyType.ENCRYPTEDSTRING, "The password to use when authenticating to Fortify SSC"),
+    DEFECTDOJO_ENABLED("integrations", "defectdojo.enabled", "false", PropertyType.BOOLEAN, "Flag to enable/disable Fortify SSC integration"),
```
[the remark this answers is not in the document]
alitheg: Ah, I thought I caught all of these. Nice, thanks.

alitheg:
> 
> 
> Just noticed that the build is failing with 2 failing tests:
> 
> ```
> Error:    DefaultObjectGeneratorTest.testContextInitialized:35->testLoadDefaultConfigProperties:80 expected:<33> but was:<37>
> Error:    DefaultObjectGeneratorTest.testLoadDefaultConfigProperties:80 expected:<33> but was:<37>
> ```
> 
> Looks like you just need to change the expectations in them now that you have added 4 new properties.

I could also have sworn I'd done these. Had some issues between computers though! I'll update these and take a look at the assertion.

alitheg:
I've added an assertion to the positive case to check the call was made, and to the negative case I've tested that the call was made *and* the error handler is called on the uploader class.

alitheg:
I could be wrong @pmckeown, but the latest build failure feels like it's somehow lost the CODACY_PROJECT_TOKEN variable, do you agree, or do you think I've missed something?

alitheg:
Ah I see it: 

> Secrets are not passed to workflows that are triggered by a pull request from a fork.

src/test/java/org/dependencytrack/integrations/defectdojo/DefectDojoClientTest.java
```diff
@@ -88,31 +85,51 @@ public void testUploadFindingsPositiveCase() throws Exception {
...
+                DefectDojoUploader.class,
+                Mockito.CALLS_REAL_METHODS
+        );
         DefectDojoClient client = new DefectDojoClient(uploader, new URL("https://localhost/defectdojo"));
         client.uploadDependencyTrackFindings(token, engagementId, new NullInputStream(16));
+        testClient.verify(
+                request()
+                .withMethod("POST")
+                .withHeader(HttpHeaders.AUTHORIZATION, "Token " + token)
+                .withPath("/defectdojo/api/v2/import-scan/"),
```
pmckeown, line 126: Very minor but judging by the code style in place these 3 lines should be indented:
```
                request()
                        .withMethod("POST")
                        .withHeader(HttpHeaders.AUTHORIZATION, "Token " + token)
                        .withPath("/defectdojo/api/v2/import-scan/"),
```

src/test/java/org/dependencytrack/integrations/defectdojo/DefectDojoClientTest.java
```diff
@@ -88,31 +85,51 @@ public void testUploadFindingsPositiveCase() throws Exception {
...
+                VerificationTimes.exactly(1)
+        );
+        Mockito.verify(uploader, Mockito.times(1)).handleUnexpectedHttpResponse(Mockito.any(), Mockito.eq("https://localhost/defectdojo/api/v2/import-scan/"), Mockito.eq(400), Mockito.any());
```
pmckeown, line 129: I'm a bit unclear as to the usage of Mockito in this Integration test as it feels kinda wrong to use a class mocking technique and a remote system mocking technique in the same class.  

I can see that you are trying to verify that a Bad Request is received but it feels like that is testing the MockServer framework more than the code.  

Perhaps the `testClient.verify()` invocation to prove that the API is called as expected and the fact that an Exception is not raised by the client is enough?

```
    @Test
    public void testUploadFindingsNegativeCase() throws Exception {
        String token = "db975c97-98b1-4988-8d6a-9c3e044dfff2";
        String engagementId = "";
        MockServerClient testClient = new MockServerClient("localhost", 1080);
        testClient.when(
                        request()
                                .withMethod("POST")
                                .withHeader(HttpHeaders.AUTHORIZATION, "Token " + token)
                                .withPath("/defectdojo/api/v2/import-scan/")
                )
                .respond(
                        response()
                                .withStatusCode(400)
                                .withHeader(HttpHeaders.CONTENT_TYPE, "application/json")
                );
        DefectDojoUploader uploader = new DefectDojoUploader();
        DefectDojoClient client = new DefectDojoClient(uploader, new URL("https://localhost/defectdojo"));
        try {
            client.uploadDependencyTrackFindings(token, engagementId, new NullInputStream(16));
        } catch (Exception ex) {
            Assert.fail("Exception not expected", ex);
        }
        testClient.verify(
                request()
                        .withMethod("POST")
                        .withHeader(HttpHeaders.AUTHORIZATION, "Token " + token)
                        .withPath("/defectdojo/api/v2/import-scan/"),
                VerificationTimes.exactly(1)
        );
        testClient.clear(
                request()
                        .withPath("/defectdojo/api/v2/import-scan/")
        );
    }
```

src/test/java/org/dependencytrack/integrations/defectdojo/DefectDojoClientTest.java
```diff
@@ -88,31 +85,51 @@ public void testUploadFindingsPositiveCase() throws Exception {
...
+        testClient.verify(
+                request()
+                .withMethod("POST")
+                .withPath("/defectdojo/api/v2/import-scan/"),
+                VerificationTimes.exactly(1)
+        );
+        testClient.clear(
+                request()
+                    .withPath("/defectdojo/api/v2/import-scan/")
+        );
```
pmckeown, line 97: Perhaps this clean up code should be in an `@After` method?

src/test/java/org/dependencytrack/integrations/defectdojo/DefectDojoClientTest.java
```diff
@@ -88,31 +85,51 @@ public void testUploadFindingsPositiveCase() throws Exception {
         DefectDojoUploader uploader = new DefectDojoUploader();
         DefectDojoClient client = new DefectDojoClient(uploader, new URL("https://localhost/defectdojo"));
         client.uploadDependencyTrackFindings(token, engagementId, new NullInputStream(0));
+        testClient.verify(
+                request()
+                .withMethod("POST")
+                .withPath("/defectdojo/api/v2/import-scan/"),
```
pmckeown, line 91: Very minor but judging by the code style in place these 2 lines should be indented:
```
                request()
                        .withMethod("POST")
                        .withPath("/defectdojo/api/v2/import-scan/"),
```

pmckeown:
> Ah I see it:
> 
> > Secrets are not passed to workflows that are triggered by a pull request from a fork.

Hmm - looks like your Pull Request check can't pass though - if the Codacy Token is not passed to PR builds, can any pass?  

We'll have to rely on @stevespringett to help on this one I think!

src/test/java/org/dependencytrack/integrations/defectdojo/DefectDojoClientTest.java
alitheg: Ok yeah - if you think that's enough of an assertion I'll change it back to the same assertion as the other test. I felt that without testing the error, it was basically the same as the positive test case.

src/test/java/org/dependencytrack/integrations/defectdojo/DefectDojoClientTest.java
alitheg: I thought this - will change!

alitheg:
Yeah, I was thinking this! I'll change up the indentation issues too, good spots.

src/test/java/org/dependencytrack/integrations/defectdojo/DefectDojoClientTest.java
pmckeown: Yeah I think so, the client code doesn't handle exceptions but Unirest should be able to handle the 400 bad request without raising an exception.  It is essentially proving that the library is being used correctly and failing as expected.

Issue #809, opened 2020-10-19
407 Proxy Authentication Required


jagusztinl:
The defect may already be reported! Please search for the defect before creating one.

### Current Behavior:
Behind corporate proxy, healthcheck unable to run: 407 Proxy Authentication Required

### Steps to Reproduce:
start dependency track behind  authenticated proxy

### Expected Behavior:
HEALTHCHECK CMD wget --no-proxy -q -O - http://127.0.0.1:8080/api/version

### Environment:
  
  - Dependency-Track Version: all
  - Distribution: [ Docker | Executable WAR | Traditional WAR ] Docker
  - BOM Format & Version: N/A
  - Database Server: [ H2 | MSSQL | MySQL | PostgreSQL ] N/A
  - Browser:  N/A

### Additional Details:
Please change the Dockerfile, makes no sense to use proxy for health check
(e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

stevespringett:
Closing. Implemented in v4.0 - to be released soon.

PR #826, opened 2020-11-06
don't use a proxy for HEALTHCHECK

ssproessig:
in order to make it work in corporate Dockers that must set the `http_proxy` environment variable for their containers


stevespringett:
Thanks for the PR.

This resolves #809

Issue #855, opened 2020-11-30
Caused by: org.postgresql.util.PSQLException: ERROR: syntax error at or near "FROM"


rvsoni:
The defect may already be reported! Please search for the defect before creating one.

### Current Behavior:
```
2020-11-30 07:03:03,925 [] WARN [org.eclipse.jetty.server.HttpChannelState] unhandled due to prior sendError
javax.servlet.ServletException: javax.jdo.JDODataStoreException: The database returned ROLLBACK, so the transaction cannot be committed. Transaction failure cause is <<ERROR: syntax error at or near "FROM"
  Position: 603>>
NestedThrowables:
org.postgresql.util.PSQLException: The database returned ROLLBACK, so the transaction cannot be committed. Transaction failure cause is <<ERROR: syntax error at or near "FROM"
  Position: 603>>
	at org.glassfish.jersey.servlet.WebComponent.serviceImpl(WebComponent.java:410)
	at org.glassfish.jersey.servlet.WebComponent.service(WebComponent.java:346)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:366)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:319)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:205)
	at org.eclipse.jetty.servlet.ServletHolder$NotAsync.service(ServletHolder.java:1452)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:791)
	at org.eclipse.jetty.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1626)
	at alpine.filters.ContentSecurityPolicyFilter.doFilter(ContentSecurityPolicyFilter.java:224)
	at org.eclipse.jetty.servlet.FilterHolder.doFilter(FilterHolder.java:193)
	at org.eclipse.jetty.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1601)
	at alpine.filters.ClickjackingFilter.doFilter(ClickjackingFilter.java:92)
	at org.eclipse.jetty.servlet.FilterHolder.doFilter(FilterHolder.java:193)
	at org.eclipse.jetty.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1601)
	at alpine.filters.WhitelistUrlFilter.doFilter(WhitelistUrlFilter.java:149)
[...]

PR #856, opened 2020-11-30
#855 Fixed by removing extra comma "," in SQL query

rvsoni:


stevespringett:
Thank you for the PR and for all the testing you're performing. It's very much appreciated.

Issue #808, opened 2020-10-19
Update bundled database drivers on Docker images


sephiroth-j:
> MySQL Connector/J 8.0 is highly recommended for use with MySQL Server 8.0, 5.7, and 5.6. Please upgrade to MySQL Connector/J 8.0. 

Current Version is [8.0.22](https://mvnrepository.com/artifact/mysql/mysql-connector-java/8.0.22)

Postgres -> [42.2.18](https://mvnrepository.com/artifact/org.postgresql/postgresql/42.2.18)

MS SQL [8.4.1](https://mvnrepository.com/artifact/com.microsoft.sqlserver/mssql-jdbc/8.4.1.jre8) or [7.4.1](https://mvnrepository.com/artifact/com.microsoft.sqlserver/mssql-jdbc/7.4.1.jre8) if you prefer to stay an the old major version.

sephiroth-j:
It should also be noted that MySQL Connector/J is licensed under GPLv2 and Postgres under BDS-2. Both require a copy of the license.

A similar discussion about Docker Images and bundled GPL software can be found at https://issues.apache.org/jira/browse/LEGAL-437. But it seems to depend on the way the Docker Image is distributed, as outlined [here](https://www.linuxfoundation.org/blog/2020/04/docker-containers-what-are-the-open-source-licensing-considerations/).

> If you are distributing the container as a whole, then you are responsible for license compliance for all of the software it contains. By contrast, if you are distributing just a Dockerfile which tells people how to build a container, and the recipients are then using your Dockerfile to obtain container layers from third-party locations, then you are perhaps not responsible for license compliance for that software.

sephiroth-j:
- [x] mysql -> #846 
- [x] postgres -> #853 
- [ ] ms sql
- [ ] use version independent filename for drivers

rvsoni:
This should get into 4.0 release, 

I am testing 4.0 Snapshot and i updated docker compose defination 2 times for postgresql jdbc driver update in relase, 

Ravi

sephiroth-j:
obsolete since #904 and version 4.1

PR #860, opened 2020-12-01
Update SQL driver and version-less filenames

sephiroth-j:
- updates the remaining MS SQL driver to 8.4.1
- **removes the version number from the files.** removes the need to adjust docker-compose files when using the "latest" tag of the docker image and the driver is updated in a future release.

references #808


sephiroth-j:
obsolete since #904

PR #878, opened 2020-12-18
Remove @JsonDeserialize annotation from Project's purl field

nscuro:
This caused Jackson to trim the serialized value instead of using the PackageURL constructor, resulting in a type mismatch error when trying to call the setter.


stevespringett:
Thank for discovering and the PR.

PR #891, opened 2021-01-11
Update Dockerfile to be OpenShift restricted SCC compliant

Homopatrol:
Set ARG for GID=0.
Also specifying the group ownership for this through chmod -R g=u and chown -R ${USERNAME}:0


k3rnelpan1c-dev:
Hi Pandora, could you check if Dependency Track 4.4.0, which includes #1303, is sufficient for you?

src/main/docker/Dockerfile
```diff
@@ -7,18 +7,19 @@ LABEL vendor="OWASP"
...
+ARG GID=0
 ARG WAR_FILENAME=dependency-track-apiserver.war
 
 # Create the directory where Dependency-Track will store its data (${DATA_DIR}) and the external library directory (${EXTLIB_DIR})
 # Create a user and assign home directory to a ${DATA_DIR}
 # Ensure UID 1000 & GID 1000 own all the needed directories
 RUN mkdir -p -m 770 ${DATA_DIR} \
     && adduser -D -h ${DATA_DIR} -u 1000 ${USERNAME} \
-    && chown -R ${USERNAME}:${USERNAME} ${DATA_DIR}
+    && chown -R ${USERNAME}:0 ${DATA_DIR}
```
savek-cc, line 18: Why not :${GID} - otherwise that new parameter isn't even used anywhere?

src/main/docker/Dockerfile
```diff
@@ -7,18 +7,19 @@ LABEL vendor="OWASP"
...
 # Ensure UID 1000 & GID 1000 own all the needed directories
 RUN mkdir -p -m 770 ${DATA_DIR} \
     && adduser -D -h ${DATA_DIR} -u 1000 ${USERNAME} \
-    && chown -R ${USERNAME}:${USERNAME} ${DATA_DIR}
+    && chown -R ${USERNAME}:0 ${DATA_DIR} 
 
 # Copy the compiled WAR to the application directory created above
 # Automatically creates the $APP_DIR directory
-COPY --chown=1000 ./target/${WAR_FILENAME} ${APP_DIR}
+COPY --chown=${USERNAME}:0  ./target/${WAR_FILENAME} ${APP_DIR}
```
savek-cc, line 22: Same as above ${USERNAME}:${GID}?

src/main/docker/Dockerfile
k3rnelpan1c-dev: The reason for this is to support arbitrary user IDs, such you can find in environments such as OpenShift.
You can read up [on it in the OpenShift Documentation](https://docs.openshift.com/container-platform/4.9/openshift_images/create-images.html#use-uid_create-images).

Regardless, since #1303 was merged that has this kind of behaviour already incorporated this PR is redundant, I think.

nscuro:
Superseded by https://github.com/DependencyTrack/dependency-track/pull/1303.

PR #918, opened 2021-02-03
ALPINE_CORS_ALLOW_METHODS list separated by commas

DrakezulsMinimalism:
The list without commas did not work for me (which is what I expected after uncommenting it).
The list with commas works in both Firefox and Chrome.


stevespringett:
Indeed. Thanks for making the correction.

PR #930, opened 2021-02-10
Add support for version policy conditions

nscuro:
This PR adds a new *Version* policy condition that allows for matching of specific component versions or version ranges:

Version Condition

Additionally, as suggested by @stevespringett in Slack, the *Coordinates* condition was extended to support version operators as well:

Coordinates Condition with Version Operator

Policy Violation with Version Operator

All numeric comparison operators are supported (`<`, `<=`, `>`, `>=`, `==`, `!=`). If a coordinate's version does not start with any of those operators, DT will fall back to the default `contains` logic.


stevespringett:
Fantastic. This is going to be very useful.

PR #965, opened 2021-03-16
Add defectdojo integration documentation

alitheg:
Adds documentation for DefectDojo integration point and puts it beside Kenna and Fortify in the side menu


alitheg:
I wrote this from the same point that I added the original integration point. Let me know if you'd prefer (to avoid conflicts) me to rebase them on top of the current work

PR #994, opened 2021-03-30
Update Dockerfile for OpenShift

namloc2001:
Hi, I am running Dependency Track 4.2.1 on OpenShift. OpenShift applies its Restricted SCC to deployments by default, whereby it assigns an arbitrary UID to the container and also assigns it GID=0. Therefore if we update Dockerfiles to chown/chgrp GID=0 and also chmod g=u (so that the group 0 has same permissions as user) we can ensure the Dockerfiles are compatible with OpenShift best practice.

Reference: "Support arbitrary user ids" https://docs.openshift.com/container-platform/4.1/openshift_images/create-images.html


namloc2001:
Covered by existing PR.

Issue #1074, opened 2021-06-12
Not able to login in dependency track using default login credential.


rsumit:
Hello 

I am new to this tool and I tried to install it on one of my VM having enough ram and cpu for dtrack to work.

Containers are working fine and I don't see any error in the logs both API and forntend  ( image attached)

In the UI  after adding the default cred  - admin:admin, nothing happens. I am not sure what I am missing here but I am attaching some images. Please let me know how can I fix this issue.
<img width="781" alt="dt - error" src="https://user-images.githubusercontent.com/32129881/121765034-cb6d3780-cb65-11eb-940b-47fedd9d2140.png">
<img width="700" alt="container" src="https://user-images.githubusercontent.com/32129881/121765080-0f603c80-cb66-11eb-92d2-55b2c6d64c92.png">

nscuro:
You have to use the external IP or hostname of your VM for the frontend container's `API_BASE_URL` environment variable, see https://github.com/DependencyTrack/dependency-track/pull/1075/files#diff-c3d30d4845cc6f240782a49f91ce0c35378d163006beee6530bf3448719db99bR92-R96

rsumit:
@nscuro   Thank you. After updating the URL it's working fine. Thanks for creating the PR with this info. I am sure it will help many like me when they go with the first server install.

West-Circle:
hi, can teach how to update the URL? any steps?

AfshinOnline:
I waited until all the dependencies from the database were downloaded in the docker logs which took a while and then tried another browser which seemed to work. I had to delete the old volumes and containers to do fresh docker pull.

donaldedwin:
If you used docker to set it up then go to the docker file and put the IP or domain name of your server as the API url. Initially it says localhost:8081/api etc. Because of that CORS will block any attempts to the make the request because it will be using an assumption that its a request from entirely another domain.

AFulgens:
The issue is very old, but still open. It also seems to be an effecitve duplicate of #656, #2205 and there is an unresolved discussion (https://github.com/DependencyTrack/dependency-track/discussions/3534) about the same issue.

A well spent™ 45 minutes _today_ (more than 4 years after this first being reported in #656!) of mine resulted in the following:

- The provided docker compose file could be built and started
- The frontend could be accessed in any browser
- The login with admin/admin resulted in an HTTP 404 from the API server and thus was never really successful

Investigation yielded the following:

- The API server started
[...]

PR #1075, opened 2021-06-12
Clarify that API_BASE_URL needs to be modified for server deployments

nscuro:
We regularly see users being confused that DT doesn't work when deploying the default `docker-compose.yml` to a server. 

Updated `docker-compose.yml` and related documentation pages to indicate that external addresses need to be used.


nscuro:
Relates to #1074

PR #1110, opened 2021-07-19
Implement repository support for Go Modules

nscuro:
Add proxy.golang.org as default repository.

Go's module proxies / mirrors are not a repository per se, but they behave like one.

outdated-components


stevespringett:
🙌🏻  Thanks for the PR

PR #1146, opened 2021-08-18
allow custom url of 'nvd.nist.gov' mirror

sulegis:
will be useful in some network blocked country

PR #1153, opened 2021-08-19
allow custom url of 'nvd.nist.gov' mirror

sulegis:
will be useful in some network blocked country
fix bad [pull1146](https://github.com/DependencyTrack/dependency-track/pull/1146)

### DIFF
--- a/docs/_posts/2021-08-07-v4.3.2.md
+++ b/docs/_posts/2021-08-07-v4.3.2.md
@@ -1,0 +1,31 @@
+---
+title: v4.3.2
+type: patch
+---
+
+**Features:**
+
+**Fixes:**
+* Resoled an issue with portfolio access control where a user belonging to multiple teams will not have access to the aggregate of all projects or components they're permitted to - [#1132](https://github.com/DependencyTrack/dependency-track/issues/1132)
+
+**Security:**
+
+**Upgrade Notes:**
+
+
+###### dependency-track-apiserver.war
+
+| Algorithm | Checksum |
+| SHA-1     | 9746e03d0bd7dc02ca1d94aa29a6445144fb7589 |
+| SHA-256   | 283282536ec276bf048428fc02aee119ff9e42f995c67cf169e2bd2a7a92cd31 |
+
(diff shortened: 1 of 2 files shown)


rgmz:
@sulegis Would it make sense to allow a custom base URL (`https://nvd.nist.gov/feeds` or `NVD_FEEDS_URL`) instead of individual URLs?

sulegis:
> @sulegis Would it make sense to allow a custom base URL (`https://nvd.nist.gov/feeds` or `NVD_FEEDS_URL`) instead of individual URLs?

commonly, not make sense. BUT，in some network Forbidenned Country(like china), 'nvd.nist.gov' is very slowly and unstable, allow custom will be very Useful.
It's not a issue，but a more flexible way, so we can use [nist-data-mirror](https://github.com/stevespringett/nist-data-mirror) or other tools to across the forbidden.

rgmz:
Oh, I agree that it is useful. I have similar network issues with a company
firewall.

I meant having a single environment variable to use for all URLs, rather
than using a separate environment variable for each feed url.

For example:
```java
private static final String NVD_FEED_URL = System.getenv("NVD_FEED_URL") ==
null ? "https://nvd.nist.gov/feeds/" : System.getenv("NVD_FEED_URL")
private static final String CVE_JSON_11_BASE_META = NVD_FEED_URL +
"json/cve/1.1/nvdcve-1.1-%d.meta"
```

sulegis:
not think a lot. just copy the style of [dependency-check-cli](https://jeremylong.github.io/DependencyCheck/dependency-check-cli/arguments.html), like   --cveUrlModified,  --cveUrlModified,--retireJsUrl. maybe more flexble?
Maybe One BaseURL custom is a Good Idea，it will be more simple.

stevespringett:
Thanks for the PR. Due to external issues such as supporting GitHub advisories and having to configure that, I've made the mirroring support part of that work as well.

In DT v4.4, configuring the NVD will look like this:

Screen Shot 2022-01-26 at 5 09 30 PM

PR #1170, opened 2021-08-24
Added a Publisher for Cisco Webex

holdebeSICKAG:
Hey,
We did some additions to be able to send notifiactions to Cisco Webex. 

We build the whole project after our changes with Netbeans, some tests for us:

 - FortifySSC Test (probably because we don't use that)
 - DefectDojo Test  (probably because we don't use that)
 - ConsolePublisher Test, didn't pass on the current master branch either. (we built with netbeans on windows)

Otherwise all tests suceeded and we are using the new publisher internally.  Would be great, if that merges into the core DependencyTrack.

Btw.  we are not happy with the style of the cswebex.peb file, but if we include \r\n into the file, the Webex Api responds with a HTTP status code of 400 - Bad Request,

If you have any questions or need further information,  feel free to ask.

Greetings from Germany!


stevespringett:
Thanks for the PR

docs/_posts/2021-08-20-v4.3.3.md
```diff
@@ -0,0 +1,33 @@
+---
```
stevespringett, line 1: This file should not be part of the PR. Please omit this file.

docs/index.md
```diff
@@ -59,7 +59,7 @@ CI/CD environments.
   * Trust boundary traversal
```
stevespringett, line 59: This file should not be part of the PR. Please omit this file.

pom.xml
```diff
@@ -225,7 +225,7 @@
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-artifact</artifactId>
-            <version>3.8.1</version>
```
stevespringett, line 228: This file should not be part of the PR. Please omit this file.

stevespringett (changes requested):
There are many modified files that are unrelated to the Cisco Webex publisher. Can you include on the the modifications necessary to implement the feature?

stevespringett:
I'm about ready to merge this in but do not have a way to test the publisher. Can you confirm this code is working?

Also, is there an easy way to test this code without having a paid WebEx account?

heintmaSICKAG:
We tested the code of this merge request and can confirm it is working.

To check the functionality it is possible to create an incoming webhook with a free webex account:
-	Login at [Webex Apphub](https://apphub.webex.com)
-	Search for the app “Incoming webhooks”
-	In the now shown dialog enter a name for your webhook and select a space
-	Click “add” and the Incoming Webhook will be added to your account. The webhook URL is shown which should be added in the “destination” of the Dependency Track notification configuration.

Hope this little description helps.

stevespringett:
Fantastic. Thanks for the step-by-step. Confirmed its working as intended. Thanks again for the PR. Much appreciated.

PR #1197, opened 2021-09-17
Parse externalReferences property from OSS Index API response

kapseliboi:
OSS Index API added recently a new property to their API responses called externalReferences which contains the references that are visible on their website. This PR adds these references to Dependency Track as well. I have never written any Java code so feedback is appreciated. Also I have not tested this as I have no clue how to do that and I do have to admit that I was too lazy to learn how to do that.

Relevant links:
https://github.com/OSSIndex/vulns/issues/193
https://ossindex.sonatype.org/rest

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java
+++ b/src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java
@@ -305,8 +305,17 @@ public class OssIndexAnalysisTask extends BaseComponentAnalyzerTask implements C
             }
         }
 
-        if (reportedVuln.getReference() != null) {
-            vulnerability.setReferences("* [" + reportedVuln.getReference() + "](" + reportedVuln.getReference() + ")");
+        final StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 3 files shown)


src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java
```diff
@@ -305,8 +305,17 @@ private Vulnerability generateVulnerability(final QueryManager qm, final Compone
...
+        final String reference = reportedVuln.getReference();
+        if (reference != null) {
+            sb.append("* [").append(reference).append("](").append(reference).append(")\n");
+        }
+        for (String externalReference: reportedVuln.getExternalReferences()) {
+            sb.append("* [").append(externalReference).append("](").append(externalReference).append(")\n");
+        }
+        final String references = sb.toString();
+        if (references.length() > 0) {
+            vulnerability.setReferences(references.substring(0, references.lastIndexOf("\n")));
```
kapseliboi, line 318: The string builder logic here is mostly just copied from NvdParser.java

stevespringett:
Looks good. Thanks for the PR

PR #1254, opened 2021-11-03
fixed packagist.org meta data parser & version comparisson

### LINKED ISSUE #1243: PHP composer version detection issues with leading `v`
### Current Behavior:

if a PHP's composer component has a version with a leading `v` it is not handled properly:
- detection of newer version fails if they have a leading `v` in them
- detection of any version fails, if version does not match completely, including the leading `v`

### Steps to Reproduce:

tested with component https://packagist.org/packages/typo3/class-alias-loader
which has versions with a leading `v` and some without a `v`.

#### to reproduce "detection of newer version fails if they have a leading `v` in them"

1. create a project in a fresh DT setup
2. upload this sbom to the new project:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<bom xmlns="http://cyclonedx.org/schema/bom/1.2" version="1">
  <components>
    <component type="library">
      <group><![CDATA[typo3]]></group>
      <name><![CDATA[class-alias-loader]]></name>
      <version><![CDATA[v1.1.3]]></version>
      <purl><![CDATA[pkg:composer/typo3/class-alias-loader@v1.1.3]]></purl>
    </component>
  </components>
</bom>
```

#### to reproduce "detection of any version fails, if version does not match completely, including the leading `v`" 

1. create a project in a fresh DT setup 
   *this one is important:* do NOT reuse the setup from the previous component SBOM
2. upload this sbom to the new project:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<bom xmlns="http://cyclonedx.org/schema/bom/1.2" version="1">
  <components>
    <component type="library">
      <group><![CDATA[typo3]]></group>
      <name><![CDATA[class-alias-loader]]></name>
      <version><![CDATA[1.1.3]]></version>
      <purl><![CDATA[pkg:composer/typo3/class-alias-loader@1.1.3]]></purl>
    </component>
  </components>
</bom>
```

### Expected Behavior:

for both reproducible examples: 
for the component https://packagist.org/packages/typo3/class-alias-loader
the latest version is `v1.1.3`, so the version hint in DT should be green, mentioning that the latest version is used.

### Environment:
  
  - Dependency-Track Version: v4.3.6
  - Distribution: Docker as of https://dependencytrack.org/docker-compose.yml
  - BOM Format & Version: CycloneDX 1.2 XML
  - Database Server: the one bundled in the docker image `dependencytrack/apiserver`
  - Browser:  FireFox

### Additional Details:

regarding PHP's composer versioning and the leading `v`: 
composer heals itself. composer might add/remove the leading `v` whenever needed. 
this means for composer components the `v1.3.0` could be synonym to `1.3.0` and vice versa.

read more: https://getcomposer.org/doc/articles/versions.md

jkowalleck:
closes #1243

Signed-off-by: Jan Kowalleck <jan.kowalleck@gmail.com>


stevespringett:
Thanks for the PR Jan.

Issue #1213, opened 2021-10-05
Support ARM64


ruckc:
The enhancement may already be reported! Please search for the enhancement before creating one.

### Current Behavior:
Dependency Track is unable to deploy on arm64 environments

### Proposed Behavior:
Change github workflow to use buildkit to build containers for commonly used architectures (linux/amd64, linux/arm64, darwin/amd64, darwin/arm64).

This enhancement would be beneficial to users/potential developers that use arm64 processors.

ruckc:
with the buildkit github actions its super simple to implement multi-arch containers, so the Temurin migration can be uncoupled

stevespringett:
This was an interesting exercise. I had a personal need to do this as I recently migrated my personal development machine to an Apple M1 Pro. What I found was that no Alpine images that included Java were available on ARM. So I had to make a switch from Alpine to Debian Slim, which I intended to do anyway (#1090), and at the same time, I had to find another OpenJDK distribution. Temurin is not available on Debian Slim and Debian packages are not yet available. So I migrated to Azul OpenJDK 17, which I'm using on my development machine anyway. So lots of changes were involved in getting an image that was supported on x86_64 and arm64.

```
03:27:51.092 INFO [Config] --------------------------------------------------------------------------------
03:27:51.093 INFO [Config] OS Name:      Linux
03:27:51.094 INFO [Config] OS Version:   5.10.47-linuxkit
03:27:51.094 INFO [Config] OS Arch:      aarch64
03:27:51.097 INFO [Config] CPU Cores:    5
03:27:51.100 INFO [Config] Max Memory:   6.2 GB (6,672,089,088.0 bytes)
03:27:51.100 INFO [Config] Java Vendor:  Azul Systems, Inc.
03:27:51.101 INFO [Config] Java Version: 17+35-LTS
03:27:51.102 INFO [Config] Java Home:    /usr/lib/jvm/zulu17-ca-arm64
03:27:51.102 INFO [Config] Java Temp:    /tmp
03:27:51.102 INFO [Config] User:         dtrack
03:27:51.102 INFO [Config] User Home:    /data/
03:27:51.102 INFO [Config] --------------------------------------------------------------------------------
03:27:51.102 INFO [Config] Initializing Configuration
03:27:51.103 INFO [Config] System property alpine.application.properties not specified
03:27:51.103 INFO [Config] Loading application.properties from classpath
03:27:51.106 INFO [Config] --------------------------------------------------------------------------------
[...]

Issue #1090, opened 2021-06-24
Migrate Docker image from Alpine to Debian Slim


stevespringett:
Alpine images can cause some incompatibility issues due to its use on musl instead of glibc.

This ticket is to track the migration of the Dockerfile from using an Alpine base image to one based on Debian slim.

Related to #1054

Frontend: https://github.com/DependencyTrack/frontend/issues/78

sephiroth-j:
@stevespringett, you might want to take a look at the "distroless" images such as "gcr.io/distroless/java-debian10" from https://github.com/GoogleContainerTools/distroless

PR #1303, opened 2021-12-19
packaging: Overhaul Containerfile and use Temurin JRE

k3rnelpan1c-dev:
### Description

I props these changes as I stumbled over this great tool while dealing with the aftermath that was the Log4Shell hell week and noticed that the Dockerfile could be improved. I did so since we would deploy dtrack to our OpenShift cluster which uses arbitrary UIDs, which is currently already supported by the fronted container yet not by the API server. And while I have made similar changes to packaging for an internal draft container to test deploy dtrack I wanted to offer these changes upstream as well.

Additionally, I noticed that there is the motivation to use the Eclipse-Temurin JVM based on the comment in https://github.com/DependencyTrack/dependency-track/issues/1213#issuecomment-978807113, so I took the liberty and migrated the Container to that alongside the other changes.

To answer questions as to why I did so via a [multi stage build](https://docs.docker.com/develop/develop-images/multistage-build/) I advice to read the following to issues that may be a long read but should explain it:

* https://github.com/adoptium/containers/issues/1 (also covers the glibc issue with Alpine and announces that there will be a Musl build of the Temurin JVM eventually :tada: )
* https://github.com/adoptium/containers/issues/54

***TLDR***: DockerHub apparently does not accept `slim` images anymore for their `Official Image` labelled images and the 'old' [`adoptopenjdk`](https://hub.docker.com/_/adoptopenjdk) too did only offer Ubuntu, CentOS7 and WindowsServerCore images. So the common consensus is to utilize multi stage builds to copy the JDK/JRE from a official image to a derived image with the same package manager (i.e. Ubuntu to Debian, CentOS to UBI or Fedora, etc.), a good example could be the [Jenkins Official Docker images](https://github.com/jenkinsci/docker/blob/master/11/debian/bullseye-slim/hotspot/Dockerfile), which use Jlink to minify the JDK and then copy it to their image. (I did the same without Jlink since the JRE does not have that nor need it to be slim)

If there are any further questions or any feedback please feel free to share them I will come back to them as soon as possible.

### Changes

* overhauls and (IMHO) simplifies/groups Dockerfile
* migrates Dockerfile to Eclipse-Temurin from Zulu JRE
* update `maven` workflow slightly 

### Issues

* relates to https://github.com/DependencyTrack/dependency-track/issues/1213#issuecomment-978807113
* relates to https://github.com/DependencyTrack/dependency-track/issues/1090
* similar to https://github.com/DependencyTrack/dependency-track/pull/994
* similar to https://github.com/DependencyTrack/dependency-track/pull/891

### Notes

I noticed that you seem to use a bash script to release new version of dtrack and want to propose moving that to CI (GitHub Actions) as well. I would be up for submitting the necessary PRs to do so if you are interested :slightly_smiling_face: 

**EDIT:** there are now official musl builds of Temurin 11, 16 and 17 available, so if you ever want to offer an alpine spin alongside the debian one you can now without any glibc hackery for the Alpine JVM variant.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -12,66 +12,79 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     env:
       DOCKER_APISERVER_REPO: dependencytrack/apiserver
       DOCKER_BUNDLED_REPO: dependencytrack/bundled
       DOCKER_TAG: snapshot
+
     steps:
     - name: Checkout code
-      uses: actions/checkout@v2.3.4
+      uses: actions/checkout@v2
+
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
+
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v1
       id: buildx
       with:
         install: true
+
     - name: Available platforms
       run: echo ${{ steps.buildx.outputs.platforms }}
+
     - name: Set up JDK
       uses: actions/setup-java@v2
       with:
-        distribution: 'adopt'
+        distribution: 'temurin'
         java-version: '11'
-    - name: Install tools
-      run: |
-        sudo apt-get install jq
-        wget -O ~/codacy-coverage-reporter-assembly.jar https://github.com/codacy/codacy-coverage-reporter/releases/download/4.0.5/codacy-coverage-reporter-4.0.5-assembly.jar
-        export DOCKLE_VERSION=$(curl --silent "https://api.github.com/repos/goodwithtech/dockle/releases/latest" | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')
-        wget -O ~/dockle.tar.gz https://github.com/goodwithtech/dockle/releases/download/v${DOCKLE_VERSION}/dockle_${DOCKLE_VERSION}_Linux-64bit.tar.gz
-        tar zxvf ~/dockle.tar.gz -C ~/
-        export TRIVY_VERSION=$(curl --silent "https://api.github.com/repos/aquasecurity/trivy/releases/latest" | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')
-        wget -O ~/trivy.tar.gz https://github.com/aquasecurity/trivy/releases/download/v${TRIVY_VERSION}/trivy_${TRIVY_VERSION}_Linux-64bit.tar.gz
-        tar zxvf ~/trivy.tar.gz -C ~/
+        cache: 'maven'
+
     - name: Build with Maven
-      run: |
+      run: |-
         mvn clean
         mvn package -Dmaven.test.skip=true -P clean-exclude-wars -P embedded-jetty -Dlogback.configuration.file=src/main/docker/logback.xml
         mvn clean -P clean-exclude-wars
         mvn package -Dmaven.test.skip=true -P embedded-jetty -P bundle-ui -Dlogback.configuration.file=src/main/docker/logback.xml
         mvn clean -P clean-exclude-wars
+
     - name: Execute unit tests
       if: github.event.inputs.skipTests != 'true'
       run: mvn test
+
     - name: Publish test coverage
       if: github.ref == 'refs/heads/master' && github.event.inputs.skipTests != 'true'
       env:
         CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
       run: java -jar ~/codacy-coverage-reporter-assembly.jar report -l Java -r target/jacoco-ut/jacoco.xml
+
     - name: Logging into Docker Hub
       if: github.ref == 'refs/heads/master'
       env:
         HUB_USERNAME: ${{ secrets.HUB_USERNAME }}
         HUB_ACCESSS_TOKEN: ${{ secrets.HUB_ACCESSS_TOKEN }}
-      run: docker login -u "$HUB_USERNAME" -p "$HUB_ACCESSS_TOKEN"
+      run: docker login -u "${HUB_USERNAME}" -p "${HUB_ACCESSS_TOKEN}"
+
     - name: Building Docker API Server container
       if: github.ref == 'refs/heads/master'
-      run: |
-        docker build -f src/main/docker/Dockerfile --build-arg WAR_FILENAME=dependency-track-apiserver.jar -t $DOCKER_APISERVER_REPO:$DOCKER_TAG --platform linux/amd64,linux/arm64 --push .
+      run: |-
+        docker build -f src/main/docker/Dockerfile \
+          --build-arg COMMIT_SHA=${GITHUB_SHA} \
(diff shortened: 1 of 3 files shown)


.github/workflows/maven.yml
```diff
@@ -12,66 +12,79 @@ on:
...
-    - name: Install tools
-      run: |
-        sudo apt-get install jq
-        wget -O ~/codacy-coverage-reporter-assembly.jar https://github.com/codacy/codacy-coverage-reporter/releases/download/4.0.5/codacy-coverage-reporter-4.0.5-assembly.jar
-        export DOCKLE_VERSION=$(curl --silent "https://api.github.com/repos/goodwithtech/dockle/releases/latest" | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')
-        wget -O ~/dockle.tar.gz https://github.com/goodwithtech/dockle/releases/download/v${DOCKLE_VERSION}/dockle_${DOCKLE_VERSION}_Linux-64bit.tar.gz
-        tar zxvf ~/dockle.tar.gz -C ~/
-        export TRIVY_VERSION=$(curl --silent "https://api.github.com/repos/aquasecurity/trivy/releases/latest" | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')
-        wget -O ~/trivy.tar.gz https://github.com/aquasecurity/trivy/releases/download/v${TRIVY_VERSION}/trivy_${TRIVY_VERSION}_Linux-64bit.tar.gz
-        tar zxvf ~/trivy.tar.gz -C ~/
```
k3rnelpan1c-dev, line 48: JQ and Docker are both already installed in a recent version in the GitHub Actions Virtual environment and since you don't seem to use `trivy`, additionally it having an official action otherwise, I cleaned up this overhead

k3rnelpan1c-dev:
@stevespringett sorry for the ping, but could I get your feedback on this and whether you would be interested in a GitHub Action based release workflow instead or as an alternative to your `release.sh`?

stevespringett:
Thanks for the PR. I'll need a few days to review.

src/main/docker/Dockerfile
```diff
@@ -1,63 +1,70 @@
-FROM debian:stable-slim
-LABEL maintainer="steve.springett@owasp.org"
-LABEL vendor="OWASP"
+FROM eclipse-temurin:11.0.13_8-jre-focal AS jre-build
```
nscuro, line 1: We should go ahead and jump to `17.0.2_8-jre-focal` directly.

src/main/docker/Dockerfile
```diff
@@ -1,63 +1,70 @@
...
 
 # Arguments that can be passed at build time
 # Directory names must end with / to avoid errors when ADDing and COPYing
+ARG COMMIT_SHA
+ARG APP_VERSION
 ARG APP_DIR=/opt/owasp/dependency-track/
 ARG DATA_DIR=/data/
-ARG USERNAME=dtrack
+ARG UID=1001
+ARG GID=1001
```
nscuro, line 12: Is there a specific reason why `1001` was chosen? The previous image used `1000`, so it's safe to assume that the permissions of most mounted volumes in current deployments are geared towards it (as in `1000:1000`). 

If we keep `1001` as the default, it should be documented in the upgrade notes. But as always, there will be quite a few users running into issues when blindly upgrading.

nscuro (commented):
LGTM overall. Have a few minor suggestions, and the `UID` situation needs clarification.

src/main/docker/Dockerfile
k3rnelpan1c-dev: ah my bad, I likely used the UID i used on another Image of mine I sourced for the changes. I will revert to `1000:1000`.
Thank you for that find!

src/main/docker/Dockerfile
```diff
@@ -1,63 +1,70 @@
-FROM debian:stable-slim
-LABEL maintainer="steve.springett@owasp.org"
-LABEL vendor="OWASP"
+FROM eclipse-temurin:11.0.13_8-jre-focal AS jre-build
+
+FROM debian:bullseye-slim
```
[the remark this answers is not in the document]
k3rnelpan1c-dev: Will do, should have done that to begin with :sweat_smile:

src/main/docker/Dockerfile
k3rnelpan1c-dev: Oh okay, so directly to the new LTS? 
I have heard of ppl running into issues at work who compiled against 11 and ran on 17.
Just want to throw out this as a note, but I will change it if you still want me to.

stevespringett: @k3rnelpan1c-dev I've been testing with Java 17, but the testing has not been extensive. Lets hold off on it for now. I'm concerned about some of the changes to reflection.

namloc2001:
@k3rnelpan1c-dev - looks really good, the only thing that is missing for full OpenShift compatibility is:
```
# Specify this so that when OpenShift restricted SCC assigns arbitrary UID - it is aware of the correct home dir to use
ENV HOME ${DATA_DIR}
```

This is what I use in my Dockerfile that uses `FROM dependencytrack/apiserver:4.3.6` as base image. (DATA_DIR is used because that is the home directory assigned to dtrack user and the chown & chmod statements make it accessible when OpenShift assigns an arbitrary UID and GID=0 to the container user under the restricted SCC).

Are you please able to add it in?

@Homopatrol FYI.

k3rnelpan1c-dev:
Hi,

first of all thank you two for this observation!
I work with OpenShift at work as well and have run/tested this changes there before I contributed them and never noticed that the user home was even listed in the logs.
However, mind me asking if any of you happen to know if this is a big issue? 
I have only a limited set of test and evaluation experience with Dependency Track on OpenShift, but as far as I was able to tell it never complained nor ran into issues caused by the improperly set home dir. The likely reason for that is that Dependency Track stores all of its data to the `/data/` dir and largely, if not entirely, ignoring the home dir despise the log output by Alpine.

namloc2001:
@k3rnelpan1c-dev it appears that the container can certainly start up and analyse an SBOM with the home directory defaulting to `/`, however I would suggest that best practice is followed for compatibility and the environment variable is added in.

Also, on a separate note to this topic, I've noticed you've amended the `HEALTHCHECK` statement. I believe that by removing the "/" but not updating the comment associated with the `CONTEXT` environment variable, you risk someone setting one without a trailing slash and breaking the healthcheck command.

k3rnelpan1c-dev:
Hey @namloc2001,
I am not opposed to adding the envVar, was just wondering :wink: 
Regarding the Context, that you are indeed right that it could mess with the Healthcheck :man_facepalming:.
Unfortunately, I messed actually messed the Healtcheck up with this PR which then needed to be fixed in a followup PR #1408 and that is now in 4.4.1 (yay for the OCI standard not supporting `HEALTHCHECK` and me not testing the Docker format xD).

All things considered, I will offer the HOME env addition in a future PR and have to check the Healthcheck again :/
Thank you for the comments and the pointer for yet another issue with the Health check :expressionless:

PR #1310, opened 2021-12-23
WIP: Add filter

wurstbrot:
As a project I want to know the SBOM of my projects including OS packages like debian/rpm and software vulnerabilities.
I want to alert on software vulnerabilities but not on OS packages or only for high/critical vulnerabilties.

This PR is there to discuss the approach.
In case you agree to this approach, I will add tests.

A future PR might include this user story:
As a project I want to filter on all attributes, e.g. CVE to filter _log4shell_ vulnerability, specific severity or CPEs.

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/FindingResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/FindingResource.java
@@ -29,9 +29,13 @@ import io.swagger.annotations.Authorization;
 import io.swagger.annotations.ResponseHeader;
 import org.dependencytrack.auth.Permissions;
 import org.dependencytrack.integrations.FindingPackagingFormat;
+import org.dependencytrack.model.Component;
 import org.dependencytrack.model.Finding;
 import org.dependencytrack.model.Project;
+import org.dependencytrack.model.Severity;
 import org.dependencytrack.persistence.QueryManager;
+import org.dependencytrack.util.VulnerabilityUtil;
+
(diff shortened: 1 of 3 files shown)


stevespringett:
Thanks for the PR and for submitting the idea. The approach looks good. If we also wanted to filter by CPE, how would that work?

wurstbrot:
Hi Steve,

in case you make the components _cpe_ accessible in org.dependencytrack.model.Finding (like purl), I can filter like in this suggestion.

Is _FindingResource_ the right place for the _filter_ method? It feels like a Controller. An option is to put it in _FindingsQueryManager_. What do you think?

Disclaimer:
Each filter uses a Lists in the filter methods (currently_filteredPurlList_, _filteredCpeList_, _minimumSeverityList_). This might be memory consuming. My java knowledge is a bit rusty, any ideas?

Remarks:
I miss a _beforeEach_ method in the test class to setup the sample component/vulnerability structure, it feels like a bit too much copy&paste.

nscuro:
Closing due to inactivity for > 1 year. Feel free to reopen or raise a new PR if you still want to see this change.

msymons:
@wurstbrot, if you are still interested in progressing your submission but still have questions then remember that the slack channel (where I see you are a member) will be a good place to ask before investing time in raising a new PR.  I say this because this issue will get locked by a bot at some point in the not so distant future.

Issue #881, opened 2020-12-22
Suport for Repository Authentication


msymons:
### Current Behavior:
Dependency-Track implemented support for:

* Configuration of "Internal Components" in v3.7 (#303)
* Configurable respositories in v3.8 (#164)

Configurable repositories were implemented as an MVP and do not yet support authentication.  This makes it impossible to make use of internal repositories that enforce authentication (Nexus Repository Manager, etc) and also of external repos that do the same (such as Oracle).

One impact of not being able to connect to an internal repository would be to not have access to "current version" information for internal components.  Current simple UI display of "current version" is already very useful, but DT v4.2 milestone contains an  issue for "Add support for component age in policy conditions" (#772) and internal components should not be excluded from such policies.

### Proposed Behavior:
Extend functionality of configurable repositories to support authentication.

nscuro:
Just curious, would basic auth be the only AuthN method that must be considered? Are there any known examples of repos that use "non-standard" methods, like custom HTTP headers, query params etc.?

stevespringett:
Docker and NPM both use realms and tokens with custom endpoints. Both also support BASIC but which can be disabled leaving only the repo-specific auth functional.

For MVP, I think BASIC should suffice. That should cover the majority of use cases. If a repo doesn't end up supporting BASIC but something else, I think DT should support it. But I also think that if a repo supports BASIC in addition to something else and BASIC is disabled, then supporting that something else will be prioritized as p3 or flagged with help wanted.

PR #1311, opened 2021-12-23
Adding basic authentication for analyzer

Marx314:
#881 might come handy.

I wasn't sure about the move of unirestGet to abstract, and the complexity level is exploding for the MavenMetaAnalyzer.

The API accept the new fields, I don't know how to update the UI.

### DIFF
--- a/src/main/java/org/dependencytrack/model/Repository.java
+++ b/src/main/java/org/dependencytrack/model/Repository.java
@@ -70,6 +70,18 @@ public class Repository implements Serializable {
     private String url;
 
     @Persistent
+    @Column(name = "USERNAME")
+    @NotBlank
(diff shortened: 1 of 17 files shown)


stevespringett:
Thanks for the PR. I think we need to store the credentials using ConfigProperty so that we get the benefit of encryption at rest.

Marx314:
@stevespringett humm I'm not exactly sure how it would be easily possible without doing some fairly large assumption.
There would be only 1 set of credentials that would be sent to all type of internal repository?
1 set par type of repository?

AbdelHajou:
Hi @Marx314 , are you still working on this?

Marx314:
@AbdelHajou not really since I didn't get any additional details.

PR #1334, opened 2022-01-15
Implement API endpoint team/self

### LINKED ISSUE #861: API endpoint similar to user/self but for teams
There are features for uploading a SBOM which require special permissions, such as PROJECT_CREATION_UPLOAD. The Jenkins plugin cannot check if this requirement is met because there is no API endpoint that could be used by non-administrative API keys.

I request a "/team/self" endpoint that provides the following subset from "/team" but is restricted to the team to which the provided API key belongs.

- name
- permissions
- uuid?

**Things to consider about this**
Such an endpoint could raise a security finding because it could be used to enumerate valid API keys using brute force. On the other hand, such possibility already exists but is limited to administrative accounts.

sephiroth-j:
returns subset of Team data
- uuid
- name
- permissions

fixes #861

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
@@ -18,6 +18,7 @@
  */
 package org.dependencytrack.resources.v1;
 
+import alpine.Config;
 import alpine.auth.PermissionRequired;
 import alpine.logging.Logger;
 import alpine.model.ApiKey;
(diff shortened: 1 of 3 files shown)


stevespringett:
Thanks for the PR. I would think the API would return all teams. Is that not the case?

sephiroth-j:
The intention was to have something like "user/self" but for a team, and therefore only the information about the team belonging to the used API key is returned.

PR #1341, opened 2022-01-20
fix: during BOM upload auto-created projects do not have a classifier

sephiroth-j:
For manually created projects or when updating a project, a classifier is required and set to the default value "APPLICATION".
However, when creating a project as part of the BOM upload, the classifier is not set.
This fix reads the classifier from the BOM and updates the project as needed. If no classifier is specified in the BOM (old version), the default "APPLICATION" is used.


stevespringett:
Nice job! Thanks for the PR.

PR #1360, opened 2022-01-31
Update README to use Helm 3 command structure

ChaosInTheCRD:
Helm v3 has been released for some time, and while some users may still be on v2, I figured it'd be worth adding an install command for v3 as currently if the current command is used it will fail.


stevespringett:
Thanks for the update and PR.

PR #1372, opened 2022-02-05
Add backslash to docs cicd multiple line curl command

francisbergin:
Hi!

This change adds the backslashes to the last curl command on the usage/cicd page. This is useful for people copying and pasting the code block.

@stevespringett Thank you so much for this great project!


stevespringett:
Thanks for the PR

Issue #338, opened 2019-05-15
VIEW_PORTFOLIO:  Display Vulnerabilities Listing for Project


msymons:
### Current Behavior:
In DT v3.4.1, a user who has only VIEW_PORTFOLIO Permission can navigate to:
* Dashboard:  displays vulnerability graphs and metrics
* Projects:  displays vulnerabilities column with correct counts
* Project -> Dependency Tab: vulnerabilities column with correct counts
* Component:  Lists vulnerabilities, and the links to each vulnerability all work.

However,  the Project screens do not have a tab that lists all vulnerabilities.  Not unless the permission VULNERABILITY_ANALYSIS is granted...  and this is not desired (grants too much).

The lack of this tab means that it can be really difficult to see the big picture in a project that has many vulnerable components (i have seen a project with 121 vulnerabilities covering 27 components)

### Proposed Behavior:
* Rename project Audit tab to "Vulnerabilities".  It should be obvious that clicking on this tab will access auditing for those who have permission to audit.
* Display this tab for both VIEW_PORTFOLIO and VULNERABILITY_ANALYSIS permissions.
* When user has VULNERABILITY_ANALYSIS Permission, perhaps display hint that audit is performed by clicking on a component.
* Do not open auditing if user with VIEW_PORTFOLIO only should click on a component.

stevespringett:
VIEW_PORTFOLIO is intended for personas that only need component inventory and license information, so non-security users.

I think I prefer having a new permission that grants only VIEW_VULNERABILITY. Used in combination with VIEW_PORTFOLIO, I think it would achieve the results you're looking for.

msymons:
A new VIEW_VULNERABILITY permission sounds like a good way solve the problem.

> VIEW_PORTFOLIO is intended for personas that only need component inventory and license information, so non-security users.

So does that mean that the security information that is currently displayed to users with this permission (eg Projects: displays vulnerabilities column) is not desired behaviour and that implementation of VIEW_VULNERABILITY should be implemented alongside changes to  VIEW_PORTFOLIO that  **removes** the vulnerability column from projects page, etc?

csansone-handy:
+1 for this

I agree that it would be very nice to see a `VIEW_VULNERABILITY` permission that while on a project specific page (`https://<domain>/project/?uuid=<uuid>`) would allow users to click a tab that lists all of the vulnerabilities affecting a specific project. 

[...]

PR #1373, opened 2022-02-05
Add `VIEW_VULNERABILITY` permission

nscuro:
This PR addresses #338 by adding the new `VIEW_VULNERABILITY` permission.

`VIEW_VULNERABILITY` grants read access to `/api/v1/analysis` and `/api/v1/finding`.
This enables users to *see* vulnerabilities a project or component is affected by, as well as the audit trail, but no do any audit decisions. A nice side-effect is that this can be used to reduce the permissions granted to CI integrations like the Jenkins plugin.

To ensure compatibility with older DT versions, all users and teams who previously had the `VULNERABILITY_ANALYSIS` permission are automatically granted the `VIEW_VULNERABILITY` permission.

I was torn whether to call this `VIEW_VULNERABILITY` or `VIEW_FINDING`. `VIEW_VULNERABILITY` fits better into the current naming scheme, but `VIEW_FINDING` is more accurate. After all, viewing vulnerabilities is already possible with `VIEW_PORTFOLIO` - it's just the findings (project<->component<->vulnerability relationships) that are hidden.

Also, we may want to implement something similar for policy violations (`VIEW_POLICY_VIOLATION`). I can implement it in this PR if desired, should be fairly straightforward.

CC @msymons

### DIFF
--- a/docs/_docs/triage/auditing-basics.md
+++ b/docs/_docs/triage/auditing-basics.md
@@ -14,6 +14,7 @@ Project auditing is the process of triaging findings on the components for each 
 and audit history performed on a project only affect the findings for said project.
 
 The **VULNERABILITY_ANALYSIS** permission is required to perform project auditing.
+The audit trail will be visible to all users with **VIEW_VULNERABILITY** permission.
 
 ![Project Auditing](/images/screenshots/audit-finding-project.png)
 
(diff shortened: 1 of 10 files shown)


stevespringett:
Nice PR. Thanks so much!

PR #1386, opened 2022-02-13
Distinguish state changes in analysis trail comments

nscuro:
This PR enriches the state change comments in the analysis trail with a prefix of which state was changed.
Without this, changes are kinda ambiguous for the untrained eye.

Such a prefix is not added for suppression changes.


stevespringett:
Thank you for the PR 😄

PR #1391, opened 2022-02-15
fix: codacy upload and JRE version

k3rnelpan1c-dev:
### Description

This PR offers a fix for 2 mistakes I caused with my earlier packaging PR (#1303).
First it fixes the failing Codacy report upload, where I missed that it used to be downloaded in the 'Tool setup' step. It now uses the official Codacy GitHub action instead of the prior downloaded Jar.
Second it fixes a staging mishap that happened likely due to late local time :man_facepalming:, where I already prepared a followup commit to bump the Container JRE to 17 as requested by the original review. With this PR I offer a revert to the 11 JRE.

### Changes

* use codacy GitHub action instead of downloaded jar
* revert Container Image JRE to 11

### Issues

* relates to and fixes #1303

### DIFF
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:17.0.2_8-jre-focal AS jre-build
+FROM eclipse-temurin:11.0.14.1_1-jre-focal AS jre-build
 
 FROM debian:bullseye-20220125-slim
 
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -56,9 +56,11 @@ jobs:
 
     - name: Publish test coverage
       if: github.ref == 'refs/heads/master' && github.event.inputs.skipTests != 'true'
-      env:
-        CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
-      run: java -jar ~/codacy-coverage-reporter-assembly.jar report -l Java -r target/jacoco-ut/jacoco.xml
+      uses: codacy/codacy-coverage-reporter-action@v1
+      with:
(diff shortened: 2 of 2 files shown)


src/main/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:17.0.2_8-jre-focal AS jre-build
+FROM eclipse-temurin:11.0.14.1_1-jre-focal AS jre-build
 
 FROM debian:bullseye-20220125-slim
```
nscuro, line 3: I think we had a misunderstanding regarding the digest pinning as well (wasn't fast enough to respond on that PR for timezone reasons as well... 😅). What I meant was pinning to the SHA256 digest:

```suggestion
FROM eclipse-temurin:11.0.14.1_1-jre-focal@sha256:246f0a07f7ba2c52b48d1879aa6f294d80619575f719ea31eaec0885560b803e AS jre-build

FROM debian:bullseye-20220125-slim@sha256:4c25ffa6ef572cf0d57da8c634769a08ae94529f7de5be5587ec8ce7b9b50f9c
```

See for example https://docs.renovatebot.com/docker/#digest-pinning for more details.

nscuro: Yeah, it'd be great if this would be a little more obvious. There are numerous open issues to indicate this stuff in DockerHub, the Docker CLI and so on, but they've been open for *years*... 😅

k3rnelpan1c-dev: I am more impressed that this also still not mentioned anywhere in the OCI standard space, but hey you never stop learning so thank you again for this heads up :smiley:

stevespringett: This is one of those pull requests that reminds me that everyday is a school day. ;-) Thanks for the collaboration everyone. Much appreciated.

PR #1405, opened 2022-02-18
Fix v440Updater for MSSQL

nscuro:
The `NAME` column of the `PERMISSION` table has a `UNIQUE` constraint, using `TOP` or `LIMIT` is thus not necessary to get only a single result.

- [ ] Test on Postgres
- [ ] Test on MSSQL


nscuro:
Tested on MSSQL by having DT v4.3.6 initialize the DB, creating mutliple users and teams with `VULNERABILITY_ANALYSIS` permission and then running DT v4.5.0-SNAPSHOT with the changes in this PR. Upgrade completed successfully and the permissions have been assigned as expected.

nscuro:
Successfully tested with Postgres as well. ✅

PR #1408, opened 2022-02-18
Fix container image health check

nscuro:
The container health check is currently failing due to the move from Alpine to Debian slim as base image:

```shell
$ docker inspect --format "{{json .State.Health }}" dtrack-apiserver | jq
{
  "Status": "starting",
  "FailingStreak": 2,
  "Log": [
    {
      "Start": "2022-02-18T10:55:15.588984439+01:00",
      "End": "2022-02-18T10:55:15.69434158+01:00",
      "ExitCode": 1,
      "Output": "/bin/sh: 1: wget: not found\n"
    }
  ]
}
```

`wget` does not come preinstalled on Debian slim. Additionally, the syntax of disabling proxies differs from the `wget` provided by Alpine vs. Debian.

Downside is that installing `wget` via apt affects the reproducability of the image built. 
We could pin `wget` to a specific version, but then we'd have to keep up with updating it regularly.

### DIFF
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -2,11 +2,6 @@ FROM eclipse-temurin:11.0.14.1_1-jre-focal@sha256:246f0a07f7ba2c52b48d1879aa6f29
 
 FROM debian:bullseye-20220125-slim@sha256:4c25ffa6ef572cf0d57da8c634769a08ae94529f7de5be5587ec8ce7b9b50f9c
 
-# Install wget for healthcheck
-RUN apt update \
-    && apt install -y wget \
-    && rm -rf /var/lib/apt/lists/*
-
 # Arguments that can be passed at build time
 # Directory names must end with / to avoid errors when ADDing and COPYing
 ARG COMMIT_SHA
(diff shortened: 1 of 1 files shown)


src/main/docker/Dockerfile
```diff
@@ -2,6 +2,11 @@ FROM eclipse-temurin:11.0.14.1_1-jre-focal@sha256:246f0a07f7ba2c52b48d1879aa6f29
 
 FROM debian:bullseye-20220125-slim@sha256:4c25ffa6ef572cf0d57da8c634769a08ae94529f7de5be5587ec8ce7b9b50f9c
 
+# Install wget for healthcheck
+RUN apt update \
+    && apt install -y wget \
+    && rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
k3rnelpan1c-dev: Oh, good call! yeah I think that moving the `RUN` in-front of the `COPY` instruction certainly would help in local development as the other layers would be cached then.
From a CI perspective there is no gain but the development experience alone would make this definitely worth it!

I would move both `COPY` instructions though, while I am not 100% sure if the multi stage build would benefit from the layer cache (I assume so though), it would group the instructions.

src/main/docker/Dockerfile
```diff
@@ -27,20 +27,26 @@ ENV TZ=Etc/UTC \
...
-# Automatically creates the $APP_DIR directory
-COPY ./target/${WAR_FILENAME} ${APP_DIR}
-
-# Create the directory where Dependency-Track will store its data (${DATA_DIR}) and the external library directory (${EXTLIB_DIR})
+# Create the directories where the WAR will be deployed to (${APP_DIR}) and Dependency-Track will store its data (${DATA_DIR})
 # Create a user and assign home directory to a ${DATA_DIR}
 # Ensure UID 1000 & GID 1000 own all the needed directories
-RUN mkdir -p -m 770 ${DATA_DIR} \
+RUN mkdir -p -m 770 ${APP_DIR} \
+    && mkdir -p -m 770 ${DATA_DIR} \
```
k3rnelpan1c-dev, line 34:
```suggestion
RUN mkdir -p -m 770 ${APP_DIR} ${DATA_DIR} \
```

src/main/docker/Dockerfile
```diff
@@ -27,20 +27,26 @@ ENV TZ=Etc/UTC \
...
     && addgroup --system --gid ${GID} dtrack || true \
     && adduser --system --disabled-login --ingroup dtrack --no-create-home --home ${DATA_DIR} --gecos "dtrack user" --shell /bin/false --uid ${UID} dtrack || true \
     && chown -R dtrack:0 ${DATA_DIR} ${APP_DIR} \
-    && chmod -R g=u ${DATA_DIR} ${APP_DIR}
+    && chmod -R g=u ${DATA_DIR} ${APP_DIR} \
+    \
+    # Install wget for health check
+    && apt update \
+    && DEBIAN_FRONTEND=noninteractive apt install -y wget \
```
k3rnelpan1c-dev, line 42:
```suggestion
    && apt-get -yqq update \
    && DEBIAN_FRONTEND=noninteractive apt-get install -yqq wget \
```
This is personal preference, but 
1. `apt` is 'non script-able' / not recomended to be run in scripts, where `apt-get` is IIRC
2. its quite nosy IMHO so `qq` (still nosy but less)

k3rnelpan1c-dev (approved):
2 final nits, but LGTM

src/main/docker/Dockerfile
nscuro: `g=u` only says "same permissions for group as for user", but it doesn't specify `rwx`. Without `-m 770`, the directory permissions are `775`.

k3rnelpan1c-dev: true, but I am wondering if we should care if `other` has read and execute rights for the folders ;)

nscuro: Probably not. When using default volumes, `other` can't access the directory on the host system anyway. And when mounted as host volume, the permissions of that host volume apply. That's for `DATA_DIR` which contains key pairs, for `APP_DIR` it doesn't matter at all. I'm going to remove the `-m` flags.

stevespringett:
Thank you both for the collab on this one. I'll push out 4.4.1 shortly

Issue #1409, opened 2022-02-18
500 when Supressing a Vulnerability


ddurham2:
### Current Behavior:
Getting an HTTP 500 when suppressing a vulnerability

### Steps to Reproduce:
Mark vulnerability as suppressed

### Expected Behavior:
Not to receive an error

### Environment:
  
  - Dependency-Track Version: 4.4.1
  - Distribution: Docker
  - BOM Format & Version: CycloneDX 1.3
  - Database Server: PostgreSQL
  - Browser: Firefox

### Additional Details:

docker log shows:
2022-02-18 21:45:57,250 ERROR [GlobalExceptionHandler] Uncaught internal server error
java.lang.NullPointerException: null
	at org.dependencytrack.resources.v1.AnalysisResource.updateAnalysis(AnalysisResource.java:162)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory.lambda$static$0(ResourceMethodInvocationHandlerFactory.java:52)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:124)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:167)
	at org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$ResponseOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:176)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.dispatch(AbstractJavaResourceMethodDispatcher.java:79)
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:475)
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:397)
[...]

PR #1417, opened 2022-02-24
Fix NPE when updating an analysis that was created with DT y= 4.3.6

nscuro:
The NPE was caused by calling `.name()` on `null` values. 
I took the chance to add `NOT_SET ->` transitions for *Justification* and *Vendor Response* as well.

Addresses #1409


nscuro:
Snyk is complaining, but I don't have the required permission to see *what* it's complaining about.

stevespringett:
Its failing on the vulnerable H2 component. But all good. Thanks for finding and fixing.

PR #1419, opened 2022-02-26
ci: optimization and release automation

k3rnelpan1c-dev:
### Description

This PR proposes a rework of the currently used GitHub Action workflow as well as to the Shell script used to release new versions of Dependency Track.
It currently does not include any particular documentation, but I can add some in case it is desired. 

### Changes

* separate test and build workflow
* add release workflow to replace `release.sh`
* pin versions of GitHub Actions used
* small Container Image optimizations (may be separated out into its own PR)
  * container home environment for arbitrary uid 
  * ignore apt-get recommended packages


stevespringett:
Thanks for the PR. I've been meaning to do this but lacked time. I'm wondering how the ci-release action would support releasing from master as well as releasing from a a versioned branch (e.g. 4.4.x). I currently have a need to release from both.

k3rnelpan1c-dev:
> Thanks for the PR. I've been meaning to do this but lacked time. I'm wondering how the ci-release action would support releasing from master as well as releasing from a a versioned branch (e.g. 4.4.x). I currently have a need to release from both.

Hi, no problem CI is my passion to an extend :)
Yes, I have both of those "environments" in mid, I just had no time to finish the workflow yet. Ran into a issue with my test runs on my fork, which I need to track down and fix before I can update the PR branch and open it for review.

k3rnelpan1c-dev:
Okay, this took way longer than expected due to several distractions and general lack of time.
Regardless, this is now ready for review from a technical point of view.
I have only one question: Assuming that I should document somewhere how to use the release workflow, is there any preferred location for this kind of documentation?

k3rnelpan1c-dev:
Well, I simply went ahead and added a very basic AsciiDoc "readme" in the `.github` folder.
Every kind of feedback is always welcome :)

stevespringett:
Thank you so much. I've gone through all the workflows and they look fantastic. I'd like to get this merged in prior to the v4.5 release and use this new release process for it. I'll likely test in my own environment, similar to what you did on your fork.

@nscuro any comments on this?

.github/workflows/ci-release.yaml
```diff
@@ -0,0 +1,175 @@
...
+        description: 'Use this to overwrite the version number to release, otherwise uses the current SNAPSHOT version (expected format x.y.z)'
+        type: string
+
+jobs:
+  prepare-release:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.varaibles.outputs.version }}
+      next-version: ${{ steps.varaibles.outputs.next-version }}
+      release-branch: ${{ steps.varaibles.outputs.release-branch }}
```
nil4, line 18: `varaibles` looks like a typo here?

k3rnelpan1c-dev: :sweat_smile: yes that is in fact a typo, however I seem to have been efficient and made it across the entire workflow, which means it still works at least.

I shall fix it though, thanks for catching that :+1:

k3rnelpan1c-dev:
> Thank you so much. I've gone through all the workflows and they look fantastic. I'd like to get this merged in prior to the v4.5 release and use this new release process for it. I'll likely test in my own environment, similar to what you did on your fork.

Glad you like this proposal :slightly_smiling_face:
FYI, be sure to alter the push location of the Image when you try this and that the push secrets are set in your test environment. That is about the only bit that is a tat tedious when testing this.

k3rnelpan1c-dev:
Oh and another thing I want to mention, I replicated a typo in the `HUB_ACCESSS_TOKEN`, which was already present in the original workflow.

Just remembered that as I am now looking at the *frontend* workflow, where it was properly spelled.
Might be worth thinking about moving that secret to the ORG level and then granting access to the repos that need that ORG secret then :wink:

.github/workflows/_meta-build.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1.2.0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1.6.0
+        id: buildx
+        with:
+          install: true
+
+      - name: Available Platforms
```
nscuro, line 79: I know this is from the current workflow, but I don't think there's a reason for keeping it.

nscuro (approved):
LGTM. Quite impressed with the GHA-fu that @k3rnelpan1c-dev pulled off here. 👍

k3rnelpan1c-dev:
thank you for the merge :smiley:

PR #1423, opened 2022-03-01
Changed port exposed from 8080 to 8081.

JDembrun:
I believe this change is necessary in order to allow both dependency track containers to run in the same task. This will expose the API service on port 8081, while the frontend service is exposed on 8080.


stevespringett:
This is currently handled by [docker-compose.yml](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/docker/docker-compose.yml) which works great with Docker Compose and Swarm.

PR #1480, opened 2022-03-17
Bugfix/mirror contention

officerNordberg:
When NVD is slow our internal process that uses the dependency-track mirror experiences issues cause by corrupted/incomplete files being downloaded when the mirroring process is taking place.  This PR improves* the chances of getting a good file by downloading to a temp location and then moving it into place.  

*The filesystem is stable but we still are running into issues with the alpine StaticResourceServlet.  Our solution is to provide the mirror via nginx for now as I don't understand alpine ins and outs. The following script is similar to what we are using to discover inconsistent mirror state if run while Mirror is updating.

` while true;  curl http://dependency-track/mirror/nvdcve-1.1-2020.json.gz |  zcat  | jq . 2>&1 > /dev/null; done`


stevespringett:
Thanks for the PR

Issue #1210, opened 2021-10-01
Decouple metrics update into standalone service


stevespringett:
Updating metrics takes a considerable about of time and takes resources away from the DT server when its processing and analyzing SBOMs.

In order to improve performance of the DT ApiServer and the metric updates themselves, a separate service should be created to perform this task.

The service may or may not use the same stack. Non-Java stack options being considered include Erlang, Python, and Go. Go may have the least overhead thus resulting in a smaller container and faster execution.

This service will run independently in a separate container. There will be no need for the DT ApiServer and the MetricsService to communicate with each other.

The MetricsUpdate service will update portfolio, project, component, and vulnerability metrics on a periodic basis, similar to how it is done now. 

Additionally, whenever the DT ApiServer consumes or processes a BOM, it will write to a new table and specify that the project is in need of a metric update. The MetricsUpdate service will periodically query (5 min???) this table looking for new projects in need of updating and perform the necessary updates.

nscuro:
Some thoughts:
- The service should stay away from ORMs as they make understanding and debugging DB issues unnecessarily hard
  - Some query tuning (selecting only the columns needed etc.) may also yield additional performance gains
- The service should have a "one shot" option. That way its execution can be controlled via [`CronJob`](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) in k8s (or standard cron)
  - Also, it doesn't have to idle when there's no work to do
- Kicking off the service for specific projects, triggered via `BOM_PROCESSED` webhook could be a useful feature
- Unsurprisingly I'd be in favor of Go regarding the tech stack, but I would also throw these in the ring:
  - .NET Core
  - Java + [Quarkus](https://quarkus.io/)
  - Java + [Micronaut](https://micronaut.io/)
- Erlang is pretty exotic, getting contributors will be a challenge

A first MVP could be worked on that implements the "one shot" mode:
* connect to DB
* calculate metrics for all projects (regardless if needed or not)
* exit

Eventually there could be multiple MVP implementations with different tech stacks before a final decision is made.

hostalp:
Also note that even the current implementation could be likely improved. Currently it seems to struggle here when filling up some collection object:
```
at java.util.AbstractCollection.addAll(java.base@11.0.12/Unknown Source)
at java.util.HashSet.<init>(java.base@11.0.12/Unknown Source)
at org.datanucleus.ExecutionContextImpl.preCommit(ExecutionContextImpl.java:4158)
[...]

Issue #1370, opened 2022-02-03
dependency-track high cpu usage


mawl:
The defect may already be reported! Please search for the defect before creating one.

### Current Behavior:
![image](https://user-images.githubusercontent.com/5493374/152354187-b3606ed8-58de-4396-903d-7593fa2db9d9.png)

![image](https://user-images.githubusercontent.com/5493374/152352785-dc042e6a-8c0d-460b-9123-07f60d7d77e5.png)

### Expected Behavior:
No high cpu usage, no cascading effect.

### Environment:
  
  - Dependency-Track Version: 4.3.6 apiserver, 4.3.1 frontend
  - Distribution: docker-compose
  - Database Server: PostgreSQL 13.5
  - ~ 1500 Projects imported

### Additional Details:

Looks like https://github.com/DependencyTrack/dependency-track/issues/264

This is found in the logs during every cpu usage increasement:

`2022-02-03 04:24:28,504 [] INFO [org.dependencytrack.tasks.MetricsUpdateTask] Executing metrics update on vulnerability database
2022-02-03 04:24:39,924 [] INFO [org.dependencytrack.tasks.scanners.InternalAnalysisTask] Internal analysis complete
2022-02-03 04:24:39,929 [] INFO [org.dependencytrack.tasks.scanners.NpmAuditAnalysisTask] Starting Node Audit analysis task
2022-02-03 04:25:04,837 [] INFO [org.dependencytrack.tasks.scanners.NpmAuditAnalysisTask] Node Audit analysis complete
2022-02-03 04:25:04,843 [] WARN [org.dependencytrack.tasks.scanners.OssIndexAnalysisTask] An API username or token has not been specified for use with OSS Index. Using anonymous access
2022-02-03 04:25:04,843 [] INFO [org.dependencytrack.tasks.scanners.OssIndexAnalysisTask] Starting Sonatype OSS Index analysis task
2022-02-03 04:25:16,616 [] INFO [org.dependencytrack.tasks.scanners.OssIndexAnalysisTask] Analyzing 2 component(s)
2022-02-03 04:25:16,616 [] INFO [org.dependencytrack.tasks.scanners.OssIndexAnalysisTask] Sonatype OSS Index analysis complete
2022-02-03 04:25:16,662 [] INFO [org.dependencytrack.tasks.VulnerabilityAnalysisTask] Completed analysis of 2 components in project: c5caee7c-a6c2-4b2c-b3fb-ec6d2a5908f9
[...]

PR #1481, opened 2022-03-19
Tweak `MetricsUpdateTask` performance

nscuro:
I was profiling the `MetricsUpdateTask` class due to some folks reporting performance concerns and managed to identify a few screws we could tighten to make the task more resource efficient.

#### Setup

* DT 4.4.2 running on Temurin 11 on macOS.
* Postgres database running in Docker, using the `postgres:14.2-alpine` image.
* 50 projects, all using the DT API Server 4.4.2 BOM. 7850 components in total.
* `MetricsUpdateTask` running in a `JerseyTest` without any resources deployed. The task is running pretty much isolated.
* Because no other tasks are running, no data is changing, thus the metrics update task won't perform any `INSERT`s, just `UPDATE`s.
* Using *Async Profiler* in IntelliJ.

#### Results

The current implementation is performing a lot of allocations in the `updateComponentMetrics` method. Profiling results showed that the biggest chunk is allocated in [`AbstractAlpineQueryManager.persist`](https://github.com/stevespringett/Alpine/blob/v2-dev/alpine-infra/src/main/java/alpine/persistence/AbstractAlpineQueryManager.java#L425). More specifically, in DataNucleus' [`preCommit`](https://github.com/datanucleus/datanucleus-core/blob/datanucleus-core-5.2.10/src/main/java/org/datanucleus/ExecutionContextImpl.java#L4153) method, where DN maintains its [L1 cache](https://www.datanucleus.org/products/accessplatform_4_1/jdo/pm.html#level1_cache):

Screenshot 2022-03-19 at 23 17 22

This is also where a lot of CPU time is spent:

Screenshot 2022-03-19 at 23 16 20

The DN docs say WRT the L1 cache:

> Each `PersistenceManager` maintains a cache of the objects that it has encountered (or have been "enlisted") during its lifetime. [...]
> Objects are placed in the L1 cache (and updated there) during the course of the transaction. This provides rapid access to the objects in use in the users application and is used to guarantee that there is only one object with a particular identity at any one time for that `PersistenceManager`. **When the `PersistenceManager` is closed the L1 cache is cleared.**

Because DT currently uses a single `PersistenceManager` for the entire portfolio metrics update run, the L1 cache keeps on growing, and DN has to do a lot of work in order to maintain it. Performance deteriorates with every project being processed. So far, this matches what @hostalp was reporting in https://github.com/DependencyTrack/dependency-track/issues/1210#issuecomment-936023060.

Another factor here is what [`AbstractAlpineQueryManager.persist`](https://github.com/stevespringett/Alpine/blob/v2-dev/alpine-infra/src/main/java/alpine/persistence/AbstractAlpineQueryManager.java#L425) does with objects passed to it:

* Start a transaction
* Make the object *persistent* (it may or may not be *transient* until now)
* Commit the transaction (this *detaches* the object again)
* Refresh the detached object, making it *persistent* again

<a href="https://www.datanucleus.org/products/accessplatform_4_2/jdo/object_lifecycle.html">Screenshot 2022-03-19 at 23 49 16</a>

Especially for larger objects, that's quite a bit of work being performed. In cases like metrics updates, we only want to `INSERT` or `UPDATE` something, but we don't care about the object after that, we don't need it to be *persistent*.

There are cases in the current logic where `persist` is called even if it isn't necessary, causing some additional overhead. One of these cases is when updating the `lastInheritedRiskScore` field of `Component`, when the metrics themselves didn't change (see [here](https://github.com/DependencyTrack/dependency-track/blob/4.4.2/src/main/java/org/dependencytrack/tasks/MetricsUpdateTask.java#L502-L505)). Semantically, this is (almost?) always a no-op. But in the background, the persistence layer is producing garbage.

When working on objects that already are *persistent* we could achieve the desired effect using something like this as well:

```java
qm.getPersistenceManager().currentTransaction().begin();
component.setLastInheritedRiskScore(last.getInheritedRiskScore());
qm.getPersistenceManager().currentTransaction();
```

That safes us the unnecessary calls to `makePersistent` and `refresh`.

### Tweaks

I didn't want to change too much here, so I only introduced two changes:

* Every `updateProjectMetrics` method gets its own `QueryManager`. This ensures that resources like the L1 cache are regularly cleared. We *do* want to take advantage of DN's caching to some degree, so using a different `QueryManager` for each `updateComponentMetrics` method is not reasonable. According to [DN's docs](https://www.datanucleus.org/products/accessplatform_4_1/jdo/performance_tuning.html), creating `PersistenceManager` instances is a cheap operation though.
* `lastInheritedRiskScore` fields of projects and components are only updated when they actually changed. In the majority of cases, this won't be the case and we're saving quite a bit of computing power.

#### Results

Allocations are reduced by more than 50%:

Screenshot 2022-03-19 at 23 22 31

CPU time spent is drastically reduced as well:

Screenshot 2022-03-19 at 23 21 38

With the original code, the task ran for **3m30s**. With the changes outlined above, it finished just under **2m30s**!

There are probably a few more places throughout the codebase where similar minor changes could be made to achieve such results.

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/MetricsUpdateTask.java
+++ b/src/main/java/org/dependencytrack/tasks/MetricsUpdateTask.java
@@ -97,10 +97,12 @@ public class MetricsUpdateTask implements Subscriber {
 
         // Iterate through all projects
         LOGGER.debug("Iterating through active projects");
-        for (final Project project: projects) {
-            try {
+        for (final Project project : projects) {
+            // Due to the large buildup of cached objects, we use a new query manager
+            // for each project. That way, resources are released in reasonable intervals.
+            try (final var pqm = new QueryManager()) {
                 // Update the projects metrics
-                final MetricCounters projectMetrics = updateProjectMetrics(qm, project.getId());
+                final MetricCounters projectMetrics = updateProjectMetrics(pqm, project.getId());
                 projectCountersList.add(projectMetrics);
             } catch (Exception e) {
                 LOGGER.error("An unexpected error occurred while updating portfolio metrics and iterating through projects. The error occurred while updating metrics for project: " + project.getUuid().toString(), e);
--- a/src/main/java/org/dependencytrack/tasks/MetricsUpdateTask.java
+++ b/src/main/java/org/dependencytrack/tasks/MetricsUpdateTask.java
@@ -361,11 +361,13 @@ public class MetricsUpdateTask implements Subscriber {
             last.setLastOccurrence(measuredAt);
             LOGGER.debug("Persisting metrics for project: " + project.getUuid());
             qm.persist(last);
-            LOGGER.debug("Updating Inherited Risk Score for project: " + project.getUuid());
             // Update the convenience fields in the Project object
-            project.setLastInheritedRiskScore(last.getInheritedRiskScore());
-            LOGGER.debug("Persisting metrics for project: " + project.getUuid());
-            qm.persist(project);
+            if (project.getLastInheritedRiskScore() != last.getInheritedRiskScore()) {
+                LOGGER.debug("Updating Inherited Risk Score for project: " + project.getUuid());
+                project.setLastInheritedRiskScore(last.getInheritedRiskScore());
+                LOGGER.debug("Persisting metrics for project: " + project.getUuid());
+                qm.persist(project);
+            }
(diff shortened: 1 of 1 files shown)


nscuro:
Will hopefully also have an impact on #1370.

PR #1486, opened 2022-03-21
Draft: Feature/matching fuzzy and removal

officerNordberg:
@stevespringett  here's the fuzzy matching fork that we are piloting internally.  I modeled it after NVD's own wildcard CPE searches and leveraged a bit of lucene's fuzzy matching to cover a few corner cases.

I also added the removal of a Vulnerability (with notification) when a version changed to a value no longer covered by the vulnerable range.  This is configurable so as not to introduce new unexpected behavior.

The JUnit test relies on an existing index so it's more of an integration than unit test.  If skip tests is ever omitted from the builds this would need to be mocked, setup or @ignored

I added the vulnerable software and cpe searching to the API because it's useful for troubleshooting but not sure if it should stay or not.


stevespringett:
Thanks much for the PR. I'll take an in-depth look at it over the next few days.

officerNordberg:
I found one issue with it that has me scratching my head. My remediation notification works on a single component version change but then erroneously fires daily. I can’t seem to identify when a vulnerability is included on a component or not when it is reanalyzed and doesn’t match the version range.  I may drop the new notification in my fork if I can’t resolve this. It seems like it should be so easy but I’m missing something 

I would also like to offer to implement pedigree including patches and variations (which might better resolve my fuzzing issue) but I would appreciate your guidance around that. 

The fuzzing may have its issues but what I’m almost certain is worth keeping is the removal of tokenization from indexing of cpe fields and the regex query creation. This now behaves much or exactly like the NIST cpe search.

stevespringett:
Hey @officerNordberg. Just following up here. Whats the status of this PR? From the ticket, its still in draft and it appears to now have a minor merge conflict.

We're getting close to releasing v4.5. If this is nearly ready, perhaps we can include it. If not, we'll wait for the next release.

officerNordberg:
I was looking for some feedback.  I can resolve the merge conflict today. As far as readiness otherwise, we have been building this fork and using this code for over a month now.  I've found and resolved some true bugs but also have tweaked the "fuzziness" a few times.  I'd possibly like to have a larger conversation about project philosophy/goals at some point.  Really, the feature might possibly be ready but without feedback I'd say skip it for v4.5

officerNordberg:
Well screw me, DCO got enabled since the last time I contributed and now I need to dance :sigh:

officerNordberg:
DCO is evil

Issue #249, opened 2018-11-30
License Change Audit


msymons:
License Management in Dependency-Track should implement an audit trail (similiar to that implemented for vulnerability management) as an enhancement.

A use-case for this will be for when support for multiple licenses is introduced (#170).  The audit trail should record what license was chosen (with optional comment), and who performed the update (and when).

stevespringett:
Do you have examples (screenshots perhaps) of what other products do in this area? License auditing is a bit out of my area of expertise.

msymons:
I have used WhiteSource and Black Duck Hub, and Nexus IQ - but none recently.

For Nexus IQ, there is public documentation:
 
- [Component License Information](https://help.sonatype.com/iqserver/reporting/application-composition-report/component-license-information)
- [License Threat Groups](https://help.sonatype.com/iqserver/managing/policy-management/license-threat-groups)

I'll try and put together some more thoughts...

hehaveri:
Here is a description of our open source license compliance needs - I wonder if these would be useful for others:

- There should be a way to define a list of compliant licenses, either as a global setting or separately per project. This is the organization's or project's OSS license policy. For example, some projects allow LGPL components while other projects cannot include LGPL code. 

- A report would show the compliance of a project with regard to the OSS license policy. This would be similar to the vulnerability dashboards, and it could highlight the projects that have license compliance issues

-  It would be nice to have another report that would list the components of a project grouped by license

- Not all components will have their license available in the CycloneDX BOM, so there should be a way to mark these components as compliant or non-compliant manually in Dependency-Track. However, as the component authors may later change the license in newer versions of the component (as the BOM is updated), then the manually entered compliance ratings may become invalid. Maybe the tool could highlight these components to indicate that the version has changed since the latest manual license compliance approval

stevespringett:
@hehaveri thank you for the use-cases. Being an OWASP project with a security focus, my first priority will be #83 which adds security-related policies. But at that point, a policy engine would have been built and integrated into Dependency-Track. I'll ensure its designed in a way to also accommodate license policies as well. 

Reporting is not something this project will ever do however. Everything is API-first, API-all-the-time. Reports themselves are not actionable and outdated the moment they're generated. The thing we're trying to build here is more about real-time consumption and being a single source of truth. We will have dashboard, a visual way to see various metrics (including policy violations), just like what you have now with other metrics, but the exporting of reports is not on the roadmap. Using the API, you could however, generate a report.

[...]

PR #1492, opened 2022-03-23
Perform policy evaluation during portfolio analysis

gpennetier:
According to this [post](https://github.com/DependencyTrack/dependency-track/issues/249#issuecomment-754725274), I have added a call to performPolicyEvaluation during the portfolio analysis. Otherwise, policy analysis is not done every 24 hours


stevespringett:
Thanks for the PR

PR #1526, opened 2022-04-09
Fix NPE in MetricsUpdateTask

nscuro:
Fixes #1525

Signed-off-by: nscuro <nscuro@protonmail.com>

PR #1527, opened 2022-04-09
Metrics update tests

nscuro:
This PR adds tests for `MetricsUpdateTask`, making it easier to debug and verify its functionality when stuff gets changed.

By writing these tests, I already spotted and squashed some bugs:
* The count of audited and suppressed findings was off, as their queries still included `project == null` expressions. Probably leftovers from the global component model before v4.
*  The count of audited findings was off, because it didn't account for suppressed findings. If a finding is suppressed, the overall finding count is reduced as well. Until now however, suppressed findings were considered "audited" as well, which could result in situations where more findings are audited than even exist.
* The audited/unaudited counts of policy violations were off, because the count of audited violations was never populated.

This PR also includes #1526.


nscuro:
Just noticed that this needs a little more work.

PR #1529, opened 2022-04-10
Project (auto)create: Add Team to ACL for ApiKey principals

### LINKED ISSUE #1435: Not allowed to upload sbom to project I create.
### Current Behavior:
As a team with BOM_UPLOAD and PROJECT_CREATION_UPLOAD permissions
When uploading a new sbom with autoCreate=true
I get an error "Access to the specified project is forbidden"
The project is created, but the sbom is not uploaded.

### Expected Behavior:
As a user or team with BOM_UPLOAD and PROJECT_CREATION_UPLOAD permissions
When uploading a new sbom with autoCreate=true
I want the newly created project to be added to my allowed list of projects
so that the bom upload succeeds.

### Environment:
  - Dependency-Track Version: v4.4.1
  - Distribution: Docker
  - BOM Format & Version: json 1.4
  - Database Server: PostgreSQL
  - Browser: Brave

valentijnscholten:
**Changes**
When Project is autocreated during BOM upload and Portfolio Access Control is enabled, the ACL of the new Project should contain the Team that created the Project. fixes #1435 

The same cannot yet be done for Users autocreating a Project as we don't know which Team should be added to the ACL. Users can be in 0 or more Teams.

**Unit tests**
I was planning on updating the unit tests to reflect this scenario. But there are no unit tests around Portfolio Access Control it seems. It would to be too big a change for me here to add them in this PR.

**Notes**
There already was some duplicate code in the `BomResource`. I didn't want to refactor too much, but tried to avoid creating more duplicate code.

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
@@ -18,6 +18,7 @@
  */
 package org.dependencytrack.persistence;
 
+import alpine.common.logging.Logger;
 import alpine.event.framework.Event;
 import alpine.model.ApiKey;
 import alpine.model.Permission;
@@ -52,6 +53,8 @@ import java.util.Map;
 import java.util.UUID;
 
 final class ProjectQueryManager extends QueryManager implements IQueryManager {
+
+    private static final Logger LOGGER = Logger.getLogger(ProjectQueryManager.class);
 
     /**
      * Constructs a new QueryManager.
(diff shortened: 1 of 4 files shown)


valentijnscholten:
Maybe we also have to add the `Team` to the ACL when creating a Project via th `ProjectResource`?

valentijnscholten:
Updated `ProjectResource` to also add the Team from ApiKey principals to the newly created Project.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -741,6 +744,32 @@ private void preprocessACLs(final Query<Project> query, final String inputFilter
...
+     * This only happens if Portfolio Access Control is enabled and the @param principal is an ApyKey
+     * For a UserPrincipal we don't know which Team(s) to add to the ACL,
+     * See https://github.com/DependencyTrack/dependency-track/issues/1435
+     * @param project
+     * @param principal
+     * @return True if ACL was updated
+     */
+    public boolean updateNewProjectACL(Project project, Principal principal) {
+        if (isEnabled(ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED) && principal instanceof ApiKey) {
+            ApiKey apiKey = (ApiKey) principal;
```
stevespringett, line 758: Is this what we want to do? If I cloned a project, I would expect all users and teams that had access to the project would also have access to the new project. This method appears to only grant access to the API key that performed the cloning.

valentijnscholten: This PR only covers creation of new projects. I was thinking about a separate PR for cloning, or do you prefer to have it combined here? It would add a field to the Clone Project request.

stevespringett: Ah correct. Different things. A separate PR for the cloning would be good.

PR #1530, opened 2022-04-10
Add tests for `MetricsUpdateTask` and squash some bugs along the way

nscuro:
This PR adds tests for `MetricsUpdateTask`, making it easier to debug and verify its functionality when stuff gets changed.

* The count of audited and suppressed findings was off, as their queries still included `project == null` expressions. Probably leftovers from the global component model before v4.
* The count of audited findings was off, because it didn't account for suppressed findings. If a finding is suppressed, the overall finding count is reduced as well. Until now however, suppressed findings were considered "audited" as well, which could result in situations where more findings are audited than even exist.
* The audited/unaudited counts of policy violations were off, because the count of audited violations was never populated.
* Suppressed policy violations were not reflected in the metrics.

This PR also includes #1526.

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/FindingsQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/FindingsQueryManager.java
@@ -58,30 +58,36 @@ public class FindingsQueryManager extends QueryManager implements IQueryManager 
 
     /**
      * Returns the number of audited findings for the portfolio.
+     * Findings that are suppressed or have been assigned the states {@link AnalysisState#NOT_SET} or {@link AnalysisState#IN_TRIAGE}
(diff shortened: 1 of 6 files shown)


nscuro:
Note that some of these bugs were not visible in the dashboard, and in consequence neither to most users.

For example, the wrong suppression count for component metrics did not lead to overall wrong suppression counts, because suppressions are queried again on the project level. So the sum on the dashboard would still be accurate, but for component metrics fetched via the API, it wasn't.

Also, per-type audited/unaudited counts of policy violations aren't displayed anywhere in the UI yet.

stevespringett:
Huge thanks for the PR

PR #1566, opened 2022-04-25
Fix unnecessary creation of audit trail entries for `Justification` and `Response`

nscuro:
Also add more tests for `AnalysisResource` and `ViolationAnalysisResource`. Additionally test notifications being sent.

Before (in v4.4.2):

Before

After (this PR):

After


stevespringett:
Good catch. Thank you.

PR #1577, opened 2022-04-29
Fix flaky `AnalysisResource` and `ViolationAnalysisResource` tests

nscuro:
The tests assert the existence of notifications, which are sent asynchronously by DT's event system. While locally this happens fast enough, it seems like it's delayed when executed in CI, causing test failures.


stevespringett:
Interesting... I was wondering about that. Thanks for identifying it. Merge when ready.

PR #1586, opened 2022-05-05
Add PATCH method to the project resource for partially updating an existing project

sephiroth-j:
This eases project management by removing the need to fetch a project at first, modifying the parts of interest and updating the project in total.
For example , it is now possible to change the `active` property from `true` to `false` without touching other properties.

Please note that I had to change the behavior of `Project#isActive` to return `null` if `active` is actually `null`. In order not to change the API response, `null` is serialized as `true` with a custom `JsonSerializer`, thus keeping the behavior the same externally.

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
@@ -23,6 +23,7 @@ import alpine.event.framework.Event;
 import alpine.persistence.PaginatedResult;
 import alpine.server.auth.PermissionRequired;
 import alpine.server.resources.AlpineResource;
+import io.jsonwebtoken.lang.Collections;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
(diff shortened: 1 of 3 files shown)


valentijnscholten:
Curious, did you try using the `isDirty` functionality built-in to JDO/Nucleus to see a write to the DB is needed?

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -392,5 +394,70 @@ public void deleteProjectInvalidUuidTest() {
         Assert.assertEquals(404, response.getStatus(), 0);
     }
 
+    @Test
+    public void patchProjectTest() {
```
nscuro, line 398: Can we split these 4 test cases into separate methods? Makes it easier to spot what went wrong in build logs, but also reduces cognitive complexity. Currently you need to keep track of whether e.g. `p1` is modified and *how* throughout the method.

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -275,6 +280,99 @@ public Response updateProject(Project jsonProject) {
...
+     * @param <T> the type of the old and new value
+     * @param source the source object that contains the new value
+     * @param target the target object that should be updated
+     * @param getter the method to retrieve the new value from {@code source}
+     * and the old value from {@code target}
+     * @param setter the method to set the new value on {@code target}
+     * @return {@code true} if {@code target} has been changed, else
+     * {@code false}
+     */
+    private <T> boolean copyIfDifferent(final Project source, final Project target, final Function<Project, T> getter, final BiConsumer<Project, T> setter) {
```
nscuro, line 367: Nitpick: It's less copying and more setting. Rename this to `setIfDifferent` maybe?

nscuro (changes requested):
Thanks for the PR! I also found myself wanting such a functionality in the past on multiple occasions. Few minor nitpicks, other than that it looks good to me.

sephiroth-j:
I have touched up the bits, please review them again.

nscuro (approved):

Issue #1396, opened 2022-02-16
Send an alert for policy violations


ribbybibby:
### Current Behavior:
It's currently possible to send alerts when an audit decision is made (i.e Analysis: Approve) but there's no alert when policy violations are found. It would be very useful to be alerted on policy violations.

### Proposed Behavior:
Send an alert when policy violations are discovered for a project.

rstmm:
Related to this, it would be helpful to define a responsible for a project in the project settings, who is getting alerts related to new vulnerabilities and policy violations in this project. Either by selecting a user or just entering a mail (which could be beneficial if there are is a mailing list for a project team for example).

A general contact for notifications of all projects (as it is now) is good too, but being able to notify the project manager or architect directly would be helpful.

stevespringett:
@rstmm Individual notification subscriptions is an excellent idea. Can you capture this in a new enhancement request please?

PR #1613, opened 2022-05-12
Added support for policy violation notification. Refactored Slack not…

stevespringett:
…ifications to support modern Slack apps. Added missing VEX consumed and processing support

Signed-off-by: Steve Springett <steve@springett.us>


stevespringett:
Resolves #1396

PR #1623, opened 2022-05-14
ci: improve pull request testing

k3rnelpan1c-dev:
### Description

A slight oversight on my previous PR resulted in pull requests only triggering the test-ci workflow, and while that sounds alright it ignores the fact that the test-ci wont build and thereby test the full build and the container builds.
This PR aims to fix this oversight.

### Changes

* only run login action and push when triggered on the default branch
* also trigger the build-ci workflow to test the general build and the container files

### Issues

* n/a


nscuro (approved):

nscuro:
Awesome, thank you!

Issue #1622, opened 2022-05-13
DefectDojo Integration Reimport Dependency Track Finding into Existing Test


yangsec888:
The enhancement may already be reported! Please search for the enhancement before creating one.

### Current Behavior:
Dependency Track (DT) has a background task for DefectDojo (DD) integration. The task will publish the DT finding into DD via its 'import-scan' API endpoint at a predefine frequency. Upon each push, a new test will be generated in DD.

### Proposed Behavior:
Add additional logic inside the same task, where the DT will try to determine if there is an existing test created from previous publishing. If no previous publishing found, then DT will create a new test via 'import-scan' endpoint. Otherwise, DT will rewrite the existing test via 'reimport-scan' endpoint instead. The new behavior will remove the burden of de-duplicating the numerous tests in the DD side, for the DD analyst.

yangsec888:
PR is now submitted at  https://github.com/DependencyTrack/dependency-track/pull/1631

PR #1631, opened 2022-05-16
Issue #1622 DefectDojo integration reimport enhancement

yangsec888:
Local dry run, with the following command work for me:

$ mvn clean package jetty:run -P enhance -DskipTests

Thank you Steve for the great project. Best regards, Sam Li


valentijnscholten:
The logic built into this PR is part of Defect Dojo standard functionality (find the latest test and reimport into that test): https://defectdojo.github.io/django-DefectDojo/integrations/importing/

If you specify an engagement_id and scan_type, it will find the latest test and will reimport into that.

src/main/java/org/dependencytrack/integrations/defectdojo/DefectDojoClient.java
```diff
@@ -68,4 +72,104 @@ public void uploadDependencyTrackFindings(final String token, final String engag
...
         }
     }
+
+    // Pulling DefectDojo 'tests' API endpoint, and retrieve a list of existing tests
+    public ArrayList getDojoTestIds(final String token) {
+        LOGGER.debug("Pulling DefectDojo Tests API ...");
+        String tests_uri = "/api/v2/tests/";
+        final UnirestInstance ui = UnirestFactory.getUnirestInstance();
+        LOGGER.debug("Make the first pagination call");
+        HttpResponse<JsonNode> response = ui.get(baseURL + tests_uri + "?limit=100")
```
valentijnscholten, line 82: This will retrieve ALL tests in defect dojo, which could be thousands. You can filter on the desired engagment by specifying a `engagement=<id>`. But as mentioned in my other comment, all this logic is not needed IIRC.

yangsec888:
> Not quite. I read the DefectDojo document and try to just replace the endpoint like you suggest. But it threw me exception, until I can provide the exact testID. My DefectDojo instance is running v2.7.1

valentijnscholten:
The API docs might be helpful. You do need a product name and engagemenr name.
https://github.com/DefectDojo/django-DefectDojo/blob/d55fb295e3eae8167af320c545445b2098422737/dojo/api_v2/views.py#L2063

src/main/java/org/dependencytrack/integrations/defectdojo/DefectDojoClient.java
yangsec888: Good point. I can add the optimization in like you suggest. Thank you.

yangsec888:
> 

Thank you for the link but I did read it many times over already. If you recall, the Dependency track integration is per project with a property data set of 4 key value pairs: Group Name - integration, Property Name - defectdojo.engagementID, Property Value - <defectDojo engagementID>, Property Type - String. There is no room to provide the additional information like you suggested here. Refer to https://docs.dependencytrack.org/integrations/defectdojo/

src/main/java/org/dependencytrack/integrations/defectdojo/DefectDojoClient.java
yangsec888: Change with suggested improvement is now in.

valentijnscholten:
You would indeed need to add some more properties, so might be to big a change for now.

nscuro:
If I understand correctly, this will be a breaking change in terms of behavior? 

There may be existing workflows that rely on new tests being created on every new upload.

yangsec888:
> 

I thought about that too. But I really double anyone with a clear mind would want to deal with hundreds of same tests (and keep on spamming with time) on the integration point. Definitely not in my shop where we already have a shortage labors in cybersec.  It seems to me most likely the original behavior is not completely thought through.

valentijnscholten:
I thought the same, but still most people do not use reimport in Defect Dojo (in general, not specific to DT).
The main reason is that they want to see over time which vulnerabilities were imported at which point in time.
I added an (re)import history function to reimport to help with that, but it's still basic.
The safest bet is to add a flag in DT to let the user choose between import and reimport?

yangsec888:
> I thought the same, but still most people do not use reimport in Defect Dojo (in general, not specific to DT). The main reason is that they want to see over time which vulnerabilities were imported at which point in time. I added an (re)import history function to reimport to help with that, but it's still basic. The safest bet is to add a flag in DT to let the user choose between import and reimport?

Agree on a flag / checkbox for user choice, if it's feasible. I thought about the option of checkbox before. But it comes down to the previous discussion - it might be too big a change for this PR. Because the current integration model doesn't provide the option / space without major revamp. In the current model, in order to preserve historical data on vulnerabilities, the user does still have the option to manually import the testing via DD GUI one at a time. So that it's ok compromise to me.

What can be done in the future: image it's a perfect world, under the 'Project Properies' page, once the user click to add 'Create Property' button for new integration, the user would then have the choice to select which integration he/she want to add-on, then the custom integration template could pop-up, where the user have the choice to set up such flag. The user could also has the optionals to additional flags / behaviors, including but not limiting to 'active', 'verified', 'push_to_jira', 'close_old_findings' etc.. This model IMHO would be more user friendly and create the value in the long term. But it does require more works both upfront and during maintenance. So it could be a major feature improvement instead of regular PR.

valentijnscholten:
Directing users to manually export the vulnerabilities and import them into DD doesn't sound too good to me.
Adding a property to a project could be done without any code/UI changes as they are setup to be flexible/dynamic?
For example `import_mode` could be `import` or `reimport`. Or maybe a boolean `reimport` `True/False`.

nscuro:
Having a feature toggle on the project level via properties rather than globally in the settings sounds like a good option to me. That would allow for gradual migration to reimports if so desired by enabling it for projects on-by-one.

yangsec888:
> Having a feature toggle on the project level via properties rather than globally in the settings sounds like a good option to me. That would allow for gradual migration to reimports if so desired by enabling it for projects on-by-one.
Hmm I do not fully get this. Are you suggesting something to the screenshot below? Won't it be both too ugly and confusing for a regular user?
Screen Shot 2022-05-26 at 3 28 10 PM

nscuro:
I don't think it will make it any more confusing.

valentijnscholten:
There might be instances where users/admins just want all integrations with DD use import or reimport. A global property (or env variable) might be helpful to simply switch a whole instance from one mode to another.

yangsec888:
Code update in per the discussion. Please let me know if you have any further concern. Best @valentijnscholten  @nscuro

nscuro (approved):
LGTM. How about you, @valentijnscholten?

valentijnscholten:
LGTM, I would like that global setting which would dictate the default setting (import or reimport), but wouldn't mind if that's left for a future PR.

yangsec888:
> LGTM, I would like that global setting which would dictate the default setting (import or reimport), but wouldn't mind if that's left for a future PR.

@valentijnscholten @nscuro  The other PR with global reimport flag is now in at:
- backend  https://github.com/DependencyTrack/dependency-track/pull/1738 
- frontend https://github.com/DependencyTrack/frontend/pull/179

Please review and let me know if you have any further concern.Thanks for your patience.

PR #1646, opened 2022-05-20
fix: latest tag logic in meta workflow

### LINKED ISSUE #1644: Latest tag on dockerhub points to snapshot instead of latest stable
### Current Behavior:

On registry dockerhub registry:
`dependencytrack/apiserver:latest` image has digest `4d9653d477fe` which corresponds to `dependencytrack/apiserver:snapshot` image

[Reference](https://hub.docker.com/r/dependencytrack/apiserver/tags)

### Expected Behavior:

`dependencytrack/apiserver:latest` should point to latest stable version, namely `dependencytrack/apiserver:4.5.0` as of today

k3rnelpan1c-dev:
### Description

> this is the backport PR

This PR aims to fix and introduce the `latest` tag logic from my prior CI PRs, sorry for any inconvenience that this error may have caused.

### Issues

* fixes #1644 on default branch (the default branch PR is here #1645)

### DIFF
--- a/.github/workflows/_meta-build.yaml
+++ b/.github/workflows/_meta-build.yaml
@@ -6,6 +6,11 @@ on:
         required: false
         default: "snapshot"
         description: "the version that should be set/used as tag for the container image"
+      publish-container:
+        type: boolean
+        required: false
+        default: false
+        description: "publish and scan the container image once its build"
     secrets:
       registry-0-usr:
         required: true
@@ -26,11 +31,6 @@ jobs:
           java-version: '11'
           cache: 'maven'
 
-      - name: Overwrite Version in pom.xml
-        if: ${{ ! contains(inputs.app-version, 'snapshot') }}
-        run: |-
-          mvn versions:set -DnewVersion=${{ inputs.app-version }}
-
       - name: Build with Maven
         run: |-
           mvn clean
@@ -38,12 +38,16 @@ jobs:
           mvn clean -P clean-exclude-wars
           mvn package -Dmaven.test.skip=true -P enhance -P embedded-jetty -P bundle-ui -Dlogback.configuration.file=src/main/docker/logback.xml
           mvn clean -P clean-exclude-wars
+          mvn cyclonedx:makeBom
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v3.0.0
         with:
           name: assembled-wars
-          path: target/*.jar
+          path: |-
+            target/*.jar
+            target/bom.json
+            target/bom.xml
 
   build-container:
     runs-on: ubuntu-latest
@@ -78,7 +82,7 @@ jobs:
 
       - name: Login to Docker.io
         uses: docker/login-action@v2.0.0
-        if: ${{ github.event_name == 'push' }}
+        if: ${{ inputs.publish-container }}
         with:
           registry: docker.io
           username: ${{ secrets.registry-0-usr }}
@@ -102,12 +106,12 @@ jobs:
             COMMIT_SHA=${{ github.sha }}
             WAR_FILENAME=dependency-track-${{ matrix.distribution }}.jar
           platforms: linux/amd64,linux/arm64
-          push: ${{ github.event_name == 'push' }}
+          push: ${{ inputs.publish-container }}
           context: .
           file: src/main/docker/Dockerfile
 
       - name: Run Trivy Vulnerability Scanner
-        if: ${{ github.event_name == 'push' }}
+        if: ${{ inputs.publish-container }}
         uses: aquasecurity/trivy-action@0.3.0
         with:
           image-ref: docker.io/dependencytrack/${{ matrix.distribution }}:${{ inputs.app-version }}
@@ -117,7 +121,7 @@ jobs:
           vuln-type: 'os'
 
       - name: Upload Trivy Scan Results to GitHub Security Tab
-        if: ${{ github.event_name == 'push' }}
+        if: ${{ inputs.publish-container }}
         uses: github/codeql-action/upload-sarif@v2
         with:
           sarif_file: 'trivy-results.sarif'
--- a/.github/workflows/_meta-build.yaml
+++ b/.github/workflows/_meta-build.yaml
@@ -84,12 +84,19 @@ jobs:
           username: ${{ secrets.registry-0-usr }}
           password: ${{ secrets.registry-0-psw }}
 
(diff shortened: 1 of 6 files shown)


nscuro:
https://github.com/DependencyTrack/dependency-track/blob/65208029a5b2b321146195c186980984c8599fdc/.github/workflows/_meta-build.yaml#L98

Means releases from the `4.5.x` branch will never be pushed, no?

k3rnelpan1c-dev:
good catch!

.github/workflows/_meta-build.yaml
```diff
@@ -78,29 +78,36 @@ jobs:
...
-            docker.io/dependencytrack/${{ matrix.distribution }}:${{ inputs.app-version }}
-            docker.io/dependencytrack/${{ matrix.distribution }}:latest
+          tags: ${{ steps.tags.outputs.tags }}
           build-args: |-
             APP_VERSION=${{ inputs.app-version }}
             COMMIT_SHA=${{ github.sha }}
             WAR_FILENAME=dependency-track-${{ matrix.distribution }}.jar
           platforms: linux/amd64,linux/arm64
-          push: ${{ github.ref == 'refs/heads/master' }}
+          push: ${{ github.event_name == 'push' }}
```
nscuro, line 105: Sorry for all the nitpicking, but now I think we may push to `snapshot` from `4.5.x` *and* `master`. 

What we need is something along the lines of:
* `master`: Push releases and snapshots
* `x.x.x`: Push releases only

I feel like it'd make sense to adopt something similar like we discussed in https://github.com/DependencyTrack/frontend/pull/141, which is creating releases based on tags. Thoughts?

k3rnelpan1c-dev: you are not nitpicking, you are reviewing, and I am thankful for that!

you are in fact correct that would happen + another flaw would be if my assumption that the event name gets passed to the meta workflow would not hold true.

nscuro (changes requested):
Marking as "Request Changes" so it doesn't get merged accidentally.

k3rnelpan1c-dev:
Okay, tiny problem with the separated workflows: No matter how I change it using the default `GITHUB_TOKEN` cannot be used to realise it :frowning_face: 

> _Note on GitHub Actions:_ You can use the default token which is provided in  the secret _GITHUB_TOKEN_. However releases done with this token will NOT trigger release events to start other workflows.
> If you have actions that trigger on newly created releases, please use a generated token for that and store it in your repository's secrets (any other name than GITHUB_TOKEN is fine).

So to say that I am stuck with that idea is quite accurate, unless we opt to setup a "Dependency-Track-Bot" account that will serve as PAT donor to create the release and commits to the projects on release.

*re-posting here so we can discuss in an open PR instead of a closed one*

nscuro:
Hmmm... But [triggering on new tags](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs) should work.

Creating a tag should be all that's needed to kick off the release pipeline. Once the tag is created, all following workflows, jobs and steps operate on the correct commit.

k3rnelpan1c-dev:
Just like regular commits that are done by the 'GitHub Actions Bot' don't trigger regular push events, the same applies to tag pushes unfortunately :disappointed:

nscuro:
Damn. But yeah, makes sense. 😅

How about this:
 - We decouple `prepare-release` to be run locally for now
 - This can be done with `mvn release:prepare` which will set the appropriate release version, create and commit a tag, and set a new SNAPSHOT version
 - Let the workflow trigger on new tags

We can still create a dedicated bot user and add a workflow that runs `mvn release:prepare` for us later.

k3rnelpan1c-dev:
> Damn. But yeah, makes sense. sweat_smile
> 
> How about this:
> 
>     * We decouple `prepare-release` to be run locally for now
> 
>     * This can be done with `mvn release:prepare` which will set the appropriate release version, create and commit a tag, and set a new SNAPSHOT version
> 
>     * Let the workflow trigger on new tags
> 
> 
> We can still create a dedicated bot user and add a workflow that runs `mvn release:prepare` for us later.

NGL, I had a similar thought, but not including any Maven release plugin :sweat_smile: 

We can still have 2 separate workflows as we would have (and what I already have ready in my fork, but on a different branch).

1. creates the GitHub Release and does the work to branch off where needed
2. assembles the `war` files and container to then upload to the GitHub Release

Gained benefits: 
* you could still do the first part manually if you really wanted to, however the build would never involve an environment outside of GitHub.
* you don't need a separate PAT to trigger the second workflow

Lost benefits:
* you would have to trigger 2 workflows manually to finish a release

nscuro:
Thank you for putting in all this work @k3rnelpan1c-dev. 

I took a gander around your fork (the state on `master` looked like it's the most current). Few questions:

* You currently have `ci-publish` being triggered on new releases. Given the limitations you described earlier, I assume this works because you use a bot user w/ PAT to create  releases?
* If we don't use a bot user to create a release, we'd have to run `ci-publish` manually from the release branch, which would be fine from my POV. However, because `ci-release` commits a new `SNAPSHOT` version after creating the release, we'd then build and publish a commit "after" our actual release commit. So we'd be building a `SNAPSHOT`?

k3rnelpan1c-dev:
No worries! My motto: I caused the mess, I clean it up.

Yes the `master` in my fork is where I test to be as close to the real thing as I can be.

1. Yes the `ci-publish` workflow works cause I use my bot-user's PAT.
2. not quite, if I where to alter the workflow to run via a manual trigger as well, then you would trigger the `ci-publish` workflow with the `git-tag` to publish from. Thereby it would check out the exact tag the release was bound to and update the release that would be associated with that exact tag.

nscuro:
Thanks for clarifying, sounds good to me.

k3rnelpan1c-dev:
Okay, then I shall alter the changes as we discussed and get them into this PR.

> note to self be sure to update the docs xD

k3rnelpan1c-dev:
so this and #1645 are now ready for review.
I ran a couple of test scenarios on my fork again and so far was unable to find any glaring problems.

I tried to update the workflow documentation to the best of my ability, but feedback is very welcome :slightly_smiling_face:

nscuro:
Thanks so much @k3rnelpan1c-dev! Looks good so far, I'll play around with this later today on my fork.

nscuro (approved):

PR #1660, opened 2022-05-25
Run Tests CI only for pushes or PRs to default- or release branches

nscuro:
The workflow is currently run twice for PRs. Tests take quite a while, and we really only care about tests on our main branches.

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -2,9 +2,15 @@ name: Tests CI
 
 on:
   push:
+    branches:
+      - 'master'          # Default branch
+      - '[0-9]+.[0-9]+.x' # Release branches
     paths-ignore:
       - '**/*.md'
   pull_request:
+    branches:
+      - 'master'          # Default branch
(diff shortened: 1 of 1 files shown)


k3rnelpan1c-dev:
May I use this opportunity to ask: would a DB service container help for test performance, or is it more the NIST mirroring that holds back the tests? (*I haven't actively looked at the test cases yet, so I am just guessing)

nscuro:
We use the in-memory database H2 for tests right now. There is also no NIST mirroring going on during tests. The time it takes is mostly due to the sheer amount of tests that create and destroy servlet contexts and re-initialize the in-memory database.

The latter will be sped up a bit with https://github.com/stevespringett/Alpine/pull/387 and https://github.com/DependencyTrack/dependency-track/pull/1652.

k3rnelpan1c-dev:
I see! Not sure if exchanging the DB for a service DB container would be much of a help then. Additionally, this would alter the tests to integration tests more than pure unit tests in a way.
Thank you for the context :+1:

nscuro:
Keep them improvement ideas comin', we're thankful for all of them!

PR #1663, opened 2022-05-27
Change NuGet package metadata URL

AbdelHajou:
According to the API index at https://api.nuget.org/v3/index.json, the registration3 endpoint which was used for retrieving package version metadata has been discontinued. This wasn't a big issue for Dependency-Track because this endpoint is only used for retrieving the publish timestamp of the package version. This PR just fixes the failing test and makes sure metadata can be correctly retrieved for newer package versions. The registration5-semver1 is backwards compatible, so this shouldn't cause any issues.


AbdelHajou:
After some more testing I realized these endpoints should be determined dynamically. This makes it more future-proof and allows support for different package repositories. For example, NuGet.org uses registration5-semver1 as registration URL and Azure Artifacts uses registrations2 or registrations2-semver2. These endpoints can be retrieved using the v3/index.json file.

Closing this PR and will provide a new PR soon.

PR #1675, opened 2022-05-30
Automatically discover NuGet server endpoints

AbdelHajou:
The NugetMetaAnalyzer currently assumes default hard-coded query endpoints that used to work with an older version of NuGet.org. The registration3 endpoint is now discontinued and no longer works for newer package versions. I've changed this default version to the endpoint used by NuGet.org (registration5-semver1) and added a mechanism for automatically determining the endpoints to use based on the index.json file of the NuGet repository. This allows support for different NuGet package repositories, like Azure Artifacts and Artifactory.

### DIFF
--- a/src/test/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzerTest.java
+++ b/src/test/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzerTest.java
@@ -22,16 +22,11 @@ import com.github.packageurl.PackageURL;
 import org.dependencytrack.model.Component;
 import org.dependencytrack.model.RepositoryType;
 import org.junit.Assert;
+import org.junit.Test;
 
 public class NugetMetaAnalyzerTest {
 
-    //@Test
-    /*
-     * Noticed that Microsoft NuGet is broken for package 3.13.0, which as of this moment, is the latest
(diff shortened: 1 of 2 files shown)


AbdelHajou:
Note, some package repositories aren't supported yet because they require authentication. I'm working on this and will provide a separate PR.

src/test/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzerTest.java
```diff
@@ -22,16 +22,11 @@
...
 public class NugetMetaAnalyzerTest {
 
-    //@Test
-    /*
-     * Noticed that Microsoft NuGet is broken for package 3.13.0, which as of this moment, is the latest
-     * version available. However, the registration3 API returns a 404 when making queries for it and the call
-     * to index.json for the nuget package returns 3.12.0 as the latest version. This is breaking DT unit tests.
-     * Commenting out for now.
-     */
+    @Test
```
nscuro, line 29: This test doesn't cover the new `initializeEndpoint` functionality. It'd be great if we had another test that covers it.

nscuro (changes requested):
Thanks for the PR, @AbdelHajou. Please add a test that covers the new functionality, then we're good to go!

PR #1687, opened 2022-06-05
Add support for private package repositories

### LINKED ISSUE #881: Suport for Repository Authentication

### LINKED ISSUE #1321: Repository: Add api key / username password support
Hey. First of all. Thank you for creating Dependency-Track, it has for certain made visualizing security risks across IT-fields and security, a different ballgame; Going from opinions, to actual facts makes the conversations and solving the issues more goal oriented.

Secondly. I'm not entirely sure how important it is to use the package repository directly or if it's just the meta files that are the goal. Still learning the ropes in dependency-track, so it's highly likely I've misunderstood something.

### Current Behavior:
Currently the repository assumes any repository does not require username and password. This is a problem if you want to use an internal package repository that require permission. The internal package repository, might be an internal package proxy.

### Proposed Behavior:
Add username and password to the Create / Edit repository view.

### Why is this a feature that we're looking for?

Having to make sure that every source is available for dependency-track, instead of just pointing to the same package repository source as the developers use, makes it an extra point of failure. Also there's the matter of the network infrastructure in an enterprise. It's preferable that all package repository requests goes through the same package proxy (in our case ProGet and Artifactory).

### LINKED ISSUE #1618: Add support for private Azure Artifacts packages
### Current Behavior:
The system only supports public package repositories. When scanning for outdated components, private components are ignored and are not shown as outdated. 
image

### Proposed Behavior:
My company uses Azure Artifacts to store private packages. Azure Artifacts is a universal package repository, meaning it can store packages for multiple ecosystems (currently nuget, npm, maven and python). Unfortunately there is no standardized authentication protocol for private package repositories, otherwise it would be possible to implement this functionality for every private package repository. However, I think Azure Artifacts would be a great first step. 

Azure Artifacts can be accessed through the Azure DevOps REST API. For access the user would need to provide a Personal Access Token with the "Packaging (read)" scope. OAuth 2.0 authorization is supported as well, but for this the user would need to configure OIDC integration with Azure AD as well. 

One limitation of the API is that it doesn't support querying for packages by name, which is what Dependency-Track currently does for all other package repositories. Instead there is an endpoint for querying all packages within an Azure DevOps organization. The existing RepositoryMetaAnalyzer scans each package one by one, which would not be possible for this API. A workaround for this would be to add a task that retrieves the latest versions for packages periodically and stores these in the database. The RepositoryMetaAnalayzer could then query a single component via the database.

The implementation for this should be made as generic as possible so that it would be possible for other contributors to add support for other (private) package repositories. If you agree that this feature would be an enhancement to the project, I would like to work on implementing it. I've already developed a prototype that demonstrates that it's feasible.

AbdelHajou:
This PR adds support for adding private package repositories that require Basic authentication. Dependency-Track users can now specify a username and password when adding an internal package repository. The username and password will then be added to the Authorization header for every request to a private package repository.

Repository passwords are encrypted using Alpine's DataEncryption module. After the user has set the password, it will only be decrypted right before making a request to a package repository. Users can not read/edit a repository password, it can only be overwritten.

I've mainly tested this feature using a private Azure Artifacts NuGet repository. While testing, I ran into two other issues:
1. The performVersionCheck tried to retrieve the latest version from NuGet by getting the last element in the versions array. However, Azure Artifacts does not sort this array like NuGet.org does. I changed the method to find the latest version using version comparison.
2. The date format performLastPublishedCheck uses does not match the date format Azure Artifacts returns. I now try parsing the date with multiple formats until the correct format is found.

I was unsure how to add integration tests for this feature, since it requires setting up a private package repository specifically for these tests. Tips on how to automatically test this feature are more than welcome.

Front-end changes are in DependencyTrack/frontend#164

This PR closes #881 and duplicates of this issue:
closes #1321 , closes #1618 

Signed-off-by: AbdelH <Abdel.Hajou@infosupport.com>


nscuro:
Awesome PR, @AbdelHajou! 🚀

> I was unsure how to add integration tests for this feature, since it requires setting up a private package repository specifically for these tests.

You can use `MockServer` instead, like it's done in https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/integrations/fortifyssc/FortifySscClientTest.java. It should allow you to test that basic auth headers are sent, but also that `NugetMetaAnalyzer` behaves as expected when encountering the different timestamp formats.

AbdelHajou:
@nscuro I added an integration test that tests the authentication, the timestamp parsing and the endpoint discovery

nscuro (approved):

PR #1697, opened 2022-06-09
Tweak metrics update efficiency

nscuro:
Continuation of #1481. Relates to #1210.

---

This PR contains a refactoring of the `MetricsUpdateTask` to improve its resource efficiency.

Calculating portfolio metrics in general is more I/O than CPU heavy and benefits greatly from parallelization. For this to work out OK, the footprint of the work being parallelized should be as slim as possible in order to keep the impact on overall resource consumption manageable. Speed of metrics updates also mainly depends on how beefy the database server is.

I investigated on doing all this entirely in the database using views and stored procedures, but ultimately decided that it'd be too much of a hassle to maintain and test. It would've been feasible if we only supported a single database, but that is not the reality today. 

Changes in this PR come down to:

* Usage of raw SQL queries and projection classes instead of using the JDO API
  * ✅ We have full control over what queries get executed which *is* more verbose, but also allows for further tuning
  * ✅ We save the time spent by DataNucleus to compile JDOQL to SQL and maintain its cache
  * ✅ We select only fields we need, reducing memory allocations
* The savings in resource usage made it feasible to parallelize project metrics calculation
  * ✅ We use a temporary thread pool of size up to `{cpuCoreCount}`, ensuring that we don't block event processing
    * I was not able to find best practices for how large the pool should be. We can tweak this depending on user feedback, and eventually make it configurable later.
* Introduction of integration tests w/ testcontainers using all databases we support to ensure that queries are functional
  * ⚠️ Running all tests now requires Docker

To clarify: This *will* increase total CPU usage due to introduction of thread pools.

---

Test setup identical to that outlined in #1481. The following test class was profiled:

```java
public class MetricsUpdateTaskBenchmarkTest {

    @Test
    public void test() {
        final Properties jdoProps = JdoProperties.get();
        jdoProps.setProperty(PropertyNames.PROPERTY_CONNECTION_URL, "jdbc:postgresql://localhost:5432/dtrack");
        jdoProps.setProperty(PropertyNames.PROPERTY_CONNECTION_DRIVER_NAME, org.postgresql.Driver.class.getName());
        jdoProps.setProperty(PropertyNames.PROPERTY_CONNECTION_USER_NAME, "dtrack");
        jdoProps.setProperty(PropertyNames.PROPERTY_CONNECTION_PASSWORD, "dtrack");

        final var pmf = (JDOPersistenceManagerFactory) JDOHelper.getPersistenceManagerFactory(jdoProps, "Alpine");
        PersistenceManagerFactory.setJdoPersistenceManagerFactory(pmf);

        new MetricsUpdateTask().inform(new MetricsUpdateEvent(MetricsUpdateEvent.Type.PORTFOLIO));
    }

}
```

CPU samples and memory allocations with current `master` (ran for 2m30s):

Screenshot 2022-06-09 at 20 57 48

Screenshot 2022-06-09 at 20 58 25

With this PR (ran for 30s):

Screenshot 2022-06-09 at 20 51 33

Screenshot 2022-06-09 at 20 52 18

For some reason running the code in IntelliJ is significantly slower than running it in a container.
Using a Docker Compose setup with Postgres, metrics for a portfolio with 1000 projects (1460000 total components) completed in 01m28s. Granted, this is with pretty much no network latency and flash storage.

---

I don't consider this to be the ultimate solution to improving metrics updates. But it's yet another step towards making it less of a pain point for large portfolios.


nscuro:
*Sigh* I missed to sign-off 3 commits and now Git is throwing merge conflicts at me when rebasing... I'll have to recreate this PR.

PR #1699, opened 2022-06-09
Include component repository metadata in component API

AbdelHajou:
I'm developing an application that gives upgrade advice based on Dependency-Track results. I noticed that the project components endpoint returns a list of all components with their repository metadata (latest version), but the endpoint for a specific component doesn't include the repo metadata. The API for a specific component should be at least as detailed (if not more) as the endpoint for a list of components.

Signed-off-by: AbdelH <Abdel.Hajou@infosupport.com>

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
@@ -40,6 +40,8 @@ import org.dependencytrack.model.Component;
 import org.dependencytrack.model.ComponentIdentity;
 import org.dependencytrack.model.License;
 import org.dependencytrack.model.Project;
+import org.dependencytrack.model.RepositoryMetaComponent;
+import org.dependencytrack.model.RepositoryType;
 import org.dependencytrack.persistence.QueryManager;
 import org.dependencytrack.util.InternalComponentIdentificationUtil;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
```diff
@@ -118,7 +120,11 @@ public Response getComponentByUuid(
             if (component != null) {
                 final Project project = component.getProject();
                 if (qm.hasAccess(super.getPrincipal(), project)) {
+                    final PackageURL purl = component.getPurl();
+                    final RepositoryMetaComponent metaComponent = qm.getRepositoryMetaComponent(
+                            RepositoryType.resolve(purl), purl.getNamespace(), purl.getName());
```
nscuro, line 125: That will raise a NPE when `purl` is `null`. Not every component has a PURL, some may have CPE, SWID tag or none of them at all. There should be a test to cover this scenario as well.

AbdelHajou:
@nscuro I've found a workaround for the problem I was having, so this functionality is no longer needed for me. If you still think it would be an enhancement to the API I will process your feedback, otherwise we can close this PR.

nscuro:
@AbdelHajou Thanks for the update. I don't think there's a need for this otherwise. It seems like `repositoryMeta` was more of a workaround (thus it being `transient`) and not intended to be used outside of the component listing in a project.

PR #1702, opened 2022-06-10
Mattermost Publisher support

stephan-strate:
Hey, I added support to send notifications to a Mattermost webhook.
We would be happy to see it being integrated into Dependency Track. Thanks in advance.
Looking forward to your review :)


src/main/resources/templates/notification/publisher/mattermost.peb
```diff
@@ -0,0 +1,5 @@
+{
+    "username": "Dependency Track",
```
nscuro, line 2: Should this be configurable? Will a user with name `Dependency Track` have to exist, or is this just a display name and can be arbitrary? I'm not familiar with Mattermost.

stephan-strate: This is just a display name. We could also not define it at all and let the webhook configure it.

nscuro:
Thanks for the PR @stephan-strate!

Issue #931, opened 2021-02-11
Support for Google OSV database


sameer1046:
### Current Behavior:
The tool doesn't have support for Google open source vulnerability database 
https://osv.dev/list

### Proposed Behavior:
The tool should be able to get data from Google osv database.

stevespringett:
Good suggestion. It's unfortunate osv doesn't support [Package URL](https://github.com/package-url/purl-spec). I'd recommend creating a ticket for them to support it. Relying on 'name' and 'ecosystem' is fragile and going to be unpredictable. Once osv supports Package URL, I'd be happy to add support for it.

VinodAnandan:
https://github.com/google/osv/issues/64

VinodAnandan:
@stevespringett OSV now supports PURL. Please find the below links for more details 

https://security.googleblog.com/2021/06/announcing-unified-vulnerability-schema.html
https://tinyurl.com/vuln-json
https://osv.dev/docs/#operation/OSV_QueryAffected

VinodAnandan:
`curl -X POST -d '{"package": {"purl": "pkg:maven/org.apache.struts:struts2-core"}, "version": "2.0.14"}' https://api.osv.dev/v1/query`

pombredanne:
@VinodAnandan FWIW, `pkg:maven/org.apache.struts:struts2-core` may not be the proper purl for this package, but rather `pkg:maven/org.apache.struts/struts2-core`

VinodAnandan:
Thanks @pombredanne

pombredanne:
@VinodAnandan sure thing! ... note that when VulnerableCode will be ready for usage, it will also have OSV content and everything is keyed by Package URL there... so this could be come a natural additional data source for DT

VinodAnandan:
The purl issue has been fixed. Thanks to @oliverchang

"` curl -X POST -d '{"package": {"purl": "pkg:maven/org.apache.struts/struts2-core@2.0.14"}}' https://api.osv.dev/v1/query `"

sahibamittal:
OSV maintains GCS bucket with all aggregated vulnerabilities which can be scheduled and mirrored in dependency-track database https://github.com/google/osv/blob/master/README.md#data-dumps
I'd like to work on it.

sahibamittal:
Draft PR for discussion -> https://github.com/DependencyTrack/dependency-track/pull/1703

nscuro:
[...]

Issue #1642, opened 2022-05-19
Add support for CDX vulnerability references


stevespringett:
### Current Behavior:
Currently, each vulnerability is unique by its source and identifier. However, different sources have different identifiers for the same vulnerability. This leads to duplicate vulnerabilities, increased risk scores, and extra effort in auditing. 

### Proposed Behavior:
CycloneDX supports vulnerability references, which are aliases that describe the same vulnerability across multiple sources of vulnerability intelligence. Dependency-Track should adopt this model and be able to:

* present the vulnerability data in its de-duped form (with aliases that cross-reference).
* take aliases into consideration with performing risk scores
* provide aliases on all REST endpoints that provide vulnerability data (including Findings API)

msymons:
* How will alerts be handled?  I am assuming that (say) NEW_VULNERABILITY will be linked to the CDX Vulnerability Reference,  thus helping addressing one aspect of the current alert spam?
* Will "Attributed On" timestamps still be tracked for the individual sources of vulnerability intelligence?  Once the aliasing is in place I think that this might allow the potential for the enterprising API user to pull data that could show (say) what percentage of the time different analysers were the first to identify a specific vulnerability. 
* Taking aliases into consideration when performing risk scores will be fantastic.  I have two Dependency-Track deployments with identical projects.  One server has GHSA enabled and one does not.  The GHSA one has a portfolio risk score that is 10,000 higher than the other.  This enhancement will drop that delta by a large amount... and then the remaining difference should reflect the vulns that only GHSA intelligence is currently successfully  identifying.  
* But what will happen when different analysers give different severity scores? As of today,  CVE-2022-25647 has a score of 7.7 on GHSA and the originating CNA (Snyk),  but a score of 7.5 on NVD.
* Will internal vulnerabilities (introduced in v4.4.0) be "aliasable"?  ie manually identify "CVE-XXX" and DT then creates alias to matching CDX vulnerability ID.
* How will auditing be handled in general?
* How will upgrade to DT implementing this enhancement handle existing audit decisions?  Some existing vulns might be suppressed simply as a way to handle the duplication that this enhancement is intended to address.   In such a situation, one would not want to have the new aliases accidentally suppressed.

stevespringett:
Support for aliasing for GHSA is working upon mirroring of the data. The next step is to de-dupe and update scoring logic. Aliases are already provided in the REST API.

[...]

PR #1703, opened 2022-06-10
Issue #931 : Support for Google OSV

sahibamittal:

Issue: https://github.com/DependencyTrack/dependency-track/issues/931

Google OSV Repo -> https://github.com/google/osv

Frontend Changes: https://github.com/DependencyTrack/frontend/pull/170

1. Download vulnerabilities from Google OSV data dumps (via https https://github.com/google/osv#data-dumps). 
NOTE: Alternative way is to use Google client library to retrieve from bucket. 
2. Extract json files, parse and map data to dependency-track vulnerability objects.
3. Save in database using query manager.

**TBD**: Mapping of OSV data to dependency-track data.
In Github advisory, we receive version range object whereas, in google OSV, we get an array of versions which can be used as range but on comparing (say, for ghsa id GHSA-vvw4-rfwf-p6hx), lower/upper ranges are bit different.

Objects from Github Advisory: https://docs.github.com/en/graphql/reference/objects#securityvulnerability
Objects from Google OSV: https://osv.dev/list?page=21&ecosystem=Maven

Example for difference between github and google for GHSA-vvw4-rfwf-p6hx :  
Github: https://github.com/advisories/GHSA-vvw4-rfwf-p6hx
Google: attachment json 
[GHSA-vvw4-rfwf-p6hx.json.zip](https://github.com/DependencyTrack/dependency-track/files/8879154/GHSA-vvw4-rfwf-p6hx.json.zip)


sahibamittal:
@nscuro

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    private VulnerableSoftware mapVulnerabilityToVulnerableSoftware(final QueryManager qm, final OSVVulnerability vuln) {
+
+        List<String> versions = vuln.getVersions();
+        String versionStartIncluding = null;
+        String versionEndIncluding = null;
+        if (versions != null) {
+            versionStartIncluding = versions.get(0);
+            versionEndIncluding = versions.get(versions.size() - 1);
```
nscuro, line 159: Looking at the example advisory GHSA-vvw4-rfwf-p6hx from OSV that you posted, it seems like we can't rely upon the fact that the latest included version is always the last in the array. 

```json
    "versions": [
        "9.0.0.M1",
        "9.0.0.M10",
        "9.0.0.M11",
...
        "9.0.37",
        "9.0.38",
        "9.0.4",
        "9.0.5",
        "9.0.6",
        "9.0.7",
        "9.0.8"
      ],
```

The possibility of there being arbitrary suffixes (and possibly prefixes) makes sorting a little tricky. 

It may be easier to use the `ranges` object instead. But can we rely on it being always present?

```json
    "ranges": [
        {
          "type": "ECOSYSTEM",
          "events": [
            {
              "introduced": "9.0.0-M1"
            },
            {
              "fixed": "9.0.39"
            }
          ]
        }
    ],
```

`introduced` would map to `versionStartIncluding` and `fixed` to `versionEndExcluding`.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,199 @@
...
+        vs.setVulnerable(true);
+        vs.setPurlType(vuln.getPackageEcosystem());
+        vs.setPurlNamespace(vuln.getPackageName());
+        vs.setPurlName(purl);
+        vs.setVersionStartIncluding(versionStartIncluding);
+        vs.setVersionEndIncluding(versionEndIncluding);
+        return vs;
+    }
+
+//      WAY 2 : to use google storage library instead of downloading data from https
```
nscuro, line 178: Curious: What would be the benefit of using the library?

nscuro (commented):
The direction you're headed in looks good, @sahibamittal.

> [...] but on comparing (say, for ghsa id https://github.com/advisories/GHSA-vvw4-rfwf-p6hx), lower/upper ranges are bit different.

Can you elaborate a bit? I am not able to spot a difference in the ranges, but it's totally possible that I'm not looking close enough.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
sahibamittal: I found a case where for a affected package (for a purl), ranges have multiple introduced-fixed pairs like : 
`"ranges": [
        {
          "type": "ECOSYSTEM",
          "events": [
            {
              "introduced": "2.2"
            },
            {
              "fixed": "2.2.26"
            },
            {
              "introduced": "3.2"
            },
            {
              "fixed": "3.2.11"
            },
            {
              "introduced": "4.0"
            },
            {
              "fixed": "4.0.1"
            }
          ]
        }
      ]`

What should be versionStartIncluding and versionEndExcluding in this case?
@nscuro @VinodAnandan

sahibamittal: OSV's API methods need whole list of packages being used as input and returns same json with same ranges format.
Yes, we can create multiple VulnerableSoftware instances, one for each range for a vulnerability object. This seems like a reliable option for now.

stevespringett: @sahibamittal First off. I want to thank you for such a quality PR. Very much appreciated.

Were you able to create multiple VulnerableSoftware instances yet? One for each range?

stevespringett:
@nscuro Anything outstanding prior to merge?

nscuro:
@stevespringett I haven't had a chance to review the latest changes, and haven't tested the new functionality yet. I'll try to get it done tonight.

stevespringett:
It doesn't appear that authentication is required in order to download the OSV ecosystem data. In the configuration, OSV is disabled by default. 

Is this what the default should be? Are there reasons why OSV should not be enabled by default?

sahibamittal:
@nscuro I tried to run the container but it is showing 'Connection timed out' error at URL.openStream in OSVDownloadTask.
And so, it's not downloading the data. Do you know or faced this during github mirroring?

sahibamittal:
> It doesn't appear that authentication is required in order to download the OSV ecosystem data. In the configuration, OSV is disabled by default.
> 
> Is this what the default should be? Are there reasons why OSV should not be enabled by default?

Yes, it should be enabled by default. The property value for enabled flag is coming as true from front-end. But, I'll change default to true in backend as well.

nscuro:
> @sahibamittal I tried to run the container but it is showing 'Connection timed out' error at URL.openStream in OSVDownloadTask. And so, it's not downloading the data. Do you know or faced this during github mirroring?

Are you connected to a corporate network of some sorts? Those networks typically require usage of a proxy to be able to reach external hosts. I'm not super sure if `URL.openStream` honors system proxy configuration, but I assume it doesn't.

It may be a good idea to use [`HttpClientPool`](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/common/HttpClientPool.java) to acquire a HTTP client that does honor proxy configuration. You can find an example usage in `NistMirrorTask`:

https://github.com/DependencyTrack/dependency-track/blob/0aa65172c37d3b299644f158a432003e808ad0db/src/main/java/org/dependencytrack/tasks/NistMirrorTask.java#L211-L218

You may need to then configure the proxy, see "Proxy Configuration" section at https://docs.dependencytrack.org/getting-started/configuration/

For running the entire DT application, you don't need to build the container. I'd recommend to launch it via Maven, see: https://github.com/DependencyTrack/dependency-track/blob/a5037fa42e73343cab513df14af5475cbc67896c/DEVELOPING.md#debugging

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,174 @@
...
+    private Vulnerability mapAdvisoryToVulnerability(final QueryManager qm, final OSVAdvisory advisory) {
+
+        final Vulnerability vuln = new Vulnerability();
+        vuln.setSource(Vulnerability.Source.GOOGLE);
+        vuln.setVulnId(String.valueOf(advisory.getId()));
+        vuln.setTitle(advisory.getSummary());
+        vuln.setDescription(advisory.getDetails());
+        vuln.setPublished(Date.from(advisory.getPublished().toInstant()));
+        vuln.setUpdated(Date.from(advisory.getModified().toInstant()));
+
```
nscuro, line 118: There are some more fields we could ingest, e.g. the CVSSv3 vector and credits.

nscuro (changes requested):

nscuro:
One issue I noticed is that GHSA vulnerabilities are duplicated. We'll have vulnerabilities with `vulnId` GHSA-* with `source=GITHUB`, and `source=GOOGLE`. As a consequence, the internal analyzer will report such GHSA vulns twice. The OSS Index scanner does not suffer from this, b/c it checks for existing CVEs:

https://github.com/DependencyTrack/dependency-track/blob/0aa65172c37d3b299644f158a432003e808ad0db/src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java#L253-L254

@stevespringett, will #1642 help here?

stevespringett:
@nscuro #1642 will help to an extent, yes. Still not sure how we're going to visualize this though.

sahibamittal:
> One issue I noticed is that GHSA vulnerabilities are duplicated. We'll have vulnerabilities with `vulnId` GHSA-* with `source=GITHUB`, and `source=GOOGLE`. As a consequence, the internal analyzer will report such GHSA vulns twice. The OSS Index scanner does not suffer from this, b/c it checks for existing CVEs:
> 
> https://github.com/DependencyTrack/dependency-track/blob/0aa65172c37d3b299644f158a432003e808ad0db/src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java#L253-L254
> 
> @stevespringett, will #1642 help here?

Can we may be disable github advisories and make OSV default as vulnerability source, to avoid duplication? since github source is a subset of OSV.

nscuro:
A viable option for sure, especially since that would also remove the necessity of configuring a PAT for GHSA mirroring.

Few things to check:
* Does OSV contain *all* GitHub advisories, or only a subset?
* How quickly are GitHub advisories ingested into OSV? A significant delay (say >1 day) would be suboptimal.

We'll also need to think of the migration scenario: Folks will already have lots of GHSA vulns in their portfolio, and they'll have audited them as well. We can't just remove the "old" GHSA vulns when upgrading, as audit trails will be lost.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        Event.dispatch(new IndexEvent(IndexEvent.Action.COMMIT, Vulnerability.class));
+    }
+
+    private Vulnerability mapAdvisoryToVulnerability(final QueryManager qm, final OSVAdvisory advisory) {
+
+        final Vulnerability vuln = new Vulnerability();
+        vuln.setSource(Vulnerability.Source.GOOGLE);
+        vuln.setVulnId(String.valueOf(advisory.getId()));
+        vuln.setTitle(advisory.getSummary());
```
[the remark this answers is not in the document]
sahibamittal: OSV provides only summary and details fields (https://ossf.github.io/osv-schema/#summary-details-fields). It says "_It is recommended that this field be kept short, on the order of no more than 120 characters_" but its no hard rule so it can be longer. So yes, we can do either hard cut or customise the title.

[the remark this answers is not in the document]
sahibamittal: I'm making it hard cut to length 253 and then appending 2 dots, if length of summary is exceeding max length 255.

src/main/java/org/dependencytrack/parser/osv/GoogleOSVAdvisoryParser.java
```diff
@@ -0,0 +1,122 @@
...
+                String lower = event.optString("introduced", null);
+                if(lower != null) {
+                    osvVulnerability.setLowerVersionRange(lower);
+                    k += 1;
+                }
+                event = rangeEvents.getJSONObject(k);
+                String upper = event.optString("fixed", null);
+                if(upper != null) {
+                    osvVulnerability.setUpperVersionRange(upper);
+                    k += 1;
```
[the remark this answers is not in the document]
sahibamittal: I'll fix it now and will add tests for different ranges.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
sahibamittal: What would be credits here? OSV have credits with name and contact https://ossf.github.io/osv-schema/#credits-fields

stevespringett:
@nscuro to my knowledge, OSV only contains a subset of GHSA. They only include the "reviewed" vulnerabilities (approx 7500). They do not include the unreviewed vulnerabilities (>170K). In the future, I want to make a configurable option for the GitHub Advisories integration for enabling reviewed and unreviewed independently. This would allow folks to use OSV for all reviewed GHSAs and use the GitHub Advisory integration for all unreviewed vulnerabilities.

stevespringett:
Does OSV have any vulnerabilities that are specific to them? For example, Sonatype OSS Index, if what OSS Index found was a CVE, then the source is set to the NVD. If it was a sonatype specific finding, then the source is set to OSSINDEX.

We need to do the same here. If OSV finds a CVE, GHSA, etc, then the source needs to be set to the NVD and GITHUB accordingly, since they are the source of the vulnerability, not OSV. OSV is just acting as an interface.

But if OSV does have their own findings (which I think they do with OSS Fuzz), then OSV would only be the source for those things. This will also prevent the duplicates from occurring, and will prevent duplicate records being mirrored in the internal database.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
nscuro: A comma-separated list of names should suffice. Using credits is currently only being done for VulnDB, see

https://github.com/DependencyTrack/dependency-track/blob/a3b42a6fe4d19fc91122887dba0841f26fa7c7c4/src/main/java/org/dependencytrack/parser/vulndb/ModelConverter.java#L113-L131

I don't think we need the contact information and can safely discard it.

VinodAnandan:
OSV does have its own findings (e.g: https://osv.dev/vulnerability/OSV-2022-217 ) . OSV also supports alias field https://ossf.github.io/osv-schema/#aliases-field. I believe that the OSV data may contain more useful information ("PURLs, expand version ranges to a full list of affected versions, etc." ) @oliverchang please correct me if I'm wrong.

I think duplicate findings can be correlated and displayed as a unified finding from multiple sources? I think in this way, we can identify any data issues with a particular source.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public void updateDatasource(final OSVAdvisory advisory) {
+        LOGGER.info("Updating datasource with Google OSV advisories");
+        try (QueryManager qm = new QueryManager()) {
+
+            LOGGER.debug("Synchronizing Google OSV advisory: " + advisory.getId());
+            final Vulnerability synchronizedVulnerability = qm.synchronizeVulnerability(mapAdvisoryToVulnerability(qm, advisory), false);
+            final List<VulnerableSoftware> vsList = new ArrayList<>();
+            for (OSVVulnerability osvVulnerability: advisory.getVulnerabilities()) {
+                VulnerableSoftware vs = mapVulnerabilityToVulnerableSoftware(qm, osvVulnerability);
```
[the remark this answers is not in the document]
sahibamittal: Should we add another try-catch as its already inside try block?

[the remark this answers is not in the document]
nscuro: That'd be my suggestion, yes. 

The wrapping `try (QueryManager qm = new QueryManager()) {` block doesn't catch any exceptions, it just ensures that the `QueryManager` is closed. At the moment, an exception in `mapVulnerabilityToVulnerableSoftware` will bubble up to the try-catch block in `inform`. 

Say mapping a vulnerability of the 10th advisory causes an error. Then the entire task is effectively canceled right now.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    private void unzipFolder(ZipInputStream zipIn) throws IOException {
+
+        BufferedReader reader;
+        GoogleOSVAdvisoryParser parser = new GoogleOSVAdvisoryParser();
+        ZipEntry zipEntry = zipIn.getNextEntry();
+        while (zipEntry != null) {
+
+            reader = new BufferedReader(new InputStreamReader(zipIn));
```
nscuro, line 86: `reader` should be closed.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,202 @@
...
+                    LOGGER.info("Updating datasource with Google OSV advisories for ecosystem " + ecosystem.getValue());
+                    String url = "https://osv-vulnerabilities.storage.googleapis.com/" + ecosystem.getValue() + "/all.zip";
+                    request = new HttpGet(url);
+                    try (final CloseableHttpResponse response = HttpClientPool.getClient().execute(request)) {
+                        final StatusLine status = response.getStatusLine();
+                        if (status.getStatusCode() == 200) {
+                            try (InputStream in = response.getEntity().getContent()) {
+                                ZipInputStream zipInput = new ZipInputStream(in);
+                                unzipFolder(zipInput);
+                            }
```
nscuro, line 67: `zipInput` should be closed as well.

```suggestion
                            try (InputStream in = response.getEntity().getContent();
                                 ZipInputStream zipInput = new ZipInputStream(in)) {
                                unzipFolder(zipInput);
                            }
```

stevespringett:
> Will the GHSA mirror task override the vulnerabilities created by OSV, and the other way around?

That's a very good point... Since OSV is not the authoritative source for GHSA, I would expect the behavior of the mirror task to skip mirroring of them, if GHSA mirroring is enabled.

For example, it is possible that someone enabled GHSA in 4.5, and will disable it in 4.6 in favor of OSV. In this scenario, we still want to update existing GHSA vulnerabilities and add new ones since the system is no longer mirroring them.

But if GHSA is enabled, we want GitHub to be the authoritative source, so the OSV mirror task should skip mirroring them. This should be true for any non-authoritative source supported by OSV.

VinodAnandan:
IMO, we should give importance to the quality of data over the origin. Some of the sources enhance/enrich the data. We should capture the origin and source attributes separately, one vulnerability may have one origin and multiple sources. So even if a particular source updates the vulnerability data, we can identify it.

stevespringett:
I wouldn't know where to start in trying to determine quality of data without making a lot of assumptions.  If we assume OSV is always better, I don't know when or how often we'll be wrong in that assumption. DT currently does not assume that OSSI has better data than the NVD for CVEs, we simply do not try to mirror things that the NVD has already mirrored. Most people will see a CVE or GHSA and go directly to the source. If DT has different data than the source, they may question the reliability of our data. If capturing the data from multiple providers for the same vulnerability is desired, then that would be an additional enhancement we can work on.

VinodAnandan:
I think capturing the data from multiple sources is important, it may help to compare it between different sources. Some sources may have the correct component, version mapping and others may have a better severity, description, references, etc.

nscuro:
> I think capturing the data from multiple sources/providers is important, it may help to compare it between different providers

As Steve pointed out we should have this discussion in a separate issue. I feel like there are some implications here that need to be well thought through. If we decide to ingest as much (mostly redundant) data as possible, then that'll be a design decision that affects other existing sources as well.

For this PR, let's go with the current approach, without clashing with native NVD and GHSA mirroring.

sahibamittal:
A. **Update in logic of severity calculation**. OSV object structure has severity info at 4 levels. New logic will pick the severity on below priority:
1. Advisory/Vulnerability level CVSS vector
2. Advisory/Vulnerability level severity string value
3. Affected package level CVSS vector
4. Affected package level string value

In case 3,4 above, one Advisory/Vulnerability will have list of affected packages and thus, list of severities. It will then pick the highest severity.

B. **Regarding avoiding clashing with native NVD and GHSA mirroring**, if advisory ID is either GHSA or CVE based and is already existing in DT, that advisory will not be mapped to a new vulnerability in DT database. But in this case, we can compare the affected packages (/components/ranges) and add missing ones as vulnerable softwares under that existing vulnerability?

nscuro:
> But in this case, we can compare the affected packages (/components/ranges) and add missing ones as vulnerable softwares under that existing vulnerability?

Yes, that should work. This way we still profit from OSV's (potentially) improved matching without interfering with the GHSA mirror task.

src/main/java/org/dependencytrack/parser/osv/GoogleOSVAdvisoryParser.java
```diff
@@ -0,0 +1,207 @@
...
+                if(k < rangeEvents.length()) {
+                    event = rangeEvents.getJSONObject(k);
+                    String upper = event.optString("fixed", null);
+                    if(upper != null) {
+                        osvVulnerability.setUpperVersionRange(upper);
+                        k += 1;
+                    }
+                }
+                osvVulnerabilityList.add(osvVulnerability);
+            }
```
nscuro, line 148: This logic is currently causing infinite loops when mirroring the `Linux` or `OSS-Fuzz` ecosystems.

Take https://osv.dev/vulnerability/GSD-2021-1000677 as example:

```json
"ranges": [
    {
        "type": "GIT",
        "repo": "https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/",
        "events": [
            {
                "introduced": "a7d42ddb3099727f58366fa006f850a219cce6c8"
            },
            {
                "limit": "c757c1f1e65d89429db1409429436cf40d47c008"
            }
        ]
    }
],
```

`rangeEvents` is of length `2`. We enter the `while` loop, find the `introduced` property and increment `k` by `1`. It doesn't have a `fixed` property though, so `k` is never incremented to `2` to break the loop.

Another point is that Dependency-Track currently can not deal with commits instead of versions. I wonder if we should skip ranges of type `GIT`?

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,277 @@
...
+        if (purl == null) return null;
+
+        VulnerableSoftware vs = qm.getVulnerableSoftwareByPurl(vuln.getPurl(), versionEndExcluding, versionStartIncluding);
+        if (vs != null) {
+            return vs;
+        }
+        vs = new VulnerableSoftware();
+        vs.setVulnerable(true);
+        vs.setPurlType(vuln.getPackageEcosystem());
+        vs.setPurl(vuln.getPurl());
```
nscuro, line 249: Sorry I didn't spot this earlier, only noticed it while testing right now: Instead of setting the entire PURL, you need to set the PURL components separately, e.g.

```java
vs.setPurlType(vuln.getPackageEcosystem());
vs.setPurlNamespace(...);
vs.setPurlName(...);
```

Otherwise the `InternalAnalysisTask` won't pick them up as of now. Here's the code used by `InternalAnalysisTask` to query for applicable `VulnerableSoftware`s:

https://github.com/DependencyTrack/dependency-track/blob/a3b42a6fe4d19fc91122887dba0841f26fa7c7c4/src/main/java/org/dependencytrack/persistence/VulnerableSoftwareQueryManager.java#L197-L220

Here's how it's done for GHSA:

https://github.com/DependencyTrack/dependency-track/blob/a3b42a6fe4d19fc91122887dba0841f26fa7c7c4/src/main/java/org/dependencytrack/tasks/GitHubAdvisoryMirrorTask.java#L283-L285

nscuro (commented):
Some early feedback. Great work so far! But I'll need a little bit more time to complete the review and do some testing.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,277 @@
...
+    public void inform(Event e) {
+
+        if (e instanceof GoogleOSVMirrorEvent && this.isEnabled) {
+
+            for (Ecosystem ecosystem : Ecosystem.values()) {
+                LOGGER.info("Updating datasource with Google OSV advisories for ecosystem " + ecosystem.getValue());
+                String url = "https://osv-vulnerabilities.storage.googleapis.com/" + ecosystem.getValue() + "/all.zip";
+                request = new HttpGet(url);
+                request.setHeader("Accept", "*/*");
+                request.setHeader("User-Agent", "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.11 (KHTML, like Gecko) Chrome/23.0.1271.95 Safari/537.11");
```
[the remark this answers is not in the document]
sahibamittal: I added those to avoid the connection error in corporate network, its not required really, I'll delete it.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    public VulnerableSoftware mapVulnerabilityToVulnerableSoftware(final QueryManager qm, final OSVVulnerability vuln) throws MalformedPackageURLException {
+
+        String versionStartIncluding = vuln.getLowerVersionRange();
+        String versionEndExcluding = vuln.getUpperVersionRange();
+
+        if (vuln.getPurl() == null) return null;
+
+        PackageURL purl = new PackageURL(vuln.getPurl());
```
nscuro, line 241: OSV is providing invalid PURLs for scoped NPM components, causing lots of parsing errors right now

```
2022-06-30 20:07:57,627 INFO [OSVDownloadTask] Updating datasource with Google OSV advisories for ecosystem npm
2022-06-30 20:07:57,819 ERROR [OSVDownloadTask] Error while mapping vulnerable software pkg:npm/@hapi/hoek
com.github.packageurl.MalformedPackageURLException: The PackageURL name specified is invalid
	at com.github.packageurl.PackageURL.validateName(PackageURL.java:280)
	at com.github.packageurl.PackageURL.parse(PackageURL.java:541)
	at com.github.packageurl.PackageURL.<init>(PackageURL.java:68)
	at org.dependencytrack.tasks.OSVDownloadTask.mapVulnerabilityToVulnerableSoftware(OSVDownloadTask.java:241)
	at org.dependencytrack.tasks.OSVDownloadTask.updateDatasource(OSVDownloadTask.java:130)
	at org.dependencytrack.tasks.OSVDownloadTask.unzipFolder(OSVDownloadTask.java:102)
	at org.dependencytrack.tasks.OSVDownloadTask.inform(OSVDownloadTask.java:75)
```

I raised an issue for OSV here: https://github.com/google/osv/issues/473

For this PR though, we should handle parsing errors here. An invalid PURL  should be treated as if no PURL was provided at all (just like the `null` check above).

src/main/java/org/dependencytrack/parser/osv/GoogleOSVAdvisoryParser.java
```diff
@@ -0,0 +1,221 @@
...
+                OSVVulnerability osvVulnerability = createOSVVulnerability(vulnerability);
+                JSONObject event = rangeEvents.getJSONObject(k);
+                String lower = event.optString("introduced", null);
+                if(lower != null) {
+                    osvVulnerability.setLowerVersionRange(lower);
+                    k += 1;
+                }
+                if(k < rangeEvents.length()) {
+                    event = rangeEvents.getJSONObject(k);
+                    String upper = event.optString("fixed", null);
```
nscuro, line 156: We'll need to handle `limit` and `last_affected` events as well:

image

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,277 @@
...
+        reader.close();
+    }
+
+    public void updateDatasource(final OSVAdvisory advisory) {
+
+        try (QueryManager qm = new QueryManager()) {
+
+            LOGGER.debug("Synchronizing Google OSV advisory: " + advisory.getId());
+            // Set to avoid duplicates
+            Set<VulnerableSoftware> vsList = new HashSet<>();
```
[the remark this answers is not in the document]
sahibamittal: Deduplication is required in case the vulnerability with some affected components already exist and we get to parse same vulnerability from OSV but OSV one has few additional affected packages. We dont want to duplicate them.
But yes if Set is not helping here, we'll compare the coming new vulnerable software to any existing one based on its purl info.

src/main/java/org/dependencytrack/tasks/OSVDownloadTask.java
```diff
@@ -0,0 +1,284 @@
...
+        String versionEndIncluding = vuln.getUpperVersionRangeIncluding();
+
+        try {
+            PackageURL purl = new PackageURL(vuln.getPurl());
+            VulnerableSoftware vs = qm.getVulnerableSoftwareByPurl(purl.getType(), purl.getNamespace(), purl.getName(),
+                    versionEndExcluding, versionEndIncluding, null, versionStartIncluding);
+            if (vs != null) {
+                return vs;
+            }
+            if (vuln.getPurl() == null) return null;
```
nscuro, line 243: This `null` check should happen before `new PackageURL` is called. ATM all advisories without PURL cause an error to be logged.

```
2022-07-01 22:19:37,499 ERROR [OSVDownloadTask] Error while mapping vulnerable software null
com.github.packageurl.MalformedPackageURLException: Invalid purl: Contains an empty or null value
	at com.github.packageurl.PackageURL.parse(PackageURL.java:478)
	at com.github.packageurl.PackageURL.<init>(PackageURL.java:68)
	at org.dependencytrack.tasks.OSVDownloadTask.mapVulnerabilityToVulnerableSoftware(OSVDownloadTask.java:237)
	at org.dependencytrack.tasks.OSVDownloadTask.updateDatasource(OSVDownloadTask.java:125)
	at org.dependencytrack.tasks.OSVDownloadTask.unzipFolder(OSVDownloadTask.java:99)
	at org.dependencytrack.tasks.OSVDownloadTask.inform(OSVDownloadTask.java:72)
	at alpine.event.framework.BaseEventService.lambda$publish$0(BaseEventService.java:99)
```

nscuro:
As requested by @VinodAnandan, I made some changes to address some outstanding issues.

* Fix https://github.com/DependencyTrack/dependency-track/pull/1703#discussion_r912232154
* Prevent infinite loops when parsing version ranges
* Be a little bit more strict when parsing range events (always require an `introduced` event as specified in the spec)
* For GitHub advisories, handle `<` constraints in addition to `<=`
* Rename `OsvVulnerability` to `OsvAffectedPackage` to avoid confusion with other vulnerability types

What I noticed and did not address yet:

1. Once OSV and NVD mirroring completes for the first time (OSV takes ~20-30sec for me in total), there's a massive performance degradation in the OSV mirror task on following executions. The culprit being the `VulnerabilityQueryManager#getVulnerabilityByVulnId` method. More specifically the fetching of associated `VulnerableSoftware` instances (this happens implicitly). It takes multiple seconds per execution, and is executed for every single vulnerability. I'm counting a total of `352642` rows in my local `VULNERABLESOFTWARE` table. It's possible that there's an important index missing, but I need more time to investigate. Also, this bottleneck was not introduced in this PR, it just surfaces now.

2. Parsing version ranges can really only be done on a best-effort basis. The `affected[].ranges[].events` array in OSV is not guaranteed to be sorted in any meaningful way. We just assume that now, because we have no other choice. Order of events is not required to answer the "is my version affected?" question (see [evaluation algorithm](https://ossf.github.io/osv-schema/#evaluation)), but because we're mirroring the database instead of using the OSV API we're always in danger of ingesting wrong or ambiguous data. 

Point 2 bothers me the most. While my local testing looked good WRT identifying vulnerabilities, I think it's safe to say that OSV will not yield the accurate data we were hoping for, at least not right now.

If we end up merging this, we should make it really clear that this is a preview feature. Not due to the implementation, but because mapping from OSV to our data model is error-prone. May be worth investigating whether just using OSV's API would be a better approach for now. That way we can use their documented evaluation algorithm, and only mirror vulnerabilities ad-hoc, like it's done for OSS Index already.

VinodAnandan:
I think there are pros and cons for each approach (individual calls vs batch/mirroring). Considering the current state (New, relatively small VDB, No SLA, etc.) of OSV.dev I believe the batch/mirroring approach is better for now, but we can revisit it later if needed. Also, the mirroring processing can be improved if we can incrementally get the data.
 
I don't believe the affected data problem is just associated with OSV.dev, any vulnerability database may contain inaccurate affected data, and we can't expect perfect affected data from any source. The DT should support correcting the affected data over time by removing the append-only restriction. DT should associate the affected data to a source and the source be allowed to update the data if needed. Otherwise it may result in more FPs even if the vulnerability data source eventually corrects the data.

nscuro:
@VinodAnandan Agreed. We need a way to track which source added a `VulnerableSoftware` entry, to then be able to remove it again once it's not reported by any source anymore. Corrections in VDBs happen frequently, and we need to consider that.

> I don't believe the affected data problem is just associated with OSV.dev, any vulnerability database may contain inaccurate affected data, and we can't expect perfect affected data from any source.

+1. However, my main concern with OSV isn't that their data isn't super accurate, it's more that me may not be able to correctly parse / interpret the data they're providing (see the ranges issue above).

nscuro:
Thanks @oliverchang, https://github.com/google/osv.dev/issues/485 would be really nice to have!

nscuro (approved):
Thanks again for all the work you put into this, @sahibamittal!

I'm going to merge this now, and I logged #1815 to address the `VulnerableSoftware` situation.

We still need documentation for the OSV datasource (similar to the [GHSA](https://docs.dependencytrack.org/datasources/github-advisories/) one). The docs should also state that this is a preview feature. Could you provide a PR for this, or do you want us to do it?

sahibamittal:
Thanks Niklas! 
I'll work on the documentation. 👍🏼

Issue #1717, opened 2022-06-15
Make project search by tags case insensitive


Mvld3r:
### Current Behavior:
Project search by tags is currently case-sensitive.
For example, `/v1/project/tag/mytag` is different from `/v1/project/tag/MyTag`

### Proposed Behavior:
Make project search by tags case insensitive

huayaoyue6:
Also,there is no way to find how many tags in system.

PR #1723, opened 2022-06-16
Make project search by tags case insensitive

Mvld3r:
Addresses https://github.com/DependencyTrack/dependency-track/issues/1717

Signed-off-by: ege-dw <enora.germond@deveryware.com>


nscuro:
Thanks for the PR @Mvld3r! I agree that tag naming and searching should be streamlined. However, this will be a breaking change for folks who currently rely on case sensitivity. I'd like to schedule this for 4.7 so we can keep changes in 4.6 manageable.

nscuro (approved):

Issue #1726, opened 2022-06-17
author field of components not stored in database


stephan-wolf-ais:
The defect may already be reported! Please search for the defect before creating one.

### Current Behavior:
The author field the components in BOM files not stored properly in database.

### Steps to Reproduce:
create a BOM of a npm project with cyclonedx for node. packages.json contains an author field wich is also used in BOM files

### Expected Behavior:
The Author should be stored in database

### Environment:
  
  - Dependency-Track Version: 4.5.0
  - Distribution:  Docker
  - BOM Format & Version: xml/json version 1.3
  - Database Server: any
  - Browser: any

PR #1727, opened 2022-06-17
store author field from BOM in database

stephan-wolf-ais:
This fixes issue #1726

### DIFF
--- a/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
+++ b/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
@@ -91,7 +91,7 @@ public class ModelConverter {
             component = new Component();
             component.setProject(project);
         }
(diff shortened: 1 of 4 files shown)


nscuro (approved):
Thanks for the PR @stephan-wolf-ais! LGTM.

@stevespringett Was the `author` field excluded on purpose up until now?

stevespringett:
> Was the author field excluded on purpose up until now?

No, not excluded on purpose. As the object model for CDX has expanded over time, the fields supported by DT has not kept up. There are other component fields that are also not supported yet.

nscuro:
OK, I filed #1733 for us to take a closer look as to which fields are currently not supported.

I'll merge this PR, however the UI won't show it right now. @stephan-wolf-ais, if you want to see the author in the frontend as well, please submit a PR for the frontend repository as well. Otherwise we'll do it when we tackle #1733.

stephan-wolf-ais:
I will try to make a PR for the frontend as well but it will take some time.

Issue #1759, opened 2022-06-28
Dependency track takes more than 30 minutes to analyze and Dependency tracker jenkins aborts due to polling time limti.


jayachathu:
The defect may already be reported! Please search for the defect before creating one.

### Current Behavior:
Jenkins Dependecy tarck jobs fails since server takes more 30 minutes analyze the project.

### Steps to Reproduce:
Run jenkins Dependecy track job in Jenkins pipeline for java project which has more than 1000 components(These are java modules) 

### Expected Behavior:
Jenkins PipleLine should not fail.

### Environment:
  
  - Dependency-Track Version: 'latest'
  - Distribution: [ Docker ]
  - BOM Format & Version: "bomFormat": "CycloneDX",
    "specVersion": "1.4",
    "version": 1,
  - Database Server: PostgreSQL 
  - Browser: Chrome

### Additional Details:

22-06-28 06:37:51,463 [] INFO [org.dependencytrack.tasks.MetricsUpdateTask] Completed portfolio metrics update
2022-06-28 06:37:56,340 [] INFO [org.dependencytrack.tasks.MetricsUpdateTask] Completed metrics update for project: cd6feaf1-c6df-4c15-a7b4-f118151a6605
2022-06-28 06:52:29,046 [] INFO [org.dependencytrack.tasks.BomUploadProcessingTask] Processing CycloneDX BOM uploaded to project: cd6feaf1-c6df-4c15-a7b4-f118151a6605
2022-06-28 06:56:59,202 [] INFO [org.dependencytrack.tasks.BomUploadProcessingTask] Processing CycloneDX dependency graph for project: cd6feaf1-c6df-4c15-a7b4-f118151a6605
2022-06-28 07:04:14,351 [] INFO [org.dependencytrack.tasks.BomUploadProcessingTask] Processed 18477 components and 0 services uploaded to project cd6feaf1-c6df-4c15-a7b4-f118151a6605
2022-06-28 07:22:08,107 [] INFO [org.dependencytrack.tasks.scanners.InternalAnalysisTask] Starting internal analysis task
2022-06-28 07:22:42,573 [] INFO [org.dependencytrack.tasks.scanners.InternalAnalysisTask] Internal analysis complete
2022-06-28 07:22:42,579 [] WARN [org.dependencytrack.tasks.scanners.OssIndexAnalysisTask] An API username or token has not been specified for use with OSS Index. Using anonymous access
[...]

PR #1772, opened 2022-07-08
Prevent event queue saturation when processing BOMs with large component quantities

nscuro:
Previously, a `RepositoryMetaEvent` was dispatched for every component in the uploaded BOM. For a BOM with 20k components, 20k events were dispatched.

This was done *before* the `VulnerabilityAnalysisTask` is dispatched. Due to the sheer number of `RepositoryMetaEvent`s, vulnerability analysis would be drastically delayed. Even worse, event processing in the entire platform would essentially halt because DT would be busy with processing `RepositoryMetaEvent`s.

To prevent such situations, a single `RepositoryMetaEvent` is now dispatched for the entire project for which the BOM was uploaded. This ensures that the event subsystem remains available for other tasks.

Addresses #1759

### DIFF
--- a/src/main/java/org/dependencytrack/event/RepositoryMetaEvent.java
+++ b/src/main/java/org/dependencytrack/event/RepositoryMetaEvent.java
@@ -20,18 +20,56 @@ package org.dependencytrack.event;
 
 import alpine.event.framework.Event;
 import org.dependencytrack.model.Component;
+import org.dependencytrack.model.Project;
+
+import java.util.Optional;
+import java.util.UUID;
 
 public class RepositoryMetaEvent implements Event {
 
-    private Component component;
+    public enum Type {
+        PORTFOLIO,
+        PROJECT,
+        COMPONENT
+    }
 
-    public RepositoryMetaEvent() { }
+    private final Type type;
+    private final UUID target;
+
+    public RepositoryMetaEvent() {
+        this(Type.PORTFOLIO, null);
+    }
+
+    /**
+     * @since 4.6.0
+     */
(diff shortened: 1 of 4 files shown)


nscuro:
Only downside I can fathom right now is that repository meta analysis will take a little longer in total, due to the switch from parallelized to sequential processing. But I think that tradeoff is worth it.

stevespringett:
As an alternative, what if DT only dispatched `RepositoryMetaEvent` for new components?

Taking the BOM with 20K components as an example, if the same 20K components are unchanged upon every BOM upload, these events will still be kicked off and DT will have to check its cache.

Instead, if DT could only fire RepositoryMetaEvent when a new component is added?

RepositoryMetaEvent is dispatched every 24 hours anyway. By doing this, we could dramatically reduce database calls.

nscuro:
> Instead, if DT could only fire RepositoryMetaEvent when a new component is added?

That will certainly help for existing projects, but not for new ones. It'd also mean that there's no way to force a refresh of repository meta anymore. 

We could optionally dispatch `RepositoryMetaEvent`s with batches of components (say 10-250, not really sure what a good size would be).

nscuro:
Hmmm, I just noticed that we don't cache repository meta data at all right now. That alone will most likely yield a significant improvement. I'll get this sorted and also see how your suggestion of only considering "new" components performs.

nscuro:
Cancelling this for now as stated in https://github.com/DependencyTrack/dependency-track/issues/1759#issuecomment-1242768566. The missing caching for repository meta data has been logged in https://github.com/DependencyTrack/dependency-track/issues/1943 and is scheduled for 4.7.

Issue #1768, opened 2022-07-07
Policy: add CWE in conditions


sahibamittal:
### Proposed Behavior: Policy conditions to include CWE(s).

CWEs can be associated with multiple vulnerabilities and would be a good option for filtering.

Backend PR: https://github.com/DependencyTrack/dependency-track/pull/1773
Frontend: https://github.com/DependencyTrack/frontend/pull/184

<img width="1422" alt="cwe" src="https://user-images.githubusercontent.com/5709882/178459514-c422a6cf-9b13-46ac-9ee7-892c032d3204.png">

sahibamittal:
@nscuro @VinodAnandan

PR #1773, opened 2022-07-11
Issue-1768-policy-condition-cwe

sahibamittal:

Issue: https://github.com/DependencyTrack/dependency-track/issues/1768
Front-end: https://github.com/DependencyTrack/frontend/pull/184

**New feature**: Policy conditions to include CWE(s).

### DIFF
--- a/src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
@@ -1,0 +1,94 @@
+ /*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Steve Springett. All Rights Reserved.
+ */
+package org.dependencytrack.policy;
+
+ import alpine.common.logging.Logger;
+ import org.apache.commons.collections4.CollectionUtils;
+ import org.dependencytrack.model.Component;
+ import org.dependencytrack.model.Policy;
+ import org.dependencytrack.model.PolicyCondition;
+ import org.dependencytrack.model.Vulnerability;
+ import org.dependencytrack.parser.common.resolver.CweResolver;
+
+ import java.util.ArrayList;
+ import java.util.Arrays;
+ import java.util.List;
+
+/**
+ * Evaluates the Common Weakness Enumeration of component vulnerabilities against a policy.
+ */
+public class CwePolicyEvaluator extends AbstractPolicyEvaluator {
+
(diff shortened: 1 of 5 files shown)


sahibamittal:
cwe

src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
```diff
@@ -0,0 +1,94 @@
...
+                    violations.add(new PolicyConditionViolation(condition, component));
+                }
+            }
+        }
+        return violations;
+    }
+
+    public boolean matches(final PolicyCondition.Operator operator, final List<Integer> vulnerabilityCwes, final String conditionValue) {
+
+        if (conditionValue == null || vulnerabilityCwes == null || "*".equals(conditionValue.trim())) {
```
nscuro, line 68: Should the condition really match when either `conditionValue` or `vulnerabilityCwes` are `null`? 

At least `vulnerabilityCwes` is `null` in some cases, for example when a CVE was identified by OSS Index before the NIST mirror task synced that particular vulnerability.

src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
```diff
@@ -0,0 +1,94 @@
...
+            return true;
+        }
+        if (PolicyCondition.Operator.IS == operator) {
+            final Integer cweId = CweResolver.getInstance().parseCweString(conditionValue);
+            if (cweId != null) {
+                return vulnerabilityCwes.contains(cweId);
+            }
```
nscuro, line 75: The behavior of `IS` does not really differ from that of `CONTAINS_ANY` with just one CWE specified.

I propose to either

* Make `IS` even stricter; Make it behave like "contains only" / "contains exactly"
* Remove `IS` because all desired behaviors are already covered by  `CONTAINS_ANY` and `CONTAINS_ALL`

src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
```diff
@@ -0,0 +1,94 @@
...
+        } else {
+            List<Integer> cweIdsToMatch = new ArrayList<>();
+            List<String> conditionCwes = Arrays.asList(conditionValue.split(","));
+            conditionCwes.replaceAll(String::trim);
+            conditionCwes.stream().forEach(cwe -> {
+                Integer id = CweResolver.getInstance().parseCweString(cwe);
+                if(id != null)
+                    cweIdsToMatch.add(id);
+            });
+            if (PolicyCondition.Operator.CONTAINS_ANY == operator) {
```
nscuro, line 85: Should there be a check here whether `conditionCwes` is empty? Perhaps we should log a warning in that case, because it indicates a misconfiguration?

nscuro (changes requested):

src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
sahibamittal: Yes, it should return false when conditionValue or vulnerabilityCwes are null but true when condition is '*'.

src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
sahibamittal: IS was for a single CWE match, but again yes, CONTAINS_ANY also fulfils the requirement. We dont need "contains only" / "contains exactly" really. I'll remove the IS operator.

src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
sahibamittal: That would be case if user types in condition value as " , " or just space or empty. In that case, we return false. I've added a empty check on cweIdsToMatch.

Issue #1775, opened 2022-07-12
Policy: add component hash in conditions


sahibamittal:
A component may have multiple hash values of different algorithm types:
_MD5("MD5"),
SHA1("SHA-1"),
SHA_256("SHA-256"),
SHA_384("SHA-384"),
SHA_512("SHA-512"),
SHA3_256("SHA3-256"),
SHA3_384("SHA3-384"),
SHA3_512("SHA3-512"),
BLAKE2b_256("BLAKE2b-256"),
BLAKE2b_384("BLAKE2b-384"),
BLAKE2b_512("BLAKE2b-512"),
BLAKE3("BLAKE3");_

### Proposed Behavior: Policy conditions to include a component hash filter, with algo types in dropdown. 

Backend: https://github.com/DependencyTrack/dependency-track/pull/1776
Frontend: https://github.com/DependencyTrack/frontend/pull/185

<img width="1422" alt="HASH" src="https://user-images.githubusercontent.com/5709882/178535930-b52728e5-bc0d-4c3a-8406-ebd9190bb7f7.png">

Issue #1553, opened 2022-04-19
Allow policies with hash support


VinodAnandan:
### Proposed Behavior:
Sometimes companies will use patched dependencies with the same pURL. These dependencies will have a different hash, there should be a hash-based policy feature that should override the advisories (suppress the vulnerabilities that are already fixed).

PR #1776, opened 2022-07-12
Issue-1775 : policy condition - component hash

sahibamittal:

Issue: https://github.com/DependencyTrack/dependency-track/issues/1775
https://github.com/DependencyTrack/dependency-track/issues/1553

HASH

### DIFF
--- a/src/main/java/org/dependencytrack/policy/ComponentHashPolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/ComponentHashPolicyEvaluator.java
@@ -1,0 +1,108 @@
+/*
+ * This file is part of Dependency-Track.
+ *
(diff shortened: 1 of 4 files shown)


src/main/java/org/dependencytrack/policy/ComponentHashPolicyEvaluator.java
```diff
@@ -0,0 +1,108 @@
...
+        return new Hash(
+                def.optString("algorithm", null),
+                def.optString("value", null)
+        );
+    }
+
+    private boolean matches(Hash hash, Component component) {
+
+        if (hash == null)
+            return true;
```
nscuro, line 76: Similarly to our discussion in #1773, shouldn't the condition not *not* match when `hash` is null?

PR #1813, opened 2022-07-23
Use datastore-native limiting in queries where only a single result is desired

nscuro:
Translates into `LIMIT 1`, `TOP 1` etc. depending on the datastore.

Results in queries being executed faster in the datastore (at least when no index is used), reduces the amount of unnecessary data being loaded into RAM.


nscuro:
Noticed while investigating on #1759 that `ComponentQueryManager#matchSingleIdentity` took ages to execute for each component. In my test scenario the query returned between 6 and 12 results, of which only the first was actually used. I haven't measured the impact of this change, but subjectively it was drastic.

A similar result is to be expected for `MetricsQueryManager#getMostRecent*Metrics`, where currently *all* rows are fetched as well.

nscuro:
Ok, measured it now. In both cases, using a BOM with 20466 components and multiple duplicates per component:

1. Upload BOM
2. Wait for all related processing to complete
3. Upload BOM again
4. Use VisualVM CPU sampler to see where processing time is spent
5. Wait for conversion of components (which calls `ComponentQueryManager#matchSingleIdentity`) to complete
6. Make screenshot

Before (`master`):

before

After (this branch):

after

Issue #1542, opened 2022-04-12
Allow users to configure intervals for periodic tasks


AbdelHajou:
### Current Behavior:
Currently most scheduled tasks are performed at a fixed interval that is hard-coded. For example, vulnerability analysis is performed every 24 hours after an initial delay of 6 hours.

### Proposed Behavior:
Some organizations may have stricter rules for how often components are scanned for vulnerabilities. We would like to scan for vulnerabilities twice a day. 

I think it would be a nice improvement to allow the Dependency-Track administrator to configure the intervals for some tasks, like vulnerability analysis and repository metadata analysis.

I'm not sure whether configuring this via the dashboard would be the best route, or to just configure this via environment variables in the docker-compose deployment.

If you agree that this would be a nice addition, I would like to work on this issue.

stevespringett:
> We would like to scan for vulnerabilities twice a day.

Thats nice, however, in most cases OSS Index only updates their feed once a day. The only exceptions to this are for things like when log4j happened. In those cases, they push out updates sooner.

I'm not sure how often GitHub updates their advisories.

I think there are two things required here:
1. Make the scans interval configurable
2. Make the ComponentAnalysisCache configurable
  2.1 Separate out repository from vulnerability cache

msymons:
>in most cases OSS Index only updates their feed once a day

From recent Sonatype announcement regarding upcoming changes to OSSI:

> 1.	**OSSI will be updated faster**. OSSI users will now get updated vulnerability information as quickly as our enterprise users do, as evidenced with Log4Shell

AbdelHajou:
@msymons That's great news, do you have a link to this announcement?

msymons:
wrt Sonatype announcement,  this was sent out via email on 8th March 2022.   Steve copied it into DT Slack the same day.

So as not to add noise to this enhancement issue, I (or someone else) will update the Slack channel when more information is available.

syalioune:
@AbdelHajou Are you working on this enhancement request ? or anyone else ? We are very interested by this feature in my organization and I should have time to get it done in the next weeks if need be.

syalioune:
@stevespringett I see that the issue is assigned to you, are you actively working on it ? If not, I would like to provide an implementation. Does not seem like @AbdelHajou is on the subject either. Thanks.

AbdelHajou:
@syalioune I’m on vacation so I wont have time to work on this anytime soon

syalioune:
[...]

PR #1841, opened 2022-07-29
Make background tasks periodicity configurable

syalioune:
Enhancement request described in #1542.

Signed-off-by: Alioune SY <sy_alioune@yahoo.fr>

Please find below related PRs to review :

- Frontend : [PR 214](https://github.com/DependencyTrack/frontend/pull/214)

**Design decisions :**

- Task are not rescheduled after period update because it would require updates on Alpine framework to be able to cancel a given task. It is also tricky because, if the task ran 5min ago, we don't want to run it again after a reschedule. With a restart, we don't have to handle all those edge cases
- I've introduced _scanner.analysis.cache.validity.period_ configuration property used in [BaseComponentAnalyzerTask](https://github.com/syalioune/dependency-track/blob/feature/configurable_period_for_tasks/src/main/java/org/dependencytrack/tasks/scanners/BaseComponentAnalyzerTask.java#L68) instead of the previous hard coded value. It can be configured through API by experienced users but not through the new UI as it is not really related to task scheduling
- For now, I haven't add a check that _scanner.analysis.cache.validity.period_ is strictly inferior to _task-scheduler.component.analysis.cache.clear.period_  as the first one is not really exposed to the end users
- No specific documentation update as the UI seems self-descriptive

Please find a sneak peak of the UI below

image

### DIFF
--- a/src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
+++ b/src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
@@ -71,17 +71,17 @@ public enum ConfigPropertyConstants {
     ACCESS_MANAGEMENT_ACL_ENABLED("access-management", "acl.enabled", "false", PropertyType.BOOLEAN, "Flag to enable/disable access control to projects in the portfolio"),
     NOTIFICATION_TEMPLATE_BASE_DIR("notification", "template.baseDir", SystemUtils.getEnvironmentVariable("DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY", System.getProperty("user.home")), PropertyType.STRING, "The base directory to use when searching for notification templates"),
     NOTIFICATION_TEMPLATE_DEFAULT_OVERRIDE_ENABLED("notification", "template.default.override.enabled", SystemUtils.getEnvironmentVariable("DEFAULT_TEMPLATES_OVERRIDE_ENABLED", "false"), PropertyType.BOOLEAN, "Flag to enable/disable override of default notification templates"),
-    TASK_SCHEDULER_LDAP_SYNC_PERIOD("task-scheduler", "ldap.sync.period","21600000", PropertyType.NUMBER, "Sync period (in milliseconds) for LDAP"),
-    TASK_SCHEDULER_GHSA_MIRROR_PERIOD("task-scheduler", "ghsa.mirror.period","86400000", PropertyType.NUMBER, "Mirror period (in milliseconds) for Github Security Advisories"),
-    TASK_SCHEDULER_OSV_MIRROR_PERIOD("task-scheduler", "osv.mirror.period","86400000", PropertyType.NUMBER, "Mirror period (in milliseconds) for OSV database"),
-    TASK_SCHEDULER_NIST_MIRROR_PERIOD("task-scheduler", "nist.mirror.period","86400000", PropertyType.NUMBER, "Mirror period (in milliseconds) for NVD database"),
-    TASK_SCHEDULER_VULNDB_MIRROR_PERIOD("task-scheduler", "vulndb.mirror.period","86400000", PropertyType.NUMBER, "Mirror period (in milliseconds) for VulnDB database"),
-    TASK_SCHEDULER_PORTFOLIO_METRICS_UPDATE_PERIOD("task-scheduler", "portfolio.metrics.update.period","3600000", PropertyType.NUMBER, "Update period (in milliseconds) for portfolio metrics"),
-    TASK_SCHEDULER_VULNERABILITY_METRICS_UPDATE_PERIOD("task-scheduler", "vulnerability.metrics.update.period","3600000", PropertyType.NUMBER, "Update period (in milliseconds) for vulnerability metrics"),
-    TASK_SCHEDULER_PORTFOLIO_VULNERABILITY_ANALYSIS_PERIOD("task-scheduler", "portfolio.vulnerability.analysis.period","86400000", PropertyType.NUMBER, "Launch period (in milliseconds) for portfolio vulnerability analysis"),
-    TASK_SCHEDULER_REPOSITORY_METADATA_FETCH_PERIOD("task-scheduler", "repository.metadata.fetch.period","86400000", PropertyType.NUMBER, "Metadada fetch period (in milliseconds) for package repositories"),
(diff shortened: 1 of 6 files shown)


src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -69,7 +70,18 @@ public enum ConfigPropertyConstants {
...
     ACCESS_MANAGEMENT_ACL_ENABLED("access-management", "acl.enabled", "false", PropertyType.BOOLEAN, "Flag to enable/disable access control to projects in the portfolio"),
     NOTIFICATION_TEMPLATE_BASE_DIR("notification", "template.baseDir", SystemUtils.getEnvironmentVariable("DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY", System.getProperty("user.home")), PropertyType.STRING, "The base directory to use when searching for notification templates"),
-    NOTIFICATION_TEMPLATE_DEFAULT_OVERRIDE_ENABLED("notification", "template.default.override.enabled", SystemUtils.getEnvironmentVariable("DEFAULT_TEMPLATES_OVERRIDE_ENABLED", "false"), PropertyType.BOOLEAN, "Flag to enable/disable override of default notification templates");
+    NOTIFICATION_TEMPLATE_DEFAULT_OVERRIDE_ENABLED("notification", "template.default.override.enabled", SystemUtils.getEnvironmentVariable("DEFAULT_TEMPLATES_OVERRIDE_ENABLED", "false"), PropertyType.BOOLEAN, "Flag to enable/disable override of default notification templates"),
+    TASK_SCHEDULER_LDAP_SYNC_PERIOD("task-scheduler", "ldap.sync.period","21600000", PropertyType.NUMBER, "Sync period (in milliseconds) for LDAP"),
```
nscuro, line 74: Nitpick: These properties should use "cadence" (as used for previously configurable tasks) or more ideally "interval", because they describe reoccurring events. A period describes a block of time, and interval the time between periods.

nscuro (commented):
LGTM, but I'd like to have @stevespringett check whether this matches what he envisioned in https://github.com/DependencyTrack/dependency-track/issues/1542#issuecomment-1096948921

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
syalioune: Thanks @nscuro. I agree that cadence is more accurate.
I'll wait feedbacks from Steve before implementing the changes.

stevespringett: All of these configurable task scheduler properties should not be in milliseconds. At the lowest, I would think minutes would be appropriate, however, I also think hour may work as well.

src/main/java/org/dependencytrack/tasks/TaskScheduler.java
```diff
@@ -22,19 +22,9 @@
...
-import org.dependencytrack.event.GitHubAdvisoryMirrorEvent;
-import org.dependencytrack.event.OsvMirrorEvent;
-import org.dependencytrack.event.InternalComponentIdentificationEvent;
-import org.dependencytrack.event.KennaSecurityUploadEventAbstract;
-import org.dependencytrack.event.MetricsUpdateEvent;
-import org.dependencytrack.event.NistMirrorEvent;
-import org.dependencytrack.event.PortfolioVulnerabilityAnalysisEvent;
-import org.dependencytrack.event.RepositoryMetaEvent;
-import org.dependencytrack.event.VulnDbSyncEvent;
+import org.dependencytrack.event.*;
```
stevespringett, line 27: Please replace with individual class imports. Do not use wildcards.

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
stevespringett: I also think we need absolute minimum allowed. For example, I would not want GitHub mirroring to take place every minute. If this is going to be user configurable, we need to provide guardrails that allows flexibility in configuration without the ability to cause the system harm - which would increase support tickets.

stevespringett (changes requested):

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
syalioune: If we go that way, I think that hours would be more appropriate. The properties being positive numbers, it's fairly easy to set a minimum of one hour without having to configure it (property > 0).
I suppose that should be ok for now.
Should more control be given on the minimum allowed, it should be configurable through alpine, DTrack system property or DTrack environment variable according to me.

src/main/java/org/dependencytrack/tasks/TaskScheduler.java
syalioune: Will do ! I think that one should go also in DEVELOPING.md with a coding style guide

valentijnscholten:
I suggest interval. Cadence, to my understanding, is more like the number of times per period, i.e twice a day. Not sure if cadence = 24 is intuitively understood as "every 24 hours"? interval=24h is more clear.

nscuro:
> I suggest interval. Cadence, to my understanding, is more like the number of times per period, i.e twice a day. Not sure if cadence = 24 is intuitively understood as "every 24 hours"? interval=24h is more clear.

I agree, but the downside is that cadence is already used for other tasks (Fortify and DefectDojo sync for example). If we were to switch to interval now, we should change the existing names as well. And that would be a breaking change, which is something we'd really like to avoid.

src/main/java/org/dependencytrack/tasks/TaskScheduler.java
nscuro: Would be even better if we had a way to enforce this sort of code style requirements in CI. [checkstyle](https://checkstyle.sourceforge.io/config_imports.html#AvoidStarImport) can do it, but there may be other options as well.

nscuro:
@stevespringett To me it looks like all your points have been addressed. Anything else you want done before we can merge?

stevespringett (approved):

stevespringett:
Awesome. Thanks for the collaboration and quality PR @syalioune. Much appreciated.

PR #1858, opened 2022-08-04
OSV - update ecosystems

sahibamittal:
New additions of missing ecosystems in OSV

For reference, you can get list of all by running `gsutil ls -lah 'gs://osv-vulnerabilities/'`

### DIFF
--- a/src/main/java/org/dependencytrack/parser/osv/model/Ecosystem.java
+++ b/src/main/java/org/dependencytrack/parser/osv/model/Ecosystem.java
@@ -19,16 +19,6 @@ public enum Ecosystem {
     NPM("npm"),
     DWF("DWF"),
     DEBIAN("Debian"),
-    DEBIAN_11("Debian:11"),
-    DEBIAN_10("Debian:10"),
-    DEBIAN_9("Debian:9"),
-    DEBIAN_8("Debian:8"),
-    DEBIAN_7("Debian:7"),
-    DEBIAN_6("Debian:6.0"),
-    DEBIAN_5("Debian:5.0"),
-    DEBIAN_4("Debian:4.0"),
-    DEBIAN_3_1("Debian:3.1"),
-    DEBIAN_3_0("Debian:3.0"),
     GITHUB_ACTIONS("GitHub Actions");
 
     private final String value;
--- a/src/main/java/org/dependencytrack/parser/osv/model/Ecosystem.java
+++ b/src/main/java/org/dependencytrack/parser/osv/model/Ecosystem.java
@@ -17,7 +17,19 @@ public enum Ecosystem {
     UVI("UVI"),
     RUST("crates.io"),
     NPM("npm"),
-    DWF("DWF");
+    DWF("DWF"),
+    DEBIAN("Debian"),
+    DEBIAN_11("Debian:11"),
+    DEBIAN_10("Debian:10"),
+    DEBIAN_9("Debian:9"),
+    DEBIAN_8("Debian:8"),
+    DEBIAN_7("Debian:7"),
+    DEBIAN_6("Debian:6.0"),
+    DEBIAN_5("Debian:5.0"),
+    DEBIAN_4("Debian:4.0"),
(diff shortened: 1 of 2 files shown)


sahibamittal:
@nscuro

nscuro:
@sahibamittal How likely is it that ecosystems change over time? Is it possible that OSV renames some of them, and can that be an issue for us?

Also, I believe you mentioned that you looked into making this a little more dynamic, so that we don't have to list all ecosystems explicitly. Is that still a viable option? With how OSV has named their Debian ecosystems, it appears like we eventually have to update this class regularly.

VinodAnandan:
@oliverchang informed that there may be C/C++ feeds coming in the second half of this year. We will send an update whenever it is ready. But we are hoping to have no other changes in the meantime.

@oliverchang Is it possible to provide an aggregated archive file with all ecosystems?

src/main/java/org/dependencytrack/parser/osv/model/Ecosystem.java
```diff
@@ -17,7 +17,19 @@ public enum Ecosystem {
...
+    DEBIAN_10("Debian:10"),
+    DEBIAN_9("Debian:9"),
+    DEBIAN_8("Debian:8"),
+    DEBIAN_7("Debian:7"),
+    DEBIAN_6("Debian:6.0"),
+    DEBIAN_5("Debian:5.0"),
+    DEBIAN_4("Debian:4.0"),
+    DEBIAN_3_1("Debian:3.1"),
+    DEBIAN_3_0("Debian:3.0"),
+    GITHUB_ACTIONS("GitHub Actions");
```
nscuro, line 32: The `value`s in this enum are inserted as-is into a URL string in `OsvDownloadTask`. The space here is thus not properly escaped and causes an exception, e.g.:

```
2022-08-05 18:59:19,866 INFO [OsvDownloadTask] Updating datasource with Google OSV advisories for ecosystem GitHub Actions
2022-08-05 18:59:19,870 ERROR [LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.IllegalArgumentException: Illegal character in path at index 57: https://osv-vulnerabilities.storage.googleapis.com/GitHub Actions/all.zip
	at java.base/java.net.URI.create(URI.java:883)
	at org.apache.http.client.methods.HttpGet.<init>(HttpGet.java:66)
	at org.dependencytrack.tasks.OsvDownloadTask.inform(OsvDownloadTask.java:67)
	at alpine.event.framework.BaseEventService.lambda$publish$0(BaseEventService.java:101)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

We should add escaping to `OsvDownloadTask` to prevent this.

nscuro (changes requested):

oliverchang:
> @oliverchang informed that there may be C/C++ feeds coming in the second half of this year. We will send an update whenever it is ready. But we are hoping to have no other changes in the meantime.
> 
> @oliverchang Is it possible to provide an aggregated archive file with all ecosystems?

Ecosystems should never get renamed in OSV. We may add new ones over time. 

Re aggregated archive file: could I understand what this is intended to solve? It seems that the underlying problem here is that an enum needs to be maintained that needs to map to OSV ecosystems. 

For Debian specifically, if you don't care about the individual `Debian:X` versions, you could just pull `Debian` which will contain everything.

VinodAnandan:
Hi @oliverchang , the main use case of the aggregate zip files is to avoid the code change and release of the DTrack whenever a new ecosystem gets introduced to OSV. This way, the users will get an aggregated archive file and it will contain a new ecosystem introduced by OSV.

nscuro (approved):
Thanks @sahibamittal!

Issue #1137, opened 2021-08-10
Track "First Seen" Date for Vulnerabilities


msymons:
### Current Behavior:
As of v4.3.2 provides:
* "Published" date for vulnerabilities on the Vulnerabilities page.
* "Attributed On" date for individual vulnerabilities, displayed on "Audit Vulnerabilities" tab of project page.

There is no way to see when a vulnerability first entered the portfolio.

If a BOM is uploaded with vulnerable component X then an attribution date will be recorded
If a new project is created every date for the next 99 days with component X then there will now be 100 attribution dates.  Every single one will be different...  and only one will be the "original" (ie, oldest) date.  The only way to know which is the oldest is to manually inspect each project,  do some heavy API parsing, or query the database directly.

### Proposed Behavior:
Record "First Seen" date for each vulnerability that affects the portfolio, and display the date on the "Vulnerabilities" page.

**Use case one**.  Compare "Published" and "First Seen" dates.  Is there a lag between the two?  To illustrate, at time of writing, CVE-2021-29425 is displayed in DT with a Published Date of 13 Apr 2021.  But after 4 months it is still not in OSS Index data.  But it will be...  some day.  But when?

**Use case two**.  If every instance of a vulnerability in the portfolio is addressed then there will be no more attribution dates for that vulnerability.  But the "First Seen" date would remain.  Thus, if the vulnerability is reintroduced to the portfolio, one would know that it has been seen before...  it is not actually "new".

**Use case three**.  Notifications configured ONLY for "NEW_VULNERABILITY" are alerting on every attribution.  This creates a lot of noise and (for me) make notifications unusable.  This was reported as #389.  Perhaps the notifications should be suppressed if "First Seen" is already populated (or something along those lines)?

stevespringett:
Fair points, but I think we need to think this through.

If I have Project A and a vulnerable component in that project. The 'first seen' would be recorded. The project fixes it, and another project (Project B) introduces the same vulnerable component a year later. Will the 'first seen' be the time when Project A seen it (which would be a year earlier than Project B used it)?.

I'm trying to understand the value this would provide? First seen, by who? First seen by the portfolio would be a year earlier. First seen by project B would be the timestamp when the vulnerable component was introduced.

PR #1864, opened 2022-08-06
Re-implement support for `NEW_VULNERABLE_DEPENDENCY` notification

### LINKED ISSUE #1611: NEW_VULNERABLE_DEPENDENCY notification not working
While working on adding policy violation support to notifications, I realized that the NEW_VULNERABLE_DEPENDENCY notification doesn't work at all and is not hooked up to anything. This is likely a result from the data model migration from 3.x to 4.x, but it's interesting that it hasn't been reported.

nscuro:
Fixes #1611

Notable change in behavior compared to 3.8.0:

`analyzeNotificationCriteria` was invoked in 3.8.0 when a new dependency was created:

https://github.com/DependencyTrack/dependency-track/blob/f37279c862a6302a2c56a19dc13d5e9a71d65790/src/main/java/org/dependencytrack/persistence/QueryManager.java#L1259

That worked because components were global and could potentially already have vulnerabilities assigned to them.

With the new component model in >= 4.0.0, new components can't have vulnerabilities assigned to them already, so we have to wait until vulnerability analysis completes before calling `analyzeNotificationCriteria`. 

The name of the new event I introduced is a little unfortunate, I'm very much open to naming suggestions 😅


syalioune:
Hello @nscuro 

That's great 👍 

Is there any way to make NEW_VULNERABLE_DEPENDENCY and NEW_VULNERABILITY mutually exclusive (i.e : knowing that the component was newly added in `NotificationUtil.analyzeNotificationCriteria` ) ?
Otherwise, it seems the two notifications would fire off on SBOM upload for any new vulnerable component.

stevespringett:
This works for me @nscuro. I think the suggestion @syalioune has would be new behavior. Entirely up to you if you want to implement that now, or in a future release. On the surface it sounds simple, but I think it may be a bit more difficult.

nscuro:
@syalioune @stevespringett Let's address that in another enhancement.

We'll need a more sturdy way to track whether a component is new. Maybe some sort of `firstSeen` / `lastSeen` combination which would generally be very useful for other reasons. If `firstSeen == lastSeen`, then the component could be considered new.

msymons:
@nscuro , you mean...  something like the enhancement #1137 that I logged last year?  =)

nscuro:
@msymons Yes! Just for components instead of vulnerabilities. And by reading the comments, the concerns also hold true in that case (what does "first seen" even mean?).

PR #1867, opened 2022-08-07
Restrict notification's affected project list when notification rule is limited to specific projects

### LINKED ISSUE #1866: Affected project list not restricted in notification's subject when notification rule is limited to certain project
### Current Behavior:

Dependency Track does not restrict the affected project list of each notification especially when the related notification rule is limited to specific projects.
Moreover, projects could be duplicated in affected projects because that list is constructed based on components and there can be duplicates in a project (e.g. : Global SBOM merge with cyclonedx-cli of several projects which may contain the same component)

### Steps to Reproduce:

- Create two projects `PROJECTA` and `PROJECTB`
- Manually add a component with a fictitious CPE (i.e. `fictitious_cpe`) in `PROJECTA` (twice, thus creating a duplicate) and `PROJECTB` (once)
- Create a notification rule limited to `PROJECTA` and NEW_VULNERABILITY event enabled
- Create an internal vulnerability impacting the fictitious CPE (i.e. fictitious_cpe)

The NEW_VULNERABILITY notification will contain `PROJECTA` and `PROJECTB` in `subject.affectedProjects` node even though the notification rule is limited to `PROJECTA`. Moreover, the project `PROJECTA` will be present twice in the set.

### Expected Behavior:

The notification only contains `PROJECTA` once.

### Environment:
  
  - Dependency-Track Version: 4.6.0-SNAPSHOT
  - Distribution: Docker
  - BOM Format & Version: CycloneDX 1.4
  - Database Server: H2

### Additional Details:

[Project](https://github.com/syalioune/dependency-track/blob/master/src/main/java/org/dependencytrack/model/Project.java) data model does not override equals method, so any `Set` property is not enforced.

Furthermore, Dependency Track does not perform component deduplication on SBOM upload, nor [cyclonedx-cli](https://github.com/CycloneDX/cyclonedx-cli) on merge. Should that be an option in either projects ?

syalioune:
Fixes #1866

Signed-off-by: Alioune SY <sy_alioune@yahoo.fr>

Project deduplication in `NotificationUtil` is debatable as we can say it is not Dependency Track job to compensate input data quality (e.g : SBOM merge with duplicates). I agree for the most part (metrics calculation, vulnerability assignment...) but for notification, I feel that it is a sound compromise.

### DIFF
--- a/src/main/java/org/dependencytrack/notification/NotificationRouter.java
+++ b/src/main/java/org/dependencytrack/notification/NotificationRouter.java
@@ -22,12 +22,12 @@ import alpine.common.logging.Logger;
 import alpine.notification.Notification;
 import alpine.notification.NotificationLevel;
 import alpine.notification.Subscriber;
-import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.dependencytrack.exception.PublisherException;
 import org.dependencytrack.model.NotificationPublisher;
 import org.dependencytrack.model.NotificationRule;
 import org.dependencytrack.model.Project;
 import org.dependencytrack.notification.publisher.Publisher;
+import org.dependencytrack.notification.vo.AnalysisDecisionChange;
 import org.dependencytrack.notification.vo.BomConsumedOrProcessed;
 import org.dependencytrack.notification.vo.NewVulnerabilityIdentified;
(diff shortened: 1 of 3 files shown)


nscuro (approved):
Thanks @syalioune!

PR #1907, opened 2022-08-24
Add LISTEN_PORT environment variable to enable configuration of HTTP …

invadersmustdie:
In order to provide more flexible configuration for container based deployments it's necessary to configure the port the application is listening on

### DIFF
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -20,6 +20,8 @@ ENV TZ=Etc/UTC \
     # The web context defaults to the root. To override, supply an alternative context which starts with a / but does not end with one
     # Example: /dtrack
     CONTEXT="/" \
+    # default port to listens on. To override set LISTEN_PORT environment variable
+    LISTEN_PORT="8080" \
     # Injects the build-time ARG "WAR_FILENAME" as an environment variable that can be used in the CMD.
     WAR_FILENAME=${WAR_FILENAME} \
     # Set JAVA_HOME for the copied over JRE
@@ -59,7 +61,7 @@ USER ${UID}
 WORKDIR ${APP_DIR}
 
 # Launch Dependency-Track
-CMD java ${JAVA_OPTIONS} -DdependencyTrack.logging.level=${LOGGING_LEVEL} -jar ${WAR_FILENAME} -context ${CONTEXT}
+CMD java ${JAVA_OPTIONS} -DdependencyTrack.logging.level=${LOGGING_LEVEL} -jar ${WAR_FILENAME} -context ${CONTEXT} -port ${LISTEN_PORT}
 
 # Specify which port Dependency-Track listens on
 EXPOSE 8080


src/main/docker/Dockerfile
```diff
@@ -59,7 +61,7 @@ USER ${UID}
 WORKDIR ${APP_DIR}
 
 # Launch Dependency-Track
-CMD java ${JAVA_OPTIONS} -DdependencyTrack.logging.level=${LOGGING_LEVEL} -jar ${WAR_FILENAME} -context ${CONTEXT}
+CMD java ${JAVA_OPTIONS} -DdependencyTrack.logging.level=${LOGGING_LEVEL} -jar ${WAR_FILENAME} -context ${CONTEXT} -port ${LISTEN_PORT}
 
 # Specify which port Dependency-Track listens on
 EXPOSE 8080
```
nscuro, line 67: If we set the exposed port at runtime via environment variables, then the `EXPOSE` statement here will not be accurate anymore. `EXPOSE` can't be used with environment variables, so we'd have to remove it entirely.

That may be a breaking change if infrastructure components rely on containers defining `EXPOSE`. That is the case for traefik for example: https://doc.traefik.io/traefik/providers/docker/#port-detection

valentijnscholten: isn't it already quite easy to change the externally used port via docker run command or the docker compose file?

nscuro: @valentijnscholten I suppose both API server and Frontend binding to port 8080 can become an issue when trying to run them in the same k8s pod. Port mapping doesn't really help there. I guess this or something similar is what @invadersmustdie is trying to do.

invadersmustdie: That's is exactly my use-case. I'm trying to deploy dependency-track api-server + frontend within the same pod. As both containers are exposing its service on :8080 thats kind of impossible without any hacks. 

ATM I'm doing some env/arg-injection to achieve this, but I would prefer to have a dedicated env var to get there.

hack:

```
        env:
        - name: CONTEXT
          value: / -port 8081
```

kentaleksandrovrx: @invadersmustdie Can you clarify if/how the hack works. I've rebuilt the image with variations of port arguments slapped on java command 
```
-Djetty.port=${PORT}
-Djetty.http.port=${PORT}
-port=${PORT}
-Dserver.port=${PORT}
```
But none of these appear to be respected

nscuro: @kentaleksandrovrx It's (ab-)using the fact that the `CONTEXT` environment variable is injected into the `Dockerfile`'s `CMD` line:

https://github.com/DependencyTrack/dependency-track/blob/e9304da3beba4776784da9104edcefbf6da0b32f/src/main/docker/Dockerfile#L62

The Dependency-Track JAR accepts a `-port` argument that defines on which port the embedded Jetty will listen on.

Note that this will cause the `HEALTHCHECK` command to fail, because `CONTEXT` is used in there, too: https://github.com/DependencyTrack/dependency-track/blob/e9304da3beba4776784da9104edcefbf6da0b32f/src/main/docker/Dockerfile#L68

nscuro:
FWIW making this work would be easier on the frontend side.

You can simply bind-mount a `ConfigMap` with your custom NGINX config to `/etc/nginx/conf.d/default.conf`. In the config, specify a different port than `8080`:

* https://github.com/DependencyTrack/frontend/blob/4.7.0/docker/Dockerfile.alpine#L33
* https://github.com/DependencyTrack/frontend/blob/4.7.0/docker/etc/nginx/conf.d/default.conf#L2

src/main/docker/Dockerfile
invadersmustdie: I see your arguments and I also the disadvantages of this hacky solution :) I will close this PR. Thanks for the feedback

PR #1912, opened 2022-08-25
Adding support for vulnerability aliases

stevespringett:
Implements #1642


src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -467,4 +484,82 @@ public List<Project> getProjects(Vulnerability vulnerability) {
...
+
+    public synchronized VulnerabilityAlias synchronizeVulnerabilityAlias(VulnerabilityAlias alias) {
+        final Map<String, Object> params = new HashMap<>();
+        String filter = "";
+        if (alias.getCveId() != null) {
+            filter += "(cveId == :cveId)";
+            params.put("cveId", alias.getCveId());
+        }
+        if (alias.getSonatypeId() != null) {
+            if (filter != null) filter += " && ";
```
sahibamittal, line 496: This will always be false as filter is set as an empty string. 
Need to change it to `if (!filter.isEmpty())`

sahibamittal:
Hey @stevespringett - This is good, I tested it as well. 
But, its missing code to add aliases in OSVdownloadTask, like below code snippet needs to be added in method '**mapAdvisoryToVulnerability**'

`if (advisory.getAliases() != null) {
            for (int i=0; i<advisory.getAliases().size(); i++) {
                String alias = advisory.getAliases().get(i);
                final VulnerabilityAlias vulnerabilityAlias = new VulnerabilityAlias();
                vulnerabilityAlias.setOsvId(advisory.getId());
                if(alias.startsWith("CVE")) {
                    vulnerabilityAlias.setCveId(alias);
                } else if (alias.startsWith("GHSA")) {
                    vulnerabilityAlias.setGhsaId(alias);
                }
                qm.synchronizeVulnerabilityAlias(vulnerabilityAlias);
            }
}`

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -467,4 +484,82 @@ public List<Project> getProjects(Vulnerability vulnerability) {
...
+    @SuppressWarnings("unchecked")
+    public List<VulnerabilityAlias> getVulnerabilityAliases(Vulnerability vulnerability) {
+        final Query<VulnerabilityAlias> query;
+        if (Vulnerability.Source.NVD.name().equals(vulnerability.getSource())) {
+            query = pm.newQuery(VulnerabilityAlias.class, "cveId == :cveId");
+        } else if (Vulnerability.Source.OSSINDEX.name().equals(vulnerability.getSource())) {
+            query = pm.newQuery(VulnerabilityAlias.class, "sonatypeId == :sonatypeId");
+        } else if (Vulnerability.Source.GITHUB.name().equals(vulnerability.getSource())) {
+            query = pm.newQuery(VulnerabilityAlias.class, "ghsaId == :ghsaId");
+        } else if (Vulnerability.Source.VULNDB.name().equals(vulnerability.getSource())) {
```
sahibamittal, line 558: Need addition of osv id as well for google datasource:
`else if (Vulnerability.Source.OSV.name().equals(vulnerability.getSource())) {
            query = pm.newQuery(VulnerabilityAlias.class, "osvId == :osvId");
        } `

nscuro:
Should we offer some sort of server-side deduping for the `/finding` API? Right now it'd be the client's responsibility of deduping findings based on the aliases. But maybe we should offer a basic solution as well to improve usability.

It should be an explicit opt-in, or potentially an entirely different endpoint like `/finding/<PROJECT_UUID>/unique`.
The format could be such that `finding[].vulnerability` refers to the vulnerability as defined by the authoritative source (e.g. CVE -> NVD), with an `aliases` or `references` field populated with either:

* entire vulnerability objects of the aliases
* UUID or source-vulnId pairs of the aliases (this is how it's done right now)

In the later case, user's can then just fetch the additional info if desired.

stevespringett:
> Should we offer some sort of server-side deduping for the /finding API?

Yes, I'd like to. Currently, the findings api is being augmented with aliases for each finding so the client can de-dup.

If we choose to do this, we need to take into consideration the algorithm:
* The most desired approach would be to favor CVE's over any of the alternative identifiers. This assumption should not be hard-coded.
* There will be occasions where a CVE does not exist, yet there are aliases between say GHSA and OSSINDEX. Need to figure out how to handle this case.
* There will be occasions where a CVE does not exist initially, but a OSSINDEX finding does. That OSSINDEX finding could be audited. At a later time, a CVE may be created and now there's a mapping. This happened with log4j and is likely the norm for high-profile vulnerabilities. I don't think we want to de-dup any finding that has an existing audit.

There may be other cases as well. These are the ones that I could immediately think of. But I do think this should be implemented in a different PR as some thought will need to be made into the design and configuration of the approach.

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
stevespringett: Good catch. Resolved in https://github.com/DependencyTrack/dependency-track/pull/1912/commits/9d200158d2c6686d56ba74f6d15b34f88f550060

nscuro:
> There will be occasions where a CVE does not exist initially, but a OSSINDEX finding does. That OSSINDEX finding could be audited.

True. Just brainstorming here, how about we rewire analyses to be associated with a `VulnerabilityAlias` instead of a `Vulnerability`? Would require a more complex migration, but would drastically simplify this case.

stevespringett:
> how about we rewire analyses to be associated with a VulnerabilityAlias instead of a Vulnerability?

VulnerabilityAliases do not exist for most vulnerabilities, only the ones that have aliases. This approach would need to take into consideration a future requirement we'll have to correct aliases in the event OSSI (or another provider) originally says a CVE is X and later come correct their own data and say its Y. This automated correction is not currently in place, but will need to be.

stevespringett:
@nscuro @sahibamittal thoughts on merge? Is it ready?

src/main/java/org/dependencytrack/tasks/MetricsUpdateTask.java
```diff
@@ -426,14 +427,33 @@ MetricCounters updateProjectMetrics(final QueryManager qm, final long oid) {
...
+                            || (alias.getInternalId() != null && dedupedVulnIdentifiers.contains(Vulnerability.Source.INTERNAL.name() + "|" + alias.getInternalId()))
+                            || (alias.getCveId() != null && dedupedVulnIdentifiers.contains(Vulnerability.Source.NVD.name() + "|" + alias.getCveId()))
+                            || (alias.getSonatypeId() != null && dedupedVulnIdentifiers.contains(Vulnerability.Source.OSSINDEX.name() + "|" + alias.getSonatypeId()))
+                            || (alias.getGhsaId() != null && dedupedVulnIdentifiers.contains(Vulnerability.Source.GITHUB.name() + "|" + alias.getGhsaId()))
+                            || (alias.getVulnDbId() != null && dedupedVulnIdentifiers.contains(Vulnerability.Source.VULNDB.name() + "|" + alias.getVulnDbId())))) {
+                        found = true;
+                    }
+                }
+                if (!found) {
+                    counters.updateSeverity(vuln.getSeverity());
```
nscuro, line 453: This will be first-come-first-served WRT the severity. `getAllVulnerabilities` does not do any explicit sorting, resulting in the first vulnerability that was ingested to dictate the severity. This will result in different metrics across different DT instances.

Should this be more deterministic?

nscuro (commented):
It's a bit confusing that GSD and Snyk are mentioned in a few places, although those sources aren't yet supported and we don't have an entry in the `Vulnerability.Source` enum for them yet. 

I'd prefer it if we left those out for now.

sahibamittal:
@stevespringett - For endpoint to getAllVulnerabilities for a project ("/project/{uuid}"), we're missing this explicit setting of aliases to vulnerability object in **VulnerabilityQueryManager**. 
Need to do same in method '**getVulnerabilities**' being called by this endpoint.

nscuro:
We may also want to include the aliases in the notification subjects, which currently have their own mapping logic:

https://github.com/DependencyTrack/dependency-track/blob/a3b42a6fe4d19fc91122887dba0841f26fa7c7c4/src/main/java/org/dependencytrack/util/NotificationUtil.java#L260-L292

nscuro:
@stevespringett I reckon you are super busy ATM, would you like me to take on the remaining outstanding issues here?

src/main/java/org/dependencytrack/tasks/MetricsUpdateTask.java
stevespringett: We have to include GSD since its included in OSV. But we can leave our Snyk for the time being.

stevespringett:
@sahibamittal I cannot find a place where aliases are missing.

stevespringett:
> We may also want to include the aliases in the notification subjects, which currently have their own mapping logic:
> 
> https://github.com/DependencyTrack/dependency-track/blob/a3b42a6fe4d19fc91122887dba0841f26fa7c7c4/src/main/java/org/dependencytrack/util/NotificationUtil.java#L260-L292

This should likely be an additional enhancement as it will have to involve changes to all of the notification templates as well.

nscuro:
> This should likely be an additional enhancement as it will have to involve changes to all of the notification templates as well.

Aye. I raised https://github.com/DependencyTrack/dependency-track/issues/1992 for it.

nscuro:
@stevespringett Is this ready to go from your side?

Issue #1903, opened 2022-08-23
OSV : replace ecosystem hardcoding with list of ecosystems managed by Google


sahibamittal:
### Current Behavior: 
For OSV mirroring, we hard code the list of ecosystems to get vulnerabilities.

### Proposed Behavior: 
OSV has enabled a list of ecosystems they maintain -> https://osv-vulnerabilities.storage.googleapis.com/ecosystems.txt

Issue: https://github.com/google/osv.dev/issues/619

### Proposed UI interaction: 
While enabling the OSV mirroring in UI, user can select (list or all) ecosystems to mirror. This will reduce load if all ecosystems are not required.

sahibamittal:
@VinodAnandan @nscuro

nscuro:
Oh wow, that was quick. 😄

PR #1927, opened 2022-09-01
Issue 1903 : removed ecosystem hardcoding

sahibamittal:
Issue: https://github.com/DependencyTrack/dependency-track/issues/1903


sahibamittal:
@VinodAnandan @nscuro

src/main/java/org/dependencytrack/resources/v1/OsvEcosytemResource.java
```diff
@@ -0,0 +1,57 @@
...
+import org.dependencytrack.tasks.OsvDownloadTask;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.List;
+
+@Path("/v1/ecosystem")
```
nscuro, line 38: This path is misleading. It's OSV-specific, this needs to be reflected here.

We should also consider that there may be similar endpoints for other integrations in the future. So something like `/v1/osv/ecosystem` may still be too specific. Maybe something like `/v1/integration/osv/ecosystem`?

src/main/java/org/dependencytrack/resources/v1/OsvEcosytemResource.java
```diff
@@ -0,0 +1,57 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(
+            value = "Returns a list of all ecosystems in OSV",
+            response = String.class,
+            responseContainer = "List"
+    )
+    @ApiResponses(value = {
+            @ApiResponse(code = 401, message = "Unauthorized")
+    })
+    @PermissionRequired(Permissions.Constants.ACCESS_MANAGEMENT)
```
nscuro, line 52: This should be `SYSTEM_CONFIGURATION` instead.

src/main/java/org/dependencytrack/resources/v1/OsvEcosytemResource.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.List;
+
+@Path("/v1/ecosystem")
+@Api(value = "ecosystem", authorizations = @Authorization(value = "X-Api-Key"))
+public class OsvEcosytemResource extends AlpineResource {
```
nscuro, line 40: There should be a test for this class.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -297,4 +308,26 @@ private boolean isVulnerabilitySourceClashingWithGithubOrNvd(String source) {
...
+        ArrayList<String> ecosystems = new ArrayList<>();
+        String url = "https://osv-vulnerabilities.storage.googleapis.com/" + "ecosystems.txt";
+        request = new HttpGet(url);
+        try (final CloseableHttpResponse response = HttpClientPool.getClient().execute(request)) {
+            final StatusLine status = response.getStatusLine();
+            if (status.getStatusCode() == 200) {
+                try (InputStream in = response.getEntity().getContent();
+                     Scanner scanner = new Scanner(in, StandardCharsets.UTF_8.name())) {
+                    while (scanner.hasNextLine()) {
+                        ecosystems.add(scanner.nextLine().trim());
```
nscuro, line 322: Checking for blank lines may be a good idea here. Just so that we don't run into issues when the file ends with an empty line for example.

src/main/java/org/dependencytrack/persistence/DefaultObjectGenerator.java
```diff
@@ -289,4 +284,20 @@ private void loadDefaultNotificationPublishers() {
...
+            LOGGER.info("Synchronizing OSV ecosystems to datastore");
+            List<String> ecosystems = OsvDownloadTask.getEcosystems();
+            if(!ecosystems.isEmpty()) {
+                ecosystems.forEach(ecosystem -> {
+                    LOGGER.debug("Creating config property: " + OsvDownloadTask.OSV_CONFIG_GROUP + " / " + ecosystem);
+                    qm.createConfigProperty(OsvDownloadTask.OSV_CONFIG_GROUP, ecosystem, "false", IConfigProperty.PropertyType.BOOLEAN, ecosystem);
+                });
+            }
+        }
+    }
```
nscuro, line 302: Do we really need this? Without having tested it, the code looks like it should be able to work just fine without the ecosystem properties being pre-populated.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -39,25 +38,36 @@
...
-import static org.dependencytrack.util.VulnerabilityUtil.*;
+import static org.dependencytrack.util.VulnerabilityUtil.normalizedCvssV3Score;
+import static org.dependencytrack.util.VulnerabilityUtil.normalizedCvssV2Score;
 
 public class OsvDownloadTask implements LoggableSubscriber {
 
     private static final Logger LOGGER = Logger.getLogger(OsvDownloadTask.class);
 
     private static final String OSV_BASE_URL = "https://osv-vulnerabilities.storage.googleapis.com/";
+    public static final String OSV_CONFIG_GROUP = "osv-ecosystems";
```
nscuro, line 55: Please adhere to the existing naming scheme (group: `vuln-source`), see https://github.com/DependencyTrack/dependency-track/blob/7685a29143dbb23f6d63af20ca4845790d81023b/src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java#L50-L51

Maybe instead of creating separate properties for every ecosystem, how do you feel about making this a single property? It could then be a string of enabled ecosystems, e.g.:
```
vuln-source.osv.ecosystems.enabled=Debian;Go;Maven
```

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -297,4 +308,26 @@ private boolean isVulnerabilitySourceClashingWithGithubOrNvd(String source) {
         return Vulnerability.Source.GITHUB.toString().equals(source)
                 || Vulnerability.Source.NVD.toString().equals(source);
     }
+
+    public static List<String> getEcosystems() {
+        ArrayList<String> ecosystems = new ArrayList<>();
+        String url = "https://osv-vulnerabilities.storage.googleapis.com/" + "ecosystems.txt";
```
nscuro, line 314:
```suggestion
        String url = "https://osv-vulnerabilities.storage.googleapis.com/ecosystems.txt";
```

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -39,25 +38,36 @@
...
 public class OsvDownloadTask implements LoggableSubscriber {
 
     private static final Logger LOGGER = Logger.getLogger(OsvDownloadTask.class);
 
     private static final String OSV_BASE_URL = "https://osv-vulnerabilities.storage.googleapis.com/";
+    public static final String OSV_CONFIG_GROUP = "osv-ecosystems";
     private final boolean isEnabled;
-    private HttpUriRequest request;
+    private final List<String> ecosystems;
+    private static HttpUriRequest request;
```
nscuro, line 58: Sharing request objects will lead to race conditions and all other sorts of issues.

Because this field is shared for downloading `ecosystems.txt` and `all.zip`, what will happen if an `all.zip` file is being downloaded while someone requests `/api/v1/ecosystem`?

It's best to scope request objects to where they're used.

nscuro (changes requested):

src/main/java/org/dependencytrack/persistence/DefaultObjectGenerator.java
sahibamittal: Right, we can remove the default generator here. We'll need a new endpoint in ConfigPropertyResource to **upsert** the config property so that frontend can call that endpoint for osv ecosystem selection. 
Currently, it just updates existing config properties. Anyway we'll need to create a config property (from frontend) whenever user selects a newly added ecosystem from OSV.

sahibamittal:
Screenshot 2022-09-06 at 11 46 30 a m

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -297,4 +310,28 @@ private boolean isVulnerabilitySourceClashingWithGithubOrNvd(String source) {
...
+        String url = OSV_BASE_URL + "ecosystems.txt";
+        HttpUriRequest request = new HttpGet(url);
+        try (final CloseableHttpResponse response = HttpClientPool.getClient().execute(request)) {
+            final StatusLine status = response.getStatusLine();
+            if (status.getStatusCode() == 200) {
+                try (InputStream in = response.getEntity().getContent();
+                     Scanner scanner = new Scanner(in, StandardCharsets.UTF_8.name())) {
+                    while (scanner.hasNextLine()) {
+                        if(!scanner.nextLine().isBlank()) {
+                            ecosystems.add(scanner.nextLine().trim());
```
nscuro, line 325: The method currently does not return all available ecosystems, but instead:

```
["DWF","Debian:10","Debian:3.0","Debian:4.0","Debian:6.0","Debian:8","GSD","Go","JavaScript","Maven","OSS-Fuzz","Pub","RubyGems","crates.io"]
```

Further, an exception is logged every time it's called:

```
2022-09-06 19:03:42,890 ERROR [OsvDownloadTask] Exception while executing Http request for ecosystems
java.util.NoSuchElementException: No line found
	at java.base/java.util.Scanner.nextLine(Scanner.java:1651)
	at org.dependencytrack.tasks.OsvDownloadTask.getEcosystems(OsvDownloadTask.java:325)
	at org.dependencytrack.resources.v1.OsvEcosytemResource.getAllEcosystems(OsvEcosytemResource.java:54)
```

The issue is that `scanner.nextLine()` advances the underlying "pointer" on each invocation, so you're always skipping a line here, and in the end you're trying to advance the scanner beyond the last line.

```suggestion
                        final String line = scanner.nextLine();
                        if(!line.isBlank()) {
                            ecosystems.add(line.trim());
```

nscuro (changes requested):

nscuro (approved):

Issue #1861, opened 2022-08-05
Migrate from NVD data feeds to NVD API 2.0


olafz:
The NVD announced changes to the API and Data Feeds, see https://nvd.nist.gov/General/News/changes-to-feeds-and-apis.

One of the changes is a rate-limit on the API:

> In October 2021, [the NVD announced](https://nvd.nist.gov/general/news/API-Key-Announcement) the availability of API keys and changes to its API rate limits. Users who request and activate a key may include it as a parameter of their request’s URL string.
> 
> Beginning immediately, users transmitting requests without a key will see a reduction in the number of requests they can make in a rolling 60 second window. Users transmitting requests that include their API key will see no change in service and may continue to make requests at the current rate. New users may [request an API key here](https://nvd.nist.gov/developers/request-an-api-key).

It would be nice to be able to enter a NVD API key in DependencyTrack.

valentijnscholten:
To use the API key, DT would first need to switch to use the API instead of the data feeds.

nscuro:
Thanks @olafz, I modified the title of your issue to reflect that we indeed have to migrate away from our legacy data feed consumption to the NVD's new API, which will also include API key usage.

If I got it right, we have to do this until 2024:

> In late 2022 the NVD will release the 2.0 version of its APIs.
> [...]
> Approximately 6 months after the release of the 2.0 APIs the NVD will retire all RSS feeds. **Approximately 12 months after the release of the 2.0 APIs the NVD will also retire all remaining data feeds**.

Obviously it'd be beneficial to migrate way earlier than that.

stevespringett:
The announcement from the NVD will lead to a number of issues DT will have to deal with.

1) Mirroring time will increase, especially for default installations without an NVD API key
2) The announcement did not provide details about the v2 API. I don't want to use the v1 APIs at this time since, per their documentation, dictionaryCpes "may become truncated" without any specific guidance for how to avoid other than "Reducing the resultsPerPage may prevent the data from being truncated" which isn't very helpful.
3) The current API uses the CVE JSON v4.0 spec, which is now old. v5.0 is the current spec with v5.1 in development.
4) DT will need to track internally where the mirroring left off and how to keep the mirror up to date without having to re-mirror everything

Marck:
Hi all, are there any updates on this?

nscuro:
[...]

Issue #13, opened 2015-01-13
Add ThreadFix integration


stevespringett:
When viewing vulnerabilities, provide an option for a user to create an issue in ThreadFix. This feature will require a system-wide configuration option in which an admin can specify the URL and API KEY to use when connecting to ThreadFix.

stevespringett:
This feature will no longer be implemented since a remote importer is planned. #49

PR #1929, opened 2022-09-02
Change compareVersion behaviour

florentulve:
NA in VulnSoftware only match when target version is ANY


nscuro:
Thanks for the PR @florentulve! Makes sense to me at first glance. 

@officerNordberg, as you dabble with CPEs a lot, may I ask for you opinion on this?

officerNordberg:
I'm torn by this.  The test case is an exact problem that we've wrestled with.  I chose to handle this with a strategy similar to @nscuro 's project for managing audit results via code and policies.  I like the results this code change gives but it does not match the CPE matching spec as I read it. https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf So either I'm misreading it or we'd be choosing to go against the spec because NIST has terrible data for older CVEs.

ChaosInTheCRD:
Hey :) I am probably misunderstanding, but shouldn't

```
        if (LogicalValue.NA.getAbbreviation().equals(vs.getVersion()) && LogicalValue.ANY.equals(targetVersion) ) {
```

instead be

```
        if (LogicalValue.NA.getAbbreviation().equals(targetVersion) && LogicalValue.ANY.equals(targetVersion) ) {
```

Reason being is that in this instance, the first if argument doesn't compare at all to the target version...

ChaosInTheCRD:
Oh of course, I think I get it. Is the first argument in the `if` saying "if there is an NA in the VulnSoftware", then let me know of the vulnerability.

Correct me if I am wrong, but do I not want to be given back ALL of the VulnSoftwares if the `targetVersion` is ANY? By definition, I am saying "my target version could be anything", and therefore Dependency Track could not possibly be able to distinguish for me which vulns I do and don't care about

florentulve:
Could we simply drop the first lines of compareVersion which handles logical value NA cause I've noticed CPE matching seems to be handled correctly lower by compareAttributes ?

syalioune:
Following discussion https://github.com/DependencyTrack/dependency-track/discussions/2188, I think that an agreement should be reached.

In any case, there is indeed a bug as comparing `cpe:2.3:a:xiph:speex:1.2:-:*:*:*:*:*:*` to `cpe:2.3:a:xiph:speex:1.2:-:*:*:*:*:*:*` lead to a no match because https://github.com/DependencyTrack/dependency-track/blob/836da60abf648e3fc091d724aef3b3d0071ba7b4/src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java#L201-L203 happens before https://github.com/DependencyTrack/dependency-track/blob/836da60abf648e3fc091d724aef3b3d0071ba7b4/src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java#L216

can detect the exact match of `target` between vulnerable software and component.

syalioune:
Just like @officerNordberg my understanding as per https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf is that `cpe:2.3:a:apache:http_server:-:*:*:*:*:*:*:*` should not match `cpe:2.3:a:apache:http_server:2.4.53:*:*:*:*:*:*:*`

melba-lopez:
@syalioune @florentulve  This PR has been inactive for some time.  Can I proceed with closing this? otherwise, can you update your branch?

syalioune:
Hello @melba-lopez 
Sorry for the late answer. I think the _blocking_ point on this PR is not a technical one but a functional one. 

With :

- Source (NVD) : `cpe:2.3:a:apache:http_server:-:*:*:*:*:*:*:*`
- Target (Component CPE) : `cpe:2.3:a:apache:http_server:*:*:*:*:*:*:*:*`

we fall under the fifth case of [NISTIR 7696 Table 6-2](https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf)

image

**Component CPE is therefore a superset of NVD vulnerable software and thus it should not be a match.**

**If the above is correct and strict spec enforcement is what we want to achieve**, then that PR should be updated/merged. Otherwise, it should be closed.

Jasper-Ben:
Hi all,

maybe a bit OT for this merge request, but @syalioune directed me here.

I have been doing a deep dive into the CPE Name Matching specification, due to incomplete component CPE information leading to false negatives (attributes containing "ANY" values), as (for example) described at https://github.com/DependencyTrack/dependency-track/issues/2580. While component SHOULD have a information-complete CPE, it struck me odd, that a "ANY" value in an target attribute should lead to a no-match.

For example: Why would an CVE CPE with attribute TargetHW=x86 not be a match for an otherwise identical component CPE which has TargetHW="ANY"? Logically, the CVE SHOULD be a match for the component since any other decision would lead to false negatives.

It then struck me: The root cause for false negatives is the decision that DT does not consider CPE name relations being a SUBSET as a match. If we would consider SUBSET relations a match (that is cases 5, 13 and 15), the problem of false negatives due to incomplete data will go away, as described in https://github.com/DependencyTrack/dependency-track/issues/2988.

The flip-side however is, that inaccurate component CPEs will lead to false positives. This is to be expected. If your component reports TargetHW as "ANY" but in reality only runs on x86, then it will logically cause false positives for a CVE CPE with TargetHW set as arm64. That is why accurate component CPEs matters.

Jasper-Ben:
I guess what I'm trying to say is that we should maybe take a step back and re-evaluate the current matching behavior as a whole.

Jasper-Ben:
I might be stating the obvious, but are we aware that there is CPE Name Matching Pseudocode in the NIST standard itself?

nscuro:
Thanks all for the research and critical looks at this area!

The dilemma that I personally have with altering the existing logic is that it's *very* hard to estimate the impact it will have on users. Will they be flooded by false positives due to bad CPE data in both the SBOMs they ingest and the NVD? The fact that we have not nearly enough tests around this (https://github.com/DependencyTrack/dependency-track/issues/2243) just makes this worse.

The NVD's CPE data is also notoriously bad, which to my understanding is what resulted in DT's "non-standard" behavior as it is today. I am very much in favor of revisiting our matching logic, and harden it with silly amounts of tests. And personally, I'd love it to be aligned with the specification as much as possible.

As a side note, the fact that we have lots of different interpretations (including mine, which I haven't posted) of something that is considered to be an official specification in this thread alone is slightly amusing.

Jasper-Ben:
Hi @nscuro,

> The dilemma that I personally have with altering the existing logic is that it's very hard to estimate the impact it will have on users. Will they be flooded by false positives due to bad CPE data in both the SBOMs they ingest and the NVD?

I understand your concern, however I for one (and of course I can only speak for myself) would rather have false positives to deal with than false negatives lulling me into a false sense of security. And as it currently stands there definitely are loads of false negatives the way matching is implemented right now.

> The NVD's CPE data is also notoriously bad, which to my understanding is what resulted in DT's "non-standard" behavior as it is today.

I obviously did not have a lot of dealing with CPE so far, is this really the case? How do other projects deal with this? Are we sure it isn't just a case of e.g. interpreting old CPE 2.2 data (wrong)?

As it currently stands, the CPE matching mechanism is a hard blocker for our use-case (Yocto embedded linux), due to a lack of complete CPE data. I have also raised this issue [there](https://lists.openembedded.org/g/openembedded-core/topic/100962508#186724), but even if this is resolved on Yocto's site (which will be quite some pain), I fear this is a huge issue for other use-cases as well.

Jasper-Ben:
If we cannot agree on how SUBSETs are handled, maybe we could introduce that as a configurable option (similar to fuzzy)?

Since the subset case only occurs when the component (target) CPE is incomplete, aka uses "ANY" where CVE (source) CPE doesn't, the subset question boils down to "How do I treat incomplete component CPEs" with two options:

a) no-match with the risk of having false negatives
b) match with the risk of having false positives

So this can be put behind a switch that is easily understood by the user.

nscuro:
@Jasper-Ben Thanks, I really do appreciate your engagement.

I think there's a misunderstanding here regarding the project's position, perhaps because I failed to bring across what I intended to say.

Quoting your comment from https://github.com/DependencyTrack/dependency-track/issues/2994#issuecomment-1701221042:

> I hope the maintainers of DT will reconsider their stance on this. Frankly what concerns me is that DT seemingly is not fazed by the fact that their tool that should provide the solid foundation for determining the safety of your product actively hides security issues in favor of the "UX" of not having potential false positives reported.

Apologies if my comment came across as disregarding your concerns. DT definitely should not hide legitimate vulnerabilities from users. I totally agree that this is a no-go. And based on the various issues raised on this topic, the current situation is definitely sub-optimal. 

What I *tried* to communicate is:
* I'm not too familiar with this part of the code
* Decisions were made in the past that appear to deviate from the specification
* The test coverage on this area is very sparse
* Touching this central piece of code without proper test coverage that ensures it's behaving as intended is high risk

I raised https://github.com/DependencyTrack/dependency-track/issues/2243 to improve the test situation, and have more confidence in making changes here. But I simply did not yet have the time to get this done.

FWIW, the code was originally ported from DependencyCheck. DC's code [has since evolved](https://github.com/jeremylong/DependencyCheck/blob/4446f2b14ffa736523ebe213079e73c7ecdc2ec3/core/src/main/java/org/owasp/dependencycheck/dependency/VulnerableSoftware.java#L223-L293), but those changes never made it to DT it seems. What immediately jumps to the eye is that DC considers subsets to be a match, whereas DT does not.

@Jasper-Ben, with what you've learned so far by digging into this topic, would you mind double-checking the current DC implementation? Is that more aligned with what you expect to see?

> [...] and we (i.e. the company I work for) might be able to commit resources into getting this back on track. But as it stands I am not confident in getting the ball rolling, if we cannot agree on a desired output and end up spending resources on something that will end up not be merged back into DT.

We'd very much welcome this, and given proper test coverage I don't see us not accepting this contribution. 

> should we maybe organize a call regarding the general topic of "CPE matching"?

Happy to discuss this further if needed! Please feel free to reach out directly via email (see GitHub profile), or OWASP Slack (username there is `nscur0`).

officerNordberg:
> Happy to discuss this further if needed! Please feel free to reach out directly via email (niklas.duester@owasp.org, steve.springett@owasp.org), or OWASP Slack (username there is nscur0).

Not sure where I figure into all this @nscuro but I feel "attacked" 😉 or at least the semi-responsible person in the crosshairs.  I'd be open to direct harassment as well. drew.thompson@owasp.org My 2 cents, besides the fact that CPE matching in DT could very well use some improvements, is mostly that I've been working HARDER to sanitize and normalize my data as it all comes back to "what does NVD say?"  If I'm sending garbage CPEs I get garbage results.  Also true, unfortunately is, if I send perfect CPEs I may also get garbage results because some of NVD's CPE inclusion, especially years ago, is atrocious.

Jasper-Ben:
👋 @nscuro,

> The dilemma that I personally have with altering the existing logic is that it's very hard to estimate the impact it will have on users.

sorry, I guess I interpreted this wrong. To me it had the flair of "current situation might not be optimal, but would rather not touch". I'm happy to hear that you are open to re-evaluating the situation. 🙂 

@officerNordberg happy to directly harass you as well 😉 

> would you mind double-checking the current DC implementation?

Sure! will give it a spin and reach out to you guys (probably tomorrow).

Jasper-Ben:
I had a look at how DependencyCheck does the matching.

For the most part (we will look at the exceptions later) it relies on the following library: https://github.com/stevespringett/CPE-Parser.

[The CPE-Parser library does comparison as follows](https://github.com/stevespringett/CPE-Parser/blob/master/src/main/java/us/springett/parsers/cpe/Cpe.java#L607C1-L639C6):

```
    protected static boolean compareAttributes(String left, String right) {
        //the numbers below come from the CPE Matching standard
        //Table 6-2: Enumeration of Attribute Comparison Set Relations
        //https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf

        if (left.equalsIgnoreCase(right)) {
            //1 6 9 - equals
            return true;
        } else if (LogicalValue.ANY.getAbbreviation().equals(left)) {
            //2 3 4 - superset (4 is undefined - treating as true)
            return true;
        } else if (LogicalValue.NA.getAbbreviation().equals(left) 
                && LogicalValue.ANY.getAbbreviation().equals(right)) {
            //5 - subset
            return true;
        } else if (LogicalValue.NA.getAbbreviation().equals(left)) {
            //7 8 - disjoint, undefined
            return false;
        } else if (LogicalValue.NA.getAbbreviation().equals(right)) {
            //12 16 - disjoint
            return false;
        } else if (LogicalValue.ANY.getAbbreviation().equals(right)) {
            //13 15 - subset
            return true;
        }
        //10 11 14 17
        if (containsSpecialCharacter(left)) {
            Pattern p = Convert.wellFormedToPattern(left.toLowerCase());
            Matcher m = p.matcher(right.toLowerCase());
            return m.matches();
        }
        return false;
    }
```
So this isn't 100% equivalent to the NIST standard, as some undefined cases be treated as match (or no-match), as stated by the author:

>  This does not follow the spec precisely because ANY compared to NA is
    classified as undefined by the spec; however, in this implementation ANY
    will match NA and return true.

But this shouldn't do too much "damage" and other than that, this is pretty much aligns with how I understand the specification.

[However, for the "Part" attribute the `compareAttributes` function works differently](https://github.com/stevespringett/CPE-Parser/blob/e0db903e37a4da6e20da3f156200b0007b8f32ea/src/main/java/us/springett/parsers/cpe/Cpe.java#L584C30-L584C47):

```
    protected static boolean compareAttributes(Part left, Part right) {
        if (left == right) {
            return true;
        } else if (left == Part.ANY) {
            return true;
        }
        return false;
    }
```

This part confuses me, as I do not see the value or necessity of this special treatment. In fact, this will again create matching issues if the component CPE does not contain "Part" information. I would also argue that this creates a deviation from the CPE name matching specification, which states:

image

Maybe the original author @jeremylong might be able give some insights into this decision.


Dependencycheck then additionally changes the matching behavior of the library for the following attribute comparisons:

1. ["Version" attribute](https://github.com/jeremylong/DependencyCheck/blob/4446f2b14ffa736523ebe213079e73c7ecdc2ec3/core/src/main/java/org/owasp/dependencycheck/dependency/VulnerableSoftware.java#L382):

Dependencycheck extends the version matching to add support for Version Ranges. I haven't looked at the code in detail (yet), but my understanding is that this checks the component version against the "Up to (excluding)" value of the CVE, to check whether the component version is still affected:

image

2. [ "Update" attribute](https://github.com/jeremylong/DependencyCheck/blob/4446f2b14ffa736523ebe213079e73c7ecdc2ec3/core/src/main/java/org/owasp/dependencycheck/dependency/VulnerableSoftware.java#L223):

Again, haven't looked into the code in detail (yet), but this works around some inaccuracies in CPE data, e.g. `a1` vs `alpha1`.

The idea behind these two changes in behavior for CVE matching seem reasonable to me, although we should take a closer look at the implementation.


In conclusion, to answer your question @nscuro:

> @Jasper-Ben, with what you've learned so far by digging into this topic, would you mind double-checking the current DC implementation? Is that more aligned with what you expect to see?

Generally, the matching specification of DC aligns much better with my understanding of the specification. The only concern I have is the separate "Part" attribute matching, which again I currently do not understand why it exists in the first place and will cause false negatives if the components "Part" attribute contains an "ANY" value.

jeremylong:
@Jasper-Ben the comparison for the `Part` is effectively identical to the other `compareAttributes` - the difference is [Part](https://github.com/stevespringett/CPE-Parser/blob/e0db903e37a4da6e20da3f156200b0007b8f32ea/src/main/java/us/springett/parsers/cpe/values/Part.java#L30) is an `enum` as opposed to a `String` - so not as many cases need to be checked.

Jasper-Ben:
Hi @jeremylong, thanks for the quick reply!

Correct me if I'm wrong, but my understanding is that since left is the CVE CPE, the code for "Part" will return `false` for cases 5, 13 and 15 from table 5-2 whereas these should be a match. I think this should be correct:

```
    protected static boolean compareAttributes(Part left, Part right) {
        if (left == right) {
            return true;
        } else if (left == Part.ANY || right == Part.ANY) {
            return true;
        }
        return false;
    }
```

jeremylong:
Looks like you are correct - but the number of times you would encounter this scenario would be really rare.

Jasper-Ben:
> Looks like you are correct - but the number of times you would encounter this scenario would be really rare.

Unfortunately this is exactly the scenario we are running into 😅

jeremylong:
Unless I'm misreading the issue - the problem is not the with the `Part` (i.e., `a`, `o`, `h`, `*`, `-`). It would be really rare to have the NVD put a `*` in the `Part` portion of the CPE.

Jasper-Ben:
> Unless I'm misreading the issue - the problem is not the with the `Part` (i.e., `a`, `o`, `h`, `*`, `-`). It would be really rare to have the NVD put a `*` in the `Part` portion of the CPE.

IMO the biggest issue in dependencytrack matching is this: https://github.com/DependencyTrack/dependency-track/issues/2988. In dependencycheck / your library that is done correctly (except for the "Part" attribute).

In addition, in our specific use-case (yocto embedded linux) we would additionally run into the "Part" attribute issue, as due to historic reasons yocto currently does not offer enough information about the component it is building. Until recently, they had "a" hardcoded, which would cause false negatives for operating system components, such as the linux kernel. In order to resolve this, I actually submitted a patch that changes this to ANY. While far from ideal, properly fixing this would require fundamental changes in yocto, so I thought this to be the lesser evil until this is [properly resolved in yocto itself](https://lists.openembedded.org/g/openembedded-core/topic/100962508#186724)

Jasper-Ben:
> It would be really rare to have the NVD put a * in the Part portion of the CPE.

To clearify: If I understand left and right in your library correctly, the issue would not arise when NVD puts * in a CVE Part, rather the false negative happens when the component Part is *.

Jasper-Ben:
@jeremylong FYI: I opened a PR to fix the part comparison: https://github.com/stevespringett/CPE-Parser/pull/180

jeremylong:
We might be thinking about the use of `matches` vs `matchedBy` differently - but whatever; you put a PR in to fix the bug. We'll need Steve to release a new version.  

I've always used it as `[cpe:search:string].matches([cpe:actual:cpe:for:a:library])`, and flip the CPEs for the `matchedBy`.

nscuro:
Thanks again for the great research @Jasper-Ben. I was not aware that the `cpe-parser` library we're apparently already using already includes all this logic, and a more recent / correct version of it as well. I wonder why DT is not using that, but instead rolls its own copy. It seems to me that at least part of the solution here is to ditch the outdated local copy of the logic in DT, and instead use what's provided by `cpe-parser` where possible. The test coverage in `cpe-parser` is *way* more assuring than what we have.

---

> @officerNordberg In my experience the largest source of false negatives is still the indexer.

Should be noted though that Lucene indexes are "only" used for fuzzy matching.

> [...] I suspect that another exists that corrupts the index.

Index writes / updates / deletes, as well as commits, are executed in a single thread. Are we talking full-blown corruption (e.g. exception on modification / read), or inconsistencies (e.g. index documents missing)?

> You can’t match what’s not there. I’m going to try setting the the threshold down to 5 % or lower and setting the interval to a half day. If that doesn’t work I may just setup a job to hit the re-index endpoint regularly

[Keeping search indexes consistent with the system of record](https://www.confluent.io/blog/using-logs-to-build-a-solid-data-infrastructure-or-why-dual-writes-are-a-bad-idea/) is hard, and TBH, DT currently does not do a good job of handling this today. I don't want to deviate too much from the original discussion here, but I think mid-term we may be able to resolve all these indexing issues by simply leveraging the DB's capabilities (e.g. [PostgreSQL's GIN index](https://about.gitlab.com/blog/2016/03/18/fast-search-using-postgresql-trigram-indexes/) or [FTS](https://www.crunchydata.com/blog/postgres-full-text-search-a-search-engine-in-a-database)).

Jasper-Ben:
> It seems to me that at least part of the solution here is to ditch the outdated local copy of the logic in DT, and instead use what's provided by cpe-parser where possible.

Agreed, why do the same work twice :+1: 

Personally, I am wondering how we want to deal with the custom modifications DC does to the checking mechanism for the "Version" and "Update" attributes. It seems to be something we would want to do in DT as well, which again would lead to maintaining separate code (although admittedly at a smaller scale). Mid-term, maybe we can move this custom logic into the CPE-Parser library itself, e.g. into its own function. For a quick short-term fix, code-duplication should be fine I guess.

jeremylong:
The Version and Update are evaluated in DC here: https://github.com/jeremylong/DependencyCheck/blob/3871517289d2cb0cf2c75cbe6ec229d280da8b8d/core/src/main/java/org/owasp/dependencycheck/dependency/VulnerableSoftware.java#L46

I think some of the `todo` comments are old and were not removed when the function was implemented.

Jasper-Ben:
Since a new version of the CPE-Parser library was released yesterday, which fixes the "Part" comparison we can now go full steam ahead on aligning the CPE matching. Does anyone feel up to it?

If not, I could probably take a look at this, but I am not familiar with the project structure and my Java skills are quite rusty, thus it would take me a while.

nscuro:
@Jasper-Ben Happy to work on it, but am a little flooded right now so it's possible I won't get to it fully before next week.

Jasper-Ben:
@nscuro That would be awesome! I don't think I would be able to finish it at an earlier date either.

Jasper-Ben:
:wave: @nscuro, any status update? Did you find the time to work on this?

nscuro:
Hey @Jasper-Ben, thanks for the ping. I did not have time yet, but did not forget about this either and still planning to do it ASAP.

nscuro:
One additional observation:

Before performing the actual CPE matching, DT fetches `VulnerableSoftware` "candidates" to match against from the database. This is done using the `part`, `vendor`, and `product` parts of the component's CPE:

https://github.com/DependencyTrack/dependency-track/blob/2a0e2d664b8e884859600bd2996e45c63d55776e/src/main/java/org/dependencytrack/tasks/scanners/InternalAnalysisTask.java#L141-L142

I'm assuming this is done to narrow down the `VulnerableSoftware`s to match against as much as possible, as matching on the other fields involves a bit more work.

The query however only considers equality:

https://github.com/DependencyTrack/dependency-track/blob/2a0e2d664b8e884859600bd2996e45c63d55776e/src/main/java/org/dependencytrack/persistence/VulnerableSoftwareQueryManager.java#L260-L263

Which means that wildcards in those three CPE parts are not expanded.
* `cpe:2.3:*:busybox:busybox:1.34.1:*:*:*:*:*:*:*` would only match if the NVD provides a CPE where the `part` part is *exactly*  `*`. It will not match when the NVD's CPE is `cpe:2.3:a:busybox:busybox:1.34.1:*:*:*:*:*:*:*`. 
* Similarly, a component with CPE `cpe:2.3:o:*:linux_kernel:4.19.139:*:*:*:*:*:*:*` will not match with the NVD's CPE `cpe:2.3:o:linux:linux_kernel:*:*:*:*:*:*:*:*`, because the `vendor` part is not *exactly* the same.

The solution being to use `LIKE` conditions for the query, whenever a given part contains (unescaped) wildcards.

This behavior is responsible for at least #2580, #2994, and one case of #2988.

nscuro:
While comparing findings reported by the NVD vs. what DT reports, I found more wrinkles.

---

Looking at `cpe:2.3:o:linux:linux_kernel:5.15.71:*:*:*:*:*:*:*` (from https://github.com/DependencyTrack/dependency-track/issues/2580#issuecomment-1691334865), the NVD *does* [report](https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&isCpeNameSearch=true&seach_type=all&query=cpe:2.3:o:linux:linux_kernel:5.15.71:*:*:*:*:*:*:*) significantly more CVEs (835 vs. 405 reported by DT).

*However*, the NVD search results include CVEs where the Linux Kernel is merely mentioned as "Running On" configuration, and the vulnerable components is completely different:

https://nvd.nist.gov/vuln/detail/CVE-2016-8963

image

I'd argue that DT is doing the right thing in *not* reporting those.

---

Among the vulnerabilities found by the NVD, but not by DT, is CVE-2023-42756.

This *is* a legitimate vulnerability of the Kernel, but for some reason DT did not have any CPEs and version ranges associated with it after mirroring the NVD.

Turns out this is due to older data overriding newer data during NVD mirroring. We still parse the [JSON feeds](https://nvd.nist.gov/vuln/data-feeds). We start with the `modified` feed, and then proceed with the yearly feeds:

https://github.com/DependencyTrack/dependency-track/blob/69d1248d19bcf826bf006fee02e151fd9c869822/src/main/java/org/dependencytrack/tasks/NistMirrorTask.java#L134-L142

It has been this way since DT 4.5 (introduced via #1557).

The problematic part is that the `modified` feed contains fresher data than, say, the `2023` feed:

```shell
jq '.CVE_Items[] | select(.cve.CVE_data_meta.ID == "CVE-2023-42756")' nvdcve-1.1-2023.json
```

```json
{
  "cve": {
    "data_type": "CVE",
    "data_format": "MITRE",
    "data_version": "4.0",
    "CVE_data_meta": {
      "ID": "CVE-2023-42756",
      "ASSIGNER": "secalert@redhat.com"
    },
    "problemtype": {
      "problemtype_data": [
        {
          "description": []
        }
      ]
    },
    "references": {
      "reference_data": [
        {
          "url": "https://access.redhat.com/security/cve/CVE-2023-42756",
          "name": "https://access.redhat.com/security/cve/CVE-2023-42756",
          "refsource": "MISC",
          "tags": []
        },
        {
          "url": "https://seclists.org/oss-sec/2023/q3/242",
          "name": "https://seclists.org/oss-sec/2023/q3/242",
          "refsource": "MISC",
          "tags": []
        },
        {
          "url": "https://bugzilla.redhat.com/show_bug.cgi?id=2239848",
          "name": "https://bugzilla.redhat.com/show_bug.cgi?id=2239848",
          "refsource": "MISC",
          "tags": []
        }
      ]
    },
    "description": {
      "description_data": [
        {
          "lang": "en",
          "value": "A flaw was found in the Netfilter subsystem of the Linux kernel. A race condition between IPSET_CMD_ADD and IPSET_CMD_SWAP can lead to a kernel panic due to the invocation of `__ip_set_put` on a wrong `set`. This issue may allow a local user to crash the system."
        }
      ]
    }
  },
  "configurations": {
    "CVE_data_version": "4.0",
    "nodes": []
  },
  "impact": {},
  "publishedDate": "2023-09-28T14:15Z",
  "lastModifiedDate": "2023-09-28T14:29Z"
}
```

```shell
jq '.CVE_Items[] | select(.cve.CVE_data_meta.ID == "CVE-2023-42756")' nvdcve-1.1-modified.json
```

```json
{
  "cve": {
    "data_type": "CVE",
    "data_format": "MITRE",
    "data_version": "4.0",
    "CVE_data_meta": {
      "ID": "CVE-2023-42756",
      "ASSIGNER": "secalert@redhat.com"
    },
    "problemtype": {
      "problemtype_data": [
        {
          "description": [
            {
              "lang": "en",
              "value": "CWE-362"
            }
          ]
        }
      ]
    },
    "references": {
      "reference_data": [
        {
          "url": "https://access.redhat.com/security/cve/CVE-2023-42756",
          "name": "https://access.redhat.com/security/cve/CVE-2023-42756",
          "refsource": "MISC",
          "tags": [
            "Patch",
            "Third Party Advisory"
          ]
        },
        {
          "url": "https://seclists.org/oss-sec/2023/q3/242",
          "name": "https://seclists.org/oss-sec/2023/q3/242",
          "refsource": "MISC",
          "tags": [
            "Exploit",
            "Mailing List",
            "Patch",
            "Third Party Advisory"
          ]
        },
        {
          "url": "https://bugzilla.redhat.com/show_bug.cgi?id=2239848",
          "name": "https://bugzilla.redhat.com/show_bug.cgi?id=2239848",
          "refsource": "MISC",
          "tags": [
            "Exploit",
            "Issue Tracking",
            "Patch",
            "Third Party Advisory"
          ]
        }
      ]
    },
    "description": {
      "description_data": [
        {
          "lang": "en",
          "value": "A flaw was found in the Netfilter subsystem of the Linux kernel. A race condition between IPSET_CMD_ADD and IPSET_CMD_SWAP can lead to a kernel panic due to the invocation of `__ip_set_put` on a wrong `set`. This issue may allow a local user to crash the system."
        }
      ]
    }
  },
  "configurations": {
    "CVE_data_version": "4.0",
    "nodes": [
      {
        "operator": "OR",
        "children": [],
        "cpe_match": [
          {
            "vulnerable": true,
            "cpe23Uri": "cpe:2.3:o:linux:linux_kernel:6.6:rc1:*:*:*:*:*:*",
            "cpe_name": []
          },
          {
            "vulnerable": true,
            "cpe23Uri": "cpe:2.3:o:linux:linux_kernel:6.6:rc2:*:*:*:*:*:*",
            "cpe_name": []
          },
          {
            "vulnerable": true,
            "cpe23Uri": "cpe:2.3:o:linux:linux_kernel:*:*:*:*:*:*:*:*",
            "versionEndExcluding": "6.6",
            "cpe_name": []
          }
        ]
      },
      {
        "operator": "OR",
        "children": [],
        "cpe_match": [
          {
            "vulnerable": true,
            "cpe23Uri": "cpe:2.3:o:redhat:enterprise_linux:9.0:*:*:*:*:*:*:*",
            "cpe_name": []
          }
        ]
      },
      {
        "operator": "OR",
        "children": [],
        "cpe_match": [
          {
            "vulnerable": true,
            "cpe23Uri": "cpe:2.3:o:fedoraproject:fedora:38:*:*:*:*:*:*:*",
            "cpe_name": []
          }
        ]
      }
    ]
  },
  "impact": {
    "baseMetricV3": {
      "cvssV3": {
        "version": "3.1",
        "vectorString": "CVSS:3.1/AV:L/AC:H/PR:L/UI:N/S:U/C:N/I:N/A:H",
        "attackVector": "LOCAL",
        "attackComplexity": "HIGH",
        "privilegesRequired": "LOW",
        "userInteraction": "NONE",
        "scope": "UNCHANGED",
        "confidentialityImpact": "NONE",
        "integrityImpact": "NONE",
        "availabilityImpact": "HIGH",
        "baseScore": 4.7,
        "baseSeverity": "MEDIUM"
      },
      "exploitabilityScore": 1.0,
      "impactScore": 3.6
    }
  },
  "publishedDate": "2023-09-28T14:15Z",
  "lastModifiedDate": "2023-10-02T17:57Z"
}
```

As of DT 4.7, we automatically prune CPEs and version ranges when they're no longer reported (#2098). 
So what is happening is, we first mirror the correct, current data, only to override it with stale data a bit later.

A short-term "fix" would be to simply mirror the `modified` feed at last again.

NVD write in their [docs](https://nvd.nist.gov/vuln/data-feeds):

> After performing a one-time import of the complete data set using the compressed JSON vulnerability feeds, the "modified" feeds should be used to keep up-to-date. 

This is not what DT is doing (and I'm not sure why). It will download and ingest the yearly feeds when they change.

Perhaps we should just migrate to the CVE API 2.0 (#1861) instead of putting too much effort into fixing the legacy mirroring behavior.

Jasper-Ben:
:wave: @nscuro Thanks again for looking into this!

> However, the NVD search results include CVEs where the Linux Kernel is merely mentioned as "Running On" configuration, and the vulnerable components is completely different [...]. I'd argue that DT is doing the right thing in not reporting those.

Completely agree! DT should only report on components actually present.

> So what is happening is, we first mirror the correct, current data, only to override it with stale data a bit later.

Nice catch!

> A short-term "fix" would be to simply mirror the modified feed at last again. [...]
> Perhaps we should just migrate to the CVE API 2.0 (https://github.com/DependencyTrack/dependency-track/issues/1861) instead of putting too much effort into fixing the legacy mirroring behavior.

Agreed, although we should use the mentioned short-term workaround for the immediate fix.

nscuro:
@Jasper-Ben I think I resolved most of the issues in https://github.com/DependencyTrack/dependency-track/pull/3070. Currently working on thorough testing, which includes digging up ancient issues that caused DT's matching logic to be how it is right now. Work is close to be ready for review.

nscuro:
Anyone in this thread willing to review https://github.com/DependencyTrack/dependency-track/pull/3070, or give those changes a spin? I can push container images to Docker Hub if necessary.

Jasper-Ben:
> Anyone in this thread willing to review #3070, or give those changes a spin? I can push container images to Docker Hub if necessary.

Sure thing! Just give me a image :+1:

nscuro:
Image pushed to [`dependencytrack/apiserver:3070-fix-cpe-matching`](https://hub.docker.com/layers/dependencytrack/apiserver/3070-fix-cpe-matching/images/sha256-c6943be4a1c69e3b705d49be666b6c7072758f39f385fd8b25f65876c0538a9d?context=explore) 🐳 

Note, please use `dependencytrack/frontend:snapshot` when testing with frontend, as the API server image is built on top of `4.9.0-SNAPSHOT`. You can use [this](https://github.com/DependencyTrack/dependency-track/blob/master/dev/docker-compose.yml) Compose file if you don't have another one handy already.

nscuro:
Thanks for the feedback @Jasper-Ben, I'll try to reproduce and include test cases for those. Can you share the SBOM you used for testing?

Jasper-Ben:
> Thanks for the feedback @Jasper-Ben, I'll try to reproduce and include test cases for those. Can you share the SBOM you used for testing?

Sure thing!
[bom.txt](https://github.com/DependencyTrack/dependency-track/files/12880860/bom.txt) (I know it is not perfect, e.g. ':' in component names, but that is a different story :sweat_smile:)

Also I have following VEX file which resolves some of the CVEs I expect to find. Might be helpful for cross-referencing.
[vex.txt](https://github.com/DependencyTrack/dependency-track/files/12880890/vex.txt)

nscuro:
At least in the test, the NGINX CPE is matching, both with and without version range ("up to (excluding)"):

https://github.com/nscuro/dependency-track/blob/2db9cd46d45f80d442a0362492f63952e59b514b/src/test/java/org/dependencytrack/tasks/scanners/InternalAnalysisTaskCpeMatchingTest.java#L225-L228

Can you confirm that CVE-2021-3618 is present in your DT instance? And if so, what do you see when clicking *View Details* -> *Affected Components*? The CPEs from the NVD should be listed there.

nscuro:
For reference, this is what I'm seeing in a new instance:

image

And DT ends up finding this CVE, along with 1170 other vulnerabilities for this BOM:

image

I noticed however that the internal analysis took longer than usual, I am assuming this is happening due to the extensive use of ANY on the CPEs included in the BOM, which requires DT to fetch more data from the DB.

---

As for the CPU vulnerabilities:

* CVE-2023-20583 has the CPE `cpe:2.3:h:amd:*:*:*:*:*:*:*:*:*`
* CVE-2022-24436 has the CPE `cpe:2.3:h:intel:*:*:*:*:*:*:*:*:*`

One of the components it's matching against currently has the CPE `cpe:2.3:*:*:xrandr:1.5.1:*:*:*:*:*:*:*`.
So let's look at how this is getting matched:

| Attribute | Source | Target | Result |
| :-------- | :------- | :----- | :--------- |
| Part | `h` | ANY | SUPERSET (`#13`) |
| Vendor | `intel` | ANY | SUPERSET (`#13`) |
| Product | ANY | `xrandr` | SUPERSET (`#3`) |
| Version | ANY | `1.5.1` | SUPERSET (`#3`) |

The problem is that by using ANY for Part and Vendor in the target, it's bypassing the only attributes that reliably filter out false positives. Granted, specifying ANY for Product in the source is just asking for trouble and simply can't be correct on the NVD side.

It comes down to a combination of:

* Target CPE being too imprecise
* NVD data being unreliable sometimes

Jasper-Ben:
> At least in the test, the NGINX CPE is matching, both with and without version range ("up to (excluding)"):
> 
> https://github.com/nscuro/dependency-track/blob/2db9cd46d45f80d442a0362492f63952e59b514b/src/test/java/org/dependencytrack/tasks/scanners/InternalAnalysisTaskCpeMatchingTest.java#L225-L228
> 
> Can you confirm that [CVE-2021-3618](https://github.com/advisories/GHSA-r9r5-jxp7-whr4) is present in your DT instance? And if so, what do you see when clicking _View Details_ -> _Affected Components_? The CPEs from the NVD should be listed there.

Now the CVE (and many more) are found in the project. Maybe the scan was not finished yet, or my manual triggering of a re-scan fixed it. :+1:

Jasper-Ben:
> For reference, this is what I'm seeing in a new instance:
> image
> 
> And DT ends up finding this CVE, along with 1170 other vulnerabilities for this BOM:
> image
> 
> I noticed however that the internal analysis took longer than usual, I am assuming this is happening due to the extensive use of ANY on the CPEs included in the BOM, which requires DT to fetch more data from the DB.
> 
> As for the CPU vulnerabilities:
> 
>     * [CVE-2023-20583](https://github.com/advisories/GHSA-52f4-rvv8-64gj) has the CPE `cpe:2.3:h:amd:*:*:*:*:*:*:*:*:*`
> 
>     * [CVE-2022-24436](https://github.com/advisories/GHSA-m7hv-7hvq-q3xx) has the CPE `cpe:2.3:h:intel:*:*:*:*:*:*:*:*:*`
> 
> 
> One of the components it's matching against currently has the CPE `cpe:2.3:*:*:xrandr:1.5.1:*:*:*:*:*:*:*`. So let's look at how this is getting matched:
> Attribute 	Source 	Target 	Result
> Part 	`h` 	ANY 	SUPERSET (`#13`)
> Vendor 	`intel` 	ANY 	SUPERSET (`#13`)
> Product 	ANY 	`xrandr` 	SUPERSET (`#3`)
> Version 	ANY 	`1.5.1` 	SUPERSET (`#3`)
> 
> The problem is that by using ANY for Part and Vendor in the target, it's bypassing the only attributes that reliably filter out false positives. Granted, specifying ANY for Product in the source is just asking for trouble and simply can't be correct on the NVD side.
> 
> It comes down to a combination of:
> 
>     * Target CPE being too imprecise
> 
>     * NVD data being unreliable sometimes

That makes sense. Thanks for the break-down. Then these are "valid" false positives due to imprecise data. That is then something that needs improvement on the component CPE generating side of things.

Jasper-Ben:
@nscuro here is a vex to suppress the hardware CVEs for the provided SBOM. I will continue to check the remaining CVEs :slightly_smiling_face:

[vex_mod.txt](https://github.com/DependencyTrack/dependency-track/files/12892707/vex_mod.txt)

Jasper-Ben:
I just cross-referenced ~15 components and their remaining CVEs against the NIST CPE search. Pleased to share that DT found ALL vulnerabilities that the CPE search returned. In addition, DT even found a handful CVEs that NIST search falsely did NOT include. Really, really happy with the result! :partying_face:

nscuro:
Awesome, thanks for checking @Jasper-Ben! 

> DT even found a handful CVEs that NIST search falsely did NOT include

While that's great, I'd love to understand *why*, as we're not doing anything special (AFAICT). Can you share a few of those CVEs?

Jasper-Ben:
> While that's great, I'd love to understand _why_, as we're not doing anything special (AFAICT). Can you share a few of those CVEs?

NIST CPE search will only return results, if a CPE entry explicitly exists in their DB. E.g. searching for CPE `cpe:2.3:*:*:libxslt:1.1.35:*:*:*:*:*:*:*` will not return results on the NIST CPE search.

image

DT however will find the valid CVE CVE-2022-29824 for this version of libxslt.

I also noticed that there are _some_ false positives for CVEs, probably due to bad CVE data. e.g. unzip 6.0 reports CVE-2008-0888 in DT.

nscuro:
> NIST CPE search will only return results, if a CPE entry explicitly exists in their DB.

Oh wow, that's a bummer. But good to know that DT behaves as expected here.

---

Thanks for all the input, and apologies that resolving this took so long. I'll merge #3070 which will also close this PR. I think we are in a way better position now when it comes to reproducing and fixing any upcoming issues in CPE matching.

Jasper-Ben:
> Thanks for all the input, and apologies that resolving this took so long.

No worries, thanks for getting it fixed!

>  I think we are in a way better position now when it comes to reproducing and fixing any upcoming issues in CPE matching.

Agreed! Do you have an ETA for the next release? Would rather not deploy a development image in production :wink:

nscuro:
> Do you have an ETA for the next release?

Later today. I am in the process of finalizing the release.

PR #1938, opened 2022-09-08
Fix : OSV improve handling duplicates

sahibamittal:
**Current behaviour**: OSV download task checks if there is any existing vulnerability with ID same as one of aliases of incoming new vulnerability (with new vulnerability ID) and if alias exists we dont add this incoming vuln in DT but only add new affected packages in existing vulnerability.
**Proposed change**: Even if alias exists in DT, we add this new incoming vuln (with new ID) as aliases will be handled separately now.


sahibamittal:
@nscuro @VinodAnandan

PR #1939, opened 2022-09-08
OSV fix : removed alias check

sahibamittal:

**Current behaviour:** OSV download task checks if there is any existing vulnerability with ID same as one of aliases of incoming new vulnerability (with new vulnerability ID) and if alias exists we dont add this incoming vuln in DT but only add new affected packages in existing vulnerability.
**Proposed change**: Even if alias exists in DT, we add this new incoming vuln (with new ID) as aliases will be handled separately now. (https://github.com/DependencyTrack/dependency-track/pull/1912)


sahibamittal:
@nscuro @VinodAnandan

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -423,6 +423,7 @@ public List<Vulnerability> getVulnerabilities(Project project, boolean includeSu
             );
             for (final Vulnerability componentVuln: componentVulns) {
                 componentVuln.setComponents(Collections.singletonList(pm.detachCopy(component)));
+                componentVuln.setAliases(getVulnerabilityAliases(componentVuln));
```
nscuro, line 426: @sahibamittal The aliases need to be detached, otherwise we'll be leaking resources.

```suggestion
                componentVuln.setAliases(new ArrayList<>(pm.detachCopyAll(getVulnerabilityAliases(componentVuln))));
```

nscuro (approved):

Issue #1174, opened 2021-08-26
Component update is broken


vmvarga:
### Current Behavior:
Currently hashes for some components are stored in uppercase format and it's accepted by UI
![image](https://user-images.githubusercontent.com/38255353/131505274-e2dbd8c1-c1eb-4b23-a30b-618a119f0378.png)

In case you need to update component API call is failing with 400 error
![image](https://user-images.githubusercontent.com/38255353/130919931-c5f8fa50-e5f6-44d5-8c8e-3fcd7ad2b7ea.png)

This happens due to API is not accepting uppercase SHA, after converting to lower component is updated
![image](https://user-images.githubusercontent.com/38255353/130920458-d9cc90db-ec01-4fac-a4f0-a821d4b3bfc5.png)

### Expected Behavior:
Convert hash to lower before processing

cdaven:
I was just about to write a new issue about this.

Issue #1182, opened 2021-08-31
Many licenses not identified and manual update failing


rstmm:
### Current Behavior:
I scanned some of our projects to test if dependency track fulfills our needs. Those involved C#.NET and JS/TS projects. The BoM was created with cyclonedx-dotnet and cyclonedx-node-module. The C# projects partly use a private nuget repository, which was given to cycloneDX as well as github auth token. The results got imported into dependency track.
Dependency Track lists for many of the projects barely any licenses. It does list for most of them the "latest version" column, so it seems to recognize the libraries, but not licenses.
In the worst project, out of 34 non-custom dependencies (for the custom ones the information is probably not available as needed, so i excluded them), only 4 licenses could be recognized. But it listed "latest versions" for all those 34 dependencies, so it generally was able to query information about them. 

From the missing 30 licenses:
17 there was no info or dead URLs in BOM file, no result expected here.
1 contains no license info in BOM, but NuGet contains a license name but a dead license link. I think the license name could be accessible maybe?
1 contained no license info in BOM, but NuGet contained Name and link to a valid license file, which should be possible to extract?
1 contained a <url>license entry in BOM to a BSD-2-Clause license. License should be recognizable.
10 contained a <url> license entry in BOM to a GitHub MIT License file. License should be recognizable.

Furthermore, when trying to manually update licenses of the packages (select in dropdown of package detail popover, then click Update button), an error 400 is shown without further info. Inspecting the network requests shows:
`[{"message":"The SHA-512 hash must be a valid 128 character HEX number","messageTemplate":"The SHA-512 hash must be a valid 128 character HEX number","path":"sha512","invalidValue":"A868C852CDF10EF56CF166808B88D077CA3FFA0DF4BD514007A06EC788073E1157D7B39DC3703E87B87DAF47EFD79A7A576EFE4D3C388BE9ECF9F5CB5E4CDB4B"}]`
Somehow updating the license seems to try to update also the Hash value (which is the same as returned by the system, it got not changed. The value seems valid to me anyways.)

### Steps to Reproduce:
I copy some of the BOM sections, so you can have a look what it looks like:
```
    <component type="library">
      <publisher>Microsoft</publisher>
      <name>DocumentFormat.OpenXml</name>
      <version>2.9.1</version>
      <description>[...]</description>
      <hashes>
[...]

PR #1944, opened 2022-09-11
Issue 1174: Fix component update due to uppercase hashes

awegg:
Predefined RegexSequences only validate lower case hash notations. While this is usually the case it is not mandatory as hex values can be upper and lower case. Furthermore, tools like cyclonedx-dotnet provide uppercase hash values and boms generated by it can be imported but components can't be changed in UI. API will return an error for failed hash validation

Fixes: #1174, Partially: #1182 

Signed-off-by: awegg <alexander@weggerle.de>

### DIFF
--- a/src/main/java/org/dependencytrack/model/Component.java
+++ b/src/main/java/org/dependencytrack/model/Component.java
@@ -147,49 +147,49 @@ public class Component implements Serializable {
     @Persistent
     @Index(name = "COMPONENT_MD5_IDX")
     @Column(name = "MD5", jdbcType = "VARCHAR", length = 32)
-    @Pattern(regexp = RegexSequence.Definition.HASH_MD5, message = "The MD5 hash must be a valid 32 character HEX number")
+    @Pattern(regexp = "^[0-9a-fA-F]{32}$", message = "The MD5 hash must be a valid 32 character HEX number")
     private String md5;
 
     @Persistent
     @Index(name = "COMPONENT_SHA1_IDX")
     @Column(name = "SHA1", jdbcType = "VARCHAR", length = 40)
-    @Pattern(regexp = RegexSequence.Definition.HASH_SHA1, message = "The SHA1 hash must be a valid 40 character HEX number")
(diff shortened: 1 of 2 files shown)


stevespringett:
Thanks for identifying the issue and the PR.

nscuro:
I think a preferable way to tackle this would be to perform the `toLowerCase` conversion directly at ingestion, e.g.:

https://github.com/DependencyTrack/dependency-track/blob/60fc8678b7ea205c6af50c3f0ffe60d2518f480f/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java#L128-L146

Because most hash-related logic in DT does not consider that values can be of different casing, just accepting uppercase values will not yield the desired improvement.

awegg:
The idea was to be tolerant what is accepted (upper/lowercase) but be specific what to store or return in the APIs (lowercase which is more or less the standard for hash values).

If I would convert to lowercase in the ModelConverter, I would also need to consider the API usage:
https://github.com/DependencyTrack/dependency-track/blob/10d94467007a1446fc7f89f8c62e5a90a8c862de/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java#L267-L274

and
https://github.com/DependencyTrack/dependency-track/blob/10d94467007a1446fc7f89f8c62e5a90a8c862de/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java#L348-L355

there are also functions to clone and update components in the `ComponentQueryManager` where I am not sure how they are used and if there are corner cases where the source or transient component are not coming from the API or ModelConverter.

Somehow I have the feeling doing this in the components set methods is bullet proof. You can't miss some corner cases or future additions and the `toLowerCase` will be called also for the code snippet you referenced. 

But I'm new to DT, so if you say the place you pointed out would be better, I will update my PR. Please also advise in this case if I should update the 4 other places where hash values are set.

nscuro:
@awegg For some reason I read the PR as if the `toLowerCase` calls where done in the getters, not the setters. My bad, sorry for the confusion. You're totally right.

stevespringett:
@nscuro are we ready to merge this in?

nscuro (approved):

nscuro:
@stevespringett Yup, merging coming right up.

PR #1977, opened 2022-09-22
1182: Add support for License URL

awegg:
Licenses may come with an URL as specified in:
https://cyclonedx.org/docs/1.4/xml/#type_licenseType This is especially useful if there is no license ID given and the URL is often a good indicator to manually assign a license in the UI.

Discussion in #1182

image

Signed-off-by: awegg <alexander@weggerle.de>


nscuro (changes requested):
Thanks for the PR, @awegg! And sorry for the long delay. 😅

For model conversion, you addressed one way (CycloneDX -> DT), but not the other (DT -> CycloneDX, used for BOM exports for example). Please add it here:

https://github.com/DependencyTrack/dependency-track/blob/64cf10f8bec2b73165d7ace0f7bbee95c6e4096f/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java#L239-L251

awegg:
Thanks for pointing me to that location @nscuro. I wasn't aware of this. 
After I've updated the ModelConverter like you suggested, I've tested the export and imported it into a new project. All the license ULRs were imported and the export contained the urls obviously. 

Let me know if there is something else I should change :-)

nscuro (approved):
Thanks @awegg!

PR #1999, opened 2022-10-03
Fix PURL mapping of GHSA vulnerabilities for PHP

### LINKED ISSUE #1998: PURLs for PHP vulnerabilities from GHSA are constructed incorrectly
### Current Behavior:

Vulnerabilities mirrored from GitHub Advisories that affect PHP components are not correlated to components during vulnerability analysis, when the package name advertised by GitHub Advisories contains slashes.

This is the case for https://github.com/advisories/GHSA-52m2-vc4m-jj33 for example, where the reported package name is `twig/twig`.

### Steps to Reproduce:

1. Enable GitHub Advisories integration
2. Restart DT and wait for the GHSA mirroring to complete
3. Upload a BOM containing a component with PURL `pkg:composer/twig/twig@2.14.11`
4. Observe that the component is not reported as vulnerable for GHSA-52m2-vc4m-jj33

### Expected Behavior:

DT should be able to correlate PHP vulnerabilities from GitHub Advisories with components by PURL.

### Environment:
  
  - Dependency-Track Version: 4.5.0
  - Distribution: N/A
  - BOM Format & Version: N/A
  - Database Server: N/A
  - Browser: N/A

### Additional Details:

This is how the data from the advisory is stored in DT right now:

| PURL\_VERSION | PURL\_NAMESPACE | PURL\_NAME | VERSIONSTARTINCLUDING | VERSIONSTARTEXCLUDING | VERSIONENDINCLUDING | VERSIONENDEXCLUDING |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| null | null | twig/twig | 1.0.0 | null | null | 1.44.7 |
| null | null | twig/twig | 2.0.0 | null | null | 2.15.3 |
| null | null | twig/twig | 3.0.0 | null | null | 3.4.3 |

When performing vuln analysis, DT will construct a PURL from the parts above. However, in PURL, special characters in the package name must be URL encoded, which includes slashes. So at runtime the PURL will turn out to be `pkg:composer/twig%2Ftwig@1.0.0` etc., thus probably failing all comparisons.

We handle such cases for Maven, NPM and Go packages already, we just need to do it for PHP components as well:

https://github.com/DependencyTrack/dependency-track/blob/e5402d3e307355e6af6eb2269e7c38d9c22bcbfe/src/main/java/org/dependencytrack/tasks/GitHubAdvisoryMirrorTask.java#L311-L329

Once fixed, there should be an upgrade procedure to automatically correct existing data. This was done in 4.5.0 already for Go advisories:

https://github.com/DependencyTrack/dependency-track/blob/a3b42a6fe4d19fc91122887dba0841f26fa7c7c4/src/main/java/org/dependencytrack/upgrade/v450/v450Updater.java#L132-L147

nscuro:
Fixes #1998

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/GitHubAdvisoryMirrorTask.java
+++ b/src/main/java/org/dependencytrack/tasks/GitHubAdvisoryMirrorTask.java
@@ -59,6 +59,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import static org.dependencytrack.model.ConfigPropertyConstants.VULNERABILITY_SOURCE_GITHUB_ADVISORIES_ACCESS_TOKEN;
 import static org.dependencytrack.model.ConfigPropertyConstants.VULNERABILITY_SOURCE_GITHUB_ADVISORIES_ENABLED;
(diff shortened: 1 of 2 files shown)


VinodAnandan:
LGTM @nscuro, thanks for the PR!!

VinodAnandan (approved):

Issue #2005, opened 2022-10-04
Include affected project(s) in notification mails


rkg-mm:
### Current Behavior:
Emails sent for notifications do not include the project for which the notification was sent for, neither in title nor in body.

### Proposed Behavior:
* The subject should indicate for which project a mail was sent
* The body should show for which project a mail was sent (with link)
* Other affected projects should be included as additional information (as long as they are covered by the notification rule)

We will create a PR for this

PR #2006, opened 2022-10-04
Include affected project(s) in notification mails

rbt-mm:
#2005 

Currently, it's hard to tell from an email notification which projects are affected by a new-found vulnerability because the subject line is always the same and the body does not show any of the affected projects.

Shows the affected project in the subject line of an email notification to make the emails more distinguishable and to make it easier for a user to gather important information like project name and version quickly.

Changes the default email template for the notification groups 'NEW_VULNERABILITY_IDENTIFIED' and 'PROJECT_AUDIT_CHANGE' to list the project of the affected component and to list other projects, if these are also affected.


src/main/resources/templates/notification/publisher/email.peb
```diff
@@ -8,7 +8,18 @@ Severity:          {{ subject.vulnerability.severity }}
...
+Version:           {{ subject.component.project.version }}
+Description:       {{ subject.component.project.description }}
+Project URL:       {{ baseUrl }}/projects/{{ subject.component.project.uuid }}
+{% if  notification.subject.affectedProjects|length > 1%}
+--------------------------------------------------------------------------------
+
+Other affected projects:
+{% for affectedProject in notification.subject.affectedProjects %}{% if not (affectedProject.uuid == subject.component.project.uuid) %}
+Project:           [{{ affectedProject.name }} : {{ affectedProject.version }}]
+Project URL:       {{ baseUrl }}/project/?uuid={{ affectedProject.uuid }}
```
nscuro, line 21: The URL is not handled by anything. The UUID is a path parameter.

```suggestion
Project URL:       {{ baseUrl }}/project/{{ affectedProject.uuid }}
```

src/main/resources/templates/notification/publisher/email.peb
```diff
@@ -34,9 +45,18 @@ Severity:          {{ subject.vulnerability.severity }}
...
```
nscuro, line 58: Same as above, the URL needs to be adjusted.

```suggestion
Project URL:       {{ baseUrl }}/project/{{ affectedProject.uuid }}
```

nscuro (changes requested):
Thanks for the PR, @rbt-mm! Small adjustments needed in the template file, otherwise this looks good for inclusion in 4.6.

nscuro (approved):

PR #2007, opened 2022-10-04
Bugfix get test resource not working with whitespace in path

rbt-mm:
Fixes test cases which use `getResource` if a file path includes white space.
Converts the returned URL from `.getResource()` to a URI to decode escaped whitespace, and by that enables whitespace in a file path.


nscuro:
Awesome, thanks @rbt-mm!

Issue #84, opened 2018-01-02
Add UI metaphors for parent/child projects and components


stevespringett:
The datamodel and corresponding REST resources support parent/child relationships for projects and components. Currently, the UI does not have a way to express this.

This ticket is to create UI metaphors for parent/child projects and components.

rkg-mm:
FYI, we are working on this topic and will provide a PR as proposal in next days probably

stevespringett:
Need to revert https://github.com/DependencyTrack/frontend/pull/272 and create a better UI experience.

rkg-mm:
@stevespringett  @nscuro 
We tried to implement this using the TreeGrid over the last days. In theory it can work, but we have no luck once we try to implement lazy loading of children, which in my opinion is definitively necessary.

The main problem is: The plugin doesn't handle a situation well, when additional data is appended to the tree. If you append data (-> child elements), the complete view resets and starts from base level again, not expanding back to the level you had opened.
Trying to manually re-open is a struggle to get the timing with events right AND you need to handle issues with scrolling etc., and have kind of a "jumping" of the table, since its for a moment collapsed again, then uncollapsing etc (especially in deeper levels you will see the table changing and jumping between states).

Any other plugins you know that could be worth a shot? Or could we stick with our version for now and if anybody has an idea to get this done better it can still be exchanged since its mainly frontend, and data is not affected?

stevespringett:
We can always fork TreeGrid, modify it to meet our requirements, and use that. It doesn't change a whole lot, so it should not be complicate to keep synced with the upstream.

https://github.com/wenzhixin/bootstrap-table/commits/develop/src/extensions/treegrid

PR #2010, opened 2022-10-04
Show hierarchical view in project list

rbt-mm:
#84 

Introducing new backend APIs which supports the hierarchical view in the frontend.

Added new project-API and project-query methods to support the display of the already existing parent-child-relationship between projects in the UI.

Introducing new parent-child relationship logic:

- Prevents that inactive projects can be selected as parents
- A project cannot be set to inactive if it has an active child
- A parent project cannot select one of his children as a parent

Project List Hierarchical View Expanded

[Frontend PR](https://github.com/DependencyTrack/frontend/pull/272)

Signed-off-by: RBickert [rbt@mm-software.com](mailto:rbt@mm-software.com)


nscuro:
Thanks for your work on this and the PR @rbt-mm! 

The 4.6.0 milestone is already quite full, and we're in the process of getting that out of the door. As this is a larger change, I'd like to include this in 4.7 instead. We're hoping to release 4.7 much sooner than we did 4.6, so hopefully there won't be too much of a delay. Will this work for you?

rkg-mm:
> Thanks for your work on this and the PR @rbt-mm!
> 
> The 4.6.0 milestone is already quite full, and we're in the process of getting that out of the door. As this is a larger change, I'd like to include this in 4.7 instead. We're hoping to release 4.7 much sooner than we did 4.6, so hopefully there won't be too much of a delay. Will this work for you?

Yes that's fine for us

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -88,6 +88,11 @@ public PaginatedResult getProjects(final boolean includeMetrics, final boolean e
         var filterBuilder = new ProjectQueryFilterBuilder()
                 .excludeInactive(excludeInactive);
 
+        if (onlyRoot){
+            filterBuilder.excludeProjectsThatAreChildren();
+            pm.getFetchPlan().addGroup(Project.FetchGroup.ALL.name());
```
nscuro, line 93: The fetch plan should be set for `query`, otherwise it may have unintended impact on other queries performed by this `pm`.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -213,6 +223,11 @@ public PaginatedResult getProjects(final Team team, final boolean excludeInactiv
                 .excludeInactive(excludeInactive)
                 .withTeam(team);
 
+        if (onlyRoot){
+            filterBuilder.excludeProjectsThatAreChildren();
+            pm.getFetchPlan().addGroup(Project.FetchGroup.ALL.name());
```
nscuro, line 228: The fetch plan should be set for `query`, otherwise it may have unintended impact on other queries performed by this `pm`.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -820,4 +877,103 @@ public boolean hasAccessManagementPermission(final ApiKey apiKey) {
...
+            } else {
+                filterBuilder = filterBuilder.withFuzzyName(filterString);
+            }
+        }
+
+        final String queryFilter = filterBuilder.buildFilter();
+        final Map<String, Object> params = filterBuilder.getParams();
+
+        preprocessACLs(query, queryFilter, params, false);
+        pm.getFetchPlan().addGroup(Project.FetchGroup.ALL.name());
```
nscuro, line 908: The fetch plan should be set for `query`, otherwise it may have unintended impact on other queries performed by this `pm`.

src/main/java/org/dependencytrack/persistence/ProjectQueryFilterBuilder.java
```diff
@@ -71,6 +72,18 @@ ProjectQueryFilterBuilder withFuzzyNameOrExactTag(String name, Tag tag) {
...
 
+    ProjectQueryFilterBuilder excludeProjectsThatAreChildren(){
+        filterCriteria.add("parent == null");
+        return this;
+    }
+
+    ProjectQueryFilterBuilder withParent(UUID uuid){
+        params.put("uuid", uuid);
+
+        filterCriteria.add("parent.uuid == :uuid");
```
nscuro, line 83: Using `:parentUuid` would be better to avoid any conflicts if the same query has a `:uuid` parameter already.

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -296,7 +309,12 @@ public Response updateProject(Project jsonProject) {
                     if (name == null) {
                         jsonProject.setName(project.getName());
                     }
-                    project = qm.updateProject(jsonProject, true);
+                    try {
+                        project = qm.updateProject(jsonProject, true);
+                    } catch (IllegalArgumentException e){
+                        LOGGER.error(e.getMessage());
```
nscuro, line 315: As this is a user error that administrators can't do anything about, it should not be logged as error. At the very least it should be `warn`, but I think really this should be `debug`. The response already tells the user what's wrong.

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -464,4 +487,64 @@ public Response cloneProject(CloneProjectRequest jsonRequest) {
...
+    @ApiResponses(value = {
+            @ApiResponse(code = 404, message = "The UUID of the project could not be found")
+    })
+    @PermissionRequired(Permissions.Constants.VIEW_PORTFOLIO)
+    public Response getProjectsWithoutDescendantsOf(@ApiParam(value = "The UUID of the project which descendants will be excluded", required = true)
+                                @PathParam("uuid") String uuid) {
+        try (QueryManager qm = new QueryManager(getAlpineRequest())) {
+            final Project project = qm.getObjectByUuid(Project.class, uuid);
+            if (project != null) {
+                final PaginatedResult result = qm.getProjectsWithoutDescendantsOf(project);
```
nscuro, line 543: There should be a `qm.hasAccess` check here, even though the project itself is not returned by this endpoint.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -820,4 +877,103 @@ public boolean hasAccessManagementPermission(final ApiKey apiKey) {
...
+            }
+        }
+
+        final String queryFilter = filterBuilder.buildFilter();
+        final Map<String, Object> params = filterBuilder.getParams();
+
+        preprocessACLs(query, queryFilter, params, false);
+        result = execute(query, params);
+
+        result.setObjects(result.getList(Project.class).stream().filter(p -> !isChildOf(p, project.getUuid())).toList());
```
nscuro, line 947: Horribly inefficient, but it does indeed seem like it's not possible to do this kind of recursive checking in JDO :(

nscuro (changes requested):
Awesome job on this PR, @rbt-mm! And sorry for the long wait.

Some nitpicks, and please rebase your PR with the current `master` branch. There have been some breaking changes in the Alpine SNAPSHOT version you developed against, but the build should succeed after a rebase. :)

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
rbt-mm: I'm really not a fan of this inefficient approach but I couldn't find anything either in JDO which could solve this recursive check, so I had to stick with this approach. I was hoping that eventually someone else will find a better solution for this ;)

nscuro (approved):

PR #2011, opened 2022-10-05
Fix vulnerability aliases not being considered during metrics calculation

nscuro:
This functionality got lost when resolving merge conflicts in #1912. It has now been adapted to the new metrics calculation method.


VinodAnandan (approved):
LGTM, thanks @nscuro !!

Issue #1608, opened 2022-05-11
Individual notification subscriptions to projects/tags/groups


rkg-mm:
### Current Behavior:
Currently, the admin can create notifications in the backend for all projects for single mail addresses or other channels. With a bigger amount of projects this is problematic, since one person can't handle all projects. Creating different notifications for individual projects is possible but takes effort and for each version must be remembered to add the new version to specific notification(s) again.

### Proposed Behavior:
It would be very helpful, if people could subscribe (or be subscribed) to individual projects, or even better: to projects matching specific criteria.
Use Cases:
1. Project Manager/Architect can subscribe to a project. Important: It must be matched by name since he usually would want notifications on every single version, not only one specific version.
2. Business Unit leader / BU security responsible can subscribe to a tag representing his business unit (alternatively: namespace/group)
3. Lead architect/legal department can subscribe to all projects for policy violations but only for license violations

Important: Especially when doing matching by tags etc. the project permissions need to be respected, to ensure you won't be notified for projects you are not allowed to see.

Additionally:
I think it would make sense to be able to define one user as main project responsible in project settings. (E.g. dealing with 25+ projects of different departments I always need to look up who is responsible if I need to contact someone). This responsible person should have by default a notification configured automatically for this project. New project versions should take over the same responsible automatically.

PR #2013, opened 2022-10-05
Include a projects children in an alert rules limitations

rbt-mm:
[#1608](https://github.com/DependencyTrack/dependency-track/issues/1608#issuecomment-1220774127)

Currently, when creating a limited alert rule you need to select every single version of a project. This does not work well for projects which create new versions regularly.
By automatically including every active child of a project to a notification rule, it is simpler to limit all existing or every future version of a project to a notification rule.

[Frontend PR]()

### DIFF
--- a/.github/workflows/_meta-build.yaml
+++ b/.github/workflows/_meta-build.yaml
@@ -25,7 +25,7 @@ jobs:
         uses: actions/checkout@v3.0.2
 
       - name: Set up JDK
-        uses: actions/setup-java@v3.5.0
+        uses: actions/setup-java@v3.5.1
         with:
           distribution: 'temurin'
           java-version: '17'
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -48,7 +48,7 @@ jobs:
         uses: actions/checkout@v3.0.2
 
       - name: Set up JDK
-        uses: actions/setup-java@v3.5.0
+        uses: actions/setup-java@v3.5.1
         with:
           distribution: 'temurin'
           java-version: '17'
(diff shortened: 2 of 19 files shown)


rbt-mm:
Hey @nscuro, I'm not quite sure why this build failed. The workflow runs successfully in our repository and I don't think that I've changed anything which would throw a `MojoExecutionException`.
Do you know if something is wrong with my changed code or is it just a build fail? And if so, how can I restart the workflows to pass the checks?

nscuro:
No worries @rbt-mm, it's just GitHub being GitHub again. When trying to download the frontend ZIP, the GitHub API returned a 503, causing the build to fail. 

I don't think contributors can manually trigger workflows, but you could always contact us maintainers (as you did), or push an empty commit to trigger another workflow execution.

nscuro (approved):
Thanks for the PR @rbt-mm! 

Overall this looks good to me. But I was wondering whether this should be an option that users can opt-in to / opt-out of? Just because there is a hierarchy, it doesn't necessarily mean I'm interested in all notifications affecting child projects.

rkg-mm:
> Thanks for the PR @rbt-mm!
> 
> Overall this looks good to me. But I was wondering whether this should be an option that users can opt-in to / opt-out of? Just because there is a hierarchy, it doesn't necessarily mean I'm interested in all notifications affecting child projects.

Yes that's something I was also thinking about, but for simplicity decided against it for the first attempt. But the question is:
1. Would this be a setting per notification rule, for all projects in that rule? (simple, add a checkbox and an additional field in the rule)
2. Or a setting per project in each notification rule (requires a field in the mapping table, not sure if datanucleus supports that without big changes, and also would have to refactor the project List to show an additional checkbox per project, plus APIs to update etc.)

If you think option 1) is enough we can easily do that, for 2) we would have to figure out how and will take some more time.

nscuro:
I think option 1 should suffice.

nscuro (approved):

PR #2014, opened 2022-10-05
Include vulnerability aliases in notifications

nscuro:
Based on #2011 because I needed some methods introduced in that PR.

Note: I've ran into an issue where aliases would be discarded when detaching the vulnerability "owning" them. Because the `Vulnerability#aliases` field is transient, DataNucleus just drops it during detachment. I worked around this for now by refreshing the aliases again whenever it's needed, but ultimately a more general solution would be good.


VinodAnandan (approved):
LGTM, thanks @nscuro !

Issue #2019, opened 2022-10-06
Add an attribute in notification content of NewVulnerabilityIdentified


sahibamittal:
### Current Behavior:

When a new vulnerability is detected, a notification is dispatched with NewVulnerabilityIdentified object in content containing info about vulnerability, component and affectedProjects. This notification is alerted with group named as 'NEW_VULNERABILITY' which can be result of a Bom upload process or a periodic analysis. It is hard to differentiate when notifications collide and are coming from either Bom upload or background periodic scan.

### Proposed Behavior:

Add a new attribute say analysisLevel (BOM_UPLOAD_ANALYSIS, PERIODIC_ANALYSIS) and based on type of event (mom upload or portfolio scheduled analysis), we can assign this attribute in notification content which would help us to differentitate between them. This would be helpful in api server with many projects, to take actions on notifications accordingly.

PR #2027, opened 2022-10-10
Issue 2019 - add new attribute in notification content

sahibamittal:

Issue: https://github.com/DependencyTrack/dependency-track/issues/2019


src/main/java/org/dependencytrack/model/AnalysisLevel.java
```diff
@@ -0,0 +1,7 @@
+package org.dependencytrack.model;
+
+public enum AnalysisLevel {
```
nscuro, line 3: I think it'd be good to rename this to `VulnerabilityAnalysisLevel`, because the other `Analysis*` classes in this package are related to auditing.

src/main/java/org/dependencytrack/event/VulnerabilityAnalysisEvent.java
```diff
@@ -85,4 +85,13 @@ public VulnerabilityAnalysisEvent project(final Project project) {
     public Project getProject() {
         return project;
     }
+
+    private String analysisLevel;
```
nscuro, line 89: This field should use the enum type, not `String`.

src/main/java/org/dependencytrack/notification/vo/NewVulnerabilityIdentified.java
```diff
@@ -29,11 +29,13 @@ public class NewVulnerabilityIdentified {
     private final Vulnerability vulnerability;
     private final Component component;
     private final Set<Project> affectedProjects;
+    private final String analysisLevel;
```
nscuro, line 32: This field should use the enum type, not `String`.

src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java
```diff
@@ -52,12 +52,13 @@ public abstract class AbstractVulnerableSoftwareAnalysisTask extends BaseCompone
      * @param component the component being analyzed
      */
     protected void analyzeVersionRange(final QueryManager qm, final List<VulnerableSoftware> vsList,
-                                       final String targetVersion, final String targetUpdate, final Component component) {
+                                       final String targetVersion, final String targetUpdate, final Component component,
+                                       final String analysisLevel) {
```
nscuro, line 56: The `analysisLevel` parameter should use the enum type, not String.

src/main/java/org/dependencytrack/tasks/scanners/InternalAnalysisTask.java
```diff
@@ -47,6 +47,8 @@ public AnalyzerIdentity getAnalyzerIdentity() {
         return AnalyzerIdentity.INTERNAL_ANALYZER;
     }
 
+    private String analysisLevel;
```
nscuro, line 50: This field should use the enum type, not String.

src/main/java/org/dependencytrack/tasks/scanners/BaseComponentAnalyzerTask.java
```diff
@@ -84,7 +84,8 @@ protected boolean isCacheCurrent(Vulnerability.Source source, String targetHost,
         }
     }
 
-    protected void applyAnalysisFromCache(Vulnerability.Source source, String targetHost, String target, Component component, AnalyzerIdentity analyzerIdentity) {
+    protected void applyAnalysisFromCache(Vulnerability.Source source, String targetHost, String target, Component component,
+                                          AnalyzerIdentity analyzerIdentity, String analysisLevel) {
```
nscuro, line 88: The `analysisLevel` parameter should use the enum type, not String.

src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java
```diff
@@ -72,6 +72,8 @@ public class OssIndexAnalysisTask extends BaseComponentAnalyzerTask implements C
     private String apiUsername;
     private String apiToken;
 
+    private String analysisLevel;
```
nscuro, line 75: This field should use the enum type, not String.

src/main/java/org/dependencytrack/tasks/scanners/VulnDbAnalysisTask.java
```diff
@@ -47,6 +47,7 @@ public class VulnDbAnalysisTask extends BaseComponentAnalyzerTask implements Sub
     private static final Logger LOGGER = Logger.getLogger(VulnDbAnalysisTask.class);
     private static final String TARGET_HOST = "https://vulndb.cyberriskanalytics.com/";
     private static final int PAGE_SIZE = 100;
+    private String analysisLevel;
```
nscuro, line 50: This field should use the enum type, not String.

src/main/java/org/dependencytrack/util/NotificationUtil.java
```diff
@@ -63,7 +63,7 @@ public final class NotificationUtil {
      */
     private NotificationUtil() { }
 
-    public static void analyzeNotificationCriteria(QueryManager qm, Vulnerability vulnerability, Component component) {
+    public static void analyzeNotificationCriteria(QueryManager qm, Vulnerability vulnerability, Component component, String analysisLevel) {
```
nscuro, line 66: The `analysisLevel` parameter should use the enum type, not String.

src/test/java/org/dependencytrack/tasks/BomUploadProcessingTaskTest.java
```diff
@@ -28,14 +28,9 @@
...
 import org.dependencytrack.event.NewVulnerableDependencyAnalysisEvent;
 import org.dependencytrack.event.VulnerabilityAnalysisEvent;
-import org.dependencytrack.model.Classifier;
-import org.dependencytrack.model.Component;
-import org.dependencytrack.model.ConfigPropertyConstants;
-import org.dependencytrack.model.Project;
-import org.dependencytrack.model.Severity;
-import org.dependencytrack.model.Vulnerability;
-import org.dependencytrack.model.VulnerableSoftware;
+import org.dependencytrack.model.*;
```
nscuro, line 31: Please expand this import so all imported classes are visible.

nscuro (changes requested):
Looks good, but a few minor nitpicks. I made some redundant comments that cover the same thing, but in different places.

You'll also need to extend the `NotificationUtil#toJson` method for `NewVulnerabilityIdentified` so that the new field is actually included in notification template data:

https://github.com/DependencyTrack/dependency-track/blob/68db0e71530aea94dc9a9d2382166ac13f119c61/src/main/java/org/dependencytrack/util/NotificationUtil.java#L302-L318

This is also where you'll want to eventually convert the enum value to a `String`.

The new field should then be added to the notification templates as well: https://github.com/DependencyTrack/dependency-track/tree/master/src/main/resources/templates/notification/publisher

nscuro (approved):

Issue #1515, opened 2022-04-07
api server is unavailable via frontend gui


rybas-dv:
### Current Behavior:
Use default quickstart from https://docs.dependencytrack.org/getting-started/deploy-docker/
# Downloads the latest Docker Compose file
curl -LO https://dependencytrack.org/docker-compose.yml

# Starts the stack using Docker Compose
docker-compose up -d

i cant login after deploy 
<img width="1431" alt="Снимок экрана 2022-04-07 в 13 00 41" src="https://user-images.githubusercontent.com/33660490/162174518-6b7db493-def3-489a-b88c-80c3e810234f.png">

docker exec -it sast_dtrack-frontend_1 sh
i try:
/app $ curl -I localhost:8080
HTTP/1.1 200 OK
Server: nginx/1.20.2
Date: Thu, 07 Apr 2022 10:03:09 GMT
Content-Type: text/html
Content-Length: 3655
Last-Modified: Thu, 17 Feb 2022 21:48:16 GMT
Connection: keep-alive
ETag: "620ec2a0-e47"
Accept-Ranges: bytes

/app $ curl -I localhost:8081
curl: (7) Failed to connect to localhost port 8081 after 0 ms: Connection refused

after that i used in my compose file
- API_BASE_URL=http://dtrack-apiserver:8080

/app $ curl -I dtrack-apiserver:8080
HTTP/1.1 200 OK
Date: Thu, 07 Apr 2022 10:06:57 GMT
Last-Modified: Sun, 13 Feb 2022 10:28:54 GMT
Content-Type: text/html
Accept-Ranges: bytes
Content-Length: 393

but the behavior hasn't changed, I also can't log in.

### Steps to Reproduce:
Use default quickstart from https://docs.dependencytrack.org/getting-started/deploy-docker/

### Expected Behavior:
providing ui functions

### Environment:
  - Ubuntu:
    VERSION="18.04.5 LTS (Bionic Beaver)
    Docker:  19.03.8
    compose: 1.29.2
  - Red Hat Enterprise Linux
     VERSION="8.3 (Ootpa)"
     Docker: 19.03.15
    compose: 1.29.2
  - Dependency-Track Version: dependencytrack/frontend:latest(4.4.0); dependencytrack/apiserver:latest(4.4.2)
  - Distribution: Docker
  - BOM Format & Version:
  - Database Server: H2
  - Browser: Version/15.3 Safari/605.1.15

### Additional Details:

[...]

PR #2036, opened 2022-10-11
Improved docker-compose documention for API_BASE_URL changes to take effect.

cesarhernandezgt:


cesarhernandezgt:
Thanks for the feedback @valentijnscholten , I agree this is a hint related to docker I faced while doing a non localhost installation that others have also eventually resolved and based on 

- https://github.com/DependencyTrack/dependency-track/issues/1515 
- https://github.com/DependencyTrack/dependency-track/discussions/2002 
- https://github.com/DependencyTrack/dependency-track/discussions/1982#discussioncomment-3744506

Before creating the PR, I wondered if https://github.com/DependencyTrack/dependency-track/blob/master/docs/_docs/getting-started/deploy-docker.md would be a better place for this information.

nscuro:
We added this part to the `docker-compose.yml` file for this exact reason, because folks struggled with setting the correct base URL:

```yml
# NOTE:
#   * This URL must be reachable by the browsers of your users.
#   * The frontend container itself does NOT communicate with the API server directly, it just serves static files.
#   * When deploying to dedicated servers, please use the external IP or domain of the API server.
```

I agree with @valentijnscholten that the necessary recreation of containers after environment variables have been changed is not specific to Dependency-Track, so I don't really see a need for the addition proposed in this PR.

cesarhernandezgt:
Thanks for the feedback @nscuro and @valentijnscholten .

It seems the feedback so far is that the information provided by this PR is out of the scope of Dependency track itself and more into the Docker Compose knowledge  the users should already have when setting up Depdendency track docker setup. 

I'll close this PR.

PR #2040, opened 2022-10-12
Docker: Add optional EXTRA_JAVA_OPTIONS environment variable

valentijnscholten:
Sometimes it's useful to pass extra JVM options into the DT Java process.

While it is possible to override the `JAVA_OPTIONS`, this would mean the existing default value from `Dockerfile` has to be copied and maintained.

This PR adds a `EXTRA_JAVA_OPTIONS` environment variable that can be specified at runtime, while benefitting from the existing defaults in the `Dockerfile`.

### DIFF
--- a/docs/_docs/getting-started/deploy-docker.md
+++ b/docs/_docs/getting-started/deploy-docker.md
@@ -155,7 +155,7 @@ services:
     # - DEFAULT_TEMPLATES_OVERRIDE_ENABLED=false
     # - DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY=/data
     #
-    # Optional environmental variables to provide more JVM arguments to the API Server JVM, i.e. "-XX:ActiveProcessorCount=8
+    # Optional environmental variables to provide more JVM arguments to the API Server JVM, i.e. "-XX:ActiveProcessorCount=8"
     # - EXTRA_JAVA_OPTIONS=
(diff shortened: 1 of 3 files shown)


docs/_docs/getting-started/deploy-docker.md
```diff
@@ -150,10 +150,14 @@ services:
     # - ALPINE_METRICS_ENABLED=true
     # - ALPINE_METRICS_AUTH_USERNAME=
     # - ALPINE_METRICS_AUTH_PASSWORD=
-    # 
+    #
     # Optional environmental variables to enable default notification publisher templates override and set the base directory to search for templates
     # - DEFAULT_TEMPLATES_OVERRIDE_ENABLED=false
     # - DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY=/data
+    #
+    # Optional environmental variables to provide more JVM arguments to the API Server JVM, i.e. "-XX:ActiveProcessorCount=8
```
nscuro, line 158: There's a quote missing at the end:

```suggestion
    # Optional environmental variables to provide more JVM arguments to the API Server JVM, i.e. "-XX:ActiveProcessorCount=8"
```

valentijnscholten: fixed

nscuro (approved):

Issue #365, opened 2019-06-11
Add Support for Snyk


stevespringett:
Once Package URL is implemented by Snyk, enhance Dependency-Track to be able to support the analysis of SBoMs using Snyk as a source of vulnerability intelligence.

Snyk API access requires a paid plan (standard, pro, or enterprise)

Depends on: https://github.com/snyk/snyk/issues/210

sahibamittal:
@stevespringett I've been working on Snyk vulnerability analysis integration upon bom upload in DT. 
It will act as an analyser similar to OSS index. It'll call Snyk API for every component purl scanned.

<img width="1416" alt="Screenshot 2022-08-02 at 10 31 56 a m" src="https://user-images.githubusercontent.com/5709882/182342567-98a9389f-7296-4098-adb9-8fd714c3744d.png">

Project vulnerabilities in DT will look like: 

<img width="1416" alt="Screenshot 2022-08-02 at 10 35 11 a m" src="https://user-images.githubusercontent.com/5709882/182343179-05d2ab44-8fbb-44ac-8782-f4d14317384e.png">

**Branch with code changes -> https://github.com/sahibamittal/dependency-track/tree/snyk-integration**

stevespringett:
Thats fantastic. Does Snyk have an API that we can use to send multiple components in a request, similar to OSS Index where we sent 100 components at a time?

sahibamittal:
No sadly for now, this is an initial version supporting two packages (maven, npm). It now enables querying the API with purl as url encoded. Currently, there are only two endpoints:

1. List all vulnerabilities of a package by purl specification
GET /packages/{purl_specification}/vulnerabilities

2. List all vulnerabilities of a package
GET /vulns/{package_manager_name}/{package_name}/{package_version} 

https://apidocs.snyk.io/?version=2022-04-04%7Eexperimental#get-/packages/-purl_specification-/vulnerabilities

nscuro:
The initial implementation (#2070) will attempt to work around the missing batching functionality by distributing the analysis requests to a configurable amount of threads (static thread pool, default size of 10). Will need to observe how well this works in production.

PR #2059, opened 2022-10-17
Issue-365 : Snyk integration

sahibamittal:
**Issue:** https://github.com/DependencyTrack/dependency-track/issues/365

**Frontend changes:** https://github.com/DependencyTrack/frontend/pull/286


src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,444 @@
...
+ */
+public class SnykAnalysisTask extends BaseComponentAnalyzerTask implements Subscriber {
+
+    private String ORG_ID;
+    private String API_BASE_URL;
+    private String BASE_URL;
+    private String PATH = "/rest/orgs/";
+    private static final String API_ENDPOINT = "/issues?version=2022-09-15";
+
+    private static final ThreadPoolExecutor executor = (ThreadPoolExecutor) Executors.newFixedThreadPool(10);
```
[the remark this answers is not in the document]
nscuro: On the other hand, spawning an executor with >= 2 threads for each invocation of `SnykAnalysisTask` will cause an uncontrollable amount of threads to be created if many `SnykAnalysisTask`s are running in parallel.

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,459 @@
...
+                    scannerSnykCvssSource.getGroupName(), scannerSnykCvssSource.getPropertyName()
+            );
+            if (property != null && ConfigProperty.PropertyType.STRING == property.getPropertyType()) {
+                return property.getPropertyValue();
+            }
+        }
+        return scannerSnykCvssSource.getDefaultPropertyValue();
+    }
+
+    public List<VulnerableSoftware> parseVersionRanges(final QueryManager qm, final String purl, final JSONArray ranges) {
```
nscuro, line 393: If we mirror the affected ranges, the next time a component matching this criteria will be scanned, it will be picked up by the internal analyzer, not the Snyk one. This will also cause the attribution in the UI to show as "INTERNAL", not "SNYK". 

For now, I'd prefer it if we leave the mirroring of `VulnerableSoftware` out, and just rely on using the API and caching.

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,459 @@
...
+                                qm.persist(vsList);
+                                Vulnerability synchronizedVulnerability = qm.synchronizeVulnerability(vulnerability, false);
+                                synchronizedVulnerability.setVulnerableSoftware(vsList);
+                                qm.persist(synchronizedVulnerability);
+                                addVulnerabilityToCache(component, synchronizedVulnerability);
+                                final Component componentPersisted = qm.getObjectByUuid(Component.class, component.getUuid());
+                                if (componentPersisted != null) {
+                                    NotificationUtil.analyzeNotificationCriteria(qm, synchronizedVulnerability, componentPersisted, vulnerabilityAnalysisLevel);
+                                    qm.addVulnerability(synchronizedVulnerability, componentPersisted, this.getAnalyzerIdentity());
+                                    LOGGER.info("Snyk vulnerability added : " + synchronizedVulnerability.getVulnId() + " to component " + component.getName());
```
nscuro, line 333: Please remove or change to `debug`.

src/main/java/org/dependencytrack/tasks/VulnerabilityAnalysisTask.java
```diff
@@ -166,7 +172,7 @@ private void performAnalysis(final Subscriber scanTask, final VulnerabilityAnaly
             }
             // Clear the transient cache result for each component.
             // Each analyzer will have its own result. Therefore, we do not want to mix them.
-            event.getComponents().forEach(c -> c.setCacheResult(null));
+            event.getComponents().parallelStream().forEach(c -> c.setCacheResult(null));
```
nscuro, line 175: I don't think this is a good idea. See https://stackoverflow.com/a/20375622

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,459 @@
...
+
+/**
+ * Subscriber task that performs an analysis of component using Snyk vulnerability REST API.
+ */
+public class SnykAnalysisTask extends BaseComponentAnalyzerTask implements Subscriber {
+
+    private String ORG_ID;
+    private String API_BASE_URL;
+    private String BASE_URL;
+    private String PATH = "/rest/orgs/";
```
nscuro, line 63: `PATH` should be `static final`.

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,459 @@
...
+import static org.dependencytrack.util.JsonUtil.jsonStringToTimestamp;
+
+/**
+ * Subscriber task that performs an analysis of component using Snyk vulnerability REST API.
+ */
+public class SnykAnalysisTask extends BaseComponentAnalyzerTask implements Subscriber {
+
+    private String ORG_ID;
+    private String API_BASE_URL;
+    private String BASE_URL;
```
nscuro, line 62: `BASE_URL` and `ORG_ID` should be local variables, as they're only ever used in `inform`.

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,459 @@
...
+                    Instant endThread = Instant.now();
+                    duration += Duration.between(startThread, endThread).toMillis();
+                }
+            }
+
+            paginatedComponents.nextPage();
+        }
+    }
+
+    public void handle(Component component, JSONObject object, int responseCode, String statusText) {
```
nscuro, line 210: As @mehab already pointed out, there's a lot of stuff going on in this class, and this method in particular. I'd suggest moving the parsing logic to dedicated classes in the `org.dependencytrack.parser.snyk` package.

nscuro (changes requested):
As Snyk has a commercial API, it'd be good to at least have a test with a mocked API endpoint of the whole scan being performed for >= 1 component(s). For example:

* Create project
* Create component
* Mock a Snyk response indicating that the component is vulnerable to SNYK-XYZ
* Run the analyzer
* Assert that a vulnerability for SNYK-XYZ has been created
* Assert that aliases have been reported correctly
* Assert that the component has been marked as vulnerable to SNYK-XYZ

You can use `MockServer` to simulate API responses from Snyk, see https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzerTest.java for example.

The test could then later be used to verify whether DT is compatible with older / newer versions of the Snyk API, without requiring a token.

src/main/java/org/dependencytrack/tasks/VulnerabilityAnalysisTask.java
mehab: this was not intended to be sent in PR before. I will revert this

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,444 @@
...
+        return url;
+    }
+
+    /**
+     * Analyzes a list of Components.
+     *
+     * @param components a list of Components
+     */
+    public void analyze(final List<Component> components) {
+        final Pageable<Component> paginatedComponents = new Pageable<>(PAGE_SIZE, components);
```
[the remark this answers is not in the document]
japurva1502: I am working on it

src/main/java/org/dependencytrack/tasks/VulnerabilityAnalysisTask.java
mehab: removed

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,444 @@
...
+        while (!paginatedComponents.isPaginationComplete()) {
+            final List<Component> paginatedList = paginatedComponents.getPaginatedList();
+            //Starting with number of threads as 10
+            int numThreads = 10;
```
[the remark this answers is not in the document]
japurva1502: done..will add in PR

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
mehab: changed this to debug

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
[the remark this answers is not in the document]
mehab: done

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,459 @@
...
+                    int k = 0;
+                    while (trackComponent < paginatedList.size() && k < numThreads) {
+                        temp.add(paginatedList.get(trackComponent));
+                        k += 1;
+                        trackComponent += 1;
+                    }
+
+                    Runnable analysisUtil = () -> {
+                        for (final Component component : temp) {
+                            final String snykUrl = API_BASE_URL + parsePurlToSnykUrlParam(component.getPurl()) + API_ENDPOINT;
```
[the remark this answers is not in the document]
mehab: changed to URLEncoder.encode(component.getPurlCoordinates().toString(), StandardCharsets.UTF_8)

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,459 @@
...
+                            if (component.getPurl() != null) {
```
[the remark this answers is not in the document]
mehab: I have added component.isCapable before adding component to the array that is being proccessed by analysisUtil

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
mehab: working on this one

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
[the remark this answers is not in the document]
japurva1502: Done

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
japurva1502: done

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
japurva1502: done

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,467 @@
...
+    public void handle(Component component, JSONObject object, int responseCode, String statusText) {
+
+        try (QueryManager qm = new QueryManager()) {
+            if (responseCode == 200) {
+                String purl = null;
+                final JSONObject metaInfo = object.optJSONObject("meta");
+                if (metaInfo != null) {
+                    purl = metaInfo.optJSONObject("package").optString("url");
+                    if (purl == null) {
+                        purl = parsePurlToSnykUrlParam(component.getPurl());
```
nscuro, line 222: This will URL encode the component's PURL, I don't think you want that here.

```suggestion
                        purl = component.getPurlCoordinates().toString();
```

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,467 @@
...
+
+                        }
+                    }
+                }
+                updateAnalysisCacheStats(qm, Vulnerability.Source.SNYK, ApiBaseUrl, component.getPurl().toString(), component.getCacheResult());
+            } else if (responseCode == 404) {
+                Vulnerability vulnerability = new Vulnerability();
+                addVulnerabilityToCache(component, vulnerability);
+                updateAnalysisCacheStats(qm, Vulnerability.Source.SNYK, ApiBaseUrl, component.getPurl().toString(), component.getCacheResult());
+                handleUnexpectedHttpResponse(LOGGER, ApiBaseUrl, responseCode, statusText);
```
nscuro, line 281: The 404 response is already handled, there is no need to invoke the unexpected response handler anymore.

nscuro (changes requested):

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
mehab: done

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
sahibamittal: So we don't need method _parsePurlToSnykUrlParam_ at all?

src/main/java/org/dependencytrack/tasks/scanners/SnykAnalysisTask.java
```diff
@@ -0,0 +1,467 @@
...
+                Instant startThread = Instant.now();
+                executor.execute(analysisUtil);
+
+                Instant endThread = Instant.now();
+                duration += Duration.between(startThread, endThread).toMillis();
+            }
+        }
+        try {
+            //System.out.println("countDownLatch" + countDownLatch.getCount());
+            countDownLatch.await();
```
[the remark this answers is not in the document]
japurva1502: Added timeout for 60 minutes

src/main/java/org/dependencytrack/resources/v1/ConfigPropertyResource.java
```diff
@@ -137,4 +135,34 @@ public Response updateConfigProperty(List<ConfigProperty> list) {
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(
+            value = "Creates/updates a config property",
+            response = ConfigProperty.class
+    )
+    @ApiResponses(value = {
+            @ApiResponse(code = 401, message = "Unauthorized")
+    })
+    @PermissionRequired(Permissions.Constants.SYSTEM_CONFIGURATION)
+    public Response upsertsConfigProperty(ConfigProperty json) {
```
nscuro, line 151: Just checked the frontend, this endpoint is not used by it and can be removed. I think this may be a leftover from a previous PR?

PR #2073, opened 2022-10-21
grafana: replace heap usage gauage

valentijnscholten:
The current heap usage gauge is not working correctly. This is because Grafana doesn't support variables in data transformations.
And we're using a transformation to dynamically set the Max value of the gauge, but this fails when using the $instance variable. It only works if the instance name is hardcoded (or left out). It currently is set hardcoded to `apiserver:8080` which probably won't work for most realworld deployments.

This PR switches from a gauge to two stat panels giving similar information. The colours are always green as again if we wanted thresholds and a max value, we would run into the same problems. 

image

If you want colours for heap usage, you can use the micrometer dashboard which shows the percentage of heap used in green - orange - red styles.

### DIFF
--- a/docs/files/grafana-dashboard.json
+++ b/docs/files/grafana-dashboard.json
@@ -201,21 +201,24 @@
           "color": {
             "mode": "thresholds"
           },
-          "mappings": [],
+          "decimals": 2,
+          "mappings": [
+            {
+              "options": {
+                "match": "null",
+                "result": {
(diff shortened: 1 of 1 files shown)


nscuro (approved):
Thanks @valentijnscholten!

valentijnscholten:
For future reference: This is latest state of variable support in Transformations: https://github.com/grafana/grafana/issues/46428

Issue #2109, opened 2022-11-01
Add reported-by for vulnerability aliases


sahibamittal:
### Current Behavior: 

A vulnerability may have different aliases reported by different data sources. But, we don't store the info of who is providing a specific alias info for that vulnerability. 

### Proposed Behavior: 

Add 'Reported By' info for each alias of vulnerability. This would also reflect in notification content and can be useful for internal filters on duplicated vulnerabilities (from different data sources).

nscuro:
Coming in v5 as per https://github.com/DependencyTrack/dependency-track/blob/main/docs/adr/014-new-alias-schema.md

PR #2111, opened 2022-11-01
Issue-2109 : add reportedBy for vulnerability alias

sahibamittal:

Issue: https://github.com/DependencyTrack/dependency-track/issues/2109


nscuro:
Thanks for the PR @sahibamittal. Did you check whether this change results in duplicate aliases being reported or displayed in the frontend? I suspect there will be some areas where deduplication would need to be done. Also the notification logic was not touched, so this does currently not address your original intention of

> This would also reflect in notification content 

---

@stevespringett What are your thoughts on this? Apparently some orgs have a need to know which source is responsible for a given alias association.

nscuro:
> Why does this PR have alias information as part of the persisted object model? The design has these concerns separated.

I'm afraid I don't understand, care to elaborate? The alias information was `@PersistenceCapable` from the very beginning, this PR merely adds a field.

stevespringett:
My bad. I though the PR modified Vulnerability.java, but it's modifying VulnerabilityAlias.java. It was really confusing.  Dyslexic.

I assume this data will only be available via the API, correct?

nscuro:
> I assume this data will only be available via the API, correct?

Yeah. Although to my understanding the primary use case is to have this data available in notifications. That way downstream systems consuming events from DT can make a decision as to which alias declarations they want to "trust".

src/main/java/org/dependencytrack/model/NotificationAlias.java
```diff
@@ -0,0 +1,54 @@
+package org.dependencytrack.model;
```
nscuro, line 1: License header is missing.

src/main/docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM eclipse-temurin:17.0.4.1_1-jre-focal@sha256:55c1ff40a91898c456dc5a874a66db17b6aa1b4aef8d9cac1ad3b194575638a1 AS jre-build
 
-FROM debian:bullseye-20221004-slim@sha256:b46fc4e6813f6cbd9f3f6322c72ab974cc0e75a72ca02730a8861e98999875c7
+FROM debian:bullseye-20221024-slim@sha256:e8ad0bc7d0ee6afd46e904780942033ab83b42b446b58efa88d31ecf3adf4678
```
nscuro, line 3: This change seems unrelated, why is it here?

nscuro (changes requested):
The notifications documentation will need to be updated to reflect the new field: https://docs.dependencytrack.org/integrations/notifications/#new_vulnerability

Also, the alias synchronization logic needs to be updated to consider the `reportedBy` field: https://github.com/DependencyTrack/dependency-track/blob/84fd49d9b0ff436ebd3e2ad360a0e519cce5564f/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java#L489-L542

At the moment, the `reportedBy` field is discarded if `existingAlias` is not `null`. Which effectively means that only ever one `reportedBy` source will be captured.

src/main/docker/Dockerfile
sahibamittal: I guess it got added in some commit automatically, let me change it back

sahibamittal:
> 

So, for model VulnerabilityAlias, should there be list of aliasReportedBy instead? or should we add check for aliasReportedBy as well while searching for existingAlias like below?
```
if (alias.getCveId() != null) {
            filter += "(cveId == :cveId && aliasReportedBy == :aliasReportedBy)";
            params.put("cveId", alias.getCveId());
            params.put("aliasReportedBy", alias.getAliasReportedBy());
        }
```

nscuro:
Hm. The current data model is not very attribution-friendly. Sorry in advance for the wall of text.

> should there be list of aliasReportedBy instead? or should we add check for aliasReportedBy as well while searching for existingAlias

I'd avoid using collections for primitive type values. DN will create additional DB tables to implement this data model (refer to [docs](https://www.datanucleus.org/products/accessplatform/jdo/mapping.html#one_many_relations)), which seems like too much overhead for simple types. A collection would make sense if we wanted to store more details than just the source.

`reportedBy` should always be part of the query, so the filter construction will look somewhat like this:

```java
final Map<String, Object> params = new HashMap<>();
String filter = "(aliasReportedBy == :reportedBy)";
params.put("reportedBy", alias.getAliasReportedBy());
if (alias.getCveId() != null) {
    filter += " && (cveId == :cveId)";
    params.put("cveId", alias.getCveId());
}
if (alias.getSonatypeId() != null) {
    filter += " && (sonatypeId == :sonatypeId || sonatypeId == null)";
    params.put("sonatypeId", alias.getSonatypeId());
}
// ...
```

---

Did some testing with how this behaves. Consider the following test case:

```java
@Test
public void test() {
    // GitHub reports GHSA-123 to alias CVE-123
    final var aliasA = new VulnerabilityAlias();
    aliasA.setCveId("CVE-123");
    aliasA.setGhsaId("GHSA-123");
    aliasA.setAliasReportedBy(Vulnerability.Source.GITHUB);
    qm.synchronizeVulnerabilityAlias(aliasA);

    // OSV reports the same
    final var aliasB = new VulnerabilityAlias();
    aliasB.setCveId("CVE-123");
    aliasB.setGhsaId("GHSA-123");
    aliasB.setAliasReportedBy(Vulnerability.Source.OSV);
    qm.synchronizeVulnerabilityAlias(aliasB);

    // OSV reports another CVE alias for GHSA-123
    final var aliasC = new VulnerabilityAlias();
    aliasC.setGhsaId("GHSA-123");
    aliasC.setCveId("CVE-321");
    aliasC.setAliasReportedBy(Vulnerability.Source.OSV);
    qm.synchronizeVulnerabilityAlias(aliasC);

    // OSV reports another OSV alias for CVE-321
    final var aliasD = new VulnerabilityAlias();
    aliasD.setCveId("CVE-321");
    aliasD.setOsvId("OSV-321");
    aliasD.setAliasReportedBy(Vulnerability.Source.OSV);
    qm.synchronizeVulnerabilityAlias(aliasD);

    // Snyk reports SNYK-123 to alias GHSA-123
    final var aliasE = new VulnerabilityAlias();
    aliasE.setGhsaId("GHSA-123");
    aliasE.setSnykId("SNYK-123");
    aliasE.setAliasReportedBy(Vulnerability.Source.SNYK);
    qm.synchronizeVulnerabilityAlias(aliasE);

    qm.getPersistenceManager().newQuery(VulnerabilityAlias.class).executeList()
            .forEach(alias -> System.out.println(alias));
}
```

With the current logic (`synchronizeVulnerabilityAlias` still untouched), we get this output:

```
VulnerabilityAlias{cveId='CVE-123', ghsaId='GHSA-123', osvId='null', snykId='SNYK-123', aliasReportedBy=GITHUB}
VulnerabilityAlias{cveId='CVE-321', ghsaId='GHSA-123', osvId='OSV-321', snykId='null', aliasReportedBy=OSV}
```

Note how the alias relationships have been merged, and the `reportedBy` is not an accurate representation of what was reported.

If we include `reportedBy` in the `synchronizeVulnerabilityAlias` query, we get this output:

```
VulnerabilityAlias{cveId='CVE-123', ghsaId='GHSA-123', osvId='null', snykId='null', aliasReportedBy=GITHUB}
VulnerabilityAlias{cveId='CVE-123', ghsaId='GHSA-123', osvId='null', snykId='null', aliasReportedBy=OSV}
VulnerabilityAlias{cveId='CVE-321', ghsaId='GHSA-123', osvId='OSV-321', snykId='null', aliasReportedBy=OSV}
VulnerabilityAlias{cveId='null', ghsaId='GHSA-123', osvId='null', snykId='SNYK-123', aliasReportedBy=SNYK}
```

Now, only two of the alias relationships reported by OSV are merged into one (`CVE-123` <-> `GHSA-123` <-> `OSV-321`).

Given there are more duplicated rows now (2 rows describe the `CVE-123` <-> `GHSA-123` relationship, all 4 of them refer to `GHSA-123`), clients like the frontend will need a more sophisticated de-duplication logic (or the API server always de-dupes and outputs a model like [`NotificationAlias`](https://github.com/DependencyTrack/dependency-track/pull/2111/files#diff-1c18a891c1efa62025e6077bdd4fa789114c9e020bb036d3cba823631e9a0285)). At the moment, `CVE-123` will be displayed twice when viewing `GHSA-123` in the UI.

And of course this only works as long as we only track the source name. If we also need timestamps etc., then the model will become more complex.

@stevespringett, does this make sense? Or are we on a entirely wrong path?

stevespringett:
The way the VulnerabilityAlias table was designed, and synchronizeVulnerabilityAlias works, is that a single row would contain all the aliases from the various sources. In the current design, there should never be duplicate rows., So the test case of aliasA and aliasB should never happen. In fact, the code should prevent it from happening. which is why only two rows are (correctly) being returned. 

I would not recommend altering the structure and functionality of this table. 

Adding a reportedBy alias to this table will completely alter the design, synchronization, and the number of rows this table is expected to contain. And it will alter the responses the API is currently returning, potentially breaking compatibility with users that have already integrated with the API.

I would recommend finding an alternate method of recording the attributions outside of this class. Possibly in a log table as its rarely going to be used.

stevespringett:
One additional thought., This table is designed to capture all aliases for the same vulnerability. In the real world, there will be vulnerabilities that will have a CVE, GHSA, SNYK, and Sonatype ID. The test example only has two. Once you have three or more, the reportedBy functionality no longer works at a row level anyway.

sahibamittal:
How about a separate attribution table similar to [FindingAttribution](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/model/FindingAttribution.java)  or [AffectedVersionAttribution](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/model/AffectedVersionAttribution.java)?

PR #1788, opened 2022-07-16
Allow search in project tag view

s-spindler:
This PR fixes / enhances the behavior described in https://github.com/DependencyTrack/dependency-track/issues/405.

PR #2121, opened 2022-11-04
Refactoring in persistence

s-spindler:
Thought I'd come back to clean up a little bit of the changes I did in #1788.

I did some refactoring in other places that looked unnecessarily complex to me. Let me know if these should move to separate PRs. As is, this PR might be easier to review per commit...


nscuro (approved):
Thanks for the PR @s-spindler, sorry for the long wait!

Issue #1632, opened 2022-05-17
Permissions do not follow the evolution of the project versions


step50:
### Current Behavior:
The permissions do not follow the evolution of the project versions

### Steps to Reproduce:
- Create a project with N teams in version 1.0.0
- menu "Portfolio Access Control", the N teams are affected to the project (section 'Project access')
- menu "Project", we have one line for this project
- menu "Project Detail", select 'Add version' and create a new version for this project 1.0.1  with the all include parameters (tags, properties, components, services, audit history)
- menu "Project", we have now 2 lines for this project. The first line with the version 1.0.0 and a second with 1.0.1
- menu "Portfolio Access Control", the N teams are affected only to the project with the version 1.0.0 (section 'Project access') , not the project with the version 1.0.1

### Expected Behavior:
- Create a project with N teams in version 1.0.0
- menu "Portfolio Access Control", the N teams are affected to the project (section 'Project access')
- menu "Project", we have one line for this project
- menu "Project Detail", select 'Add version' and create a new version for this project 1.0.1  with the all include parameters (tags, properties, components, services, audit history). Need to add a parameter for permissions ?
- menu "Project", we have now 2 lines for this project. The first line with the version 1.0.0 and a second with 1.0.1
- menu "Portfolio Access Control", the N teams are affected to the project with the version 1.0.0 (section 'Project access') AND the the version 1.0.1

OR

- Create a project with N teams in version 1.0.0
- menu "Portfolio Access Control", the N teams are affected to the project (section 'Project access')
- menu "Project", we have one line for this project
- menu "Project Detail", select 'Add version' and create a new version for this project 1.0.1  with the all include parameters (tags, properties, components, services, audit history). Need to add a parameter for permissions ?
- menu "Project", we have ONLY 1 line for this project. The choice to select a specific version is available in the project itself
- menu "Portfolio Access Control", the N teams are affected to the project (no permission per version ?)

Question : maybe the same behavior for automatic creation by maven for example ?

### Environment:
  
  - Dependency-Track Version: 4.3.2
  - Distribution: Docker
  - BOM Format & Version: All
  - Database Server: PostgreSQL
  - Browser: Chrome / FireFox

### Additional Details:

valentijnscholten:
duplicate of #1534.

PR #2122, opened 2022-11-05
Allow cloning of ACL during Project cloning

### LINKED ISSUE #1534: Cloning a project should optionally clone the ACL
### Current Behavior:
When cloning a project, the ACL for Portfolio Access Control cannot selected to be included in the clone.

### Proposed Behavior:
Allow the ACL to be cloned similar to Properties, Dependencies, Components, Services, Audit History

valentijnscholten:
Adds an `includeACL` parameter to the project clone api. And an extra `/acl/enabled` endpoint for the front-end to query ACL enabled status.

Fixes #1534 and option 1 of #1632

Front-end PR linked below.

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
@@ -27,6 +27,8 @@ import alpine.model.UserPrincipal;
 import alpine.persistence.PaginatedResult;
 import alpine.resources.AlpineRequest;
 import com.github.packageurl.PackageURL;
+
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.dependencytrack.auth.Permissions;
 import org.dependencytrack.event.IndexEvent;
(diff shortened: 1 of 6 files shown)


valentijnscholten:
@nscuro Do you want to keep the new `/api/v1/acl/enabled` endpoint in place, even though the front-end doesn't use it anymore?

nscuro:
Let's remove that endpoint. It's hard to keep track of which endpoint is used, best to remove it now.

valentijnscholten:
I'm not sure if it really a requirement that the API can only have methods used by the front-end. 
But my feeling was to keep the endpoint as it might be useful for some clients / scripts to be able to know if ACL is enabled or not. Without needing SYSTEM_CONFIGURATION permissions to query the system configuration properties. But the same could be said of other system properties.

nscuro:
Yeah I get that this could be useful. However, as this information *can* indeed be queried already (as you pointed out, it requires the `SYSTEM_CONFIGURATION` permission), I don't see a reason to have this endpoint right now. 

We can still add it back in if someone really needs it.

valentijnscholten:
OK. Removed. You're bringing down my LOC score today ;-)
Maybe a more generic solution could be implemented in the future if needed. I remember me considering the option to make  part of the config properties available to non-admin users as read-only.

nscuro:
> You're bringing down my LOC score today ;-)

Non-existent code can't have bugs, which is always a win in my book! :p

Anyway, thanks for all your work, it really is appreciated.

nscuro (approved):

PR #2124, opened 2022-11-05
Add Dependency-Track as default project

### LINKED ISSUE #1559: Add Dependency-Track as default project
### Current Behavior:
When starting a fresh instance of Dependency-Track, there are no projects in the portfolio.

### Proposed Behavior:
Add Dependency-Track as default project to every new portfolio, with the SBOM that has been generated during the build (should be in the resources folder). 

When a new critical CVE-vulnerability is reported, companies want to know whether any of the applications within their application landscape are affected by this vulnerability. Dependency-Track is part of this application landscape, so I think it should be part of the portfolio. 

Maybe the project can be made read-only, but this would take more effort to implement. I’m also not sure whether the front-end and API server should be added as separate projects, or all components should be added to the same project.

nscuro:
Draft PR to get the discussion going as to how this feature should behave.

When building, the BOM generated at build-time (`target/bom.json`) is merged with the manually-maintained service BOM (`src/main/resources/services.bom.json`) using the CycloneDX CLI. Because this build step requires the CLI to be present, it is disabled by default, and must be enabled by passing the `-Dbom-merge.skip=false` property to the Maven command.

The BOM is moved to `.well-known/sbom` inside the final JAR (this has been the case for a while already). During startup, we try to locate and parse the BOM from there, map the `metadata/component` node to a DT project, create the project if it doesn't already exist, and emit a `BomUploadEvent` for it. This will trigger the usual analysis tasks.

The entire behavior is enabled by default, but can be disabled via `default.project.disable=true` (`DEFAULT_PROJECT_DISABLE=true`). Additionally, tags for the default project may be provided using the `default.project.tags` (`DEFAULT_PROJECT_TAGS`) property.

As it is now, the project will only be created if it doesn't already exist. Upgrading to a newer DT version will cause a new project version to be created (based on the version of the embedded BOM). It will behave like a poor-man's version of subscribing to a service from which the DT BOM would be acquired automatically.

Looking for input and opinions of all sorts. Will add documentation and tests once we agree on how to proceed.

Closes #1559

### DIFF
--- a/.github/workflows/_meta-build.yaml
+++ b/.github/workflows/_meta-build.yaml
@@ -31,12 +31,20 @@ jobs:
           java-version: '17'
           cache: 'maven'
 
+      - name: Setup CycloneDX CLI
+        run: |
+          mkdir -p "$HOME/.local/bin"
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+          wget -O "$HOME/.local/bin/cyclonedx" https://github.com/CycloneDX/cyclonedx-cli/releases/download/v0.24.2/cyclonedx-linux-x64
+          echo "ef0d3b31d176e02bc594f83e19cfcea053c6bc5b197351f71696e189390f851d $HOME/.local/bin/cyclonedx" | sha256sum -c
+          chmod +x "$HOME/.local/bin/cyclonedx"
+
       - name: Build with Maven
         run: |-
           mvn clean
-          mvn package -Dmaven.test.skip=true -P enhance -P embedded-jetty -Dlogback.configuration.file=src/main/docker/logback.xml
+          mvn package -Dmaven.test.skip=true -P enhance -P embedded-jetty -Dbom-merge.skip=false -Dlogback.configuration.file=src/main/docker/logback.xml
           mvn clean -P clean-exclude-wars
-          mvn package -Dmaven.test.skip=true -P enhance -P embedded-jetty -P bundle-ui -Dlogback.configuration.file=src/main/docker/logback.xml
+          mvn package -Dmaven.test.skip=true -P enhance -P embedded-jetty -P bundle-ui -Dbom-merge.skip=false -Dlogback.configuration.file=src/main/docker/logback.xml
           mvn clean -P clean-exclude-wars
           mvn cyclonedx:makeBom
 
--- a/pom.xml
+++ b/pom.xml
@@ -103,6 +103,9 @@
         <plugin.retirejs.breakOnFailure>false</plugin.retirejs.breakOnFailure>
         <!-- SonarCloud properties -->
         <sonar.exclusions>src/main/webapp/**</sonar.exclusions>
+        <!-- BOM merging -->
+        <bom-merge.cyclonedx-cli.path>cyclonedx</bom-merge.cyclonedx-cli.path>
+        <bom-merge.skip>true</bom-merge.skip>
     </properties>
 
     <repositories>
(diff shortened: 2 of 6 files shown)


valentijnscholten:
Shouldn't the BOM also include the front-end dependencies? Not completely related to this PR maybe, but a valid question.

stevespringett:
@valentijnscholten. No, the frontend dependencies are already distributed with the frontend and available at .well-known/sbom. So we distribute the api server sbom with the api server and the frontend sbom with the frontend.  I guess a case can be made for the bundled distribution, however, based on some of the architecture changes we're making in the future, I think the bundled distro will eventually be end-of-life, so likely should not worry about it.

valentijnscholten:
Ok, didn't notice/check the SBOM in the frontend.

valentijnscholten:
I guess what I meant is shouldn't the front-end SBOM also be uploaded into the default project?

stevespringett:
Ah yes I guess it should be uploaded. Possibly to a front end project, but yes.  Im not sure how the api server will get it though since it doesn't know where the frontend is deployed. Will need some thought, especially since the frontend version and upgrade cycle are independent of the api server.

syalioune:
My two cents :

- The default project should be partially read-only. SBOM or component update should not be possible even for administrators. Vulnerability triage should however be possible by administrators
- Should any change to `default.project.tags` impact existing project versions ? I would say yes
- Regarding the frontend's default project, since the apiserver can't easily locate the SBOM, it could be the responsibility of the frontend to check if the default project exists (name and correct version) and upload the SBOM using REST API. The upload should be restricted to sessions with the appropriate admin rights. Tricky part is concurrency : N admin connecting at the same time after rollout and thus triggering N uploads...

nscuro:
@syalioune Interesting. What I get from your points is that you'd expect DT to continuously maintain its own projects?

I guess the idea of a default project started out to just have something in the portfolio when DT initially launches, making discoverability of features a bit easier. I like the "continuous dogfeeding" idea though, maybe something for a later enhancement? 🤔

---

> The default project should be partially read-only. SBOM or component update should not be possible even for administrators. Vulnerability triage should however be possible by administrators

That will require a more nuanced ACL system than what we currently have. At the moment we just differentiate between "can access" and "can't access", but we do not have fine-grained read/write permissions within that.

---

I'm torn on the frontend part, and potentially other services when we introduce them. It'd be nice to have, but would either require:

* The API server to know about all services and their versions being deployed, or
* All services to publish their BOMs to the API server

I don't like the thought of the frontend performing automated actions in the name of the user.

stevespringett:
@nscuro we may want to think about the ability to use Project Koala, which is already planned to be implemented in DT, as a way to inform DT instances of the SBOM and VEX for given versions of Dependency-Track.

Supplying the SBOM is great and all, but we also need to inform the community of vulnerable components that are not exploitable in the context of DT.

nscuro:
@stevespringett Perhaps we should prioritize this as it would be (to my knowledge) an industry first.

nscuro:
I figure this whole topic may be bigger and ultimately we prefer a solution that involves usage of the BOM exchange API.

However, I'd like to have the BOM merging part in 4.7 so that our BOM includes the services we defined in `services.bom.json`. I'll raise another PR for that.

nscuro:
Closing in favor of project Koala. Thanks everyone for your input!

msymons:
And as for what "Koala" is,  it is cute and cuddly and is the implementation of [SBOM Exchange API](https://cyclonedx.org/news/owasp-cyclonedx-launches-sbom-exchange-api-standardizing-sbom-distribution/)

PR #2128, opened 2022-11-06
Add reanalyze button for projects

### LINKED ISSUE #2116: Add an "Analyze now" button to Projects
### Current Behavior:
Currently projects are re-analyzed every 24 hours (or as configured by the admin). 

### Proposed Behavior:
Sometimes there is a need to do an immediate analysis, for example when a important vulnerability has come out, or you're preparing a release. In this scenario it might help to have an Analyze Now button to analyze a project right now. 

### Notes
- Currently a user can already do this by downloading the SBOM and immediately reuploading it
- Reanalyzing "now" might still not lead to the perfect result due to the 12h (default) analysis cache. Should the reanalyze now button bypass the cache? Should there be two options? Should there be a separate clear cache option (for admins)?
-

valentijnscholten:
This PR adds `/api/v1/findings/project/{uuid}/analyze` to trigger a reanalysis of a project's components.

This will perform an asynchronous analyis, returning a token to check for completion.
The vulnerability analysis is identical to the scheduled vulnerability analysis, so any cached results that haven't expired will be used.

Front-end part with button on Project view is linked below.

Fixes #2116 

Questions:
- Does it have enough value if it's using the cache?

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/FindingResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/FindingResource.java
@@ -156,7 +156,7 @@ public class FindingResource extends AlpineResource {
             final Project project = qm.getObjectByUuid(Project.class, uuid);
             if (project != null) {
                 if (qm.hasAccess(super.getPrincipal(), project)) {
-                  LOGGER.info("Analyzing Project " + project.getUuid());
+                  LOGGER.info("Analysis of project " + project.getUuid() + " requested by " + super.getPrincipal().getName());
 
                   final List<Component> detachedComponents = qm.detach(qm.getAllComponents(project));
                   final Project detachedProject = qm.detach(Project.class, project.getId());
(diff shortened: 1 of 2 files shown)


src/main/java/org/dependencytrack/resources/v1/FindingResource.java
```diff
@@ -128,4 +136,42 @@ public Response exportFindingsByProject(@PathParam("uuid") String uuid) {
...
+    })
+    @PermissionRequired(Permissions.Constants.VIEW_VULNERABILITY)
+    public Response analyzeProject(
+            @ApiParam(value = "The UUID of the project to analyze", required = true)
+            @PathParam("uuid") String uuid) {
+        try (QueryManager qm = new QueryManager()) {
+            final Project project = qm.getObjectByUuid(Project.class, uuid);
+            if (project != null) {
+                if (qm.hasAccess(super.getPrincipal(), project)) {
+                  LOGGER.info("Analyzing Project " + project.getUuid());
```
nscuro, line 159: We probably want to log *who* triggered the analysis so abuse can be detected:

```suggestion
                  LOGGER.info("Analysis of project " + project.getUuid() + " requested by " + super.getPrincipal().getName());
```

nscuro (changes requested):

nscuro (approved):
Thanks @valentijnscholten!

Issue #2136, opened 2022-11-08
Project version retention policy


antonsulim0000:
### Current Behavior:
There is no standard automated way of stale project version deletion

### Proposed Behavior:
Implement general retention policy that keeps N most recently uploaded versions of the project

Relates to [BOM Retention Policy](https://github.com/DependencyTrack/dependency-track/issues/877)
See discussion in [#proj-dependency-track](https://owasp.slack.com/archives/C6R3R32H4/p1664865428095589)

MedUnes:
I have been struggling long time automating this feature through CI/CD pipelines, was even more tedious as pagination logic had to be implemented as well to make sure all version are tracked. Bash script wasn't the best for complex logic handling. I ended up packing that into this mini cli tool here.
https://github.com/MedUnes/dtrack-cli

for now it allows to list all version, flag all but a given version as inactive, and also flag a specific version as latest. (useful just after uploading an sbom file)

last but not least, it also allows uploading an sbom file in one liner.

nscuro:
Shipping in v5.

Issue #877, opened 2020-12-17
BOM Retention Policy


msymons:
### Current Behavior:
Per post in Slack by @stevespringett regarding what happens to BOMs when they are uploaded to Dependency-Track:

> The original BOMs are dismissed. When BOMs are uploaded, they are validated, processed, and discarded. When the BOMs are processed, a lot of the data (but not all) available in the BOM is applied to the DT object model. When a user requests a BOM be created for a project...  the BOM is dynamically constructed from the DT object model. So it’s not original, but you can get a BOM back.

...and then:

> I have been thinking about an option to keep BOMs that have been uploaded along with a retention policy.

### Proposed Behavior:

An option to keep BOMs at have been uploaded, along with a retention policy, would be very useful indeed.

First of all, I think that the BOM API endpoint (ie upload) would need to be extended to implement a parameter to tell DT whether or not to keep the uploaded BOM.  The default should be "false".  ie,  Keeping a BOM needs to be a sin of commission not omission.

`/v1/bom/cyclonedx/project/{uuid}` would need an optional parameter(s) in order to extract a kept BOM.

In my environment, we use Jenkins multi-branch scripted pipeline jobs and perform nightly builds of trunk/master that uploads a BOM.   ie, version is alway SNAPSHOT.  Something that needs no retention at all.  Implementation of this enhancement should allow tweaking of the  pipelines such that release builds could generate release BOMs.  It is these that we would want to explicitly tell DT to keep.

We have three environments: Integration, Staging, Production.   Anything that is released is automatically deployed to Integration.  However, not everything that is deployed to Integration gets promoted to Staging, and even fewer versions get promoted from there to Production.  This is worth mentioning because it has ramifactions for two or three things in DT:

* When a BOM is uploaded you might now you want to keep it but you do not have any idea (yet) of how long you will want to keep it
* We thus need a way to tag/label retained BOMs dynamically... and do so at any time after initial upload. This would allow us  to indicate what environment a BOM can be "found" in. We already do scripting that extracts all this info from our environments so it would not be too much extra work to update DT if DT allowed such to be done.  Note that there is currently no API endpoint for changing tags (but what is needed in this situation might be "something like a tag but not a tag" )
* A retained BOM might be in Integration AND Staging AND Production all at the same time.
* Should it be possible to upload old BOMs (if you have them to hand) to flesh things out?
[...]

PR #2138, opened 2022-11-08
Feature: implement general project retention policy

antonsulim0000:
### Current Behavior:
There is no standard automated way of stale project version deletion

### Proposed Behavior:
Implement general project retention policy that keeps N most recently uploaded versions of the project

Solves [Project retention policy enhancement request](https://github.com/DependencyTrack/dependency-track/issues/2136)
Relates to [BOM Retention Policy](https://github.com/DependencyTrack/dependency-track/issues/877)
See discussion in [#proj-dependency-track](https://owasp.slack.com/archives/C6R3R32H4/p1664865428095589)


src/main/java/org/dependencytrack/event/ApplyRetentionPolicyEvent.java
```diff
@@ -0,0 +1,10 @@
+package org.dependencytrack.event;
+
+import alpine.event.framework.Event;
+
+/**
+ * Defines an event used to apply the general retention policy
+ */
+public class ApplyRetentionPolicyEvent implements Event {
```
stevespringett, line 8: Knowing that a BOM retention policy is coming, I think this class name should be changed to `ApplyProjectRetentionPolicyEvent` so that its clear it is project-related.

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -25,6 +25,7 @@ public enum ConfigPropertyConstants {
 
     GENERAL_BASE_URL("general", "base.url", null, PropertyType.URL, "URL used to construct links back to Dependency-Track from external systems"),
     GENERAL_BADGE_ENABLED("general", "badge.enabled", "false", PropertyType.BOOLEAN, "Flag to enable/disable SVG badge support from metrics"),
+    GENERAL_RETENTION_POLICY("general", "retention.policy.keep.recent", null, PropertyType.INTEGER, "Number of recent project versions to keep, all by default"),
```
stevespringett, line 28: Change `retention.policy.keep.recent` to `project.retention.policy.keep.recent`

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -88,7 +89,8 @@ public enum ConfigPropertyConstants {
...
     TASK_SCHEDULER_REPOSITORY_METADATA_FETCH_CADENCE("task-scheduler", "repository.metadata.fetch.cadence", "24", PropertyType.INTEGER, "Metadada fetch cadence (in hours) for package repositories"),
     TASK_SCHEDULER_INTERNAL_COMPONENT_IDENTIFICATION_CADENCE("task-scheduler", "internal.components.identification.cadence", "6", PropertyType.INTEGER, "Internal component identification cadence (in hours)"),
-    TASK_SCHEDULER_COMPONENT_ANALYSIS_CACHE_CLEAR_CADENCE("task-scheduler", "component.analysis.cache.clear.cadence", "72", PropertyType.INTEGER, "Cleanup cadence (in hours) for component analysis cache");
+    TASK_SCHEDULER_COMPONENT_ANALYSIS_CACHE_CLEAR_CADENCE("task-scheduler", "component.analysis.cache.clear.cadence", "72", PropertyType.INTEGER, "Cleanup cadence (in hours) for component analysis cache"),
+    TASK_SCHEDULER_RETENTION_POLICY_CADENCE("task-scheduler", "retention.policy.cadence", "24", PropertyType.INTEGER, "Apply cadence (in hours) for retention policy execution");
```
stevespringett, line 93: Change `retention.policy.cadence` to `project.retention.policy.cadence`

src/main/java/org/dependencytrack/tasks/ApplyRetentionPolicyTask.java
```diff
@@ -0,0 +1,75 @@
...
+import org.dependencytrack.event.ApplyRetentionPolicyEvent;
+import org.dependencytrack.model.Project;
+import org.dependencytrack.persistence.QueryManager;
+
+/**
+ * Subscriber task that applies general retention policy.
+ * <p> Keeps N most recently uploaded versions of the project.
+ * <p> By default, keeps all versions.
+ */
+public class ApplyRetentionPolicyTask implements Subscriber {
```
stevespringett, line 23: Change `ApplyRetentionPolicyTask` to `ApplyProjectRetentionPolicyTask`

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
stevespringett: Change `GENERAL_RETENTION_POLICY` to `GENERAL_PROJECT_RETENTION_POLICY`

stevespringett (changes requested):
Thanks for the PR. Just a few minor changes that I see so as to avoid confusion when other retention policies are added (such as BOM retention policy)

src/main/java/org/dependencytrack/event/ApplyRetentionPolicyEvent.java
antonsulim0000: Hi @stevespringett , thank you for the review. It makes sense, I updated the name everywhere.

valentijnscholten:
I think it would be more useful to have a retention policy mechanism that "delete project versions that haven't had a BOM upload for over XXX months" instead of "keeping N project versions based on their most recent bom upload".

antonsulim0000:
> I think it would be more useful to have a retention policy mechanism that "deletes project versions that haven't had a BOM upload for over XXX months" instead of "keeping N project versions based on their most recent bom upload".

Different retention policies are not mutually exclusive and can be plugged at further stages.

In my opinion and use case it makes more sense to keep several versions of the project to see the current state and a trend. I want to keep it for all modules. It does not matter when I built it and deployed. E.g. there are services which are deployed only several times per year (and I still want to keep them). On the other hand there are ones that are deployed several times per day (and I want to cleanup them).

nscuro:
I'm also torn on this one. As @savek-cc pointed out, retention policies highly depend on how your org works and how it chose to integrate DT into its software lifecycle. A policy may be 

```py
# Delete project versions with "critical" tag after 120 days of inactivity, when there are more than 5 versions
'critical' in tags and len(versions) > 5 and lastBomUpload < (now() - 120 days)
```

I'd like to avoid having to deal with multiple input fields or config property values for every single combination that users may want, as it gets very messy and confusing really fast.

jrevillard:
> I'm also torn on this one. As @savek-cc pointed out, retention policies highly depend on how your org works and how it chose to integrate DT into its software lifecycle. A policy may be
> 
> ```python
> # Delete project versions with "critical" tag after 120 days of inactivity, when there are more than 5 versions
> 'critical' in tags and len(versions) > 5 and lastBomUpload < (now() - 120 days)
> ```
> 
> I'd like to avoid having to deal with multiple input fields or config property values for every single combination that users may want, as it gets very messy and confusing really fast.

For me, if the functionality is not as powerful as the one that you describe above, this is useless to implement it inside DT and it's better to leave it to the clients (using the API). There are indeed so many ways to integrate DT in a company...

Best,
Jerome

nscuro:
@stevespringett Any opinions on this one?

nscuro:
From my POV this PR is unlikely to get merged. If anyone has different views, please bring them forward until next week.

I'll close this PR if there's no feedback by friday next week.

In any case, thanks @antonsulim0000 for your contribution. It is appreciated, we just have to make sure that we cater to more than one use case with features like this.

PR #2169, opened 2022-11-19
Support AuthN and AuthZ disabled

### LINKED ISSUE #653: alpine.enforce.authentication=false produces 404s
The defect may already be reported! Please search for the defect before creating one.

### Current Behavior:
When setting the alpine.enforce.authentication and authorization properties to false, the webserver returns a 404 response for all URLs.

Additionally setting the propertise as environment variables seems to have no effect (e.g. ALPINE_ENFORCE_AUTHENTICATION still show the login screen)

### Steps to Reproduce:
Set alpine.enforce.authentication to false. Sample docker-compose and properties file below

### Expected Behavior:
The depedencytrack dashboard appears with no login screen prompted

### Environment:
  
  - Dependency-Track Version: latest (digest efc65e702ee1)
  - Distribution: Docker
  - BOM Format & Version: N/A
  - Database Server: H2
  - Browser: Firefox | Google Chrome

### Additional Details:

docker-compose.yaml:
```
version: "3"
services:
  dependency-track:
    image: "owasp/dependency-track"
    command: java -XX:MaxRAMPercentage=90.0 -Dalpine.application.properties=/app/props/application.properties -DdependencyTrack.logging.level=INFO -jar dependency-track-embedded.war -context ${CONTEXT}
    ports:
      - 8080:8080
    volumes:
      - ./data:/data
      - ./props:/app/props/application.properties
```
NOTE: Use all other default properties from https://docs.dependencytrack.org/getting-started/configuration/
props/application.properties
```
############################ Alpine Configuration ###########################
# Optional
# When authentication is enforced, API keys are required for automation, and
# the user interface will prevent anonymous access by prompting for login
# credentials.
alpine.enforce.authentication=false

# Optional
# When authorization is enforced, team membership for both API keys and user
# accounts are restricted to what the team itself has access to. To enforce 
# authorization, the enforce.authentication property (above) must be true.
alpine.enforce.authorization=false
```

samrocketman:
Description
-----------

I use DependencyTrack locally through VSCode Dev Containers.  Authentication and authorization setup is tedious every time I rebuild my development environment.

This adds support for uploading SBOM while authentication and authorization is disabled.

Addressed Issue
---------------

Is related to:

* Issue https://github.com/DependencyTrack/frontend/issues/333
* Pull Request https://github.com/DependencyTrack/frontend/pull/334
* relates to #653

Additional Details
------------------

In current master branch, starting apiserver, if I set the following environment:

```yaml
ALPINE_ENFORCE_AUTHENTICATION: 'false'
ALPINE_ENFORCE_AUTHORIZATION: 'false'
```

And I attempt to upload a BOM, then I get the following error.

```
The principal does not have permission to create project.
```

I'm new to this project so still working to figure out writing tests for it.  I have manually validated it works by building it and testing it in conjunction with the frontend pull request.

My development host is Ubuntu 20.04 Linux and I built the Docker image with the following commands.

```bash
docker run -it --rm -u "$(id -u):$(id -g)" -v "$PWD:$PWD" -w "$PWD" maven:3-eclipse-temurin-17-alpine /bin/bash -exc '
  mvn package -Dmaven.test.skip=true -P enhance -P embedded-jetty -Dlogback.configuration.file=src/main/docker/logback.xml
  mvn clean -P clean-exclude-wars
  '

docker build -t apiserver:snapshot -f src/main/docker/Dockerfile .
```

Checklist
---------

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/ExtendedAlpineResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ExtendedAlpineResource.java
@@ -1,0 +1,72 @@
+/*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Sam Gleske. All Rights Reserved.
+ * Copyright (c) Steve Springett. All Rights Reserved.
+ */
+package org.dependencytrack.resources.v1;
+
+import alpine.Config;
+import alpine.model.ManagedUser;
+import alpine.server.resources.AlpineResource;
+import java.security.Principal;
+import org.dependencytrack.persistence.QueryManager;
+
+/**
+ * An extension of AlpineResource that provides all of the same features but for
+ * Dependency-Track also enables disabled authentication and authorization from
+ * Alpine.  Anonymous is treated the same as admin user or Administrators team.
+ * @since 4.7.0
+ */
+public abstract class ExtendedAlpineResource extends AlpineResource {
+    /**
+     * Returns the principal for who initiated the request.  If
+     * ALPINE_ENFORCE_AUTHENTICATION is disabled then the admin ManagedUser is
+     * returned.
+     * @return a Principal object
+     * @see alpine.model.ApiKey
+     * @see alpine.model.LdapUser
+     * @see alpine.model.ManagedUser
+     */
+    @Override
+    protected Principal getPrincipal() {
+        if (Config.getInstance().getPropertyAsBoolean(Config.AlpineKey.ENFORCE_AUTHENTICATION)) {
+            // Authentication is not enabled, try returning admin principal
+            try (QueryManager qm = new QueryManager()) {
+                final ManagedUser user = qm.getManagedUser("admin");
+                if (user != null) {
+                    return (Principal) user;
+                }
+            }
+        }
+        return super.getPrincipal();
+    }
+
+    /**
+     * Convenience method that returns true if the principal has the specified permission,
+     * or false if not.  If ALPINE_ENFORCE_AUTHENTICATION is disabled then the
+     * true will always be returned.
+     * @param permission the permission to check
+     * @return true if principal has permission assigned, false if not
+     */
+    @Override
+    protected boolean hasPermission(final String permission) {
+        if (!Config.getInstance().getPropertyAsBoolean(Config.AlpineKey.ENFORCE_AUTHENTICATION)) {
+            return true;
+        }
+        return super.hasPermission(permission);
+    }
+}
--- a/src/main/java/org/dependencytrack/resources/v1/ExtendedAlpineResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ExtendedAlpineResource.java
@@ -43,7 +43,7 @@ public abstract class ExtendedAlpineResource extends AlpineResource {
      */
     @Override
     protected Principal getPrincipal() {
-        if (Config.getInstance().getPropertyAsBoolean(Config.AlpineKey.ENFORCE_AUTHENTICATION)) {
+        if (!Config.getInstance().getPropertyAsBoolean(Config.AlpineKey.ENFORCE_AUTHENTICATION)) {
             // Authentication is not enabled, try returning admin principal
             try (QueryManager qm = new QueryManager()) {
                 final ManagedUser user = qm.getManagedUser("admin");
--- a/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
@@ -21,6 +21,7 @@ package org.dependencytrack.resources.v1;
 import alpine.Config;
 import alpine.common.logging.Logger;
 import alpine.model.ApiKey;
+import alpine.model.ManagedUser;
 import alpine.model.Team;
 import alpine.server.auth.PermissionRequired;
 import alpine.server.resources.AlpineResource;
@@ -293,6 +294,16 @@ public class TeamResource extends AlpineResource {
                     }
                 } else {
                     return Response.status(Response.Status.BAD_REQUEST).entity("Invalid API key supplied.").build();
+                }
+            }
+        }
+        // Authentication is not enabled, try to return Administrators team
+        try (QueryManager qm = new QueryManager()) {
+            final ManagedUser user = qm.getManagedUser("admin");
+            if (user != null) {
+                if(!user.getTeams().isEmpty()) {
+                    final Team team = user.getTeams().get(0);
+                    return Response.ok(team).build();
                 }
             }
         }
--- a/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
@@ -24,7 +24,6 @@ import alpine.model.ApiKey;
 import alpine.model.ManagedUser;
 import alpine.model.Team;
 import alpine.server.auth.PermissionRequired;
-import alpine.server.resources.AlpineResource;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
@@ -58,7 +57,7 @@ import java.util.List;
  */
 @Path("/v1/team")
 @Api(value = "team", authorizations = @Authorization(value = "X-Api-Key"))
-public class TeamResource extends AlpineResource {
+public class TeamResource extends ExtendedAlpineResource {
 
     private static final Logger LOGGER = Logger.getLogger(TeamResource.class);
 
--- a/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
@@ -297,14 +297,10 @@ public class TeamResource extends ExtendedAlpineResource {
             }
         }
         // Authentication is not enabled, try to return Administrators team
-        try (QueryManager qm = new QueryManager()) {
-            final ManagedUser user = qm.getManagedUser("admin");
-            if (user != null) {
-                if(!user.getTeams().isEmpty()) {
-                    final Team team = user.getTeams().get(0);
-                    return Response.ok(team).build();
-                }
-            }
+        final ManagedUser user = (ManagedUser) getPrincipal();
+        if (user != null && !user.getTeams().isEmpty()) {
+            final Team team = user.getTeams().get(0);
+            return Response.ok(team).build();
         }
         // Authentication is not enabled, but we need to return a positive response without any principal data.
         return Response.ok().build();
--- a/src/main/java/org/dependencytrack/resources/v1/UserResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/UserResource.java
@@ -34,7 +34,6 @@ import alpine.server.auth.JsonWebToken;
 import alpine.server.auth.OidcAuthenticationService;
 import alpine.server.auth.PasswordService;
 import alpine.server.auth.PermissionRequired;
-import alpine.server.resources.AlpineResource;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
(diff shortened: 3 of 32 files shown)


samrocketman (commented):
Upon self review I think I need to override `AlpineResource` methods in `TeamResource` similar to how I overrode `getPrincipal()` and `hasPermission()` methods in `UserResource`.  I'll make these changes.

I also still need to write tests which is why I left the boxes unchecked in the PR description.

samrocketman (commented):
Upon self review I noticed there are many classes extending `AlpineResource`.  I think it makes more sense to centralize overriding `getPrincipal()` and `hasPermission()` in a centralized `ExtendedAlpineResource` class and rewrite all classes to extend from the `ExtendedAlpineResource` class.

samrocketman:
I think I'm at  good stopping point for the evening.

@stevespringett let me know what you think and if this seems acceptable I'll proceed with filling out the rest of the checklist in the PR description.

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -296,6 +296,16 @@ public Response getSelf() {
                 }
             }
         }
+        // Authentication is not enabled, try to return Administrators team
+        try (QueryManager qm = new QueryManager()) {
+            final ManagedUser user = qm.getManagedUser("admin");
```
samrocketman, line 301: This could probably be simplified to calling `getPrincipal()`.

src/main/java/org/dependencytrack/resources/v1/UserResource.java
```diff
@@ -307,6 +306,13 @@ public Response getSelf() {
                 return Response.status(401).build();
             }
         }
+        // Authentication is not enabled, try returning admin username
+        try (QueryManager qm = new QueryManager()) {
+            final ManagedUser user = qm.getManagedUser("admin");
```
samrocketman, line 311: This could probably be simplified to calling `getPrincipal()`.

src/main/java/org/dependencytrack/resources/v1/ExtendedAlpineResource.java
```diff
@@ -0,0 +1,72 @@
...
+     * ALPINE_ENFORCE_AUTHENTICATION is disabled then the admin ManagedUser is
+     * returned.
+     * @return a Principal object
+     * @see alpine.model.ApiKey
+     * @see alpine.model.LdapUser
+     * @see alpine.model.ManagedUser
+     */
+    @Override
+    protected Principal getPrincipal() {
+        if (Config.getInstance().getPropertyAsBoolean(Config.AlpineKey.ENFORCE_AUTHENTICATION)) {
```
samrocketman, line 46:
```suggestion
        if (!Config.getInstance().getPropertyAsBoolean(Config.AlpineKey.ENFORCE_AUTHENTICATION)) {
```

samrocketman (commented):
I'll make a few more cleanup changes

valentijnscholten:
It could be helpful to have some kind of mechanism to load a default dataset when starting / deploying DT. This data could be:
- optional, for example only for dev work, initial setup, load tests, demo instance, ...
- predictable in terms of users / projects / apikeys being available.

I don't think this PR is the right thing to do. Even if you really want something like this it's probably better to inject something somewhere in a filter or annotation instead of modifying "all" Resource(s).

nscuro:
Most software I know of allows users to specify the initial admin PW via environment variables. That way the password change after the first launch can be skipped. This could be further extended to add an API key for the `Administrators` team, making API calls with elevated privileges possible immediately after the container has started. Would be useful for automated provisioning as well.

samrocketman:
> > Authentication and authorization setup is tedious every time I rebuild my development environment.
> 
> Could you please elaborate on this point? What exactly is tedious about it?
> 
> I suspect there is a better way to address your pain points rather than disabling authentication and authorization. Writing and maintaining code to disable a security feature feels backward to me.

Have you used VSCode with Dev Containers?  Knowing might help me explain more context.  I'll do my best with the assumption you have not.

# About VSCode Dev Containers

When you open a project with VSCode that supports Dev containers for the first time you get a prompt, "Reopen in Container?"

Under the hood what happens is the following:

* Docker or, in my case, `docker-compose` provisions a stack of services.  You can think of these services as two categories:
  * Services your development might depend on... in my case I have SonarQube, DependencyTrack, and HashiCorp Vault (because I develop against the Vault API in some cases).
  * The primary Dev Container in which all of your development tooling is installed (Gradle, Maven, JDK Java, GCC, etc).  The IDE opens your code base "remotely" inside of the Dev container.  You're editing files from within the Dev Container (not on your host).
* As development proceeds, there are often use cases for me me to rebuild my Dev environment.  Maybe I'm changing up VSCode tasks which require software to be pre-installed (e.g. let's say I added ShellCheck).  Rebuilding would reprovision everything from scratch.
* First time setup, the more setup you have to do when working in a Dev Container, then the more effort is required before you start actually being productive.

Upon reprovisioning first time setup will need to occur again.  

# Current Flow

Without this pull request

- Reopen in Dev Container and wait around 5 minutes or so for all of the Docker images to be pulled, dependent services to be provisioned, and so on.
- **Manually log into and configure DependencyTrack**.  I'm not developing DependencyTrack.  I am using it to improve security on the project I am developing.  Having to manually click through the UI means I spend roughly an additional 10 minutes of manual setup.  If I had to do this for all of my local development services then I would have quite some time before I'm actually productive.  Then I have to think about how to manage the secrets because VSCode has no straight forward way for this.  I would likely just write the token to `~/.bash_profile` as an environment variable but this is really left up to the implementer to figure out.
- I run a Dependency scan.
- I visit the Web UI and get a login portal.  Now I have to think about what credential I reset admin to when I changed password.

# Why no login?

- DependencyTrack is only accessible locally from the internal docker network.  There's no public ports.  No remote access just my personal copy of DT.
- Even a default password is inconvenient because when I go to view my dev scans in the frontend I have to face a login prompt.

# Ideal flow

- Reopen in Dev container wait around 5 minutes or so for all services and containers to be provisioned.
- Run Task for DependencyTrack scan
- Run task to open web browser and see scan results right away (no login portal).

# Mindset shift

I understand DependencyTrack is typically a centrally hosted service meant to be authenticated with and injected into the CICD pipeline.  I use it for open source development and do not have the resources for my projects to pay for hosting so instead I put more effort into local development automation.

# Example

My usage of Dev Containers is meant to get me up and running for seamless development from a Linux host regardless of the machine.

If not on Linux you can also `vagrant up` (does not work on new ARM Macs).

https://github.com/samrocketman/jervis/tree/main/.devcontainer

src/main/java/org/dependencytrack/resources/v1/ExtendedAlpineResource.java
```diff
@@ -0,0 +1,72 @@
...
+     * ALPINE_ENFORCE_AUTHENTICATION is disabled then the admin ManagedUser is
+     * returned.
+     * @return a Principal object
+     * @see alpine.model.ApiKey
+     * @see alpine.model.LdapUser
+     * @see alpine.model.ManagedUser
+     */
+    @Override
+    protected Principal getPrincipal() {
+        if (!Config.getInstance().getPropertyAsBoolean(Config.AlpineKey.ENFORCE_AUTHENTICATION)) {
```
samrocketman, line 46: I could gate this behind disable authorization instead.

samrocketman (commented):
A potential compromise?

src/main/java/org/dependencytrack/resources/v1/ExtendedAlpineResource.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Convenience method that returns true if the principal has the specified permission,
+     * or false if not.  If ALPINE_ENFORCE_AUTHENTICATION is disabled then the
+     * true will always be returned.
+     * @param permission the permission to check
+     * @return true if principal has permission assigned, false if not
+     */
+    @Override
+    protected boolean hasPermission(final String permission) {
+        if (!Config.getInstance().getPropertyAsBoolean(Config.AlpineKey.ENFORCE_AUTHENTICATION)) {
```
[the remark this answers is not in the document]
samrocketman: Would it make sense for me to open a PR to the Alpine framework?  In the case of authorization being disabled, if it is then it should always return true for hasPermission.

samrocketman:
Code aside, is there still disagreement on the basis of this feature?  I’d like authorization and authentication to be disabled as an option.  There was an issue and proposals encouraged.

This is a permission scheme I would regularly use considering my open source development style.

samrocketman:
Changed “closes” issue to “relates to” since I don’t think this would fully cover the other user’s issue but it does cover my use case.  (I’m not trying to put this behind an alternate login portal)

valentijnscholten:
> > It could be helpful to have some kind of mechanism to load a default dataset when starting / deploying DT…
> 
> This already exists. Use an external DB and do a DB import. This can also be done via docker-compose locally fairly easily.
There is no sample dataset currently. The idea I referred to is something similar to Defect Dojo where there is a sample data set. This is a database independant json filem so can be used with MySQL, Postgres, H2, etc. But it might be enough here to have a configurable admin password that you can use in your scripting/tooling.

 
> > I don't think this PR is the right thing to do. Even if you really want something like this it's probably better to inject something somewhere in a filter or annotation instead of modifying "all" Resource(s).
> 
> Can you elaborate? I don’t understand what you mean by injecting via filter. By “right thing to do” do you disagree with the idea of disabling auth or do you mean the suggested alternate implementation via filter or annotations is preferred to disable auth?
First of all I don't understand your use case, it doesn't like something commonly used and doesn't align with how DT is designed to be used. Chances are you run into more issues later on. If there's a simple flag or small code modification to Alpine it could be acceptable to have it in, but I don't think it's a good idea to modify 30+ files and the maintenance burden that comes with it.

samrocketman:
> There is no sample dataset currently. The idea I referred to is something similar to Defect Dojo where there is a sample data set. This is a database independant json filem so can be used with MySQL, Postgres, H2, etc. But it might be enough here to have a configurable admin password that you can use in your scripting/tooling.

It is not really in scope for the intent of this PR.  If you refer to my comment on how development proceeds for open source projects I contribute there's no need for authorization.

> > > I don't think this PR is the right thing to do. Even if you really want something like this it's probably better to inject something somewhere in a filter or annotation instead of modifying "all" Resource(s).

> First of all I don't understand your use case, it doesn't like something commonly used and doesn't align with how DT is designed to be used. Chances are you run into more issues later on. If there's a simple flag or small code modification to Alpine it could be acceptable to have it in, but I don't think it's a good idea to modify 30+ files and the maintenance burden that comes with it.

Tests will be written when an implementation is agreed upon for all 30 files (or whatever form the final impl is).  Tests ease maintenance burden.

The use case is simple I think.  It is basically single user mode (no authN and no authZ).  In terms of threat model, the application is offline, only available on a local host, and only used by a single user (in this case a developer running dependency scans on projects being developed on their laptop).

Steve mentioned a few times that while Alpine supports disabled auth DT does not and PRs welcome.  I don't think the concept of disabling authN and authZ is odd.

I considered opening a PR to Alpine but this is very much a DT specific modification.  I'm open to making it generic.

I do not understand what changes have been requested which is why I'm asking for more info or examples.

nscuro:
First of all I really do appreciate the discussion here, and also the effort you're putting into making your point understandable @samrocketman 🙏

On the general topic of being able to disable authN / authZ: I'm still not sold, and I still think DT should not support it. If anything, there should be improvements to get up-and-running faster, but disabling security features should not be part of it. Speaking for myself here obviously. @stevespringett your opinion would be nice to have.

Looking at your workflow, it seems like all you really need is a vulnerability scanner for your SBOM. DT is not exactly lightweight, and does a lot more than one-off scans. Depending on DT for your local workflow means you'll have a container sitting around that consumes a good chunk of RAM, just to upload an SBOM every once in a while.

Being a maintainer of DT, it kind of hurts me to say it, but maybe DT is just not the right tool for this use case. Personally, I'd look into using a CLI tool like [Bomber](https://github.com/devops-kung-fu/bomber) if I wanted to incorporate BOM scanning into my local workflow.

stevespringett:
> If anything, there should be improvements to get up-and-running faster

Completely agree @nscuro 

> but disabling security features should not be part of it

From a strategic point of view, Dependency-Track is an OWASP Flagship project. The flagship status is not a sign of maturity, rather it indicates direct support from the OWASP Board of Directors that a project is strategic to the OWASP mission. See https://owasp.org/www-committee-project/#div-promotions

With the OWASP mission in mind, I do not believe that disabling security features is something that the OWASP mission would support. I attempted to find guidance in [ASVS](https://owasp.org/www-project-application-security-verification-standard/), but was unable to find any statements or requirements that would provide support for unauthenticated users or the disabling of security features. We'd also open adopters up to future [OWASP Top 10 : Security Misconfiguration](https://owasp.org/Top10/A05_2021-Security_Misconfiguration/) issues with DT as well.

While this is not a solid NO from my point of view, I think there's no easy path forward. A lot of functionality in DT would need to also be removed, such as the ability to audit vulnerabilities. 

In general, the PRs we typically favor are the ones that help the DT community identify and reduce risk. I see this PR as potentially adding risk along with a lot more work for everyone involved. From a value perspective, I'm not seeing much. If there were some amazing use cases we were missing out on that would help organizations identify and reduce risk, then I'd be in favor of it. But I'm not seeing the argument being made in this case.

samrocketman:
That's fair.  I thought it realistic that the feature could be rejected which is why I resisted coding further and writing tests.

Sounds like there's consensus here that it's not desirable so I'll go ahead and close the PRs.

Should you change your mind feel free to ping me in an issue or something.  Thanks

samrocketman:
That's fair.  I thought it realistic that the feature could be rejected which is why I resisted coding further and writing tests.

Sounds like there's consensus here that it's not desirable so I'll go ahead and close the PRs.

Should you change your mind feel free to ping me in an issue or something.  Thanks

samrocketman:
I'll check out Niklas' suggestion to use Bomber CLI.  If it can generate browsable reports I may use it but I do like the DT frontend

samrocketman:
I don't think Bomber will do what I want in the way that I need it.  Thanks for the suggestion, though.  In the mean time I'll just maintain my own fork of dependency-track as that's what I want in my workflow.

Issue #2197, opened 2022-11-27
Allow to bypass starting requirements


nathan-mittelette:
### Current Behavior

Currently, when trying to start the dependency-track API, if your system does not meet the required requirements, the application will not start.

This behavior can force you to allocate a lot of resources.

### Proposed Behavior

I suggest keeping the current behavior but allowing the user to easily override the requirements using an environment variable.
We can create an **system.requirements.bypass** environment variable. By default, this variable will be set to false. But it can be replaced by true if we want to bypass this mechanism.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

stevespringett:
> This behavior can force you to allocate a lot of resources.

I know "a lot" is subjective, but I don't think 4.5 GB and 2 cores is "a lot". The system will not function correctly below 3GB.

https://docs.dependencytrack.org/getting-started/deploy-docker/

nathan-mittelette:
I understand, but currently in my company we are running dependency-track API with 2Gi and 1 CPU because we don't have huge projects that needs a lot of memory while doing scans. Using dependency-track with this setup isn't perfect, but it's sufficient for our use.
Maybe if we allow users to bypass the requirements with obviously an warning at the start, that might help in some cases.

nscuro:
IMO it's OK to provide the ability to opt out of this restriction.

Ideally we would have recommendations as to what workload requires (approximately) how many resources. But we don't currently.

At the moment we have users modifying DT internally to disable the check, and DT runs fine for them. There were multiple occasions in Slack where this was mentioned.

nathan-mittelette:
> At the moment we have users modifying DT internally to disable the check, and DT runs fine for them. There were multiple occasions in Slack where this was mentioned.

Yes, it's exactly what we've done in my company, and I taught that we may not be alone and it can be useful to provide an implementation to properly disable the check.

PR #2198, opened 2022-11-27
Issue 2197: Allow to bypass starting requirements

nathan-mittelette:
### Description

In this PR, I added a new environment variable called **system.requirement.check.enabled** that allow users to disable Dependency-Track requirements startup check.

### Addressed Issue

Fixes DependencyTrack/dependency-track#2197

### Checklist

- [ ] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/github-templates/CONTRIBUTING.md#pull-requests)

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -318,6 +318,11 @@ alpine.oidc.team.synchronization=false
 alpine.oidc.teams.claim=groups
 
 # Optional
+# Define whether system requirement check is enable.
+# The default value is true.
+system.requirement.check.enabled=true
+
+# Optional
 # Defines the size of the thread pool used to perform requests to the Snyk API in parallel.
 # The thread pool will only be used when Snyk integration is enabled.
 # A high number may result in a quicker excession of API rate limits,
--- a/docs/_docs/getting-started/deploy-docker.md
+++ b/docs/_docs/getting-started/deploy-docker.md
@@ -18,6 +18,8 @@ other than a modern version of Docker.
 | 4.5GB RAM   | 16GB RAM    |
 | 2 CPU cores | 4 CPU cores |
 
+> These requirements can be disabled by setting the 'system.requirement.check.enabled' property or the 'SYSTEM_REQUIREMENT_CHECK_ENABLED' environment variable to 'false'. 
+
 ### Container Requirements (Front End)
 
 | Minimum     | Recommended |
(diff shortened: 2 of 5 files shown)


valentijnscholten:
I like this as it allows users to run DT as they see fit. In my experience 2G also works fine for smaller instances which allows DT to run on cheaper VM instances.

valentijnscholten:
The word "bypass" feels weird in the context of a security product. How about DISABLE_SYSTEM_REQUIREMENT_CHECK?

nscuro:
@stevespringett Any hard objections against this? Added my 2ct in https://github.com/DependencyTrack/dependency-track/issues/2197#issuecomment-1330927879

If we decide to adapt this, there should be a reference to the config option on the system requirements page(s) in the documentation:
* https://docs.dependencytrack.org/getting-started/deploy-docker/#container-requirements-api-server
* https://docs.dependencytrack.org/getting-started/deploy-exewar/#requirements

nathan-mittelette:
> The word "bypass" feels weird in the context of a security product. How about DISABLE_SYSTEM_REQUIREMENT_CHECK?

I agree with you, I updated the property name.

nathan-mittelette:
@nscuro 

> If we decide to adapt this, there should be a reference to the config option on the system requirements page(s) in the documentation:
> 
> * https://docs.dependencytrack.org/getting-started/deploy-docker/#container-requirements-api-server
> * https://docs.dependencytrack.org/getting-started/deploy-exewar/#requirements

I updated the documentation. Do these changes work for you?

src/main/resources/application.properties
```diff
@@ -317,6 +317,11 @@ alpine.oidc.team.synchronization=false
 # will most likely need to be configured.
 alpine.oidc.teams.claim=groups
 
+# Optional
+# Define whether system requirement check is disable.
+# The default value is false.
+disable.system.requirement.check=false
```
valentijnscholten, line 323: don't want to nitpick, but I suggest {{system.requirement.check.enabled=true}}, similar to the CORS setting etc.

stevespringett:
@nscuro 
> Any hard objections against this? Added my 2ct in

No objections. But I would add a WARN statement on startup stating that its disabled and provide a URL to the docs where the minimum requirements can be found.

My concern is that Person A disables the check because they have a small DT instance without a lot of activity. After a year or more, Person B takes over managing DT which has now greatly increased the amount of BOM consumption and they're running into performance issues as a result. Person B doesn't have the backstory, but if we have a WARN in the log, they may see that.

nscuro:
> After a year or more, Person B takes over managing DT which has now greatly increased the amount of BOM consumption and they're running into performance issues as a result.

To be fair this also can happen when the instance starts out with 4GB, but usage is ramped up to a point where 8 or 16GB would be more appropriate. But I agree that at the lower end of specs the exhaustion of resources is more likely to occur, and a warning is a good thing to have.

nscuro (approved):
Thanks @nathan-mittelette! 🚀

mulder999:
To add to the thread, I noticed the max memory size is checked statically at startup. With current VM memory ballooning technics this cannot work as memory would increase with actual usage. Also the max memory gives no indication whether this memory could be requested or not by the application (thinking container world here). I believe it would be more suitable to allocate the memory needed along the way and not bother with this check at all, those that have memory issues will notice it the hardway anyway... At least now this check is optional.

Issue #1518, opened 2022-04-08
Ability to set up license policy of Null or Blank


peterkline:
The enhancement may already be reported! Please search for the enhancement before creating one.

### Current Behavior:
There is no easy way to set up a license policy where a license for a component is blank.  To get around this I have create a license policy including every license.  This is not optimal if new licenses are added, then this policy becomes obsolete

### Proposed Behavior:
Include an option when creating a policy where you can set a `License is Null` or possibly `License is not Defined`

PR #2214, opened 2022-11-30
Enable to set unresolved license for license policy

rbt-mm:
### Description

If you want to set up a license policy where the license is blank or null, it is currently only possible by creating a license policy which includes every single license. This is tedious to set up and needs to be maintained if new licenses are added.

To simplify setting up this license policy, it is now possible to choose `unresolved` as the value for a policy condition. This condition then checks, if a component does not have a license.

### Addressed Issue

#1518

### Additional Details

image

[Frontend PR](https://github.com/DependencyTrack/frontend/pull/339)

### Checklist


rbt-mm:
Yeah, I agree that it would be more clear if it was named `unresolved` instead of `undefinedLicense`.

I will change the names in the backend and frontend.

nscuro:
@rbt-mm Please rebase with `master` for the build to work again. The failure cause was fixed in https://github.com/DependencyTrack/dependency-track/pull/2241.

nscuro (approved):

PR #2240, opened 2022-12-05
Fix: Exact match between CPE with NA (-) update part

syalioune:
### Description

https://github.com/DependencyTrack/dependency-track/discussions/2188

### Addressed Issue

See https://github.com/DependencyTrack/dependency-track/pull/1929#issuecomment-1331223525

This PR fix a bug making `cpe:2.3:a:xiph:speex:1.2:-:*:*:*:*:*:*` not equal to `cpe:2.3:a:xiph:speex:1.2:-:*:*:*:*:*:*` when performing internal analysis. Special cases (`NA` and `ANY` on each part) are handled before making the simplest of comparison.

Right now, the relation 6 in the picture below from [NIST CPE matching spec](https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf) is correctly handled for CPE's update part.

cpe_comparison

### Additional Details

N/A

### Checklist
~~- [] This PR implements an enhancement, and I have provided tests to verify that it works as intended~~

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java
+++ b/src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java
@@ -198,6 +198,10 @@ public abstract class AbstractVulnerableSoftwareAnalysisTask extends BaseCompone
      * <code>false</code>
      */
     private static boolean compareUpdate(VulnerableSoftware vs, String targetUpdate) {
+
+        if (targetUpdate != null && targetUpdate.equals(vs.getUpdate())) {
+            return true;
+        }
         if (LogicalValue.NA.getAbbreviation().equals(vs.getUpdate())) {
             return false;
         }
--- a/src/test/java/org/dependencytrack/tasks/scanners/InternalAnalysisTaskTest.java
+++ b/src/test/java/org/dependencytrack/tasks/scanners/InternalAnalysisTaskTest.java
@@ -6,8 +6,12 @@ import org.dependencytrack.model.Component;
 import org.dependencytrack.model.Project;
 import org.dependencytrack.model.Vulnerability;
 import org.dependencytrack.model.VulnerableSoftware;
+import org.dependencytrack.parser.nvd.ModelConverter;
 import org.junit.Test;
+import us.springett.parsers.cpe.exceptions.CpeEncodingException;
+import us.springett.parsers.cpe.exceptions.CpeParsingException;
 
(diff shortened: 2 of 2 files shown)


nscuro:
@syalioune Please rebase with `master` for the build to work again. The failure cause was fixed in #2241.

nscuro:
LGTM, but I'm a bit nervous to merge this one as I can't estimate the impact this will have on existing behavior. Mainly due to the lack of tests we have for this functionality (I raised #2243 for that).

valentijnscholten:
I looked at the linked issues/PRs an I have a feeling about what this PR does. But if I really wanted to know I would have to study the code more. Can I suggest to add a little explanation here so that it "instantly" clear for readers what this PR does?

syalioune:
> I looked at the linked issues/PRs an I have a feeling about what this PR does. But if I really wanted to know I would have to study the code more. Can I suggest to add a little explanation here so that it "instantly" clear for readers what this PR does?

Done, ty for the suggestion.

> LGTM, but I'm a bit nervous to merge this one as I can't estimate the impact this will have on existing behavior. Mainly due to the lack of tests we have for this functionality.

Would it make it less risky if I perform the update higher up instead of deep down in the comparison methods like in the snippet below.

```java
protected void analyzeVersionRange(final QueryManager qm, final List<VulnerableSoftware> vsList,
                                       final String targetVersion, final String targetUpdate, final Component component,
                                       final VulnerabilityAnalysisLevel vulnerabilityAnalysisLevel) {
        for (final VulnerableSoftware vs: vsList) {
            boolean equalCpe = (vs.getCpe23() != null && vs.getCpe23().equals(component.getCpe())) ||
                    (vs.getCpe22() != null && vs.getCpe22().equals(component.getCpe()));
            if (equalCpe || (compareVersions(vs, targetVersion) && compareUpdate(vs, targetUpdate))) {
                if (vs.getVulnerabilities() != null) {
                    for (final Vulnerability vulnerability : vs.getVulnerabilities()) {
                        NotificationUtil.analyzeNotificationCriteria(qm, vulnerability, component, vulnerabilityAnalysisLevel);
                        qm.addVulnerability(vulnerability, component, this.getAnalyzerIdentity());
                    }
                }
            }
        }
    }
```

Worst case being to link more vuln that people are used to but it seems legit.

nscuro (approved):
@syalioune Sorry, I think I was not fully grasping the proposed solution when writing my previous comment.

You're right, checking for equality first makes sense as it's the simplest and most obvious "match" case. Thanks for the PR!

Issue #2231, opened 2022-12-04
Vulnerability details modal shows "undefined" component label


syalioune:
### Current Behavior

When viewing the details of a vulnerability having affected components with a PURL identifier without a namespace (e.g. : `pkg:pypi/aiohttp@3.5.3`), Dependency Track shows an undefined label.

![vulndetails-issue-4-6-3](https://user-images.githubusercontent.com/6144741/205505444-fc112920-b2ff-4ac3-a085-e7ba92be59a4.png)

This is due to the REST API `/v1/vulnerability/source/{source}/vuln/{vulnId}` return a payload without an `identity` node for affectedComponents.

It should be fixed in several places :

- [Mapping from VulnerableSoftware to AffectedComponent](https://github.com/DependencyTrack/dependency-track/blob/7b573b52aac533cdb30ca4fe6ab54d82478fc1c6/src/main/java/org/dependencytrack/resources/v1/vo/AffectedComponent.java#L78) : Assume that a PURL can have a null namespace
- [Set the PURL string when mapping external vuln to DT VulnerableSoftware model](https://github.com/DependencyTrack/dependency-track/blob/7b573b52aac533cdb30ca4fe6ab54d82478fc1c6/src/main/java/org/dependencytrack/tasks/GitHubAdvisoryMirrorTask.java#L299-L306) : It should be done for all mirroring tasks (GHSA, OSS, Snyk,...)

### Steps to Reproduce

1. View the affected components for a GHSA vulnerability like `GHSA-v6wp-4m6f-gcjg` or `GHSA-6757-jp84-gxfx`

### Expected Behavior

Show the PURL instead on `undefined`.

### Dependency-Track Version

4.7.0-SNAPSHOT

### Dependency-Track Distribution

Container Image

### Database Server

H2

### Browser

Google Chrome

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

PR #2250, opened 2022-12-07
Fix: AffectedComponent mapping from VulnerableSoftware in case of purl with no namespace

syalioune:
### Addressed Issue

#2231

### Checklist


nscuro (approved):
Thanks @syalioune! 🤘

PR #2256, opened 2022-12-08
grafana: add total heap size and 100% cpu lines

valentijnscholten:
Three small changes to the grafana dashboard. 
- CPU usage graph: Set max to 100% to make it more clear when cpu usage is high / close to 100%
- Heap usage: Add total heap size line to make it more clear when heap usage is close to maximum heap size
- Add Summary row around first row of panels

new:
image

old:
image

new:
image

old:
image

summary row:
image


docs/files/grafana-dashboard.json
```diff
@@ -736,6 +737,18 @@
...
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "code",
+          "expr": "sum(jvm_memory_max_bytes{instance=\"$instance\", area=\"heap\"})",
+          "hide": false,
+          "legendFormat": "__auto",
```
nscuro, line 749: Can we have a label for this one? I believe it's the `legendFormat` field, it's set to `Heap` and `Non-Heap` for the other two fields. Maybe call this one `Total` or so, to make it a bit more readable.

valentijnscholten: good suggestion, added now

nscuro (approved):
Thanks @valentijnscholten! 🚀

Issue #356, opened 2019-06-05
Add Packagist (Composer) repository support


nscuro:
Add support for the [Packagist](https://packagist.org/) repository, which is used by PHP's [Composer](https://getcomposer.org/).

The public API is [fairly straightforward](https://packagist.org/apidoc#get-package-data), except that there's no easy way to determine the latest version of a package.

b2un0:
Any progress on this?

there is also a `composer.lock` support in [jeremylong/DependencyCheck](https://github.com/jeremylong/DependencyCheckl), but [broken](https://github.com/jeremylong/DependencyCheck/issues/1743).

nscuro:
Dependency-Track already has working vuln detection for Composer components. You'll need this Composer plugin to generate a CycloneDX BOM: https://github.com/CycloneDX/cyclonedx-php-composer. 

This issue addresses the fact that DT can't currently detect the latest version of a Composer component.

PR #2272, opened 2022-12-12
Display version status in `Audit Vulnerabilities` and `Exploit Predictions` tab

rbt-mm:
### Description

Information about the version status of a component is currently only displayed in the `Components` tab of a project.

Also adding this information to the `Audit Vulnerabilities` and the `Exploit Predictions` tab of a project improves the usability of those two tabs, because it is no longer needed to switch tabs to check for version status information.

### Addressed Issue

Frontend: [#356](https://github.com/DependencyTrack/frontend/issues/356)

### Additional Details

The retrieved `Finding` objects from `/finding/project/<projectUUID>` must be appended by the latest version of a component to correctly display the information in the UI.

[Frontend PR](https://github.com/DependencyTrack/frontend/pull/357)

image

image

### Checklist


src/main/java/org/dependencytrack/persistence/FindingsQueryManager.java
```diff
@@ -278,6 +281,14 @@ public List<Finding> getFindings(Project project, boolean includeSuppressed) {
...
                 finding.getVulnerability().put("description", vulnerability.getDescription());
                 finding.getVulnerability().put("recommendation", vulnerability.getRecommendation());
+                final PackageURL purl = component.getPurl();
+                if (purl != null) {
+                    final RepositoryType type = RepositoryType.resolve(purl);
+                    if (RepositoryType.UNSUPPORTED != type) {
+                        final RepositoryMetaComponent repoMetaComponent = getRepositoryMetaComponent(type, purl.getNamespace(), purl.getName());
+                        finding.getComponent().put("latestVersion", repoMetaComponent.getLatestVersion());
+                    }
+                }
```
nscuro, line 291: It should be possible to to this via `LEFT JOIN` in the existing findings query:

https://github.com/DependencyTrack/dependency-track/blob/8f4c2b5c0d2e153363fed0ab2e865653558b7d6d/src/main/java/org/dependencytrack/model/Finding.java#L50-L81

That would save us a few database round-trips.

nscuro (changes requested):

src/main/java/org/dependencytrack/persistence/FindingsQueryManager.java
rbt-mm: The problem that I see here is that neither `COMPONENT` nor `REPOSITORY_META_COMPONENT` reference each other via a foreign key.

nscuro:
@rbt-mm Could you add a test for this in `FindingResourceTest`? Also, please update the example in the FPF documentation to include the new field: https://github.com/DependencyTrack/dependency-track/blob/master/docs/_docs/integrations/file-formats.md

nscuro (approved):
Thanks @rbt-mm! 🚀

Issue #2131, opened 2022-11-07
Error while uploading same SBOM Second time


g-sahil22:
The defect may already be reported! Please search for the defect before creating one.

### Current Behavior:
When I upload the SBOM in Dependency Track for the First time in a new project it uploads successfully without any error logs and also reports the correct number of Components, but I upload the same SBOM in the same project again it shows error logs and reports the incorrect count of components 

**Logs When I upload SBOM First time** 
![image](https://user-images.githubusercontent.com/100830439/200242584-1781c914-c999-41f0-ad48-4fdae4c1d0b0.png)
Number of Components: 1686

**Logs When I upload the SBOM Second time**
![image](https://user-images.githubusercontent.com/100830439/200242851-6245f50b-5455-48b8-a8d9-91b61591b140.png)
Number of Components: 1611

### Steps to Reproduce:

Step 1: Create a New project
Step 2: Upload SBOM in the New project
Step 3: Monitor the logs and Number of components shown in the Dependency track
Step 4: Again upload the same SBOM in the Same project 
Step 5: Monitor the logs and Number of components shows in the Dependency track

### Environment:

Report the Same number of components without any error logs when we upload SBOM second time in same project
  
  - Dependency-Track Version: 4.6.2
  - Distribution: Docker 
  - BOM Format & Version: JSON, Cyclonedx 1.4
  - Database Server: H2 and PostgreSQL
  - Browser: Chrome

jakobpinggera:
We are experiencing the same issue. Interestingly, it does not seem to be deterministic whether the issue arises. 

I uploaded the same BOM file twice (to a project that already had processed BOM files in the past). On the first try, the exception occurred. A minute later, the same file got processes successfully. 

`
2022-11-07 13:14:16,580 [] INFO [org.dependencytrack.tasks.BomUploadProcessingTask] Processing CycloneDX BOM uploaded to project: 2eaa3deb-1c38-4d24-a306-4c8d3081e393
2022-11-07 13:14:22,020 [] INFO [org.dependencytrack.tasks.BomUploadProcessingTask] Identified 0 new components
2022-11-07 13:14:22,021 [] INFO [org.dependencytrack.tasks.BomUploadProcessingTask] Processing CycloneDX dependency graph for project: 2eaa3deb-1c38-4d24-a306-4c8d3081e393
2022-11-07 13:14:22,583 [] ERROR [org.dependencytrack.tasks.BomUploadProcessingTask] Error while processing bom
[...]

PR #2278, opened 2022-12-13
Workaround for deleting duplicate components when uploading a new BOM

rbt-mm:
### Description

When uploading a new BOM to a project with duplicated components, the `BomUploadProcessingTask` might throw exceptions while trying to recursively delete the old components, resulting in a failed upload, where the components from the new BOM are added, but not all old components are deleted.

By catching the exceptions which are thrown while recursively deleting old components, the `BomUploadProcessingTask`  can finish successfully and all of the old components are correctly deleted.

### Addressed Issue

Workaround for #2131

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
@@ -18,6 +18,7 @@
  */
 package org.dependencytrack.persistence;
 
+import alpine.common.logging.Logger;
 import alpine.event.framework.Event;
 import alpine.model.ApiKey;
 import alpine.model.Team;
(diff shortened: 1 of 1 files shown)


src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -394,16 +394,20 @@ public void recursivelyDelete(Component component, boolean commitIndex) {
...
+            Event.dispatch(new IndexEvent(IndexEvent.Action.DELETE, pm.detachCopy(result)));
+            deleteAnalysisTrail(component);
+            deleteViolationAnalysisTrail(component);
+            deleteMetrics(component);
+            deleteFindingAttributions(component);
+            deletePolicyViolations(component);
+            delete(component);
+            commitSearchIndex(commitIndex, Component.class);
+        } catch (javax.jdo.JDOObjectNotFoundException | org.datanucleus.exceptions.NucleusObjectNotFoundException ignored) {
+
```
nscuro, line 408: These exceptions should still be logged, potentially with level `WARNING`. It should be obvious that there's still something going wrong unintentionally.

nscuro (changes requested):

Issue #1532, opened 2022-04-10
Deleting a Team that is part of 1 or more ACLs leads to 500 error


valentijnscholten:
### Steps to Reproduce:
Create a Team
Enable Portfolio Access Control
Create a Project
Assign the Project to the Team
Try to delete the Team
Observe 500 error in the UI
Observe exception in the logs:

```
Caused by: org.h2.jdbc.JdbcBatchUpdateException: Referential integrity constraint violation: "PROJECT_ACCESS_TEAMS_FK2: PUBLIC.PROJECT_ACCESS_TEAMS FOREIGN KEY(TEAM_ID) REFERENCES PUBLIC.TEAM(ID) (39)"; SQL statement:
DELETE FROM TEAM WHERE ID=? [23503-200]
        at org.h2.jdbc.JdbcPreparedStatement.executeBatch(JdbcPreparedStatement.java:1235)
        at com.zaxxer.hikari.pool.ProxyStatement.executeBatch(ProxyStatement.java:127)
        at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.executeBatch(HikariProxyPreparedStatement.java)
        at org.datanucleus.store.rdbms.ParamLoggingPreparedStatement.executeBatch(ParamLoggingPreparedStatement.java:366)
        at org.datanucleus.store.rdbms.SQLController.processConnectionStatement(SQLController.java:675)
        at org.datanucleus.store.rdbms.SQLController.processStatementsForConnection(SQLController.java:643)
        at org.datanucleus.store.rdbms.SQLController$1.transactionFlushed(SQLController.java:729)
        at org.datanucleus.store.connection.AbstractManagedConnection.transactionFlushed(AbstractManagedConnection.java:95)
        at org.datanucleus.store.connection.ConnectionManagerImpl$2.transactionFlushed(ConnectionManagerImpl.java:528)
        at org.datanucleus.TransactionImpl.flush(TransactionImpl.java:222)
        at org.datanucleus.TransactionImpl.commit(TransactionImpl.java:286)
        at org.datanucleus.api.jdo.JDOTransaction.commit(JDOTransaction.java:99)
```

### Expected Behavior:
Automatically delete ACL entries before deleting a Team / enable cascaded delete.

### Environment:
  
  - Dependency-Track Version: 4.5.0-SNAPSHOT https://github.com/DependencyTrack/dependency-track/commit/ad371d889606f8bbcffc584eaf9510c917bd9001
[...]

PR #2287, opened 2022-12-14
Fix : Deleting a team with existing ACL

syalioune:
### Description

Enable deleting a team linked to an ACL recursively. 

Unit tests for DELETE REST API with payload are also fixed. 

### Addressed Issue

See #1532

### Additional Details

- Frontend PR : https://github.com/DependencyTrack/frontend/pull/364

### Checklist

~~- [ ] This PR implements an enhancement, and I have provided tests to verify that it works as intended~~


src/test/java/org/dependencytrack/DTGrizzlyWebTestContainer.java
```diff
@@ -0,0 +1,200 @@
+package org.dependencytrack;
```
nscuro, line 1: This may be subjective, but I'd prefer a "simpler" approach using reflection instead. It may be dirty, but it's only for tests and saves us from having to maintain this grizzly code.

Found this on [Stack Overflow](https://stackoverflow.com/a/36311238) and it appears to work fine:

```java
public class DTGrizzlyWebTestContainerFactory implements TestContainerFactory {
    @Override
    public TestContainer create(URI baseUri, DeploymentContext deploymentContext) {
        final TestContainer testContainer = new GrizzlyWebTestContainerFactory().create(baseUri, deploymentContext);
        try {
            HttpServer server = (HttpServer) FieldUtils.readDeclaredField(testContainer, "server", true);
            server.getServerConfiguration().setAllowPayloadForUndefinedHttpMethods(true);
        } catch (IllegalAccessException e) {
            fail(e.getMessage());
        }
        return testContainer;
    }
}
```

Thoughts?

nscuro (approved):
Thanks @syalioune! 🙌

Especially nice that you went ahead and fixed all the other tests that were affected by the `DELETE` limitation! Waiting for your take on the reflection bit I commented on, but that is not a blocker.

src/test/java/org/dependencytrack/DTGrizzlyWebTestContainer.java
syalioune: Shame on me !
I actually tried that manually but quickly abandoned when I thought (without excuting really) that `testContainer.getClass().getDeclaredField("server")` would not work because I was working with the interface `TestContainer` instead of the implem class 😭

PR #2298, opened 2022-12-16
Set Jira URL for Jira notifications globally rather than in alerts

### LINKED ISSUE #2297: Set Jira URL for Jira notifications globally rather than in alerts
### Current Behavior

The Jira URL is set in the destination field, for each alert.

However the Jira credentials are set globally, which is incoherent (different Jira instances would need to have the same user and password).

### Proposed Behavior

The Jira URL should be set globally.

It is expected that users will only ever need to send Dependency-Track alert notifications to a single Jira instance, hence it makes sense to configure the Jira URL globally rather than having to fill it in each alert.

The destination field then becomes the Jira project key.

### Checklist

Mvld3r:
Signed-off-by: Mvld3r <Mvld3r@users.noreply.github.com>

### Description

Set Jira URL for Jira notifications globally rather than in alerts

### Addressed Issue

Fixes #2297.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
+++ b/src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
@@ -55,7 +55,7 @@ public abstract class AbstractWebhookPublisher implements Publisher {
         try {
             credentials = getBasicAuthCredentials();
         } catch (PublisherException e) {
-            logger.warn("An error occurred during the retrieval of credentials needed for notification publication. Skipping notification");
+            logger.warn("An error occurred during the retrieval of credentials needed for notification publication. Skipping notification", e);
             return;
         }
         if (credentials != null) {
--- a/docs/_docs/integrations/notifications.md
+++ b/docs/_docs/integrations/notifications.md
@@ -406,7 +406,7 @@ Once the alert is created it can be configured. Start with selecting from the li
 to notify on. Then specify the destination:
 - For the Email publisher: it's a comma separated list of email addresses
 - For Slack, Mattermost and Microsoft Teams: it's the incoming webhook URL generated from each respective
-- For Jira: it's the URL for the Jira instance. The Jira alert also asks for the Jira ticket type ('Task', 'Story', etc., refer to the Jira documentation for more details) to be created, along with the project key where the issue will be created
+- For Jira: it's the project key where the issue will be created. The Jira alert also asks for the Jira ticket type ('Task', 'Story', etc., refer to the Jira documentation for more details) to be created
 - For the Outbound Webhook publisher: it's a URL to which to publish the notification
 
 ![configure notification](/images/screenshots/notifications-configure.png)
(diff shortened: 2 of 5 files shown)


src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
```diff
@@ -33,7 +33,7 @@
 public abstract class AbstractWebhookPublisher implements Publisher {
 
     public void publish(final String publisherName, final PebbleTemplate template, final Notification notification, final JsonObject config) {
-        final Logger logger = Logger.getLogger(this.getClass());
+        final Logger logger = Logger.getLogger(AbstractWebhookPublisher.class);
```
nscuro, line 36: What's the reason for this change? The logger should refer to the class extending `AbstractWebHookPublisher` (e.g. `JiraPublisher`), not the abstract class itself.

src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
```diff
@@ -55,8 +55,8 @@ public void publish(final String publisherName, final PebbleTemplate template, f
         try {
             credentials = getBasicAuthCredentials();
         } catch (PublisherException e) {
-            logger.warn("An error occurred during the retrieval of credentials needed for notification publication. Skipping notification");
-            return;
+            logger.warn("An error occurred during the retrieval of credentials needed for notification publication. Skipping notification", e);
+            throw e;
```
nscuro, line 59: What's the reason for this behavior change (re-throwing the exception)?

src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
Mvld3r: You're right, I think it was a local change that I forgot to remove from my commit. I'll revert to the previous behaviour.

src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
Mvld3r: I feel that the exception stacktrace should be part of the log. Multiple exceptions can arise from `getBasicAuthCredentials()`, including exceptions resulting from the retrieval and decryption of the stored jira password, and it would help the user to pinpoint the issue.

As for rethrowing the exception or silently returning, I leave it up to you to decide.

This isn't strictly in relation to the PR title, but is on a related subject (jira publisher) and doesn't grant a separated PR imo.

nscuro: I'm with you on the logging the stacktrace part, but re-throwing the exception may have some undesirable side effects. Best case the exception is just logged twice because it will be handled again by the "unhandled exception handler", worst case other operations may be negatively affected.

nscuro (approved):
Thanks @Mvld3r! 🙌

PR #2319, opened 2022-12-20
Fix `NullPointerException` when updating a project

### LINKED ISSUE #2317: Updating a project fails when its active status is `null`
### Current Behavior

If the active status of a project is `null` and someone tries to update that project, a `NullPointerException` will be thrown because that case is currently not handled.

### Steps to Reproduce

Adding this test to `ProjectResourceTest` and executing it will throw a `NullPointerException`:

```
@Test
  public void updateProjectTestIsActiveEqualsNull() {
      Project project = qm.createProject("ABC", null, "1.0", null, null, null, true, false);
      project.setDescription("Test project");
      project.setActive(null);
      Assert.assertNull(project.isActive());
      Response response = target(V1_PROJECT)
              .request()
              .header(X_API_KEY, apiKey)
              .post(Entity.entity(project, MediaType.APPLICATION_JSON));
      Assert.assertEquals(200, response.getStatus(), 0);
      JsonObject json = parseJsonObject(response);
      Assert.assertNotNull(json);
      Assert.assertEquals("ABC", json.getString("name"));
      Assert.assertEquals("1.0", json.getString("version"));
      Assert.assertEquals("Test project", json.getString("description"));
      Assert.assertTrue(json.getBoolean("active"));
  }
```

### Expected Behavior

It should be checked if the active status of a project is `null` and no `NullPointerException` should be thrown.

### Dependency-Track Version

4.7.0

Container Image

### Database Server

H2

### Database Server Version

_No response_

### Browser

N/A

### Checklist

rbt-mm:
### Description

If the active status of a project is `null` and someone tries to update that project, a `NullPointerException` will be thrown because that case is currently not handled.

This PR checks if the active status of a project is `null` when trying to update the project, and handles the case, resulting in no `NullPointerException` being thrown.

### Addressed Issue

Closes #2317

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/notification/NotificationRouter.java
+++ b/src/main/java/org/dependencytrack/notification/NotificationRouter.java
@@ -226,7 +226,7 @@ public class NotificationRouter implements Subscriber {
             return false;
         }
         for (Project child : parent.getChildren()) {
-            if ((child.getUuid().equals(uuid) && child.isActive()) || isChild) {
+            if ((child.getUuid().equals(uuid) && Boolean.TRUE.equals(child.isActive())) || isChild) {
                 return true;
             }
             isChild = checkIfChildrenAreAffected(child, uuid);
--- a/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
@@ -487,7 +487,7 @@ final class ProjectQueryManager extends QueryManager implements IQueryManager {
         project.setVersion(version);
         project.setPurl(purl);
 
-        if (!active && project.isActive() && hasActiveChild(project)){
+        if (!active && Boolean.TRUE.equals(project.isActive()) && hasActiveChild(project)){
             throw new IllegalArgumentException("Project cannot be set to inactive, if active children are present.");
         } else {
             project.setActive(active);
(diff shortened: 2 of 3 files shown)


valentijnscholten:
if you run a project wide search for `isActive()` there are 2 or 3 other places which could do with the same safeguard.

rbt-mm:
Thanks for noticing, I think that every `isActive()` should be handled correctly now.

nscuro (approved):
Thanks @rbt-mm and @valentijnscholten! 🙌

valentijnscholten:
@nscuro This started occurring out of nowhere on my instances. There might be some codepaths in 4.7.0 that somehow set a null value in the database. So a onetime SQL query to fix those values might not be a structural solution. Any chance we can get a 4.7.1?

nscuro:
@valentijnscholten We're looking into bundling multiple of such annoyances into a bugfix release right now. I've seen a few similar requests now on other issues so I think it makes sense to backport more than just this.

SaberStrat:
Any update on a patch release to help fix this? 
I'm doing a POST on the project API in several places in our CI, and this is failing now :/. Is there any workaround that I could attempt until then? A manual SQL query against the table with the missing active values?

nscuro:
We backported this to 4.7.1, which was just released.

SaberStrat:
Thank you very much ❤️

PR #2325, opened 2022-12-23
logging: reduce periodic details to debug level

valentijnscholten:
After enabling cloudwatch logging for my DT logs, I noticed the volume was a little more than expected.
Turns out 90% of the logging is coming from periodic tasks like updating metrics and syncing advisories. 
I suggest to reduce some of these loglines to DEBUG level instead of INFO level. 
Especially for the metrics updates I don't think it's necessary to log on INFO level 2 lines for each project every hour.

image

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/GitHubAdvisoryMirrorTask.java
+++ b/src/main/java/org/dependencytrack/tasks/GitHubAdvisoryMirrorTask.java
@@ -169,7 +169,7 @@ public class GitHubAdvisoryMirrorTask implements LoggableSubscriber {
      * @param advisories the results to synchronize
      */
     void updateDatasource(final List<GitHubSecurityAdvisory> advisories) {
-        LOGGER.info("Updating datasource with GitHub advisories");
+        LOGGER.debug("Updating datasource with GitHub advisories");
         try (QueryManager qm = new QueryManager()) {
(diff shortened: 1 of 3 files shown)


nscuro:
I see why you'd want this change, but for many users these messages are the only way to tell that stuff is happening. For those users, if they want to continue to see these messages, setting the log level to "debug" will flood them with all sorts of technical information they don't care about.

Wouldn't setting [`LOGGING_LEVEL`](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/docker/Dockerfile#LL17C5-L17C18) to `WARN` be the better option for you?

valentijnscholten:
There's still a INFO message when the job starts and when it completes. I doubt anybody is looking at hundreds of log messages stating a projects metrics are being recalculated every hour.

nscuro:
Other areas where missing log messages in `INFO` level may be confusing:

* When a metrics refresh is requested via `/api/v1/metrics/project/<UUID>/refresh`
* After uploading a BOM

In both cases, there will be no indication in the logs anymore that something is happening.

valentijnscholten:
Are you stating that because you would like an updated version of this PR, or because you want to close the PR?
For now I have added a commit that solves these cases.

nscuro:
Yeah, I was implying that we need alternatives for those cases. Sorry for being unclear.

Your solution works for me. Thanks for the PR, as always!

PR #2326, opened 2022-12-23
Bug/2323: Filter portfolio metrics in portfolio access control mode to match team's accesses

### LINKED ISSUE #2323: /api/v1/metrics/portfolio are not filtered by project team ACL
### Current Behavior

To completing the gap analysis available in #1127 regarding portfolio access control: the portfolio metrics are not filtered by project team ACL

### Steps to Reproduce

1. Enable `portfolio access control` and limit the number of projects accessible
2. Call any `/api/v1/metrics/portfolio` endpoints

WRONG: The number of projects corresponds to the full list of project irrelevant of the ACL settings

### Expected Behavior

Expected: The number of projects should corresponds to the list of project granted by the ACL settings

### Dependency-Track Version

4.7.0

Container Image

### Database Server

PostgreSQL

### Database Server Version

14

### Browser

Google Chrome

### Checklist

mulder999:
### Description

Filter portfolio metrics in portfolio access control mode to match team's accesses

### Addressed Issue

Fixes #2323

### Additional Details

Sum in real time the matching project's metrics

### Checklist


valentijnscholten:
There are a lot of unrelated commits, did something go wrong when rebasing/updating from upstream?

valentijnscholten:
When I was looking at this recently I was wondering if the portfolio metrics query couldn't be just a realtime call to the database? Or 1 query that does the summing of which the result is stored as metrics? Isn't it just summing up the things from all the project metrics?

mulder999:
> There are a lot of unrelated commits, did something go wrong when rebasing/updating from upstream?

Unsure what happen exactly, this came when opening the PR on upstream. Files changed looks good though. Could try again if you wish.

mulder999:
> When I was looking at this recently I was wondering if the portfolio metrics query couldn't be just a realtime call to the database? Isn't it just summing up the things from all the project metrics? Results could still be stored for historical/graph purposes.

ACL permissions can be different per user and can further change anytime making any form of long term storage useless. Putting the sum logic in database is over complex because of a) current design of time ranges b) inherited risk score. Please notice pre-existing metrics tasks are also not performing any sum operations in DB.

Despite this, I took extra care to optimise the logic and keep the complexity as low as possible. Please review and challenge.

valentijnscholten:
At first glance it looked (and still looks) like the portfolio metrics are just summing up the values from the projects that already have been calculated. Even the risk score looks like it can be implemented in DB. But I have no knowledge of how it ended up in Java code, probably for a good reason.

As you state in your comment, this PR fires possibly lots of concurrent DB queries. If you have 200 projects, you'll get 200 queries. Might be a lot slower than 1 big query or batching them? I notice the `ProjectMetricsUpdateTask` also performs 1 query per project, but in a background process it's less noticable. The queries in this PR directly affect the response time / rendering in the UI believe.

mulder999:
> At first glance it looked (and still looks) like the portfolio metrics are just summing up the values from the projects that already have been calculated. Even the risk score looks like it can be implemented in DB. But I have no knowledge of how it ended up in Java code, probably for a good reason.

Fully correct this use pre-calculated project metrics. I have no historical context either, but I noticed DT used to support many different databases and this type of query if at all possible is likely close to the edge case for most universal persistence framework.

> 
> As you state in your comment, this PR fires possibly lots of concurrent DB queries. If you have 200 projects, you'll get 200 queries. Might be a lot slower than 1 big query or batching them? I notice the `ProjectMetricsUpdateTask` also performs 1 query per project, but in a background process it's less noticable. The queries in this PR directly affect the response time / rendering in the UI believe.

For the api called by the UI, we are doing a single request to the DB no matter what. UI is asynchronous and not affected much.

Only the current metrics is doing the parallel calls you noticed. I believe this can be fired only through api, not UI. If you have 200 projects, you would likely use ACL to limit to only a bunch of them. Those accessing all the projects likely have management rights which bypass the ACL filter and do the legacy query anyway. But I agree this query remains suboptimal, hence my comment in code.

Until someone is willing to improve further this area, we could focus on the immediate "data leak" fix part and simply return `unallowed` for this call in this mode at this time.

mulder999:
@valentijnscholten I will rewrite the current metrics to use the since yesterday (or rather 2 days to account the fact, yesterday results might still be lacking) variant. This would limit to a single DB query and allow to remove my comment on parallel calls.

valentijnscholten:
Looks smart as it avoids having to do a complexer query to get exactly the latest metrics for each project in one go.

mulder999:
Sweet. Thanks @valentijnscholten for all the comments. Let me know if you see anything else.

src/main/java/org/dependencytrack/persistence/MetricsQueryManager.java
```diff
@@ -62,33 +67,54 @@ public List<VulnerabilityMetrics> getVulnerabilityMetrics() {
         return execute(query).getList(VulnerabilityMetrics.class);
     }
 
+    /**
+     * Retrieve all the projects suitable for portfolio
+     */
+    private List<Project> getProjectsForPortfolio() {
```
nscuro, line 73: Did you do any testing of how this behaves when the principal has access to multiple thousands of projects? When I rewrote the metrics calculation in 4.6, I made sure to avoid `getAll` operations like the plague.

If you really *need* to fetch all projects at once, at the very least you should use a [fetch group](https://www.datanucleus.org/products/accessplatform_6_0/jdo/persistence.html#fetch_groups) to only retrieve the fields you need, which, looking at the code calling this method, would be only the ID or UUID.

There is a `METRICS_UPDATE` fetch group for `Project` already:

https://github.com/DependencyTrack/dependency-track/blob/7cfc466baf7548ddb758f9e7b008b2d4aaedd148/src/main/java/org/dependencytrack/model/Project.java#L92-L96

And here's how you would use it:

https://github.com/DependencyTrack/dependency-track/blob/7cfc466baf7548ddb758f9e7b008b2d4aaedd148/src/main/java/org/dependencytrack/tasks/metrics/PortfolioMetricsUpdateTask.java#L179

In any case, it would be helpful if you could do some profiling / sampling for this new functionality. I'm a fan of [VisualVM](https://visualvm.github.io/), but there are other alternatives that you could use.

mulder999: Thanks for suggestion @nscuro, will adapt to use a simplified fetch plan in this specific case.

Regarding the 1000+ projects using ACL, this seems super theoretical and highly unlikely. Still it could create somewhat useless pressure on the system, hence I suggest to simply return empty for anyone with above 100+ projects referenced through its ACL. This will solve the actual leak of information and it is still much better to have no data than data you should not see in the first place. Most users with ACL will not face this limitation. Will that work for you ?

stevespringett: @mulder999 we have installations with 10K+ projects and one adopter that needs DT to be able to support 50K+ projects. If querying all projects is a requirement, then this feature needs to be disabled by default, have a configurable way to enable it in the UI, and provide a warning if they do.

mulder999: Hi @stevespringett, specific projects are fetched only for users that have ACL enable. Nothing has changed for those not using ACL.

stevespringett: Right, but even for the installations that have ACLs enabled, this could be a big performance hit that they currently don't have today.

mulder999: @nscuro and @stevespringett I have implemented the changes requested in the backend. Please have a look. For sure teams with important/huge ACL would benefit of some additional work (namely some form of caching).

The web UI still requires a manual change to display the new config settings switch (would have dream of dynamic UI here but it did not happen) and the warning. Unsure when it will fit my bandwidth in the coming days.

valentijnscholten:
Thinking out loud here. Could/should it be a background job that gathers the metrics for each users?
I was about to say for each Team, but you can have users that are in no team. The metrics for all projects are already calculated, so just adding them up for each user should be quite fast, even for 1000 users? Not sure what kind of numbers DT aims to support without requiring enterprise funding :-)
The downside might be that on every project metrics update, all users with access to that project need their metrics updated as well.

In Defect Dojo we had these type of metrics all calculated on the fly (by the DB) and that performed quite well even for 1M vulnerabilties.

mulder999:
> Thinking out loud here. Could/should it be a background job that gathers the metrics for each users? I was about to say for each Team, but you can have users that are in no team. The metrics for all projects are already calculated, so just adding them up for each user should be quite fast, even for 1000 users? Not sure what kind of numbers DT aims to support without requiring enterprise funding :-) The downside might be that on every project metrics update, all users with access to that project need their metrics updated as well.
> 
> In Defect Dojo we had these type of metrics all calculated on the fly (by the DB) and that performed quite well even for 1M vulnerabilties.

Adding metrics computation should be ultra fast indeed. I tend to believe building and maintaining any sort of long term cache would not be useful for this usage. However a short term cache (redis alike) when the user is currently connected and navigating back and force could be useful to avoid repeating same query over and over.

nscuro:
> @valentijnscholten Could/should it be a background job that gathers the metrics for each users?

I fear that would not scale well. Metrics are already one of the main bottlenecks we have when it comes to exhausting the thread pool.

> In Defect Dojo we had these type of metrics all calculated on the fly (by the DB) and that performed quite well even for 1M vulnerabilties.

Would you be willing to share some more insights, or links to where this is done in DD? Maybe there are some improvements we can make to the data model so it performs better for these ad-hoc queries.

> @mulder999 Adding metrics computation should be ultra fast indeed. [...] However a short term cache (redis alike) when the user is currently connected and navigating back and force could be useful to avoid repeating same query over and over.

The "ultra fast" part we'd have to test. But agreed, caching would be a good way to optimize. Even such short-term storage could be realized using the database, we don't necessarily have to introduce a dependency on a 3rd party cache.

valentijnscholten:
> Would you be willing to share some more insights, or links to where this is done in DD? Maybe there are some improvements we can make to the data model so it performs better for these ad-hoc queries.

The thing in the datamodel that does look weird to me is the alias part, that's gonna give a bit of trouble I think for some usecases. I think it doesn't lend itself very well to let the database handle the counting (and other queries that need to deduplicate aliases, etc). From a vulnerability record you can't easily get all its ids/aliases as you have to do a lot of comparisons in code (or queries):

image

image

If the aliases table was more like a join table to map vuln_ids onto vulnerabilities and with `source` and `vuln_id`  I think all the joining and counting and deduplication would become a lot easier. If you truely want to deduplicate vulnerabilities based on their aliases, it might be best to model that data like that:

image

Don't have an ER modeling tool handy, but the idea is to make vulnerabilities unique. And have multiple aliases attached to the vulnerability. So if a vulnerability has multiple aliases, you have multiple records. The `vuln_id` is then CVExxxx, GHSAxxx, depending on the source. Thinking out loud there could be sources that report multiple ids/aliases for the same vulnerability, if needed those could be duplicated:

image

Just thinking out loud here, so don't pick on the details :-)

About DD: DD has a datamodel similar to the above where  a unique vulnerability is one database records with multiple aliase records (ids) attached to it in a separate table.
In DD you have Product Type -> Product -> Engagement -> Test -> Finding. So 5 tables joined together. Even with 1M findings and ~10000 tests and 1000 engagements and 100 products, this performs quite OK when doing a SELECT COUNT(ID) FROM .... JOIN ..... JOIN .... GROUP BY Product for example. IIRC the query takes about 500ms on a stock mysql running on my laptop and that mysql instance has not been tuned or anything.  

The queries in DT would look quite similar, just count a bunch of vulnerabilities and group them by project and maybe severity. Even with 10000 projects and 1M vulnerabilities it should be quite fast. Unless there's something very different from the DD case.
Even if the query takes 1s or 5s or 10s it would still be OK for the background job to populate the metrics and keep track of historical metrics. When a user logs in it should still be quite fast to just query that single table with metrics, even if the user has 10000 projects in its ACL. Just make sure the ACL is handled by WHERE/JOIN clauses and not 10000 projects loaded into memory and then serialized in a list of 10000 ids.

In DD permissions are more complex and involve some more joining for users that are not superusers/admins. Similar to the ACL part in DT.

stevespringett:
@valentijnscholten join tables with foreign keys were specifically not chosen as a design pattern due to the performance impact they will have.

valentijnscholten:
I am not convinced doing counting and alias handling in code and doing 1+N queries or N * M queries is going to be faster than 1 or 2 extra joins. There's also great benefit in simpler code and simpler queries.

melba-lopez:
@valentijnscholten its been some time since this PR has been updated.  Can this PR be closed?  If not, could you update your branch?

valentijnscholten:
This is not my PR.

melba-lopez:
I understand @valentijnscholten.   You were very active in this PR and was not sure if there were other conversations happening behind the scenes with @mulder999.  @mulder999 please let me know if you will be updating this PR or if I should close it.

mulder999:
Hi @melba-lopez, thanks for asking. The PR was complete but did not reach community consensus. Would be happy to solve conflicts if it has any chance to be considered for merge in the trunk.

mulder999:
@melba-lopez PR has been updated. Anything else needed to remove the "pending more information" tag ?

melba-lopez:
@nscuro @stevespringett i know there were several discussions previously about how this may not work for DT.  Can you chime in and let us know if there is anything else you would like to modify?

src/main/java/org/dependencytrack/persistence/MetricsQueryManager.java
mulder999: Web UI changes are implemented in frontend [PR 741](https://github.com/DependencyTrack/frontend/pull/741)

mulder999:
@melba-lopez @nscuro @stevespringett The feature is complete and can now also conveniently be enabled from the web interface (default setting: disabled).

valentijnscholten:
@mulder999 @melba-lopez @nscuro I think this feature is something that is needed to make Portfolio Access more usable in real life. Looking back at this PR I think, with the current datamodel/metrics approach, this PR has an OK approach. 
@mulder999 Would you be willing to rebase/resolve conflicts once again?
@nscuro Do you agree it could be valuable to have this PR merged?

nscuro:
Fixed for v5 as per https://github.com/DependencyTrack/hyades/issues/1680.

PR #2333, opened 2022-12-28
Validating purl-policies should now only return 1 violation

### LINKED ISSUE #1925: PackageURLPolicyEvaluator might return two violations
I think there is an issue with PackageURLPolicyEvaluator when the operator is MATCHES. The evaluator might return two violations even though the evaluator should either match or don't. There should be an else-if or break out of the loop after a match is found to avoid duplicate violations.

https://github.com/DependencyTrack/dependency-track/blob/7685a29143dbb23f6d63af20ca4845790d81023b/src/main/java/org/dependencytrack/policy/PackageURLPolicyEvaluator.java#L58-L68

This is related to issue: #1924

### LINKED ISSUE #2212: Policy with ALL operator not working
### Current Behavior

When a policy with multiple conditions is created and the operator is ALL, the evaluation assigns a policy vulnerability per each condition vulnerated. Policy evaluator doesnt seem take into account the "ALL" operator and it is working like if policy was ANY operator always.

### Steps to Reproduce

1. Create new Policy, operator all
2. create condition license group IS NOT All Licenses, being All Licenses a group containing all licenses.
3. Create condition purl MATCHES pkg:maven (because we want only check  mavenized components with empty license field)
4. Update any project containing components with purl and with license.
5. A policy violation is assigned to this component showing it violates the purl condition, because the component has a purl value matching the purl condition.

### Expected Behavior

As the ALL operator says, policy evaluation must only trigger a policy violation when all conditions are violated. Because the component has a license matching the "all licenses" group, first condition is not violated, so this policy violation should not be triggered.

Verion  Version is 4.6.3 bundled.

### Dependency-Track Version

4.6.2

Executable WAR

### Database Server

H2

### Database Server Version

_No response_

### Browser

Google Chrome

### Checklist

malice00:
### Description

Validating purl-policies should now only return 1 violation.

### Addressed Issue

Fixes #1925
Fixes #2212

### Additional Details

Discussion of the solution can be found in #2330. Small change from the proposed solution there because of an already existing check for `null` on the `purl`.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/github-templates/CONTRIBUTING.md#pull-requests)

### DIFF
--- a/src/main/java/org/dependencytrack/policy/PackageURLPolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/PackageURLPolicyEvaluator.java
@@ -56,26 +56,12 @@ public class PackageURLPolicyEvaluator extends AbstractPolicyEvaluator {
         for (final PolicyCondition condition: super.extractSupportedConditions(policy)) {
             LOGGER.debug("Evaluating component (" + component.getUuid() + ") against policy condition (" + condition.getUuid() + ")");
             if (PolicyCondition.Operator.MATCHES == condition.getOperator()) {
-                if (component.getPurl() != null) {
-                    if (component.getPurl().canonicalize().contains(condition.getValue())) {
-                        violations.add(new PolicyConditionViolation(condition, component));
-                    }
-                }
-                if (component.getPurlCoordinates() != null) {
-                    if (component.getPurlCoordinates().canonicalize().contains(condition.getValue())) {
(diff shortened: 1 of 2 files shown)


nscuro (approved):
Sorry for the long wait, @malice00. Thanks a lot for the PR!

Issue #2320, opened 2022-12-21
E-Mail notifications for LDAP Teams not working correctly


rkg-mm:
### Current Behavior

The feature for sending email alerts to complete Teams doesn't seem to work with LDAP users.
The Log shows an error, seems to be for each user in the Team that should receive an alert:

```
2022-12-21 12:47:42,609 [] INFO [org.dependencytrack.tasks.metrics.ProjectMetricsUpdateTask] Completed metrics update for project 28714eb0-3266-4033-800f-75c226921037 in 00:51:434
2022-12-21 12:47:42,715 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
2022-12-21 12:47:42,781 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
2022-12-21 12:47:42,950 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
2022-12-21 12:47:45,797 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
2022-12-21 12:47:45,800 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
2022-12-21 12:47:45,975 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
2022-12-21 12:47:46,059 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
2022-12-21 12:47:49,064 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
2022-12-21 12:47:49,459 [] ERROR [alpine.event.framework.LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: null
[...]

PR #2371, opened 2023-01-12
Fix email notifications for LDAP teams not working properly

rbt-mm:
### Description

Currently, Dependency Track does not retrieve the email addresses of LDAP users from the LDAP server. This results in no email notifications being for LDAP users of a team, which is subscribed to an alert.

By retrieving the email addresses of LDAP users before sending an email notification, all team members should successfully receive an email.  

### Addressed Issue

#2320

### Additional Details

Also fixes `AnalysisDecisionChange` email notifications not containing any analysis, project or component information, if the related notification rule is limited to specific projects.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/notification/NotificationRouter.java
+++ b/src/main/java/org/dependencytrack/notification/NotificationRouter.java
@@ -226,7 +226,7 @@ public class NotificationRouter implements Subscriber {
             return false;
         }
         for (Project child : parent.getChildren()) {
-            if ((child.getUuid().equals(uuid) && child.isActive()) || isChild) {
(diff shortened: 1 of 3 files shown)


nscuro:
@stevespringett Not too familiar with the LDAP side of things. What's the reasoning behind not persisting the email we get during LDAP sync? Having to query the LDAP server for every single notification feels like a massive overhead that we should better avoid.

stevespringett:
> What's the reasoning behind not persisting the email we get during LDAP sync

There is none. This is an oversight

nscuro:
Closing as this will be addressed in Alpine.

Issue #2374, opened 2023-01-12
relation "project_access_teams" does not exist at character 13


asoldo11:
### Current Behavior
I apologize in advance if this is not the issue with Dependency Track itself, but is actually related to Chart. If this is strictly regarded to Chart, issue can be discarded

We deployed Dependency-Track v4.7.0 via Helm chart on private Kubernetes cluster (chart repo https://github.com/evryfs/helm-charts/tree/master/charts/dependency-track). We have overridden parameters regarding dependency track application version (both frontend and apiserver), all LDAP connection parameters and Ingress definition, everything else is default.

After everything was successfully deployed and started (Kubernetes said that liveness and readiness are OK, we did not investigate the logs of particular pod), we started configuring application. We added some teams successfully, added some AD users, uploaded some BOM for projects and everything was running smooth. 

After that we tried to delete some teams but received error, with stacktrace from apiserver below:

<details>
<summary>Stacktrace from apiserver</summary>

2023-01-12 15:16:37,707 ERROR [GlobalExceptionHandler] Uncaught internal server error
javax.jdo.JDODataStoreException: Error executing SQL query "DELETE FROM PROJECT_ACCESS_TEAMS WHERE "PROJECT_ACCESS_TEAMS"."TEAM_ID" = ?".
 at org.datanucleus.api.jdo.JDOAdapter.getJDOExceptionForNucleusException(JDOAdapter.java:605)
 at org.datanucleus.api.jdo.JDOQuery.executeInternal(JDOQuery.java:456)
 at org.datanucleus.api.jdo.JDOQuery.executeWithArray(JDOQuery.java:318)
 at org.dependencytrack.persistence.QueryManager.recursivelyDeleteTeam(QueryManager.java:1278)
 at org.dependencytrack.resources.v1.TeamResource.deleteTeam(TeamResource.java:184)
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
 at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
 at java.base/java.lang.reflect.Method.invoke(Unknown Source)
 at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory.lambda$static$0(ResourceMethodInvocationHandlerFactory.java:52)
[...]

PR #2381, opened 2023-01-15
Fix: Double quoting table to avoid PG "relation does not exist"

syalioune:
### Description

c.f issue. Double quoting the table just like the column in the native sql query.

### Addressed Issue

#2374 

### Additional Details

N/A

### Checklist


stevespringett:
Thanks for the PR

Issue #2042, opened 2022-10-12
Respect hierarchy in ACL definitions


rkg-mm:
### Current Behavior:
With implementation of #84 project hierarchies get actually relevant. For notifications we introduce respecting the hierarchy by including children in a notification rule with https://github.com/DependencyTrack/dependency-track/pull/2013.
But for ACLs this is not yet possible and each version needs to be defined separately.

### Proposed Behavior:
If a team is given access to a project via ACL, any children of this project should be covered automatically, too.

tosicky:
+1

netomi:
This feature would be highly appreciated. The portfolio access management as it is right now is not that useful.

E.g. you model your projects in a hierarchical manner, where different teams have access to differnt top-level projects. Now you need to grant each team access to all child-projects of a tlp for them in order to access them.  If a new version of a child project is being created, this new project needs to be explicitly added to the list of projects the team can access.

Also not sure how it would work to autocreate new versions of a project with this feature enabled, need to test that.

nscuro:
The main issue that was blocking this, was performance concerns of traversing project hierarchies, especially when fetching many objects at once (i.e. listing all projects), see #2407.

In DT v5, we have more freedom in designing efficient queries and table structures, since we can focus on PostgreSQL instead of having to consider 4 different RDBMSes.

An initial implementation of this feature that works with the existing table structure can be found here: https://github.com/DependencyTrack/hyades-apiserver/pull/1060

As a tl;dr, it pushes the ACL check logic into a SQL function, and extends the ORM to be able to call that function.

nscuro:
Update on the above, we're further tweaking the approach to avoid the need for recursive queries altogether. Details about the approach here: https://github.com/DependencyTrack/hyades/blob/issue-1699/docs/architecture/decisions/005-materialize-project-hierarchies.md

It is more or less what @mulder999 had proposed here: https://github.com/DependencyTrack/dependency-track/pull/2407#issuecomment-1951433599

We're going to use database triggers in v5, but the same outcome *could* be achieved in v4 using [DataNucleus lifecycle listeners](https://www.datanucleus.org/products/accessplatform_6_0/jdo/persistence.html#lifecycle_listeners). It will be nowhere near as efficient and reliable, but at least it would work, I *think*.

mulder999:
[...]

PR #2407, opened 2023-01-25
 Include descendants of projects in ACL definitions

rbt-mm:
### Description

Grants team members access to every descendant of a project, which is assigned to their team via ACL.
Also displays the descendants of the assigned projects in the project list and the component search view in the frontend.

### Addressed Issue

#2042

### Additional Details

[Frontend PR]()

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
@@ -27,6 +27,7 @@ import alpine.persistence.PaginatedResult;
 import alpine.resources.AlpineRequest;
 import com.github.packageurl.MalformedPackageURLException;
 import com.github.packageurl.PackageURL;
+import org.datanucleus.api.jdo.JDOQuery;
 import org.dependencytrack.event.IndexEvent;
 import org.dependencytrack.model.Component;
 import org.dependencytrack.model.ComponentIdentity;
@@ -52,6 +53,24 @@ import javax.json.JsonArray;
 final class ComponentQueryManager extends QueryManager implements IQueryManager {
 
     private static final Logger LOGGER = Logger.getLogger(ComponentQueryManager.class);
+
+    public static final String QUERY_ACL_1 = """
+            SELECT *
+            FROM (WITH RECURSIVE "DESCENDANTS" ("ID", "NAME") AS
+                (SELECT "PROJECT"."ID",
+                "PROJECT"."NAME"
+                 FROM "PROJECT"
+            """;
+
+    public static final String QUERY_ACL_2 = """
+            UNION ALL
+                SELECT "CHILD"."ID",
+                "CHILD"."NAME"
+                FROM "PROJECT" "CHILD"
+                     JOIN "DESCENDANTS"
+                         ON "DESCENDANTS"."ID" = "CHILD"."PARENT_PROJECT_ID")
+            SELECT "DESCENDANTS"."ID", "DESCENDANTS"."NAME" FROM "DESCENDANTS") alias
+            """;
 
     /**
      * Constructs a new QueryManager.
@@ -187,7 +206,8 @@ final class ComponentQueryManager extends QueryManager implements IQueryManager 
         };
 
         final Query<Component> query = pm.newQuery(Component.class);;
-        final Map<String, Object> params = Map.of("hash", hash);
+        final Map<String, Object> params = new HashMap<>();
+        params.put("hash", hash);
         preprocessACLs(query, queryFilter, params, false);
         return execute(query, params);
     }
@@ -531,20 +551,65 @@ final class ComponentQueryManager extends QueryManager implements IQueryManager 
                     return;
                 }
             }
+            // Query every project that the teams have access to
+            final Map<String, Object> tempParams = new HashMap<>();
+            final Query<Project> queryAclProjects = pm.newQuery(Project.class);
             if (teams != null && teams.size() > 0) {
-                final StringBuilder sb = new StringBuilder();
+                final StringBuilder stringBuilderAclProjects = new StringBuilder();
                 for (int i = 0, teamsSize = teams.size(); i < teamsSize; i++) {
                     final Team team = super.getObjectById(Team.class, teams.get(i).getId());
-                    sb.append(" project.accessTeams.contains(:team").append(i).append(") ");
-                    params.put("team" + i, team);
+                    stringBuilderAclProjects.append(" accessTeams.contains(:team").append(i).append(") ");
+                    tempParams.put("team" + i, team);
                     if (i < teamsSize-1) {
-                        sb.append(" || ");
+                        stringBuilderAclProjects.append(" || ");
                     }
                 }
-                if (inputFilter != null) {
-                    query.setFilter(inputFilter + " && (" + sb.toString() + ")");
+                queryAclProjects.setFilter(stringBuilderAclProjects.toString());
+            } else {
+                if (inputFilter != null && !inputFilter.isEmpty()) {
+                    query.setFilter(inputFilter + " && false");
                 } else {
-                    query.setFilter(sb.toString());
+                    query.setFilter("false");
+                }
+                return;
+            }
+            List<Project> result = (List<Project>) queryAclProjects.executeWithMap(tempParams);
+            // Query the descendants of the projects that the teams have access to
+            if (result != null && !result.isEmpty()) {
+                final StringBuilder stringBuilderDescendants = new StringBuilder();
+                stringBuilderDescendants.append("WHERE");
+                int i = 0, teamSize = result.size();
+                for (Project project : result) {
+                    stringBuilderDescendants.append(" ID = ").append(project.getId()).append(" ");
+                    if (i < teamSize-1) {
+                        stringBuilderDescendants.append(" OR");
+                    }
+                    i++;
+                }
+                stringBuilderDescendants.append("\n");
+                final Query<Object[]> queryDescendants = pm.newQuery(JDOQuery.SQL_QUERY_LANGUAGE, QUERY_ACL_1 + stringBuilderDescendants + QUERY_ACL_2);
+                final List<Object[]> list = queryDescendants.executeList();
+                // Add queried projects and descendants to the input filter of the query
+                if (list != null && !list.isEmpty()) {
+                    final StringBuilder stringBuilderInputFilter = new StringBuilder();
+                    for (int j = 0, resultSize = list.size(); j < resultSize; j++) {
+                        stringBuilderInputFilter.append(" project.id == :id").append(j);
+                        params.put("id" + j, list.get(j)[0]);
+                        if (j < resultSize-1) {
+                            stringBuilderInputFilter.append(" || ");
+                        }
+                    }
+                    if (inputFilter != null && !inputFilter.isEmpty()) {
+                        query.setFilter(inputFilter + " && (" + stringBuilderInputFilter.toString() + ")");
+                    } else {
+                        query.setFilter(stringBuilderInputFilter.toString());
+                    }
+                }
+            } else {
+                if (inputFilter != null && !inputFilter.isEmpty()){
+                    query.setFilter(inputFilter +" && false");
+                } else {
+                    query.setFilter("false");
                 }
             }
         } else {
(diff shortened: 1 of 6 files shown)


nscuro:
@rbt-mm Did you test this on external databases? If so, on which? I *think* the SQL should be portable, but we should ensure that it's the case.

rkg-mm:
https://modern-sql.com/caniuse/with_recursive_(top_level)
Looking at this the "WITH RECURSIVE" could be an issue with SQLServer?

rbt-mm:
@nscuro I do not have any external databases setup at the moment, so I didn't test it on any of them.

nscuro:
Indeed this appears to be incompatible with MSSQL:

```
javax.jdo.JDODataStoreException: Error executing SQL query "SELECT * FROM (WITH RECURSIVE "DESCENDANTS" ("ID", "NAME") AS     (SELECT "PROJECT"."ID",     "PROJECT"."NAME"      FROM "PROJECT" WHERE ID = ?  UNION ALL     SELECT "CHILD"."ID",     "CHILD"."NAME"     FROM "PROJECT" "CHILD"          JOIN "DESCENDANTS"              ON "DESCENDANTS"."ID" = "CHILD"."PARENT_PROJECT_ID") SELECT "DESCENDANTS"."ID", "DESCENDANTS"."NAME" FROM "DESCENDANTS") alias".
	at org.datanucleus.api.jdo.JDOAdapter.getJDOExceptionForNucleusException(JDOAdapter.java:605)
	at org.datanucleus.api.jdo.JDOQuery.executeInternal(JDOQuery.java:456)
	at org.datanucleus.api.jdo.JDOQuery.executeList(JDOQuery.java:345)
	at org.dependencytrack.persistence.ProjectQueryManager.preprocessACLs(ProjectQueryManager.java:904)
	at org.dependencytrack.persistence.ProjectQueryManager.getProjects(ProjectQueryManager.java:139)
	at org.dependencytrack.persistence.QueryManager.getProjects(QueryManager.java:319)
	at org.dependencytrack.resources.v1.ProjectResource.getProjects(ProjectResource.java:93)
	...
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Incorrect syntax near the keyword 'WITH'.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:265)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1676)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatement(SQLServerPreparedStatement.java:620)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtExecCmd.doExecute(SQLServerPreparedStatement.java:540)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7620)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:3916)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:268)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:242)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.executeQuery(SQLServerPreparedStatement.java:459)
	at com.zaxxer.hikari.pool.ProxyPreparedStatement.executeQuery(ProxyPreparedStatement.java:52)
	at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.executeQuery(HikariProxyPreparedStatement.java)
	at org.datanucleus.store.rdbms.SQLController.executeStatementQuery(SQLController.java:548)
	at org.datanucleus.store.rdbms.query.SQLQuery.performExecute(SQLQuery.java:669)
	at org.datanucleus.store.query.Query.executeQuery(Query.java:2004)
	at org.datanucleus.store.rdbms.query.SQLQuery.executeWithMap(SQLQuery.java:838)
	at org.datanucleus.api.jdo.JDOQuery.executeInternal(JDOQuery.java:437)
```

rkg-mm:
@rbt-mm did set up a test environment yesterday for postgresql and mssql, and is already working on a fix

nscuro:
I will add developer documentation on how to test with external databases, but in short, if you have access to Docker and Docker Compose, this `docker-compose.yml` does the trick:

```yml
version: "3"

services:
  mssql:
    image: mcr.microsoft.com/mssql/server:2022-latest
    container_name: dt-mssql
    environment:
      ACCEPT_EULA: "Y"
      MSSQL_SA_PASSWORD: "Dtrack!234"
    ports:
    - "127.0.0.1:1433:1433"
    volumes:
    - "mssql-data:/var/opt/mssql"
    restart: unless-stopped

  mssql-init:
    image: mcr.microsoft.com/mssql/server:2022-latest
    container_name: dt-mssql-init
    command: /opt/mssql-tools/bin/sqlcmd -S dt-mssql -U sa -P 'Dtrack!234' -Q 'CREATE DATABASE dtrack'
    depends_on:
    - mssql
    restart: on-failure

volumes:
  mssql-data: { }
```

If you use the `Jetty` run configuration with IntelliJ, add these environment variables to it:

```ini
ALPINE_DATABASE_DRIVER=com.microsoft.sqlserver.jdbc.SQLServerDriver
ALPINE_DATABASE_MODE=external
ALPINE_DATABASE_PASSWORD=Dtrack!234
ALPINE_DATABASE_URL=jdbc:sqlserver://localhost:1433;databaseName=dtrack;sendStringParametersAsUnicode=false;trustServerCertificate=true
ALPINE_DATABASE_USERNAME=sa
```

nscuro:
> @rkg-mm: @rbt-mm did set up a test environment yesterday for postgresql and mssql, and is already working on a fix

I should've guessed that, of course you guys are one step ahead again 😅 Thanks!

rbt-mm:
I've tested this fix on MSSQL 2022, PostgreSQL 15.1.1 and the internal H2 database, but it should also work for MySQL, MariaDB and OracleDB.

valentijnscholten:
I understand that this might be the only way to solve this permission inheritance, but I do have some concerts around performance for bigger instances. Just thinking out loud, not trying to be negative. 
Did you try to use traversins the parent-child relationship in JDOQL?

For example (freewheeling here with the queries):

```
SELECT project FROM Project
WHERE project.accessTeams.contains(: team)
|| project.parent != null && project.parent.accessTeams.contains(:team)
|| ...
```

Or the other way around (not sure if this is possible):

```
SELECT project FROM Project
WHERE project.accessTeams.contains(: team)
|| !project.children.isEmpty() && project.childrenaccessTeams.contains(:team)
|| ...
```

This obviously has a limit to the recursion depth that could be supported. And it might also have performance implications.

The reason I am asking is because I think there would be great benefit in having the `prepareACL` code be able to generate the query filter clause without having to do database queries. In the PR 2 database queries are done just to generate the ACL filter clause. In small instances it might work fine and not be really noticable, or maybe it already is noticable. Also the ACL will probably be used in more places, for example portfolio metrics, audit vulnerability overview (new PR), calculating affected projects in the list of vulnerabilities (which already is not so fast). I have understood some supporters of DT are haveing portfolio's with 50k+ projects.

For 50k projects the first query in this PR might be not too bad. The second one might be. It might return 5k or 10k (or more) projects depending on the teams a user is in. In this case 10k * 16bits IDs = 20kb of data needs to be retrieved from the database. Then these IDs need to be transformed into strings, resulting in maybe 70k of string (10k * 3 * 16bit characters + 10k of comma's). This 70kb needs to be parsed by DataNucleus. After that the 70kb will be sent to the database, where it is parsed again. 

Maybe this is a worst case scenario, but the point is that ideally this type of scenario would best be handled inside the DB by using a single query. 

I did see if DataNucleus supports combining JDOQL with PreparedStatements, i.e. to use the results of the PreparedStatement in the JDOQL filter. But it doesn't seem to be supported. Same for [Stored Procedures](https://www.datanucleus.org/products/accessplatform_4_2/jdo/stored_procedures.html). Also no extensions to be found. In Django ORM there's something like 'Extra' which can be used ad 'extra SQL' to a query. Something like that could have been helpful to have a filter like `... project.id in (SELECT DESCENDANTS.....)`

Because we are bound to DataNucleus, there might not be a better way to do it. Unless we do something like maintaining all effective permissions for each project and each user/team to be able to have faster queries.It might be an option to put the inheritance behind a feature flag for now to see how well it performsm similar to what was suggested in #2326?

msymons:
@rbt-mm,  The description says...

> Grants team members access to every descendant of a project

rkg-mm:
@valentijnscholten It is true that this might have an impact on huge portfolios performance. But we didn't find another way that would work. Only supporting a specific amount of levels via JDOQL is not really an option I think (and is that really faster in the end? not sure).
I wonder how other systems do something like this, e.g. create and maintain ACL cache tables? The info is pretty static, only  a few events would have to update it (change ACL, create/delete project, change project parent, ?), but there is surely always a risk we miss one spot where we would need it if we try this. It's not ideal either, but would be fast performance wise. 
Would be interested if anyone has experience in this.

@msymons Not supported, since we already had performance and complexity concerns regarding the "simple" solution. I can think of use cases here, but would not know how to do this performant unless you really work with a complete different datastructure like mentioned above.

valentijnscholten:
I couldn't really think of any tools doing this, except filesystems and windows active directory. Of cour there will be more, but there some things that are "special" in the DT scenario.

- In most tools you probably will not often fetch "all" items a user has permission to. You just disply the top level elements in the tree and maybe expand step by step from there. In DT this preprocess method is used everywhere where permissions are checked, without paging. For example if you go to the list of projects it will show 10-20 projects only, but the preprocess query will possibly retrieve 1000s of project ids just to check the permissions of these 10-20 projects. Even when retrieveing a single component by hash, these 1000s of project ids are retrieved.

- DT is bound to JDO/DataNucleus. Other tools are probably more native and using their own datstructure or maybe plain SQL.

I think the problem can be solved easily in plain SQL as this PR does, but the real limitation is that with JDO we cannot use this CTE as a subquery in JDO queries. I haven't used JPA in a while, but my impression is still that it has (a lot?) more (advanced) RDBMS features out of the box. And more community contributions. For example there is the JPA criteriabuilder from Blazebit which has our exact usecase as an example in their documentation: https://persistence.blazebit.com/documentation/1.6/core/manual/en_US/#recursive-ctes 

BTW CTE support in H2 is experimental: http://www.h2database.com/html/advanced.html#recursive_queries Not sure, but it could be that many people are running DT on H2?

rkg-mm:
@nscuro  what do you think about an ACL cache DB table?

valentijnscholten:
Another option could be to as the DataNucleus team about this scenario. I recently ran into an issue/bug in DN and this was fixed (after some communication): https://datanucleus.groups.io/g/main/topic/1_n_avoidance_fetch_of/96914844.

nscuro:
@rkg-mm @valentijnscholten How would you imagine such a cache table to look like (roughly)?

Generally I think introducing a new table for ACLs is totally a valid option. In other enterprise-ish codebases I've seen, ACLs have always been modeled as a dedicated table instead of via n:m relationships.

rkg-mm:
Rough draft, how about something like this?

```
ProjectID	|	TeamID	|	InheritedFromProjectID
---------------------------------------------------
1		| SuperUser 	|
2 		| SuperUser 	| 1
2 		| Example 	|
3		| SuperUser 	| 1
3		| Example	| 2
4		| Foo		|
5		| Foo		| 4
5		| Bar		|


1 [ACL: SuperUser]
- 2 [ACL: Example]
-- 3
4 [ACL: Foo]
-5 [ACL: Bar]

```

rkg-mm:
@StL-Jim  it won't be included in 4.8 since we need to find a better solution. I don't have much resources which I could allocate from my team so unless someone else takes this over it will take some time.

melba-lopez:
thanks for being forthcoming about time constraints @rkg-mm  I have put this PR on hold and added a help wanted to see if others would be willing to pick up where you left off.

mulder999:
Recursivity can be avoided by using a slightly denormalized form, achieved simply by adding a column that points to the root element in addition to the column that points to its immediate parent.

With this approach, it becomes possible to retrieve the entire tree with a common root without recursion in a single operation. The ORM would reconstruct the structure in memory, allowing you to use the parent navigation property exactly as you would expect.

In your case, because you would like to limit the nodes to a higher depth in the tree, you would also need to store the actual depth of each node.

Bookkeeping of these two additional properties is super easy. Basically, every operation can be achieved in O(1). The migration task is also O(1) generating 'n' writes that can be batched, assuming that you can hold every node that has a parent in memory. If memory is too limited then you need to apply a divide and conquer approach.

stephan-wolf-ais:
Hi all,

Is there plan to finish this PR? Do you need help with testing or any thing else?

Regards

nscuro:
This has been implemented for v5 and works entirely without over-fetching or recursive queries. See discussion in #2042 for details.

Issue #2455, opened 2023-02-06
PR #2336 doesn't fix all cases of AND policy


malice00:
### Current Behavior

When creating a policy that has multiple of the same types of conditions and its operator set to AND, it will not show the violations.

### Steps to Reproduce

1. Create a new policy with operator AND
2. Add multiple conditions of the same type
3. Upload a BOM that should actually violate this policy -- it will not

### Expected Behavior

The policy should be violated and violations shown.

### Dependency-Track Version

4.8.0-SNAPSHOT

### Dependency-Track Distribution

Executable WAR

### Database Server

H2

### Browser

Mozilla Firefox

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

malice00:
I know what the issue is, I'll try to get a fix out when I get home from work tonight.

Issue #1924, opened 2022-08-30
PolicyEngine's ALL operator is flawed


eldis80:
The policy evaluation with operator 'All' doesn't work as expected. The problem is that for the 'All' operator to work, each evaluator should return at most one violation. This is because in the code the number of violations is compared to the number of conditions.

Here:
https://github.com/DependencyTrack/dependency-track/blob/7685a29143dbb23f6d63af20ca4845790d81023b/src/main/java/org/dependencytrack/policy/PolicyEngine.java#L92

At least I'm expecting the 'All' operator to mean that every conditions must match for a policy violation to be raised. We noticed this issue when trying to implement a policy which would include only critical application vulnerabilities and exclude critical OS (Debian/Ubuntu) vulnerabilities. There is a component openssl (3.0.2-0ubuntu1.6) for which Dependency Track has found two critical vulnerabilities and also log4j (1.2.16) with two critical vulnerabilities.

Our policy's conditions look like this:
Severity is Critical
PURL does not match pkg:deb

However, only openssl is raised (wrongly) by the policy as violation. This is because the severity evaluator returns two violations and PURL evaluator returns zero violations. Thus, the number of violations and conditions match. log4j is not raised as a violation because severity evaluator returns two violations and PURL evaluator also returns 1 violation making the total as three which is not equal to number of conditions.

I think there are two possible ways to fix this:

1. Modify every evaluator to return at most one violation
2. Modify the condition in branch for All operator to check that there is at least one violation for each condition

We are using v4.5.0 of Dependency Track

malice00:
Made a PR for this, which unfortunately will still suffer from #2212. Still looking into that one, but it might take some more time to fix. At least this issue should now show the correct component that is in violation of the policies.

PR #2460, opened 2023-02-07
Correctly fix the handling of operator 'AND' in policies

malice00:
### Description

This change correctly fixes the handling of the 'AND'-operator in policies, even if the policy consists of multiple conditions of the same type.

### Addressed Issue

#1924
#2455

### Additional Details

Seeing how the first fix did not take in consideration that multiple conditions of the same type were handled in a single evaluator, it only actually counted the number of evaluators that found violations.
This implementation actually checks how many conditions per evaluator where violated and compares those with the number of actual conditions defined for a policy.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/github-templates/CONTRIBUTING.md#pull-requests)

### DIFF
--- a/src/main/java/org/dependencytrack/policy/PolicyEngine.java
+++ b/src/main/java/org/dependencytrack/policy/PolicyEngine.java
@@ -96,7 +96,7 @@ public class PolicyEngine {
                     final List<PolicyConditionViolation> policyConditionViolationsFromEvaluator = evaluator.evaluate(policy, component);
                     if (!policyConditionViolationsFromEvaluator.isEmpty()) {
                         policyConditionViolations.addAll(policyConditionViolationsFromEvaluator);
-                        policyConditionsViolated += policyConditionViolationsFromEvaluator.stream()
+                        policyConditionsViolated += (int) policyConditionViolationsFromEvaluator.stream()
                                 .map(pcv -> pcv.getPolicyCondition()).sorted(policyConditionComparator).distinct().count();
                     }
                 }
--- a/src/main/java/org/dependencytrack/policy/PolicyEngine.java
+++ b/src/main/java/org/dependencytrack/policy/PolicyEngine.java
@@ -29,6 +29,7 @@ import org.dependencytrack.persistence.QueryManager;
 import org.dependencytrack.util.NotificationUtil;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/dependencytrack/policy/PolicyEngine.java
```diff
@@ -84,8 +93,11 @@ private List<PolicyViolation> evaluate(final QueryManager qm, final List<Policy>
                 int policyConditionsViolated = 0;
                 for (final PolicyEvaluator evaluator : evaluators) {
                     evaluator.setQueryManager(qm);
-                    if (policyConditionViolations.addAll(evaluator.evaluate(policy, component))) {
-                        policyConditionsViolated++;
+                    final List<PolicyConditionViolation> policyConditionViolationsFromEvaluator = evaluator.evaluate(policy, component);
+                    if (!policyConditionViolationsFromEvaluator.isEmpty()) {
+                        policyConditionViolations.addAll(policyConditionViolationsFromEvaluator);
+                        policyConditionsViolated += (int) policyConditionViolationsFromEvaluator.stream()
+                                .map(pcv -> pcv.getPolicyCondition()).sorted(policyConditionComparator).distinct().count();
```
nscuro, line 100: It might be a better idea to execute `sorted` and `distinct` on a stream of primitive values (here: the condition's IDs), as `equals` is not explicitly defined for `PolicyCondition`. That also avoids the custom `Comparator`. How about something like this?

```java
policyConditionsViolated += (int) policyConditionViolationsFromEvaluator.stream()
        .map(PolicyConditionViolation::getPolicyCondition)
        .map(PolicyCondition::getId)
        .sorted(Long::compareTo)
        .distinct()
        .count();
```

nscuro (commented):
Thanks for the PR @malice00, appreciate you digging through the policy functionality!

Minor suggestion on how you determine how many conditions were violated.

src/main/java/org/dependencytrack/policy/PolicyEngine.java
malice00: Can't believe I didn't see that... Guess it had been a long day.
Just wondering if we really want the double mapping and not just do that in a single line? Also, sorting long doesn't really need a comparator, so personally, I'd leave it empty...

```
policyConditionsViolated += (int) policyConditionViolationsFromEvaluator.stream()
        .map(pcv -> pcv.getPolicyCondition().getId())
        .sorted()
        .distinct()
        .count();
```
But if yours is the general coding-style used, I can gladly implement it like that. :-)

nscuro: Wasn't trying to enforce a specific coding style, I just like to give examples of what I mean 😅

Your suggestion sounds good.

malice00: Don't worry, I'm not saying you did -- it's just that the contribution documentation says to use the 'existing style' and I hadn't come across streams yet, so I wanted to make sure if that was maybe the preferred style. :smile: 

New commit pushed.

nscuro (approved):

PR #2466, opened 2023-02-07
remove superfluous ComponentMetricsUpdate event dispatch

valentijnscholten:
### Description

Remove a superfluous dispatch of a `ComponentMetricsUpdateEvent` in `OSSindexAnalysisTask`.

### Addressed Issue

The `OSSindexAnalysisTask` is emitting a `ComponentMetricsUpdateEvent` for each component that was analyzed.
This seems to be superfluous because `OSSindexAnalysisTask` is only triggered from a `VulnerabilityAnalysisTask` and that already triggers a `ProjectMetricsUpdateEvent` when analysis is complete. The latter will also update all metrics for all components in that project.

### Additional Details

If my reasoning is correct, this PR will reduce the load on the event system. The `ComponentMetricsUpdateEvent` is the most frequently ocurring event, at least in my instance (7days):

image

It should go to zero as the `ProjectMetricsUpdateEvent` doesn't emit individual `ComponentMetricsUpdateEvent` instances, it performs the metrics update inline.

### Checklist


nscuro (approved):
Good catch, and thanks for the PR!

PR #2481, opened 2023-02-10
Simplify and speed up VulnerabilityMetricsUpdateTask

valentijnscholten:
### Description

This PR simplifies the `VulnerabilityMetricsUpdateTask` by letting the database handling the counting. The runtime is reduced from ~50 seconds to about 4 seconds.

### Addressed Issue

I noticed the `VulnerabilityMetricsUpdateTask` was taking around 50 seconds to complete on my instance. The only thing that task does is to count the vulnerabilities that were published each year/month. The old code was loading all ~300k vulnerabilities into memory, in batches of 500. So this meant around 600 queries and 300000 object instantiations and other Datanucleus "overhead".

The PR now performs 2 queries to get all the counts from the database. Then it does some juggling to match up the `published` and `created` fields. And then the old metrics are replaced by the new ones.

### Additional Details

During the refactoring I found out this data is not even used by the DT front-end, so it might not be on top of the priority list. But it is exposed through the API. Either way, I had fun optimizing it. First time doing something with Datanucleus as well as Java Streams.

Question: The old code had some logic in place to keep the old database ID values in place for each Year-Month combination. This PR just removes old metrics and creates new ones to keep it simple. Not sure if there are any clients/users that expect the ID to be constant? I hope they look at the Year-Month combination...

### Checklist


src/main/java/org/dependencytrack/tasks/metrics/VulnerabilityMetricsUpdateTask.java
```diff
@@ -59,104 +62,72 @@ private void updateMetrics() throws Exception {
...
+            .flatMap(e -> Stream.of(new VulnerabilityMetrics((Integer)e.getKey(), null, e.getValue().values().stream().mapToInt(d->d.intValue()).sum(), measuredAt)));
+
+            // Get Mutable List to merge again
+            List<VulnerabilityMetrics> vulnerabilityMetrics = new ArrayList<VulnerabilityMetrics>(monthlyStream.toList());
+            vulnerabilityMetrics.addAll(new ArrayList<>(yearlyStream.toList()));
+
+            qm.synchronizeVulnerabilityMetrics(vulnerabilityMetrics);
+
+            LOGGER.info("Completed metrics update on vulnerability database in " +
+            DurationFormatUtils.formatDuration(new Date().getTime() - measuredAt.getTime(), "mm:ss:SS"));
```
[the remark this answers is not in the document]
valentijnscholten: @sonatype-lift ignore this is all over the place in DT still

valentijnscholten:
Looks like DN is generating the wrong SQL for H2, even though H2 should be supported as well as the `getYear()` and `getMonth()`methods:
```
"Syntax error in SQL statement "SELECT YEAR(A0.PUBLISHED) AS [*]YEAR,((MONTH(A0.PUBLISHED) - 1) + 1) AS MONTH,COUNT(A0.ID) AS COUNT FROM VULNERABILITY A0 WHERE A0.PUBLISHED IS NOT NULL AND A0.CREATED IS NULL GROUP BY YEAR(A0.PUBLISHED),((MONTH(A0.PUBLISHED) - 1) + 1)"; expected "identifier""
```

nscuro:
@valentijnscholten I think H2 is stumbling here because `MONTH` and `YEAR` are [reserved function names](http://www.h2database.com/html/functions.html), and you use those names for variables (`AS MONTH`).

valentijnscholten:
Yeah I'm testing a fix and adding some tests.

src/main/java/org/dependencytrack/tasks/metrics/VulnerabilityMetricsUpdateTask.java
```diff
@@ -59,104 +62,75 @@ private void updateMetrics() throws Exception {
...
+            Collection<YearMonthMetric> published = queryForMetrics(pm, "published", "created");
+
+            // Get metrics by created date regardless of value of published field
+            Collection<YearMonthMetric> created = queryForMetrics(pm, "created", null);
+
+            // Merge flat lists
+            published.addAll(created);
+
+            // Collect into nested map so we can sum the counts
+            Map<Object, Map<Object, Long>> metrics = published.stream()
```
nscuro, line 91: If you provide concrete type arguments here, you can avoid the casting to `Integer` later on:

```suggestion
            Map<Integer, Map<Integer, Long>> metrics = published.stream()
```

src/main/java/org/dependencytrack/tasks/metrics/VulnerabilityMetricsUpdateTask.java
```diff
@@ -59,104 +62,75 @@ private void updateMetrics() throws Exception {
...
+                            Collectors.groupingBy(ymm -> ymm.month, Collectors.summingLong(ymm -> ymm.count))));         
+
+            // Flatten again, but now into VulnerabilityMetrics that can be persisted
+            Stream<VulnerabilityMetrics> monthlyStream = metrics.entrySet().stream()
+            .flatMap(e -> e.getValue().entrySet().stream().flatMap(
+                v -> Stream.of(new VulnerabilityMetrics((Integer)e.getKey(), (Integer)v.getKey(), v.getValue().intValue(), measuredAt))));
+
+            // Flatten another time, for the yearly counts
+            Stream<VulnerabilityMetrics> yearlyStream = metrics.entrySet().stream()
+            .flatMap(e -> Stream.of(new VulnerabilityMetrics((Integer)e.getKey(), null, e.getValue().values().stream().mapToInt(d->d.intValue()).sum(), measuredAt)));
```
nscuro, line 103: You don't need `flatMap` (and thus also `Stream.of`) here, as you're only ever returning one element. So a `map` without the `Stream.of` part should suffice.

src/main/java/org/dependencytrack/tasks/metrics/VulnerabilityMetricsUpdateTask.java
```diff
@@ -59,104 +62,75 @@ private void updateMetrics() throws Exception {
...
+
+            // Get Mutable List to merge again
+            List<VulnerabilityMetrics> vulnerabilityMetrics = new ArrayList<VulnerabilityMetrics>(monthlyStream.toList());
+            vulnerabilityMetrics.addAll(new ArrayList<>(yearlyStream.toList()));
+
+            qm.synchronizeVulnerabilityMetrics(vulnerabilityMetrics);
```
nscuro, line 109: Can be simplified to:

```suggestion
            qm.synchronizeVulnerabilityMetrics(Stream.concat(monthlyStream, yearlyStream).toList());
```

nscuro (commented):
Great optimization, thanks @valentijnscholten!

I made some suggestions on how the code could be simplified even more, but those are really just of cosmetic nature.

> Question: The old code had some logic in place to keep the old database ID values in place for each Year-Month combination. This PR just removes old metrics and creates new ones to keep it simple. Not sure if there are any clients/users that expect the ID to be constant? I hope they look at the Year-Month combination...

To be honest I don't think there was a specific reason for keeping it, other than perhaps the assumption that deleting and inserting it all would be more expensive that updating.

IDs are never exposed to clients, so this change should not matter to users.

src/main/java/org/dependencytrack/persistence/MetricsQueryManager.java
```diff
@@ -162,34 +164,21 @@ public List<DependencyMetrics> getDependencyMetricsSince(Component component, Da
...
-            persist(metric);
-        } else {
-            delete(result);
-            persist(metric);
-        }
+    public void synchronizeVulnerabilityMetrics(List<VulnerabilityMetrics> metrics) {
+        // No need for complex updating, just replace the existing ~400 rows with new ones
+        // Unless we have a contract with clients that the ID of metric records cannot change?
+
+        pm.setProperty(PropertyNames.PROPERTY_QUERY_JDOQL_ALLOWALL, "true");
```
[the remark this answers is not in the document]
valentijnscholten: I had to set it here, otherwise it does ~400 queries

src/main/java/org/dependencytrack/tasks/metrics/VulnerabilityMetricsUpdateTask.java
valentijnscholten: ok thanks, in one of the earlier iterations it wouldn't accept this cast

src/main/java/org/dependencytrack/tasks/metrics/VulnerabilityMetricsUpdateTask.java
valentijnscholten: ok, thanks

src/main/java/org/dependencytrack/persistence/MetricsQueryManager.java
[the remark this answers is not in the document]
valentijnscholten: Looks like it's not needed with the current version of the code.

src/main/java/org/dependencytrack/persistence/MetricsQueryManager.java
```diff
@@ -162,34 +164,21 @@ public List<DependencyMetrics> getDependencyMetricsSince(Component component, Da
...
+        final Query<VulnerabilityMetrics> delete = pm.newQuery("DELETE FROM org.dependencytrack.model.VulnerabilityMetrics");
+        delete.execute();
+
+        // This still does ~400 queries, probably because not all databases can do bulk insert with autogenerated PKs
+        // Or because Datanucleus is trying to be smart as it wants to cache all these instances
+        pm.makePersistentAll(metrics);
```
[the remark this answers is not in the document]
valentijnscholten: Good idea, I thought implicitly that the whole event ran in a transaction, but it doesn't.

src/main/java/org/dependencytrack/tasks/metrics/VulnerabilityMetricsUpdateTask.java
valentijnscholten: I had that at first, but it complained about the stream already having been operated upon. But with the current version of the code it seems to work.

nscuro (approved):
Looks good to me, thanks @valentijnscholten! 

Any more pending changes you were planning to add, or is this ready to merge from your side?

valentijnscholten:
I think it's good to go.

Issue #2473, opened 2023-02-08
Edge case of duplicate ecosystems in OSV config


sahibamittal:
### Current Behavior

While selecting ecosystems for OSV mirroring, same ecosystem can be selected multiple times. This may cause duplicate downloading of one ecosystem (large files) from OSV client, which may result in repeated processing of already processed vulnerabilities.

### Proposed Behavior

Fix the selection of ecosystems to be unique list.
**From backend:** Handle the list of ecosystems to be unique.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #2491, opened 2023-02-13
Issue-2473 : fix OSV ecosystem to be unique

sahibamittal:
Signed-off-by: sahibamittal <sahiba.mittal@citi.com>

### Description

While selecting ecosystems for OSV mirroring, same ecosystem can be selected multiple times. This may cause duplicate downloading of one ecosystem (large files) from OSV client, which may result in repeated processing of already processed vulnerabilities.
Issue: https://github.com/DependencyTrack/dependency-track/issues/2473

### Addressed Issue

1. Fix the selection of ecosystems to be unique list. Handle the list of ecosystems to be unique in backend.
2. Add a missing check in switch block for SNYK in VulnerabilityAlias.java (not related to this issue).

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/OsvEcosytemResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/OsvEcosytemResource.java
@@ -34,6 +34,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import java.util.List;
+import java.util.stream.Collectors;
 
 @Path("/v1/integration/osv/ecosystem")
 @Api(value = "ecosystem", authorizations = @Authorization(value = "X-Api-Key"))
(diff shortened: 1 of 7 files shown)


VithikaS (approved):

mehab (approved):
lgtm

src/main/java/org/dependencytrack/resources/v1/OsvEcosytemResource.java
```diff
@@ -42,7 +43,7 @@ public class OsvEcosytemResource extends AlpineResource {
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(
-            value = "Returns a list of all ecosystems in OSV",
+            value = "Returns a list of all not-yet-selected ecosystems in OSV",
```
nscuro, line 46: I feel like this should either be a separate endpoint (e.g. `/api/v1/integration/osv/ecosystem/inactive`), or a different solution is needed. If the behavior of the existing endpoint is changed like this, it will be unnecessarily hard to work out which ecosystems exist overall.

Relying on this endpoint to do the de-duplication will also not prevent duplicate values from entering the system. So perhaps performing de-duplication when *saving* the property (e.g. [here](https://github.com/DependencyTrack/dependency-track/blob/4.7.1/src/main/java/org/dependencytrack/resources/v1/AbstractConfigPropertyResource.java#L57)) would be more fool-proof?

sahibamittal: Duplicate values should not be able to be selected by end-user also, and sent back to backend in configProperty. 
So, either we can write a new endpoint for this, in order to display only not-yet-selected-available ecosystems in UI or to do same we can do this subtraction of (ALL - SELECTED) in frontend logic itself?

nscuro (approved):

PR #2525, opened 2023-02-25
Disable DataNucleus L2 cache for mirroring tasks

nscuro:
### Description

This PR lowers the memory footprint of mirroring operations, by disabling the "Second Level" cache of the object relational mapper we use (DataNucleus). **The cache is only disabled for mirroring operations, other areas of DT are not impacted**.

Prior to this change, heap usage would kreep up to almost 5.5 GB while mirroring the NVD and EPSS. Even after a manual GC invocation at the end, ~2.8 GB would still remain in heap.

nvd-mirror-vanilla_manual-gc

After disabling L2 caching, overall heap usage stays lower (peaks at  ~4.2 GB), and a manual GC shows that pretty much the entire heap can be reclaimed.

Screenshot 2023-02-25 at 22 31 58

### Addressed Issue

N/A

### Additional Details

Objects that are persisted (created, updated) are put into DN's second level cache. Per default, the cache keeps soft references to objects.

https://www.datanucleus.org/products/accessplatform_6_0/jdo/persistence.html#cache_level2

While this kind of cache may be beneficial for BAU operations, it is totally overkill for our mirroring tasks. Over the course of mirroring the NVD for the first time alone, the cache will grow to over 2 million objects. The majority of those objects will never be accessed again, especially not soon enough for the cache to be of any use.

Soft references can potentially stay for a long time, and thus may inflate heap utilization.

There is a bug in DN that caused NPEs in `Vulnerability#setCwes` on newly created `Vulnerability` objects when L2 cache is disabled. The DN-internal method `dnGetcwes` would return `null`, despite the `cwes` field being initialized with `new ArrayList<>()` one line before. Not sure why this happens, but it can be prevented by setting the `cwes` field to a fully initialized list.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/QueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/QueryManager.java
@@ -29,6 +29,7 @@ import alpine.persistence.AlpineQueryManager;
 import alpine.persistence.PaginatedResult;
 import alpine.resources.AlpineRequest;
 import com.github.packageurl.PackageURL;
+import org.datanucleus.PropertyNames;
 import org.datanucleus.api.jdo.JDOQuery;
 import org.dependencytrack.event.IndexEvent;
 import org.dependencytrack.model.AffectedVersionAttribution;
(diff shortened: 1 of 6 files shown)


nscuro:
There still seem to be some excessive allocations in the NVD parsing that should be optimized. For both GHSA (first screenshot) and OSV (second screenshot) we barely even scratch the 500 MB mark.

Screenshot 2023-02-25 at 22 52 32

Screenshot 2023-02-25 at 23 10 13

src/main/java/org/dependencytrack/persistence/QueryManager.java
```diff
@@ -314,6 +315,22 @@ private CacheQueryManager getCacheQueryManager() {
...
+     * blow up heap usage.
+     *
+     * @return This {@link QueryManager} instance
+     * @see <a href="https://www.datanucleus.org/products/accessplatform_6_0/jdo/persistence.html#cache_level2">L2 Cache docs</a>
+     */
+    public QueryManager withL2CacheDisabled() {
+        pm.setProperty(PropertyNames.PROPERTY_CACHE_L2_TYPE, "none");
+        return this;
+    }
+
```
valentijnscholten, line 333: With this method on the `QueryManager` I would expect the L2 Cache to be enabled again on subsequent quermanagers created from the same `pm`. Is this the case? It looks to me as if L2 is disabled for the life of the `pm` and not the `QueryManager`? In this case where the whole tasks is supposed to have L2 disabled it makes no difference, but it could be confusing for other scenario's where only a couple of queries want to have their L2 disabled?

nscuro: > With this method on the QueryManager I would expect the L2 Cache to be enabled again on subsequent quermanagers created from the same pm. Is this the case? It looks to me as if L2 is disabled for the life of the pm and not the QueryManager?

Yes, the setting applies to the lifetime of the `PersistenceManager`. Every `QueryManager` creates a new underlying `PersistenceManager`, and closes it again when it itself is closed. So the lifetime of `QueryManager` and `PersistenceManager` is tightly coupled.

> but it could be confusing for other scenario's where only a couple of queries want to have their L2 disabled?

Hmmm... Any suggestions on making this more obvious? AFAICT for individual queries one would use

```java
var query = pm.newQuery(...);
query.setIgnoreCache(true);
```

And for writing operations one *must* use the `PersistenceManager` setting.

Issue #2424, opened 2023-01-30
DefectDojo reimport reopens closed findings


KramNamez:
### Current Behavior

Because DependencyTrack's integration sets `active: true` and `verified: true`, when a finding is reimported, it is opened again in DefectDojo. 

This adds a lot of work because we cannot remove false positives and they keep clogging up DD's UI.

### Proposed Behavior

Adding the `do_not_reactivate` flag to Dependency-Track's upload would change this behavior so that findings that got closed on the DD side stay closed.

This way, I have both plenty of time to go through all the findings and remove noise from the findings before passing them to the devs, and findings that are not going to be fixed do not keep reappearing.

The core of the change is very simple, but it is a breaking change and one that likely should be configurable, so I figured I'd open this issue first, in order to discuss the best way of implementing this.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

nscuro:
> The core of the change is very simple, but it is a breaking change

True, but I as I do not use DD myself, I'm wondering what a valid use case would be for reactivating. I'll share this issue in Slack, maybe some folks who use this integration can chime in.

valentijnscholten:
DD normally always assumes that the imported report is the source of truth. The `do_not_reactivate` flag was added recently for cases like this I assume. There's possibly a potential for false negatives if a finding was closed at some point and then later reappeared because somebody downgraded a library or something like that. I haven't tested this scenario. 
To be honest I was expecting it to be a tag on a finding to avoid reactivating a finding, which might be a safer option.
Either way I think it should be configurable and not set by default. 

It would be nice to have a 2-way integration with defect dojo so that audit decisions in DD are reflected in DT and vice versa.

lme-nca:
I created a proposed solution for this problem. The pull request I propose is just a quick fix, I also think that in the long term a feature such as: https://github.com/DependencyTrack/dependency-track/issues/1926 or something similar (as @valentijnscholten mentioned a 2-way integration with Defectdojo) would be needed.

PR #2532, opened 2023-02-27
add DefectDojo "do not reactivate" flag, fixes issue 2424

lme-nca:
### Description

This is a proposed fix for issue: [2424](https://github.com/DependencyTrack/dependency-track/issues/2424) 

### Addressed Issue
https://github.com/DependencyTrack/dependency-track/issues/2424

### Additional Details

This is a simple fix, by adding a new property a User could set the "do not reactivate" flag that is available on the Defectdojo API. This stops the finding from being reactivated in Defectdojo, even if its still available in subsequent scans.

This comes with the downside that a potentially patched vulnerability that is re-introduced by, for example a library downgrade, is reactivated. We could document this edge-case in the Defectdojo sync documentation.

### Checklist

- [] This PR fixes a defect, and I have provided tests to verify that the fix is effective


lme-nca:
Hi @nscuro I have a suggested fix for this issue, do you think this approach is ok ? if so ill go ahead and update documentation and some tests to complete the pull request.

nscuro:
The DD integration is entirely community-maintained, and I do not have a good understanding of how DefectDojo works. Given the discussions that lead to this PR, I think it's fine.

@valentijnscholten Any counterarguments?

valentijnscholten:
LGTM. I would put something like "DefectDojo: " in the PR title to make it more clear, but that's just a detail.

valentijnscholten (approved):

nscuro (approved):
Thanks @lme-nca! 🚀

Issue #2538, opened 2023-02-28
VEX import does not apply to SNYK/OSV/VULNDB sources


syalioune:
### Current Behavior

Current implementation of `CycloneDXVexImporter` contains a vulnerability source filter that seems incomplete.

```java
 if ((vuln.getSource().equals(Vulnerability.Source.NVD.name())
                        || vuln.getSource().equals(Vulnerability.Source.OSSINDEX.name())
                        || vuln.getSource().equals(Vulnerability.Source.GITHUB.name())
                        || vuln.getSource().equals(Vulnerability.Source.INTERNAL.name()))
                        && vuln.getVulnId().equals(cdxVuln.getId())) {
```

As it stands, vulnerability from `OSV, SNYK, VULNDB` cannot be audited using a VEX file.

### Steps to Reproduce

1. Enable `NVD` and `OSV` analyzers, and wait for the sync to complete
2. Create a SBOM containing components `pkg:deb/debian/puppet@2.7.0` and `pkg:alpine/py3-pillow@8.2.0` which are respectively affected by vulnerabilities `DSA-2715-1` from `OSV` and `CVE-2021-23437` from `NVD` (see attached file)
3. Upload the SBOM and ensure that the vulnerabilities are indeed found for the test project
4. Create a VEX to set vulnerabilities `DSA-2715-1, CVE-2021-23437` as `false_positive` (see attached file)
5. Check that only `CVE-2021-23437` has been audited and is suppressed

[2538-sbom.txt](https://github.com/DependencyTrack/dependency-track/files/10854568/2538-sbom.txt)
[2538-vex.txt](https://github.com/DependencyTrack/dependency-track/files/10854569/2538-vex.txt)

### Expected Behavior

Vulnerabilities from `OSV, SNYK, VULNDB` are audited upon VEX upload just like any other vulnerabilities

### Dependency-Track Version

4.8.0-SNAPSHOT

### Dependency-Track Distribution

Container Image

### Database Server

N/A

### Browser

Google Chrome

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

msymons:
Include tests to ensure that future integrations do not slip through the net.

syalioune:
[...]

PR #2549, opened 2023-03-02
Fix: Allowing VEX import to audit all vulnerability sources

syalioune:
### Description

Allowing VEX import to audit all vulnerability sources. I removed the previous vulnerability source filter. Any feedback on why they were relevant (for rollback) will be appreciated.

### Addressed Issue

See #2538 for details

### Additional Details

N/A

### Checklist


nscuro (approved):
Thanks @syalioune!

I too can't see a reason for VEX imports being limited to NVD, OSS Index, GitHub, and internal vulnerabilities.

@stevespringett, is there an edge case we're overseeing?

stevespringett:
> is there an edge case we're overseeing?

Yes. The source is a string field in CycloneDX and can be anything. For example, the source could be NVD, "National Vulnerability Database", or "Acme Inc". The intent of the code that checks the source is to ensure the source defined in the BOM matches the sources that DT supports. If its an unsupported source, it should be rejected.

In addition, the source and vulnid need to match.

For example, a vulnId of 12345 which may be from VulnDB or may be an internal vulnerability.

syalioune:
> The intent of the code that checks the source is to ensure the source defined in the BOM matches the sources that DT supports. If its an unsupported source, it should be rejected

OK got it. Then the original check did not perform the intended verification as it was done on the DB vulnerability, not the one coming from the VEX. In that case, the source would have always been correct.

To sum up :

- If VEX vulnerability does not contain source : vulnId from VEX must match an existing vulnerability bound to the project (**any source match**)
- If VEX vulnerability contain source :
  - vulnId must match an existing vulnerability bound to the project
  - the source on the VEX vulnerability must match the one on the existing vulnerability (**same source match**)
- A pre-filter could be done to eliminate all VEX vulnerabilities whose source is not supported

nscuro:
> If VEX vulnerability does not contain source : vulnId from VEX must match an existing vulnerability bound to the project (any source match)

IIUC this case should specifically *not* happen, due to the ambiguity of some vulnerability identifiers.

src/main/java/org/dependencytrack/parser/cyclonedx/CycloneDXVexImporter.java
```diff
@@ -81,6 +80,13 @@ public void applyVex(final QueryManager qm, final Bom bom, final Project project
         }
     }
 
+    private boolean shouldAuditVulnerability(org.cyclonedx.model.vulnerability.Vulnerability bomVulnerability, Vulnerability dtVulnerability) {
+        boolean result = true;
+        result = result && dtVulnerability.getVulnId().equals(bomVulnerability.getId());
+        result = result && (bomVulnerability.getSource() == null || dtVulnerability.getSource().equalsIgnoreCase(bomVulnerability.getSource().getName()));
```
nscuro, line 86: @syalioune I think the `bomVulnerability.getSource() == null` case should not cause the analysis to be applied (https://github.com/DependencyTrack/dependency-track/pull/2549#issuecomment-1493126424) and should be removed here.

nscuro (approved):

Issue #2443, opened 2023-02-03
Improve automated deployment with Docker


phoenixadb:
### Current Behavior

The password for the admin account has to be update once the application starts for the first time.
The different vulnerabilities datasources: NVD, GitHub advisories, OSS Index, Open Source Vulnerabilities, Sonatype OSS Index, Snyk need to be manually configured. 

### Proposed Behavior

To have properties inside the Docker Compose file to configure as code Dependency Track.

Example:
`
ALPINE_DEFAULT_ADMIN_NAME="admin"
ALPINE_DEFAULT_ADMIN_PASSWORD="admin"
ALPINE_DATASOURCE_NVD_MIRROR="http://hostname/mirror/nvd"
ALPINE_DATASOURCE_GHSA_PAT="mypat"
ALPINE_DATASOURCE_OSV_ENABLE=true
ALPINE_DATASOURCE_OSV_BASE_URL="https://osv-vulnerabilities.storage.googleapis.com"
ALPINE_DATASOURCE_OSV_ECOSYSTEM="JavaScript;Maven"
ALPINE_DATASOURCE_SONATYPE_OSS_INDEX_ENABLE=true
ALPINE_DATASOURCE_SONATYPE_OSS_INDEX_EMAIL="me@example.com"
ALPINE_DATASOURCE_SONATYPE_OSS_INDEX_API_TOKEN="mytoken"
ALPINE_DATASOURCE_SNYK_ENABLE=true
ALPINE_DATASOURCE_SNYK_ORGANIZATION_ID="myid"
ALPINE_DATASOURCE_SNYK_API_TOKEN="mytoken1;mytoken2"
ALPINE_DATASOURCE_SNYK_API_VERSION="2022-11-14"
ALPINE_DATASOURCE_VULNDB_ENABLE=true
ALPINE_DATASOURCE_VULNDB_BASE_URL="https://vulndb.cyberriskanalytics.com"
ALPINE_DATASOURCE_VULNDB_API_KEY="mykey"
ALPINE_DATASOURCE_VULNDB_API_SECRET="mysecret"
`

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

jorikseldeslachts:
[...]

PR #2550, opened 2023-03-02
Feature: Improve automated containerized deployment

syalioune:
### Description

Allow to configured several items through environment variables on first startup for containerized deployments :

- `ConfigProperty` values
- Default admin user and credentials

### Addressed Issue

#2443

### Additional Details

N/A

### Checklist


src/main/java/org/dependencytrack/persistence/DefaultObjectGenerator.java
```diff
@@ -244,13 +265,23 @@ private void loadDefaultConfigProperties() {
...
-            }
+            NotificationUtil.loadDefaultNotificationPublishers(qm);
+        } catch (IOException e) {
+            LOGGER.error("An error occurred while synchronizing a default notification publisher", e);
         }
     }
+
+    String generateEnvVariableName(ConfigPropertyConstants configProperty) {
+        StringBuilder sb = new StringBuilder();
+        sb.append(configProperty.getGroupName().toUpperCase().replaceAll("[\\-\\.]", "_"));
```
nscuro, line 276: Should we require a common prefix here, too? Similar to `DEPENDENCY_TRACK_ADMIN_USERNAME`.

Perhaps `DT` would be a better choice than `DEPENDENCY_TRACK`, to keep the length reasonable :D

syalioune: I thought about it and didn't add a prefix out of pure lazyness. For consistency check it would require env variable like  `DEFAULT_TEMPLATES_OVERRIDE_ENABLED` to have the prefix also.
I'll add the `DT` prefix to match alpine behaviour.

src/main/java/org/dependencytrack/persistence/DefaultObjectGenerator.java
```diff
@@ -48,6 +48,24 @@ public class DefaultObjectGenerator implements ServletContextListener {
 
     private static final Logger LOGGER = Logger.getLogger(DefaultObjectGenerator.class);
 
+    static final String DEFAULT_ADMIN_USERNAME = "admin";
+
+    static final String ADMIN_USERNAME_ENV_VARIABLE = "DEPENDENCY_TRACK_ADMIN_USERNAME";
```
valentijnscholten, line 53: Should it be made even more clear that this is just about the initial value for these variables, i.e. `DEPENDENCY_TRACK_INITIAL_ADMIN_USERNAME`? To prevent people on Slack asking why the "change of the admin password env variable doesn't work".

syalioune: Yes good idea !

marcosene (approved):

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -79,8 +79,8 @@ public enum ConfigPropertyConstants {
...
     ACCESS_MANAGEMENT_ACL_ENABLED("access-management", "acl.enabled", "false", PropertyType.BOOLEAN, "Flag to enable/disable access control to projects in the portfolio"),
-    NOTIFICATION_TEMPLATE_BASE_DIR("notification", "template.baseDir", SystemUtils.getEnvironmentVariable("DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY", System.getProperty("user.home")), PropertyType.STRING, "The base directory to use when searching for notification templates"),
-    NOTIFICATION_TEMPLATE_DEFAULT_OVERRIDE_ENABLED("notification", "template.default.override.enabled", SystemUtils.getEnvironmentVariable("DEFAULT_TEMPLATES_OVERRIDE_ENABLED", "false"), PropertyType.BOOLEAN, "Flag to enable/disable override of default notification templates"),
+    NOTIFICATION_TEMPLATE_BASE_DIR("notification", "template.baseDir", SystemUtils.getEnvironmentVariable("DT_DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY", SystemUtils.getEnvironmentVariable("DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY", System.getProperty("user.home"))), PropertyType.STRING, "The base directory to use when searching for notification templates"),
```
melba-lopez, line 82: @syalioune maybe im misinterpreting this line.  in the deploy-docker.md and docker-compose.yml you replaced:

`DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY`
with
`DT_DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY`

should this state: 
....`SystemUtils.getEnvironmentVariable("DT_DEFAULT_TEMPLATES_OVERRIDE_BASE_DIRECTORY", System.get.`....

melba-lopez: similar comment for the next line "`DEFAULT_TEMPLATES_OVERRIDE_ENABLED`" is called, but was removed earlier.

melba-lopez (commented):
@syalioune thanks for adding this feature!!  I've been thinking about how to do this on my laptop before it gets deployed.  looking forward to using it :) 

I only have 1 minor comment on 2 lines.  It may be a misinterpretation on my part, but wanted to make sure.

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
syalioune: Hello @melba-lopez 
Sorry I missed your comment.
The intent is to be backward compatible with previous releases where `DEFAULT_TEMPLATES_OVERRIDE_.*` variables were valid so the code fetches the values from (decreasing priority order) :

- `DT_DEFAULT_TEMPLATES_.*`
- `DEFAULT_TEMPLATES_.*`
- `user.home`

nscuro:
Thanks for your work on this @syalioune. 

I am going to postpone merging this though. I've had a few discussions around this and having environment variables that are only effective upon first launch of the application turn out to be rather confusing to users. It was pointed out that something akin to [what Grafana is doing](https://grafana.com/docs/grafana/latest/administration/provisioning/) would be preferable.

Keeping it under the umbrella of "provisioning" makes it more obvious what it's doing. 

It also makes more sense given that we need similar functionality for other areas of the application, e.g. alert configurations, policy definitions, and more (see for example #3101).

sylvainOL:
Hello, I've seen that #2550 has not been integrated in 4.9 or 4.10. Do you know the plan to handle it?

thanks, I'm very interested in this so we can create a small operator (mainly for configuration) dealing with Dependency Track

nscuro:
@sylvainOL There's no one actively working on this ATM. But given your plans to build an operator around DT, we'd love to hear from you how configuration should look like *ideally*. What would make your life the easiest?

sylvainOL:
Hello @nscuro,

After a quick research, I think I've got two big missing point to simplify creating a (configuration) operator:

* default admin password. We could change it the first time (if API is available) but it's preferrable to be able to give a secret / env variable for setting it.
* first API key: AFAIK, the first API key creation flow is not very linear. Same as for admin password, it would be nice to be able to retrieve it from env variable

After, as @phoenixadb (disclosure, we're on same company ;) ) said in #2443, if basic configuration (enabling the different vuln databases) could be also done via env variable it would be good.
If not, a simple API endpoint for that is mandatory of course.

And by configuration operator, I mean a basic one but if community is interested, we can opensource it (when we have something, we don't have anything ad far of today).

PR #2561, opened 2023-03-04
Add developer docs for database inspection

nscuro:
### Description

Oftentimes during development and local testing, it is necessary to inspect data in the database.

Because DT uses an embedded H2 database per default, and that database can only be used by a single process at a time, it can be hard for new contributors to figure out how to do this correctly.

This PR adds documentation around that topic to our `DEVELOPING.md`.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


DEVELOPING.md
```diff
@@ -86,6 +86,74 @@ mvn jetty:run -P enhance -Dlogback.configurationFile=src/main/docker/logback.xml
...
+* URL: `jdbc:h2:~/.dependency-track/db`
+
+> **Warning**  
+> Make sure that your database tool uses version **2** of the H2 database driver.
+> Connections using version 1 of the driver will fail!
+
+A limitation of the H2 database in `embedded` mode is that *only a single process at a time can access it*.
+If you want to inspect the database while Dependency-Track is running, you have two options:
+
+#### Use H2 in `server` mode
```
syalioune, line 109: May I suggest a paragraph on using the H2 web console as well ?
It requires minor addition on `web.xml` (see [attached file](https://github.com/DependencyTrack/dependency-track/files/10891772/web.h2.txt))  and is external-tool free

nscuro: Good idea!

We should even be able to add native support for this in Alpine. I'll update the docs and look into extending Alpine separately.

nscuro: I tried to add a `ServletContextListener` that would dynamically add the `WebServlet` of the H2 console, depending on whether it's enabled via `Config`. That part works, however H2 is using `iframe`s, and the [current CSP](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/webapp/WEB-INF/web.xml#L89) is not allowing those. 

Modifying `init-params` of filters in `ServletContextListener` is not possible, so we'd have to hardcode `frame-ancestors: 'self'` in the CSP to make it work. I don't believe that setting would compromise on security, but it feels very dirty to implement workarounds in production code to make test stuff work...

nscuro: Does this correctly display the H2 console when you browse to `/h2-console/`?

I think I mixed things up a little when I wrote my previous comment. Adding the `frame-ancestors` init param works, because it's not set in `web.xml`. The `FilterRegistration` will let you *add* params, but not modify those that are already there. [`FilterRegistration#setInitParameter`](https://docs.oracle.com/javaee/7/api/javax/servlet/Registration.html#setInitParameter-java.lang.String-java.lang.String-) will return `false` and not modify the underlying value if it already existed. [`getInitParameters`](https://docs.oracle.com/javaee/7/api/javax/servlet/Registration.html#getInitParameters--) is supposed to return an immutable map, so modifying that *shoudn't* work, but I admittedly didn't try it.

The actual problem I was having was that I couldn't modify the `WhitelistUrlFilter`'s `allowUrls` and `forwardExcludes` params. Unless they both included `/h2-console`, I was always redirected to `/index.html`.

syalioune: You're right, the `setInitParameter` provide some controls and [Jetty implement them](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-servlet/src/main/java/org/eclipse/jetty/servlet/Holder.java#L245-L260). However, in Jetty case, the `getInitParameters` just returns the inner `Hashmap`

image

All in all, the H2 console display correctly in my browser 😄. I've added the modifications to the `WhitelistUrlFilter`

```java
public void contextInitialized(final ServletContextEvent event) {
    LOGGER.info("Building H2 servlet");
    ServletContext servletContext = event.getServletContext();
    WebServlet h2WebServlet = new WebServlet();
    ServletRegistration.Dynamic registration = servletContext.addServlet("h2Console", h2WebServlet);
    registration.setInitParameter("webAllowOthers", "false");
    registration.setInitParameter("trace", "");
    registration.addMapping("/h2-console/*");
    registration.setLoadOnStartup(1);
    servletContext.getFilterRegistration("CspFilter").getInitParameters().put("frame-ancestors","'self'");
    Map<String, String> whitelistUrlParams = servletContext.getFilterRegistration("WhitelistUrlFilter").getInitParameters();
    whitelistUrlParams.put("allowUrls", "/index.html,/css,/fonts,/img,/js,/static,/favicon.ico,/api,/metrics,/mirror,/.well-known,/h2-console");
    whitelistUrlParams.put("forwardExcludes", "/api,/metrics,/mirror,/h2-console");
}
```

nscuro: Jetty is violating the spec, that's heresy! 😱

Thanks for investigating, I'll package this in a PR (hopefully) this weekend. If we have native support for the H2 console, the docs in this PR can be reduced to something much more simpler.

syalioune: > Thanks for investigating, I'll package this in a PR (hopefully) this weekend

Don't sweat it ! I'll push a PR later in the day. One less thing for you to worry about 😉

syalioune (approved):

PR #2586, opened 2023-03-09
Minor bugfix for PolicyEngine

mehab:
added bugfix for isPolicyAssignedToProjectTag to scan through all project tags

### Description

Currently the isPolicyAssignedToProjectTag method in the policyEngine class takes the first project tag and compares it with all the policy tags and if there is a match it returns true.
Expected behaviour is that all the tags of a project would be compared with all the tags of a policy and if there is a match this method would return true. This PR provides this fix

### Addressed Issue
Fixes below issue:
Currently the isPolicyAssignedToProjectTag method in the policyEngine class takes the first project tag and compares it with all the policy tags and if there is a match it returns true.

### Checklist


sahibamittal (approved):
Lgtm

nscuro (approved):
Thanks @mehab!

PR #2651, opened 2023-04-07
Include Cloud SQL database connectors

### LINKED ISSUE #2650: Support Cloud SQL connectors
### Current Behavior

Connecting to Cloud SQL over mTLS and with IAM authorization is not supported right now.
This makes it difficult and less secure to run Dependency Track in services like Cloud Run or anywhere else in the GCP ecosystem.

### Proposed Behavior

Adding support for Cloud SQL database connectors would make connecting to Cloud SQL easier and more secure.

https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory

### Checklist

dustin-decker:
### Description

Adds database connectors to facilitate securely connecting to Cloud SQL with IAM and MTLS encryption. This is the easiest way to connect securely when running Dependency Track on GCP.

This includes Cloud SQL for Postgres, MySQL, and SQL Server.

https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory

This socket factory also adds to capability to connect to databases through unix sockets: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/docs/jdbc-postgres.md#connection-via-unix-sockets

### Addressed Issue

Closes #2650 

### Checklist

### DIFF
--- a/docs/_docs/getting-started/database-support.md
+++ b/docs/_docs/getting-started/database-support.md
@@ -75,6 +75,30 @@ alpine.database.url=jdbc:mysql://localhost:3306/dtrack?autoReconnect=true&useSSL
 MySQL may erroneously report index key length violations ("Specified key was too long"), when in fact the multi-byte
 key length is lower than the actual value. **Do not use MySQL if don't know how to work around errors like this**!
 
+#### Cloud SQL
+
+Connecting to Cloud SQL with IAM and mTLS is supported using the Cloud SQL database connectors included.
+
+More information [here](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory)
+
+##### CloudSQL PostgreSQL
+
+```
+jdbc:postgresql:///<DATABASE_NAME>?cloudSqlInstance=<INSTANCE_CONNECTION_NAME>&socketFactory=com.google.cloud.sql.postgres.SocketFactory
+```
+
+##### CloudSQL Microsoft SQL Server
+
+```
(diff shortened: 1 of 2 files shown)


syalioune:
Wouldn't using [Cloud SQL Auth Proxy](https://cloud.google.com/sql/docs/mysql/sql-proxy) be easier and less invasive for Dependency Track ? I don't know much about Cloud Run but for GKE it's quite straightforward (c.f https://cloud.google.com/sql/docs/mysql/connect-kubernetes-engine)

For maintenance sake, I think it shouldn't be Dependency Track responsibility to adapt to each and every deployment context (i.e. embed connectors for all possible public or private cloud).

dustin-decker:
When running in a serverless environment such as Cloud Run there isn't an opportunity to run a side car like Cloud SQL Auth Proxy. But Cloud Run will make a unix socket available to connect to the database. A socket factory must be provided to connect to a unix socket as well. The socket factory in this PR can connect to both a unix socket and directly to the Cloud SQL instance.

https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/docs/jdbc-postgres.md#connection-via-unix-sockets

nscuro:
Including the connectors is fine I think, as we want users to be able to run DT in various cloud settings. They don't interfere with any of the JDBC connectors, so they're not *that* invasive.

The major downside here is that we as maintainers will not be able to test whether these connectors work, and continue to work.

nscuro (approved):

PR #2675, opened 2023-04-15
Add Docker Compose files for simplified local testing

nscuro:
### Description

This PR adds multiple Docker Compose files that simplify local testing in various deployment constellations.

These make it easier to test database migrations and compatibility with the different RDBMSes in general. Also simplifies testing the impact of performance critical functions with Prometheus metrics.

But to be honest the main motivation to add these files is that I always forget how to set up MSSQL locally.

### Addressed Issue

N/A

### Additional Details

Not super-sure where to best put these files.

I don't want them to cause confusion as we already have `src/main/docker/docker-compose.yml`, which is linked on the website, too. As the files in this PR are mainly for local testing, they probably should reside in some sort of `hack` or `dev` directory?

### Checklist


docker-compose.monitoring.yml
```diff
@@ -0,0 +1,65 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) Steve Springett. All Rights Reserved.
+version: "3"
```
robertlagrant, line 17: You probably don't need this - the Compose Specification has deprecated the `version` field. It will just ignore the value.

docker-compose.mssql.yml
```diff
@@ -0,0 +1,60 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) Steve Springett. All Rights Reserved.
+version: "3"
+
+services:
+  apiserver:
+    depends_on:
```
robertlagrant, line 21: This probably needs a `service_healthy` to ensure it waits for mssql to pass its healthcheck before letting the API server loose on it. E.g.
```
  depends_on:
    mssql-init:
      condition: service_completed_successfully
```

And then mssql-init would do the following:
```
  depends_on:
    mssql:
      condition: service_healthy
```

And then mssql would need a healthcheck, which I won't guess at, but [here's an example](https://github.com/Microsoft/mssql-docker/issues/133#issuecomment-1484064993).

nscuro: Thanks for the hint! I was relying in the services just restarting if the service they depend on isn't ready yet. But this is much cleaner!

robertlagrant: No worries! The other thing you could look at is using [`extends`](https://docs.docker.com/compose/extends) for sharing definitions between files, and using [profiles](https://docs.docker.com/compose/profiles) to toggle the use of metrics. Might slightly simplify the commands - e.g. you could say:
 - `docker compose -f docker-compose.postgres.yml up -d` for Postgres
 - `docker compose -f docker-compose.mssql.yml --profile metrics up` for MS SQL with metrics
 - `docker-compose --profile metrics up` for H2 with metrics

Might be a bridge too far, but it wouldn't be hard to change if you fancied it.

nscuro: Will definitely look into it! I used profiles extensively in other projects but never mixed them with overrides.

PR #2676, opened 2023-04-16
Policy Evaluator: add null check to versionMatches

### LINKED ISSUE #2135: NullPointerException during policy evaluation
### Current Behavior:
```
2022-11-03 11:55:42,213 ERROR [LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.length()" because "this.text" is null
        at java.base/java.util.regex.Matcher.getTextLength(Unknown Source)
        at java.base/java.util.regex.Matcher.reset(Unknown Source)
        at java.base/java.util.regex.Matcher.<init>(Unknown Source)
        at java.base/java.util.regex.Pattern.matcher(Unknown Source)
        at org.dependencytrack.policy.CoordinatesPolicyEvaluator.versionMatches(CoordinatesPolicyEvaluator.java:99)
        at org.dependencytrack.policy.CoordinatesPolicyEvaluator.evaluate(CoordinatesPolicyEvaluator.java:65)
        at org.dependencytrack.policy.PolicyEngine.evaluate(PolicyEngine.java:85)
        at org.dependencytrack.policy.PolicyEngine.evaluate(PolicyEngine.java:69)
        at org.dependencytrack.tasks.VulnerabilityAnalysisTask.performPolicyEvaluation(VulnerabilityAnalysisTask.java:130)
        at org.dependencytrack.tasks.VulnerabilityAnalysisTask.inform(VulnerabilityAnalysisTask.java:76)
        at alpine.event.framework.BaseEventService.lambda$publish$0(BaseEventService.java:101)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
```

https://github.com/DependencyTrack/dependency-track/blob/6c6ceadaaa86e1c1da7d8035eeaa1c950b0cb7b4/src/main/java/org/dependencytrack/policy/CoordinatesPolicyEvaluator.java#L98-L102

### Steps to Reproduce:
Might be this policy we have with no version in coordinates field:

image

### Expected Behavior:
No errors.
Having a policy matching all versions should be possible, or proper validation should be done when creating/updating policies.

### Environment:
  
  - Dependency-Track Version: 4.6.1
  - Distribution: Docker
  - BOM Format & Version: n/a
  - Database Server: PostgreSQL
  - Browser: n/a

valentijnscholten:
### Description
Fixes #2135 (at least the original opening exception in there)

### Addressed Issue
Fixes a possible nullpointer during policy evaluation.

### Additional Details
The other `matches` method in this class seems to default to `true` for when the condition or part is empty, so I added that check for `versionMatches` as well. Not sure if this is the best/final/ultimate solution. But I am not actually using policies, just wanted to get rid of this NullPointer that can interfere with processing / policy evaluations. Maybe some checks should be added to not allow policies with empty fields? Or it should be made more clear what empty fields mean and handle them better.

There's another exception mentioned in #2135 around license policy evaluation, but I don't know what should be the result if the license group is empty (true or false).

### Checklist

- [] This PR fixes a defect, and I have provided tests to verify that the fix is effective

### DIFF
--- a/src/main/java/org/dependencytrack/policy/CoordinatesPolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/CoordinatesPolicyEvaluator.java
@@ -85,6 +85,9 @@ public class CoordinatesPolicyEvaluator extends AbstractPolicyEvaluator {
     }
 
     private boolean versionMatches(final PolicyCondition.Operator conditionOperator, final String conditionValue, final String part) {
+        if (conditionValue == null && part == null) {
+            return true;
+        }
         final Matcher versionOperatorMatcher = VERSION_OPERATOR_PATTERN.matcher(conditionValue);
         if (!versionOperatorMatcher.find()) {
             // No operator provided, use default matching algorithm


nscuro:
Thanks @valentijnscholten! Could you add a test case in [`CoordinatesPolicyEvaluatorTest`](https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/policy/CoordinatesPolicyEvaluatorTest.java) so we have this behavior locked in and can detect future regressions?

nscuro:
> There's another exception mentioned in https://github.com/DependencyTrack/dependency-track/issues/2135 around license policy evaluation, but I don't know what should be the result if the license group is empty (true or false).

That error is happening when the license group referred to by the policy condition does not exist, or no longer exists (https://github.com/DependencyTrack/dependency-track/issues/2135#issuecomment-1510402355). 

IMO the correct way to handle that situation is to log a warning about the group not existing, and to not yield a policy violation. I can make this change once this PR is merged, or perhaps you want to give it a go?

valentijnscholten:
> Thanks @valentijnscholten! Could you add a test case in [`CoordinatesPolicyEvaluatorTest`](https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/policy/CoordinatesPolicyEvaluatorTest.java) so we have this behavior locked in and can detect future regressions?

I don't see an easy way to generate a testcase where the `condition` or `part` is null. And testing the private `versionMatches` method also doesn't feel right.

nscuro (approved):

Issue #2361, opened 2023-01-09
Notify when a vulnerability changes


Mvld3r:
### Current Behavior

If a vulnerability changes after it has been created (such as having its severity changed from `Unassigned` to `Critical`), it currently doesn't trigger a notification

### Proposed Behavior

Create a new notification group called `VULNERABILITY_CHANGED`.

Such a notification should be sent when a vulnerability is changed after it has been created, at least for changes that can trigger a change in the risk score (severity, CPE, ...), and list not only that something has changed, but also what changed.

If it is decided to also send notifications for changes in the vulnerability that don't impact the risk score (eg its description), users should be able to opt out of this subset of notifications (eg, having 2 groups : `VULNERABILITY_CHANGED` and `VULNERABILITY_RISK_CHANGED`).

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

prabhushan:
Hi.. any update on this feature

renegrob:
We’re hitting real-world gaps in the current notification system that caused us to miss several critical vulnerabilities:

### Observations / limitations

- `NEW_VULNERABILITY` triggers only when a vulnerability is first seen globally, not when it becomes relevant to an existing project.  
- `NEW_VULNERABLE_DEPENDENCY` only triggers when a new dependency is added, not when an existing dependency gains a newly disclosed CVE.  
- As a result, newly critical vulnerabilities on existing projects can be missed entirely unless a policy violation is manually configured.  

### Suggested improvements

1. **Project-scoped exposure events**  
   - `PROJECT_VULNERABILITY_ADDED`: fires whenever a vulnerability becomes applicable to a project for the first time, regardless of whether it’s new globally.  
   - `PROJECT_VULNERABILITY_SEVERITY_INCREASED`: fires when a vulnerability affecting a project crosses a severity threshold.  

2. **Delta-based risk notifications**  
   - `PROJECT_RISKSCORE_INCREASED` with configurable delta (e.g., ≥ 5)  
   - Captures new vulnerabilities, severity upgrades, exploit maturity changes, and policy impacts  
   - Provides a generic, future-proof way to alert when a project’s overall risk increases  

3. **Optional severity filtering**  
[...]

PR #2730, opened 2023-05-04
Feat: Add notification when severity of a vulnerability changes

Ehoky:
### Description

Feat: Add notification when severity of a vulnerability changes

### Addressed Issue

https://github.com/DependencyTrack/dependency-track/issues/2361

### Checklist


src/main/java/org/dependencytrack/util/NotificationUtil.java
```diff
@@ -378,6 +403,23 @@ public static JsonObject toJson(final NewVulnerabilityIdentified vo) {
         return builder.build();
     }
 
+    public static JsonObject toJson(final VulnerabilityUpdate vo) {
+        final JsonObjectBuilder builder = Json.createObjectBuilder();
+        if (vo.getVulnerabilityUpdateDiff() != null) {
+            builder.add("old severity", vo.getVulnerabilityUpdateDiff().getOldSeverity().toString());
+        }
```
nscuro, line 410: Thinking of future use-cases here, where perhaps we would want to capture more than just a change in severity, WDYT about having an `old` and `new` object instead? This would give us the flexibility to add more properties of interest over time, e.g. CVSS or EPSS scores.

Looking at popular solutions for Change Data Capture, like Debezium, they typically do this, too: https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-update-events

I'd imagine something like:

```json
{
  "component": {
    ...
  },
  "project": {
    ...
  },
  "vulnerability": {
    "old": {
      "severity": "MEDIUM"
    },
    "new": {
      "severity": "HIGH"
    }
}
```

Where `old` and `new` may just as well be `before` and `after`.

From the DT perspective, this form of communicating changes could be re-used in other notification types as well. WDYT?

cc @Mvld3r

src/main/java/org/dependencytrack/util/NotificationUtil.java
```diff
@@ -104,6 +106,29 @@ public static void analyzeNotificationCriteria(QueryManager qm, Vulnerability vu
...
+    public static void analyzeNotificationCriteria(QueryManager qm, Vulnerability vulnerability, VulnerabilityUpdateDiff vulnerabilityUpdateDiff) {
+        final Map<Long,Project> affectedProjects = new HashMap<>();
+        List<Component> components = vulnerability.getComponents();
+        if (components != null) {
+            int i = 0;
+            for (final Component c : components) {
+                i++;
+                if (!affectedProjects.containsKey(c.getProject().getId())) {
+                    affectedProjects.put(c.getProject().getId(), qm.detach(Project.class, c.getProject().getId()));
+                }
```
nscuro, line 118: This computation needs to happen outside of the `for` loop that dispatches notifications.

Otherwise `affectedProjects` will grow with each iteration, resulting in all components other than the very last one to not get the complete set of projects.

docs/_docs/integrations/notifications.md
nscuro, line 1: Please also include an example JSON of the new notification on this page.

melba-lopez:
Thanks for the contributions @Ehoky!  I'm having a hard time visualizing how this would actually work if 100 CVEs get updated.   Is it at a global level? a component level? would you pop up 100 alerts?  

I definitely have been wanting a notification that tells me if something in a project or about a component has changed.  If we think of some BOMs having THOUSANDS of components, or a DT with hundreds of projects, would you inundate the user with too many notifications?  Would there be a separate page to show the changes?

Issue #1905, opened 2022-08-23
DependencyTrack hangs when uploading a large SBOM to a project a second time


JayAtFujifilm:
### Current Behavior:

When we create a new project and upload a large SBOM (~7.5MB) there is no problem. However, if we then upload the SBOM again to the same project, the DepTrack API server hangs and never finishes processing. Inspection of the logfile via Docker indicates a stack overflow error (java.lang.StackOverflowError), as shown in the attached logfile ([DependencyTrackLog.txt](https://github.com/DependencyTrack/dependency-track/files/9409897/DependencyTrackLog.txt)). 

This happens even if we wait a long time (several days) to upload the second SBOM.

For smaller SBOMs we never have this problem. 

### Steps to Reproduce:

Using the DependencyTrack API:

1. Create a new project
2. Upload a large SBOM (~7.5MB)
3. Verify that processing completes and the analysis is shown in the DepTrack UI.
4. Upload the SBOM once again to the previously created project.
5. DepTrack never finishes processing.

### Expected Behavior:

Should be able to upload even large SBOMs multiple times to the same project.

### Environment:
  
  - Dependency-Track Version: 4.5.0
  - Distribution: Docker
  - BOM Format & Version: CycloneDX 1.4 in JSON
  - Database Server: PostgreSQL
  - Browser: Various

### Additional Details:

Unfortunately, we cannot send you the actual SBOM for proprietary reasons.

nscuro:
Wow, that's certainly a first. Ever seen anything like this happening before, @stevespringett?

```
2022-08-08 11:20:15,312 ERROR [LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.StackOverflowError: null
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at org.datanucleus.state.StateManagerImpl.replaceStateManager(StateManagerImpl.java:2096)
	at org.datanucleus.state.StateManagerImpl.initialiseForDetached(StateManagerImpl.java:644)
	at org.datanucleus.state.StateManagerImpl.initialiseForDetached(StateManagerImpl.java:126)
	at org.datanucleus.state.StateManagerImpl.detachCopy(StateManagerImpl.java:4932)
	at org.datanucleus.state.StateManagerImpl.detachCopy(StateManagerImpl.java:126)
	at org.datanucleus.ExecutionContextImpl.detachObjectCopy(ExecutionContextImpl.java:2741)
[...]

PR #2746, opened 2023-05-15
Fix: Detect and avoid infinite recursion during second upload of SBOM with nested duplicate

syalioune:
### Description

When a SBOM contains component with nested duplicate (componentA contains itself as a children), the second SBOM upload causes an infinite recursion.

### Addressed Issue

See #1905 for details

### Additional Details

N/A

### Checklist

- ~[ ] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)~
- ~[ ] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly~

### DIFF
--- a/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
+++ b/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
@@ -23,6 +23,7 @@ import com.github.packageurl.MalformedPackageURLException;
 import com.github.packageurl.PackageURL;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.cyclonedx.model.Bom;
 import org.cyclonedx.model.Dependency;
 import org.cyclonedx.model.Hash;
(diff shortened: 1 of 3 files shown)


syalioune:
> It sounds like a nice find, but I was curious to see the cause of the problem. And I don't understand how it could cause an infinite recursion. When `convert` is called on the child cdx component, it should stop processing there because the child doesn't have any children so `cycloneDxComponent.getComponents()` for the child component would be `null` or empty?
> (Probably not, but I'm not seeing it :-))

Please check my comment here https://github.com/DependencyTrack/dependency-track/issues/1905#issuecomment-1548296828, the scenario is described. Important thing is that the recursion problem is not on the `ModelConverter.convert`  method but on `BomUploadProcessingTask.processComponent`.

> I am also wondering what will happen, with the fix in place, if there is ComponentA having ComponentB as child which has ComponentA as child. So the "recursive" dependency is one level deeper.

In that case we're doomed 😄 . More seriously, the fix should be updated to maintain ancestors component identities instead of just the nearest parent identity.

valentijnscholten:
_Just use Neo4J_ ;-)

nscuro:
This PR is superseded by #3357. 

Still, thanks a lot @syalioune for proposing a fix!

Issue #1211, opened 2021-10-05
Support licenses without spdx ids; add scancode LicenseDB to the list of known Licenses


mawl:
Hi there,

First of all, dependency-track is a great piece of software - thanks for your work. 

### Current Behavior:
Actually only licenses with spdx ids are supported.

### Proposed Behavior:
Support licenses without spdx ids - by adding them manually (with api/ui) and/or by using https://scancode-licensedb.aboutcode.org.

OSS Review Toolkit uses scancode license ids as an additional identifier: https://github.com/oss-review-toolkit/ort/tree/master/spdx-utils/src/main/resources/licenserefs

Best Regards,
Marco

rkg-mm:
@stevespringett Since this is something I need too, to be able to use some custom licenses in policies, how would you expect this to work in its easiest way? Would an API endpoint to add a new license to the existing license-table be enough? From what I see, the LICENSEID is a required column and other code probably relies on it, so one would have to define that ID even though its not the SPDX id. Would that be fine anyway for an easy solution?

stevespringett:
> Would an API endpoint to add a new license to the existing license-table be enough?

I would not want to do this. I'm not sure if there's an SLA, rate limiting, etc in place on their service. I would rather fetch the JSON files from their GitHub repo upon database initialization.

> From what I see, the LICENSEID is a required column and other code probably relies on it, so one would have to define that ID even though its not the SPDX id. Would that be fine anyway for an easy solution?

License.licenseId is not a required field in DT. So we could add the non-SPDX licenses by name only, OR we could use the scancode license IDs that begin with LicenseRef-*

> Since SPDX seems to explicitly allow custom license-ids prefixed by "LicenseRef-", if I understand that correctly, I think that would be a valid solution. Would you agree?

Yes, that would work.

The best way to get support for this incorporated into DT would be through a pull request.

rkg-mm:
@stevespringett not sure if you got my idea, i wasn't thinking about integrating the other license sources. More useful for me is a way to just simply create my own licenses via API. Thats what I was talking about. We just created a PR for this simple solution.

nscuro:
FTR, I think the enhancement implemented in #2153 is a good compromise between wanting more flexibility and keeping the model clean. Admin users will be able to add licenses if desired, but there is no automatism to pollute the table. It does not address the original ask in this ticket though, so I'll leave this open.

PR #2769, opened 2023-05-17
Custom licenses

HagarJNode:
### Description
When uploading CycloneDx created sbom files custom licenses where not matched and updating projects removed manual matched licenses, this is solved here.

### Addressed Issue

Missing custom licenses mapping, as part of this pull:

https://github.com/DependencyTrack/dependency-track/pull/2153

And may be linked to this as well:

https://github.com/DependencyTrack/dependency-track/issues/1211

### Checklist


HagarJNode:
Damn, I think that the process signing my changes did this. I'll see if I can correct it on Monday.

HagarJNode:
@valentijnscholten looking better now?

nscuro:
@HagarJNode Thanks for the PR! 

Just for my clarification: In custom licenses you can also define a license ID. If the license ID is present in the BOM, the current matching logic should work just fine. In your case, the custom licenses do not have a `license.id` field in your BOM, but only a `license.name`, is that correct?

HagarJNode:
> @HagarJNode Thanks for the PR!
> 
> Just for my clarification: In custom licenses you can also define a license ID. If the license ID is present in the BOM, the current matching logic should work just fine. In your case, the custom licenses do not have a `license.id` field in your BOM, but only a `license.name`, is that correct?

@nscuro - yes, that's very correct.

nscuro (approved):

Issue #2693, opened 2023-04-19
Add cvssV2 mapping in Snyk vulnerability parsing


sahibamittal:
### Current Behavior

Snyk parser maps and sets only cvssV3 information for either source NVD/SNYK. 

### Proposed Behavior

It should parse cvssV2 as well in addition to cvssV3 and assign the values in object accordingly based on source priority.
Also, update the latest api version for Snyk and update the mapping if required. 

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #2770, opened 2023-05-17
Add cvssV2 mapping for Snyk

sahibamittal:
### Description

Snyk parser maps and sets only cvssV3 information for either source NVD/SNYK.
It should parse cvssV2 as well in addition to cvssV3 and assign the values in object accordingly based on source priority.
Also, update the latest api version for Snyk API.

### Addressed Issue

https://github.com/DependencyTrack/dependency-track/issues/2693

### Checklist


nscuro:
@sahibamittal Did we ever see Snyk reporting a CVSSv2 score before? Do they even track those?

Also, currently only a single score is selected if multiple are reported. Ideally what should happen is that if both CVSSv3 and CVSSv2 are provided (if that ever happens, see above), we ingest both of them, not just one. We may need to have some prioritization / sorting logic in place to handle this. There is something similar in hyades: https://github.com/DependencyTrack/hyades/blob/main/vulnerability-analyzer/src/main/java/org/hyades/client/snyk/ModelConverter.java#L72-L88

sahibamittal:
I've submitted the clarification for cvssV2 to Snyk team. Awaiting their response.

sahibamittal:
Snyk has confirmed that they do not support CvssV2 in their API.

sahibamittal:
**Closing the PR for now.**

Issue #2737, opened 2023-05-11
Add missing support for supplier and manufacturer


stevespringett:
### Current Behavior

DT does not support supplier on the component or supplier and manufacturer at the metadata level. This is an oversight as `OrganizationalEntity.java` was created for this purpose and is used by CycloneDX service to support providers. But OrganizationalEntity was never added to the BOM or component objects.

### Proposed Behavior

Add support for supplier on the component and supplier and manufacturer at the BOM metadata level.

Considering this is an NTIA minimum requirement, this needs to be added asap.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

stevespringett:
Related to #2373

leec94:
is this completed now with the merged PR? @nscuro

nscuro:
Hey @leec94, I found a few wrinkles while dabbling around locally:

1. [As per CDX spec](https://cyclonedx.org/docs/1.5/json/#components), components have only suppliers, but no manufacturers. The current implementation has both assigned to `Project` and `Component`, but `manufacturer` can never be populated for those.
2. [`metadata.manufacture`](https://cyclonedx.org/docs/1.5/json/#metadata_manufacture) and [`metadata.supplier`](https://cyclonedx.org/docs/1.5/json/#metadata_supplier) in the spec map to `Bom#manufacture` and `Bom#supplier` respectively. The current implementation however assigns `metadata.manufacture` to `Project#manufacture`, and `metadata.component.supplier` and  `metadata.supplier` are ignored entirely.
3. We need changes in the frontend to display this information. [`ComponentDetailsModal`](https://github.com/DependencyTrack/frontend/blob/master/src/views/portfolio/projects/ComponentDetailsModal.vue) and [`ProjectDetailsModal`](https://github.com/DependencyTrack/frontend/blob/master/src/views/portfolio/projects/ProjectDetailsModal.vue) must be extended with a *Supplier* tab. We have something similar for service providers in [`ServiceDetailsModal`](https://github.com/DependencyTrack/frontend/blob/9b44da76dcdf1ada12c90741b04e8238102776d0/src/views/portfolio/projects/ServiceDetailsModal.vue#L33-L58) already.

[...]

PR #2867, opened 2023-07-04
DRAFT: Issue #2737 - Fulfil NTIA minimum requirements

melba-lopez:
### Addressed Issue
#2737 

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/Component.java
+++ b/src/main/java/org/dependencytrack/model/Component.java
@@ -28,9 +28,9 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.github.packageurl.MalformedPackageURLException;
 import com.github.packageurl.PackageURL;
 import org.apache.commons.lang3.StringUtils;
+import org.cyclonedx.model.OrganizationalEntity;
 import org.dependencytrack.model.validation.ValidSpdxExpression;
 import org.dependencytrack.resources.v1.serializers.CustomPackageURLSerializer;
-
 import javax.jdo.annotations.Column;
 import javax.jdo.annotations.Element;
 import javax.jdo.annotations.Extension;
@@ -115,6 +115,20 @@ public class Component implements Serializable {
     @Size(max = 255)
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The publisher may only contain printable characters")
     private String publisher;
+
+    @Persistent /**Issue #2373, #2737 */
+    @Column(name = "MANUFACTURE", jdbcType = "VARCHAR")
+    @Serialized
+    @Size(max = 255)
+    @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The manufacture may only contain printable characters")
+    private OrganizationalEntity manufacture;
+
+    @Persistent /**Issue #2373, #2737 */
+    @Column(name = "SUPPLIER", jdbcType = "VARCHAR")
+    @Serialized
+    @Size(max = 255)
+    @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The supplier may only contain printable characters")
+    private OrganizationalEntity supplier;
 
     @Persistent
     @Column(name = "GROUP", jdbcType = "VARCHAR")
@@ -385,6 +399,13 @@ public class Component implements Serializable {
         this.publisher = publisher;
     }
 
+    public OrganizationalEntity getSupplier() { /**Issue #2373, #2737 */
+        return supplier;
+    }
+
+    public void setSupplier(OrganizationalEntity supplier) {/**Issue #2373, #2737 */
+        this.supplier = supplier;
+    }
     public String getGroup() {
         return group;
     }
--- a/src/main/java/org/dependencytrack/model/Component.java
+++ b/src/main/java/org/dependencytrack/model/Component.java
@@ -117,14 +117,14 @@ public class Component implements Serializable {
     private String publisher;
 
     @Persistent /**Issue #2373, #2737 */
-    @Column(name = "MANUFACTURE", jdbcType = "VARCHAR")
+    @Column(name = "MANUFACTURE", jdbcType = "BLOB")
     @Serialized
     @Size(max = 255)
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The manufacture may only contain printable characters")
     private OrganizationalEntity manufacture;
 
     @Persistent /**Issue #2373, #2737 */
-    @Column(name = "SUPPLIER", jdbcType = "VARCHAR")
+    @Column(name = "SUPPLIER", jdbcType = "BLOB")
     @Serialized
     @Size(max = 255)
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The supplier may only contain printable characters")
--- a/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
+++ b/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
@@ -98,6 +98,7 @@ public class ModelConverter {
         return components;
     }
 
+    /**Convert from CycloneDX to DT */
     @SuppressWarnings("deprecation")
     public static Component convert(final QueryManager qm, final org.cyclonedx.model.Component cycloneDxComponent, final Project project) {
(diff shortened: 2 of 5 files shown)


src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
```diff
@@ -348,6 +349,31 @@ public static org.cyclonedx.model.Metadata createMetadata(final Project project)
...
+                supplier.setName(project.getSupplier().getName());
+            
+                if (project.getSupplier().getUrls() != null) {
+                    supplier.setUrls(project.getSupplier().getUrls());
+                } else {
+                    supplier.setUrls(null);
+                }
+                if (project.getSupplier().getContacts() != null) {
+                    List<OrganizationalContact> contacts = new ArrayList<>();
+                    for (org.cyclonedx.model.OrganizationalContact OrganizationalContact: project.getSupplier().getContacts()) {
```
nscuro, line 364: Probably a typo, but the loop variable should be of type `org.dependencytrack.model.OrganizationalContact`, as you're iterating over supplier contacts of the DT model.

```suggestion
                    for (OrganizationalContact organizationalContact: project.getSupplier().getContacts())) {
```

src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
```diff
@@ -109,8 +109,11 @@ public void inform(final Event e) {
                         if (project.getClassifier() == null) {
                             final var classifier = Optional.ofNullable(cycloneDxBom.getMetadata())
                                 .map(org.cyclonedx.model.Metadata::getComponent)
+                                .map(org.cyclonedx.model.Metadata::getSupplier) /**Issue #2373, #2737 */
+                                .map(org.cyclonedx.model.Metadata::getManufacture)
                                 .map(org.cyclonedx.model.Component::getType)
                                 .map(org.cyclonedx.model.Component.Type::name)
+                                .map(org.cyclonedx.model.Component::getSupplier) /**Issue #2373, #2737 */
                                 .map(Classifier::valueOf)
                                 .orElse(Classifier.APPLICATION);
```
nscuro, line 118: This entire statement is only intended to extract the component's type, and map it to DT's data model (`Classifier`).

You can think of the `Optional.ofNullable(...).map(...).map(...)` statement like this, but null-safe:

```java
String cdxComponentType = cycloneDxBom.getMetadata().getComponent().getType().name();

Classifier dtClassifier;
if (cdxComponentType == null) {
    dtClassifier = Classifier.APPLICATION; // Fallback when the component has no type
} else {
    dtClassifier = Classifier.valueOf(cdxComponentType);
}

project.setClassifier(dtClassifier);
```

Thus, you should remove the `getSupplier` and `getManufacture` here, as they'll cause compilation errors.

Instead, you'll need a similar logic to what you wrote in `ModelConverter`, but instead of setting the resulting `OrganizationalEntity` on a `Component`, you'll have to set it on `project`. 

It could make sense to move the `OrganizationalEntity` mapping logic into a dedicated method in `ModelConverter`; That way you can reuse it.

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
```diff
@@ -348,6 +349,31 @@ public static org.cyclonedx.model.Metadata createMetadata(final Project project)
...
+                        OrganizationalContact contact = new OrganizationalContact();
+                        contact.setName(OrganizationalContact.getName());
+                        contact.setEmail(OrganizationalContact.getEmail());
+                        contact.setPhone(OrganizationalContact.getPhone());
+                        contacts.add(contact);
+                    }
+                    supplier.setContacts(contacts);
+                }
+                project.setSupplier(supplier);
```
nscuro, line 373: `createMetadata` is converting *from* DT's data model *to* CycloneDX, so it should not set anything on DT's `project`:

```suggestion
                cycloneComponent.setSupplier(supplier);
```

src/test/java/org/dependencytrack/tasks/BomUploadProcessingTaskTest.java
```diff
@@ -145,6 +145,7 @@ public void informTest() throws Exception {
         assertThat(components).hasSize(1);
 
         final Component component = components.get(0);
+        assertThat(component.getSupplier()).isEqualTo("Example Incorporated"); /*Issue #2373, #2737 */
```
nscuro, line 148: `getSupplier()` will return an object of type `OrganizationalEntity`, which you cannot compare with a string.

It should be something along the lines of:

```java
assertThat(component.getSupplier().getName()).isEqualTo("Example Incorporated");
assertThat(component.getSupplier().getContacts().get(0).getName()).isEqualTo("Foo");
assertThat(component.getSupplier().getContacts().get(0).getEmail()).isEqualTo("foo@bar.com");
// etc.
```

For supplier data to be present here, it needs to be added to the BOM this test is using first: https://github.com/DependencyTrack/dependency-track/blob/master/src/test/resources/bom-1.xml

nscuro (commented):
Hey @melba-lopez, added some comments to the changes made so far.

So far you've covered supplier data in `metadata.component` of the BOM, which corresponds to a project's supplier in DT.

For the component side of things in DT, you want to look at these methods:

* Convert from CycloneDX to DT: https://github.com/DependencyTrack/dependency-track/blob/561ef18d4480f987cfb492748a2be85c1d6cec94/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java#L98
* Convert from DT to CycloneDX again: https://github.com/DependencyTrack/dependency-track/blob/561ef18d4480f987cfb492748a2be85c1d6cec94/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java#L209

Happy to help if you need more clarification or get stuck!

melba-lopez:
Thanks for all the feedback @nscuro def will take all the help i can get :) hopefully i can finally start fixing this tomorrow evening.

src/main/java/org/dependencytrack/model/Component.java
```diff
@@ -115,6 +116,12 @@ public enum FetchGroup {
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The publisher may only contain printable characters")
     private String publisher;
 
+    @Persistent /**Issue #2373, #2737 */
+    @Column(name = "SUPPLIER", jdbcType = "VARCHAR")
+    @Size(max = 255)
+    @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The supplier may only contain printable characters")
+    private OrganizationalEntity supplier;
```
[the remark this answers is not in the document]
melba-lopez: I believe i fixed per your recommendation.  Will commit shortly.

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
melba-lopez: made the change :-) will commit shortly.

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
melba-lopez: made the change.  will commit shortly :)

src/test/java/org/dependencytrack/tasks/BomUploadProcessingTaskTest.java
melba-lopez: I added all fields for manufacture and supplier in this.  Committing shortly :)

src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
melba-lopez: removed code.  and created manufacture logic in the ModelConverter.  In the interest of time, and lack of understanding ;), ive not made it into a single method yet.

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
melba-lopez: when i make this change it is complaining for some reason 😭

melba-lopez:
@nscuro i made all the changes and had to fix/undo some changes i made accidentally.  not sure why it does not like your recommendation `cycloneComponent.setSupplier(supplier);`

nscuro:
This was taken over by @leec94 in #3090. Thanks @melba-lopez for your work on this! :)

Issue #2785, opened 2023-05-24
DT memory consumption grows over time and eventually CPU suddenly goes permanently to 100%


flemminglau:
### Current Behavior

We are seeing after some weeks that the memory consumption of DT gradually grows beyond 10Gb.
We also just had to restart DT as the APIServer had stopped responding in a sensible time and consuming 100% CPU.

We were also seeing timeout communicating w. the postgres server but probably this was a side effect of the lack of available CPU cycles.

### Steps to Reproduce

1. Load 107 projects
2. Leave DT running for 3 weeks uploading BOM updates for existing projects a few times a day.
3. Run scripts once a day exporting BOMs and VEXs.

### Expected Behavior

Memory consumption remains limited/constant

CPU consumption stays reasonable and never jumps to 100% permanently

### Dependency-Track Version

4.8.0

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Browser

Google Chrome

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

nscuro:
Are you [monitoring](https://docs.dependencytrack.org/getting-started/monitoring/) your instance? Any metrics that jump out in particular?

Also, how much memory did you allocate for the API server?

flemminglau:
Unfortunately no.

We have not seen the need till now.

```
  deploy:
     resources:
       limits:
         memory: 12288m
       reservations:
         memory: 8192m
```
       
so allocation is 12Gb max.

We are not attempting to limit the JVM memory consumption.

nscuro:
Also, what's the resource allocation for your database server? And did you configure database connection pooling to non-default values? https://docs.dependencytrack.org/getting-started/database-support/#connection-pooling

flemminglau:
It is not so much the fact that a lot of ram is consumed that worries us.
More the incident of suddenly consuming all available CPU resources.

I have indeed due to some past DB issue updated the DB pool setting in my compose file to:

```
      ALPINE_DATABASE_POOL_MAX_SIZE: 40
      ALPINE_DATABASE_POOL_MIN_IDLE: 10
``` 

In the logs I see a few error messages logged. Sometimes 2-3 of these:
```
[...]

PR #2872, opened 2023-07-05
Fix memory leak issue in policy evaluation

Shawyeok:
### Description

In `org.dependencytrack.policy.AbstractPolicyEvaluator`, there is a `QueryManager` created while instance initialization, https://github.com/DependencyTrack/dependency-track/blob/a7e512235c33b650decc90a8c12ceedc5234c6d9/src/main/java/org/dependencytrack/policy/AbstractPolicyEvaluator.java#L37-L41

and later overwrite by it's setter which is called in `VulnerabilityAnalysisTask` execution, therefor the older instance of `QueryManager` will never get closed, so it leads to a memory leak.
https://github.com/DependencyTrack/dependency-track/blob/a7e512235c33b650decc90a8c12ceedc5234c6d9/src/main/java/org/dependencytrack/policy/PolicyEngine.java#L88

### Addressed Issue

https://github.com/DependencyTrack/dependency-track/issues/2785#issuecomment-1621497318

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/policy/AbstractPolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/AbstractPolicyEvaluator.java
@@ -34,7 +34,7 @@ import java.util.stream.Collectors;
  */
 public abstract class AbstractPolicyEvaluator implements PolicyEvaluator {
 
-    protected QueryManager qm = new QueryManager();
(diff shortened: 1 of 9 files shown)


nscuro (approved):
Thanks @Shawyeok!

PR #2873, opened 2023-07-05
Fix memory leak issue in vex upload processing task

Shawyeok:
### Description

It's a follow up PR with #2872 , there another case may cause memory leak.

In `VexUploadProcessingTask`, there is a `QueryManager` never get call it's `close` method, it could also leads memory leak.
https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/tasks/VexUploadProcessingTask.java#L62

### Checklist


nscuro (approved):
Another great find, thanks @Shawyeok!

PR #2876, opened 2023-07-06
Support authentication for public / non-internal repositories

### LINKED ISSUE #2870: Implement global authentication for repositories
### Current Behavior

Currently, authentication is only limited to internal components in a repository. Some enterprises may not be allowed to connect directly to external repositories for external packages. In this case they would require credentials to login to artifactory / nexus repo to get access to the external package via this route.
Because authentication is only allowed for internal packages, user will not be allowed to enter credentials for retrieving external packages and thus would not be able to proceed.

### Proposed Behavior

The authentication mechanism should be a global setting for internal and external packages. This should be optional when putting in the repository details. If authentication is required, user will select the authentication required button and put in the credentials as known. Then these credentials would be used to access the repo to download packages (whether external or internal). 

### Checklist

mehab:
### Addressed Issue

Closes #2870

### Checklist


src/main/java/org/dependencytrack/model/Repository.java
```diff
@@ -86,6 +86,12 @@ public class Repository implements Serializable {
     @NotNull
     private Boolean internal; // New column, must allow nulls on existing databases
 
+    //New column to determine if authentication is required for a repository
+    @Persistent
+    @Column(name = "AUTHENTICATIONREQUIRED")
+    @NotNull
```
nscuro, line 92: As this is a new column, it should be nullable to accommodate for existing repositories that have no value set.

```suggestion
    @Column(name = "AUTHENTICATIONREQUIRED", allowsNull = "true")
```

Alternatively, we'll need an [upgrade](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade) that creates and populates this column for existing repositories.

src/main/java/org/dependencytrack/persistence/QueryManager.java
nscuro, line 1: Can we revert the whitespace changes? It's just adding unnecessary noise and makes it harder to spot what was meaningfully changed.

nscuro (changes requested):

src/main/java/org/dependencytrack/tasks/repositories/RepositoryMetaAnalyzerTask.java
```diff
@@ -155,21 +155,24 @@ private void analyze(final QueryManager qm, final Component component, final IMe
...
-                    LOGGER.debug("Building repository Metamodel from cache for "+purl);
+                    LOGGER.debug("Building repository Metamodel from cache for " + purl);
                     model.setLatestVersion(StringUtils.trimToNull(cac.getResult().getString(LATEST_VERSION)));
                     model.setPublishedTimestamp(Date.from(Instant.ofEpochMilli(cac.getResult().getJsonNumber(PUBLISHED_TIMESTAMP).longValue())));
                 } else {
                     LOGGER.debug("Analyzing component: " + component.getUuid() + " using repository: "
                             + repository.getIdentifier() + " (" + repository.getType() + ")");
 
-                    if (repository.isInternal()) {
+                    if (Boolean.TRUE.equals(repository.isAuthenticationRequired())) {
```
nscuro, line 165: For existing installations, keep in mind that `isAuthenticationRequired` will be `null`, but `isInternal` might be `true` (again, unless we add a migration that flips `isAuthenticationRequired` to `true` for all repos that are internal and have credentials configured).

```suggestion
                    if (Boolean.TRUE.equals(repository.isInternal()) || Boolean.TRUE.equals(repository.isAuthenticationRequired())) {
```

nscuro:
Apologies @mehab, I created a merge conflict here by merging #3112 (which incidentally benefits from the changes you made here!). May I ask you to fix the conflicts one last time? Thanks!

PR #2942, opened 2023-08-08
Add "Show in Dependency-Graph" Button in "Affected Projects" List

rbt-mm:
### Description

This PR adds the `Show in Dependency-Graph` button to the every project in the `Affected Projects` tab of a vulnerability, but only if the affected project has a dependency graph.
Clicking the button redirects the user to the projects dependency graph and highlights the affected component.

### Addressed Issue

DependencyTrack/frontend#533

### Additional Details

Frontend PR: DependencyTrack/frontend#573

### Checklist

### DIFF
--- a/.github/workflows/_meta-build.yaml
+++ b/.github/workflows/_meta-build.yaml
@@ -25,7 +25,7 @@ jobs:
         uses: actions/checkout@v3.5.3
 
       - name: Set up JDK
-        uses: actions/setup-java@v3.11.0
+        uses: actions/setup-java@v3.12.0
         with:
           distribution: 'temurin'
           java-version: '17'
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -48,7 +48,7 @@ jobs:
         uses: actions/checkout@v3.5.3
 
       - name: Set up JDK
-        uses: actions/setup-java@v3.11.0
+        uses: actions/setup-java@v3.12.0
         with:
           distribution: 'temurin'
           java-version: '17'
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -32,7 +32,7 @@ jobs:
         uses: actions/checkout@v3.5.3
 
       - name: Set up JDK
-        uses: actions/setup-java@v3.11.0
+        uses: actions/setup-java@v3.12.0
         with:
           distribution: 'temurin'
           java-version: '17'
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -1,6 +1,6 @@
 FROM eclipse-temurin:17.0.7_7-jre-focal@sha256:20534a45cc565ac436593164c7f986f9499aa981b5bb8d936a11fcc25e41c55a AS jre-build
 
-FROM debian:stable-slim@sha256:bfb338bbc84031a32bb0d4ba61025728ac725968b562707716c0fcc42775d26b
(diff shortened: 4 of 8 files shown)


rkg-mm:
@nscuro  any reason for not merging this?

nscuro:
@rkg-mm No particular reason other than me overlooking it. I'll have a look in the coming days. Thanks for the ping.

rkg-mm:
@nscuro  ok thanks, just wanted to make sure we didn't miss something  :D

nscuro:
I think there is a caveat here:

The modified endpoint returns projects affected by a given vulnerability. It is possible that multiple components in a project are affected, but this implementation does not cover this case and will only ever yield a single component per project.

Few options I see:

* We return the UUIDs of *all* affected components for a project. Clicking the button in the UI will then go on to highlight all of them in the graph.
* We keep it at a single component per project, but clearly indicate that it's only one of potentially multiple. In this case, we need to make sure that it's deterministically the same component UUID when requesting the endpoint multiple times.

rkg-mm:
@nscuro  you are right. Which solution do you prefer?

nscuro:
@rkg-mm I would love to see the former option, but have not looked into how much effort it is to support this in the dependency graph view.

rbt-mm:
@nscuro I think these are probably all the steps necessary to implement the new solution:

**Backend**
- Return all the UUIDs of all affected components for a project in the `Vulnerability` API endpoint.
- Modify [the method to get the dependency graph of a component](https://github.com/DependencyTrack/dependency-track/blob/c84abd77ff913334b077315f6639e241984865da/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java#L631-L680) to accept an array of components instead of just a single component.
- Modify [the SQL statement](https://github.com/DependencyTrack/dependency-track/blob/c84abd77ff913334b077315f6639e241984865da/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java#L637) to get the parent nodes of every component in the array.
    - This will create a new possible case where a searched UUID can also be the parentNode of another searched UUID. I don't think that this case will create problems, but a new test should probably be written to cover it.

**Frontend**
- Clicking the `Show in dependency graph` button must redirect to a project's dependency graph and the URL must contain the UUID of every component affected by the Vulnerability.
    - The different UUIDs in the URL need to be separated by some kind separator (e.g. `/<projectUUID>/dependencyGraph/<componentUUID1>|<componentUUID2>`) 
    - The maximum length of a URL in the address bar is 2048 characters so we shouldn't run into any problems here.
- [The code that sets the labels of the dependency graph nodes](https://github.com/DependencyTrack/frontend/blob/1ef540563d8eebc02ead5fbe8031518cce2564f4/src/views/portfolio/projects/ProjectDependencyGraph.vue#L378C1-L384C7) must be modified to check for every UUID in the URL so that every node that contains any of these UUIDs can be highlighted.
- The current message, which is displayed when the UUID in the URL cannot be found in the dependency graph, only says `Dependency could not be found in dependency graph`. Maybe for multiple UUIDs in the URL, this needs to be changed to specify which of the UUIDs couldn't be found in the dependency graph.

I think this would probably need 1 or 2 days to implement, but unfortunately I currently cannot allocate my time to implement this myself.

nscuro:
Thanks for those details @rbt-mm, very helpful! We also can't work on this within the 4.10 milestone, but should be able to in 4.11 if needed. I'm thus pushing this to 4.11.

rkg-mm:
@nscuro  I made the discussed improvements. I wasn't able to push on the original branch from @rbt-mm so I created a new PR which replaces this here: https://github.com/DependencyTrack/dependency-track/pull/3285

I also improved the performance of the tree by a lot and refactored some code to simplify it.

Issue #2519, opened 2023-02-23
Believe this is an issue ... findings change with each scan of same SBOM 


charbl:
### Current Behavior

So, we have a program using our Jenkins. They produce an SBOM using CycloneDX integrated into their build process. Actually, it's 3 SBOMS and then they use the ClycloneDX CLI to merge into one SBOM. Long story but they report the results into a parent project in DependencyTrack and we asked them to report the results into a child project. They did rerun the job and pointed to the child project. The rerun created an SBOM it was submitted to the child project. The child project reported diff results (components and vulnerabilities) ... this raised the question ... why? We ran the Jenkins job again (publishing to the child project) and the numbers changed again ... they started to align closer to the parent project numbers (still have not removed the parent analysis yet). We started to question the SBOM that were generated from the two different builds, we pulled them and did a compare .... identical aside from the metadata which had different serial numbers.

So, to really test this out, we created a dummy Jenkins job that just report the same SBOM to two different test jobs in DependencyTrack (two projects under parent project). Ran test for project one, sending over the SBOM to test project one and looked in DependencyTrack. The project one was now visible, and we compared the reported numbers to the original project ...
This is what we expected:
Components - 1749
Vulnerabilities: Crti: 11, High: 33, Med 28
Risk Score: 364

This is waht we got on first submission:
Components - 2498
Vulnerabilities: Crti: 23, High: 48, Med 37
Risk Score: 725

Now this was kind of shocking, we poked around and double checked settings and the SBOM diff.
We decided to run our test job and report to test job two .... here are the results we got:

This is waht we got on first submission:
Components - 2312
Vulnerabilities: Crti: 28, High: 43, Med 35
Risk Score: 745

Again, shocking that these two were different. And again, double check all .... same conclusion .... config looks good.
We then decided to run our two jobs and re-report the same SBOM to their DependencyTrack projects ... that already had data!

both runs now reported new numbers below for the second submission:

Job 1
Components - 1764
Vulnerabilities: Crti: 12, High: 35, Med 29
Risk Score: 387

Job 2

Components - 1754
Vulnerabilities: Crti: 11, High: 33, Med 28
Risk Score: 364

They pulled closer to the expected but still different .... I can't explain. Same SBOM ....

Thoughts? Have read all the documentation and are we doing something wrong?

### Steps to Reproduce

1. Create job in Jenkins and use the DepTrack plugin to publish SBOM (use Sync publish)
[...]

PR #2956, opened 2023-08-15
Fix #2955: Import components that are located under 'metadata.component.components'

### LINKED ISSUE #2955: Components in the SBOM under 'metadata.component.components' are not imported
### Current Behavior

I generate an SBOM from a multi-module Gradle project, using cdxgen. When I import this, my modules are not imported as components and therefore the dependency tree doesn't work.

This is most likely because cdxgen puts the sub-projects under 'metadata.component.components' instead of directly under 'components'. (See [the specs for 1.4](https://cyclonedx.org/docs/1.4/json/#metadata_component_components))

### Steps to Reproduce

1. Generate an SBOM for a Gradle project with cdxgen
2. Import the SBOM

### Expected Behavior

I expect my modules to be added just as any other components and the dependency tree to be available.

### Dependency-Track Version

4.8.2

Container Image

### Database Server

N/A

### Database Server Version

_No response_

### Browser

Mozilla Firefox

### Checklist

malice00:
### Description

Certain SBOM-generators save sub-projects or -modules as part of the component, instead of in the list of dependencies. This change makes sure to also import these as components, so a correct dependency-tree can be generated.

### Addressed Issue

Fixes #2955

### Checklist

- [X] This PR fixes a defect, ~and I have provided tests to verify that the fix is effective~

Testing was done manually, since no tests for this part of the code exists and I wasn't sure on how to add them there.

### DIFF
--- a/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
+++ b/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
@@ -92,6 +92,14 @@ public class ModelConverter {
                 }
             }
         }
+        if (bom.getMetadata() != null && bom.getMetadata().getComponent() != null && bom.getMetadata().getComponent().getComponents() != null) {
+            for (int i = 0; i < bom.getMetadata().getComponent().getComponents().size(); i++) {
+                final org.cyclonedx.model.Component cycloneDxComponent = bom.getMetadata().getComponent().getComponents().get(i);
+                if (cycloneDxComponent != null) {
+                    components.add(convert(qm, cycloneDxComponent, project));
+                }
+            }
+        }
         return components;
     }
 
--- a/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
+++ b/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
@@ -84,23 +84,32 @@ public class ModelConverter {
      */
     public static List<Component> convertComponents(final QueryManager qm, final Bom bom, final Project project) {
         final List<Component> components = new ArrayList<>();
-        if (bom.getComponents() != null) {
-            for (int i = 0; i < bom.getComponents().size(); i++) {
-                final org.cyclonedx.model.Component cycloneDxComponent = bom.getComponents().get(i);
-                if (cycloneDxComponent != null) {
-                    components.add(convert(qm, cycloneDxComponent, project));
-                }
-            }
-        }
-        if (bom.getMetadata() != null && bom.getMetadata().getComponent() != null && bom.getMetadata().getComponent().getComponents() != null) {
-            for (int i = 0; i < bom.getMetadata().getComponent().getComponents().size(); i++) {
-                final org.cyclonedx.model.Component cycloneDxComponent = bom.getMetadata().getComponent().getComponents().get(i);
-                if (cycloneDxComponent != null) {
-                    components.add(convert(qm, cycloneDxComponent, project));
-                }
-            }
+        final List<String> purls = new ArrayList<>();
+        components.addAll(convertComponents(qm, bom.getComponents(), project, purls));
+        if (bom.getMetadata() != null && bom.getMetadata().getComponent() != null) {
+            components.addAll(convertComponents(qm, bom.getMetadata().getComponent().getComponents(), project, purls));
         }
         return components;
+    }
+
+    private static List<Component> convertComponents(final QueryManager qm, final List<org.cyclonedx.model.Component> components,
+            final Project project, final List<String> purls) {
+        final List<Component> convertedComponents = new ArrayList<>();
+        if (components != null) {
+            for (org.cyclonedx.model.Component component : components) {
+                if (component != null) {
+                    boolean hasPurl = component.getPurl() != null && !component.getPurl().isBlank();
+                    if (hasPurl && purls.contains(component.getPurl())) {
+                        continue;
+                    }
+                    convertedComponents.add(convert(qm, component, project));
+                    if (hasPurl) {
+                        purls.add(component.getPurl());
+                    }
+                }
+            }
+        }
+        return convertedComponents;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
```diff
@@ -92,6 +92,14 @@ public static List<Component> convertComponents(final QueryManager qm, final Bom
                 }
             }
         }
+        if (bom.getMetadata() != null && bom.getMetadata().getComponent() != null && bom.getMetadata().getComponent().getComponents() != null) {
```
prabhu, line 95: As per the spec, such nesting is allowed even with component.components.

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
```diff
@@ -92,6 +92,14 @@ public static List<Component> convertComponents(final QueryManager qm, final Bom
...
+            for (int i = 0; i < bom.getMetadata().getComponent().getComponents().size(); i++) {
+                final org.cyclonedx.model.Component cycloneDxComponent = bom.getMetadata().getComponent().getComponents().get(i);
+                if (cycloneDxComponent != null) {
+                    components.add(convert(qm, cycloneDxComponent, project));
```
prabhu, line 99: Can we track the bom-ref to avoid duplicate components?

malice00:
I've been checking [the specs](https://cyclonedx.org/docs/1.4/json/#components_items_bom-ref) again:
1) a bom-ref is not mandatory
2) if used, it **must** be unique (although neither dependency-track nor cdxgen actually (currently) validate this)

So, yes, we could, but only if they are used and are not unique.
Maybe it's better to check the `purl` instead? Unfortunately, it's also not mandatory, but it also doesn't need to be unique!

prabhu:
> I've been checking [the specs](https://cyclonedx.org/docs/1.4/json/#components_items_bom-ref) again:
> 
>     1. a bom-ref is not mandatory
> 
>     2. if used, it **must** be unique (although neither dependency-track nor cdxgen actually (currently) validate this)
> 
> 
> So, yes, we could, but only if they are used and are not unique. Maybe it's better to check the `purl` instead? Unfortunately, it's also not mandatory, but it also doesn't need to be unique!

What do you think about the value returned by the convert method? Does it include a bom-ref?

malice00:
> What do you think about the value returned by the convert method? Does it include a bom-ref?

From what I can tell, the convert method simply copies the bom-ref it received from the SBOM.

During the conversion, DT does check if a certain component already exists for the project, however components are only persisted after the whole SBOM has been read and converted. This check is done on any of several fields (purl, swid, cpe, group/name/version-combo), which we could maybe do when converting as well?

Anyway, I just tested importing the same SBOM again and now there are no more duplicates! This is similar to something I had [already discussed](https://github.com/CycloneDX/cyclonedx-cocoapods/issues/52) with the devs of [cyclonedx-cocoapods](https://github.com/CycloneDX/cyclonedx-cocoapods). It's not 100% the same problem, but there it was decided to document this behavior, as it was not actually incorrect.

malice00:
I implemented a simple duplicate-check on `purl` -- if it is set.

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
```diff
@@ -84,15 +84,32 @@ private ModelConverter() { }
...
+    }
+
+    private static List<Component> convertComponents(final QueryManager qm, final List<org.cyclonedx.model.Component> components,
+            final Project project, final List<String> purls) {
+        final List<Component> convertedComponents = new ArrayList<>();
+        if (components != null) {
+            for (org.cyclonedx.model.Component component : components) {
+                if (component != null) {
+                    boolean hasPurl = component.getPurl() != null && !component.getPurl().isBlank();
+                    if (hasPurl && purls.contains(component.getPurl())) {
```
prabhu, line 102: Since purl is also optional, we might also need bom-ref and CPE-based checks. I suggest waiting for a response from DT developers before making this change.

nscuro:
@malice00 We recently [re-worked](https://github.com/DependencyTrack/hyades-apiserver/pull/218) the BOM ingestion area in [hyades-apiserver](https://github.com/DependencyTrack/hyades-apiserver) and will backport that soon to this code base.

Components are flattened prior to de-duplication: https://github.com/DependencyTrack/hyades-apiserver/blob/23725258b8681ada8a5fc605a06dd45350fd83ef/src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java#L213-L214 I believe this is where your change could be implemented as well.

We de-duplicate based on `ComponentIdentity`, which includes the coordinates (`group`, `name`, `version`), CPE, and PURL (and later in the process also the UUID). We also have logic in place to ensure that we do not break the dependency graph by de-duplicating: https://github.com/DependencyTrack/hyades-apiserver/blob/23725258b8681ada8a5fc605a06dd45350fd83ef/src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java#L194-L205

There was also a relevant change to how existing components are looked up in the DB (see https://github.com/DependencyTrack/dependency-track/issues/2519#issuecomment-1621889555); We made this more strict.

It may be easier to incorporate your proposed change to that logic, rather than retrofitting it onto the existing one. If you can, maybe look at the changes we made and let us know if you think your requirement can be integrated there.

malice00:
@nscuro Might take me until the weekend to get to it, but I'll give it a look!

malice00:
@nscuro Nice clean codebase, looks like I can integrate very easily! Now for a test to make sure that integration actually worked... :-)

nscuro:
Hey @malice00, apologies for the long wait. Your contribution, along with all the other enhancements we made in Hyades, is now finally being backported to Dependency-Track via #3357. Thus, I'll close this PR. Thanks again for your work on this!

Issue #2018, opened 2022-10-06
Add license text to BOM export 


wollepb:
This enhancement request is created based on https://github.com/DependencyTrack/dependency-track/issues/1990#issuecomment-1270346726

### Current Behavior:
The Json downloaded from DependencyTrack via the API endpoint `/api/v1/bom/cyclonedx/project/{uuid}`, does not contain the license text anymore. The license text is available in the input bom.xml, within DependencyTrack but not in the downloaded Json.

### Proposed Behavior:
There is an option to include license texts in the exported Json. If activated, the exported Json contains the license texts.

Issue #371, opened 2019-06-13
Add License Permissions, Restrictions, and Limitations


stevespringett:
Leverage GitHub License API and/or choosealicense.com to retrieve a list of permissions, restrictions, and limitations for all supported SPDX license IDs.

https://developer.github.com/v3/licenses/#get-an-individual-license
https://github.com/github/choosealicense.com/tree/gh-pages/_licenses

The list of permissions, restrictions, and limitations, should be displayed on the individual license page as well as the license auditing work that will be included in #249.

A ticket was created on the SPDX project to add this metadata to their XML master list.
https://github.com/spdx/license-list-XML/issues/881

stevespringett:
Another option is to simple get all licenses supported by GitHub via GraphQL:

```graphql
{ 
  licenses {
    conditions {
      description
      key
      label
    }
    description
    featured
    hidden
    id
    implementation
    key
    limitations {
      description
      key
      label
    }
    name
    nickname
    permissions {
      description
      key
      label
    }
    pseudoLicense
    spdxId
    url
  }
}
```

Although, at this time, GitHub has only analyzed 13 licenses, so its usefulness may be limited.

PR #2963, opened 2023-08-17
Added license export

### LINKED ISSUE #651: Add license attestation report
A license attestation report is a document which contains the information of OSS components used in a specific product. Usually, it identifies following information of each OSS: name, version, copyright notice, licensed under which license and the full license text.

From DT's view, I think that a specific product is represented by the project and its version combination.

This document will then be distributed with the product to fulfill the requirement of software license compliance.

It will be great to be able to download this document from DT directly.

DerBene:
### Description

Add licence attestation report functionality

### Addressed Issue

closes #651 

### Additional Details

Added it, even though we are not 100% sure if thats the right place to put it. Unit tests and documentation are missing and will be added once it's clear where the changes should be in.

### Checklist


nscuro:
Thanks for the PR @DerBene!

I am torn on the export format. I think the format should be CycloneDX, matching the other export variants. The complete license text can be included there as well, refer to https://cyclonedx.org/use-cases/#license-compliance

I don't believe DT should export to non-standard formats, especially when they're not machine-readable.

CC @stevespringett; I attached an example export: [7169d274-0cbe-4933-899d-5a6a47c2d729-license-attestation.txt](https://github.com/DependencyTrack/dependency-track/files/13267042/7169d274-0cbe-4933-899d-5a6a47c2d729-license-attestation.txt)

troy256:
Since the CycloneDX spec allows for a "licenses" attribute for each component, is there a way to include the component licenses in the BOM export?  This would be really helpful instead of having to export a completely different license document.

E.g. - 

```
{
  "version": "1.5",
  "components": [
    {
      "name": "some-software-library",
      "version": "1.0.0",
      "license": [
        "Apache-2.0"
      ]
    }
  ]
}

```
https://cyclonedx.org/docs/1.5/json/#components_items_licenses

nscuro:
@troy256 Licenses are already part of the "Inventory" BOM export. Are you encountering situations where this is not the case?

andife:
Maybe, I don't fully understand it. What is an "Inventory" BOM export? My status of knowledge is the the "license text" is not included in the export (https://github.com/DependencyTrack/dependency-track/issues/2018)

nscuro:
I fear I am not entirely clear what the requested feature is here. License data (except the entire license text, as pointed out by @andife) is already included in the inventory export. This PR does not add anything beyond that, but instead adds a new non-standard export format.

We can offer a separate download option that includes inventory *and* complete license texts, to address #2018. But from what I understand so far, #651 is already implemented via inventory export.

Removing from the 4.10 milestone because it needs clarification.

DerBene:
I don't quite know on how to proceed now. Both options (CycloneDX formatted export and the current implementation) are viable for us. What is the preferred way?

stevespringett:
Lawyers typically like to have a large text file, pdf, or similar of all basic information such as the component name and version, full license text, and copyright statements. Some additionally like to have evidence of other licenses and copyrights found in a component but may not be directly applied to a component. CycloneDX supports this, but I do not believe DT can import that yet.

Having a human readable format is fine IMO. It satisfies a typical legal use case. @MarkusGeigerDev brings up a valid point however in that there should ideally be a way to include and optionally filter on the components that have licenses that have obligations. This additional capability is captured in ticket #371.

I see this as a good MVP, but there should be a plan on implementing #371 shortly thereafter.

PR #2967, opened 2023-08-20
Fix `AffectedComponent` format for CPEs with version ranges

nscuro:
### Description

For CPEs with ranges, `version` is set to a non-empty value (mostly the wildcard `*`).

For such cases, the API would previously return a JSON object that caused the frontend to render it as e.g. `cpe:2.3:o:apple:tvos:*:*:*:*:*:*@*`, whereas it should've been `cpe:2.3:o:apple:tvos:*:*:*:*:*:* (|<15.6)` instead.

Screenshot 2023-08-20 221000

Screenshot 2023-08-20 220857

### Addressed Issue

This issue was indirectly reported in Slack: https://owasp.slack.com/archives/C6R3R32H4/p1692113964318869

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/vo/AffectedComponent.java
+++ b/src/main/java/org/dependencytrack/resources/v1/vo/AffectedComponent.java
@@ -96,15 +96,18 @@ public class AffectedComponent {
                 LOGGER.warn("Error assembling PURL", e);
             }
         }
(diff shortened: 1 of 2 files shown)


leec94 (approved):
tested and verified the changes appear. looks good!

PR #2980, opened 2023-08-23
Fix version distance policy being evaluated despite not being configured

### LINKED ISSUE #2979: Version distance policy is evaluated despite not being configured, and errors in distance calulcation are not handled
### Current Behavior

Originally reported via Slack: https://owasp.slack.com/archives/C6R3R32H4/p1692794377817589?thread_ts=1692563653.439049&cid=C6R3R32H4

The new version distance policy is being evaluated despite the user not having configured it.
The user noticed this because the respective policy evaluator was running into an unhandled exception while calculating the version distance:

```
2023-08-23 12:36:51,510 ERROR [LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
java.lang.NumberFormatException: For input string: "1u"
        at java.base/java.lang.NumberFormatException.forInputString(Unknown Source)
        at java.base/java.lang.Integer.parseInt(Unknown Source)
        at java.base/java.lang.Integer.parseInt(Unknown Source)
        at org.dependencytrack.util.VersionDistance.parseVersion(VersionDistance.java:156)
        at org.dependencytrack.util.VersionDistance.getVersionDistance(VersionDistance.java:331)
        at org.dependencytrack.policy.VersionDistancePolicyEvaluator.evaluate(VersionDistancePolicyEvaluator.java:86)
        at org.dependencytrack.policy.PolicyEngine.evaluate(PolicyEngine.java:89)
        at org.dependencytrack.policy.PolicyEngine.evaluate(PolicyEngine.java:71)
        at org.dependencytrack.tasks.PolicyEvaluationTask.performPolicyEvaluation(PolicyEvaluationTask.java:55)
        at org.dependencytrack.tasks.PolicyEvaluationTask.inform(PolicyEvaluationTask.java:44)
        at alpine.event.framework.BaseEventService.lambda$publish$0(BaseEventService.java:108)
```

### Steps to Reproduce

1. 

### Expected Behavior

The version distance policy must not be evaluated when no such policy condition is configured.

Errors in version distance calculation must be catched and logged as with level `WARN`. 

The log message should include:
* for which component the error occurred
* the input versions of the attempted distance calculation
* the encountered exception

### Dependency-Track Version

4.9.0-SNAPSHOT

Container Image, Executable WAR

### Database Server

N/A

### Database Server Version

_No response_

### Browser

N/A

### Checklist

nscuro:
### Description

This PR fixes a defect where the new version distance policy was being evaluated despite not being configured by users.

This PR additionally adds checks to policy evaluators that ensure that no expensive or potentially failing computation is being performed, *unless* the policy actually needs to be verified. This was previously not always the case, causing unnecessary database queries to be performed etc.

### Addressed Issue

Fixes #2979

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/policy/ComponentAgePolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/ComponentAgePolicyEvaluator.java
@@ -25,6 +25,7 @@ import org.dependencytrack.model.PolicyCondition;
 import org.dependencytrack.model.RepositoryMetaComponent;
 import org.dependencytrack.model.RepositoryType;
 import org.dependencytrack.persistence.QueryManager;
+
 import java.time.LocalDate;
 import java.time.Period;
 import java.time.ZoneId;
@@ -59,6 +60,11 @@ public class ComponentAgePolicyEvaluator extends AbstractPolicyEvaluator {
     public List<PolicyConditionViolation> evaluate(final Policy policy, final Component component) {
         final var violations = new ArrayList<PolicyConditionViolation>();
         if (component.getPurl() == null) {
+            return violations;
+        }
+
+        final List<PolicyCondition> policyConditions = super.extractSupportedConditions(policy);
+        if (policyConditions.isEmpty()) {
             return violations;
         }
 
--- a/src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/CwePolicyEvaluator.java
@@ -18,17 +18,17 @@
  */
 package org.dependencytrack.policy;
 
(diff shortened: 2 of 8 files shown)


leec94 (approved):
i was unable to run a test by “specifying a non-numerical value for the “epoch”", looks like only numbers are allowed.  
but i ran a second test, so I added a component with invalid version and i didn’t get any output, which is expected. 
looks like this version distance policy is only being run when specified now. good work, approved

nscuro:
Thanks @leec94! I'll try to make it a habit to include "how to test" instructions in the PR description. 😁

leec94:
a "how to test" section would be really helpful! thanks!

Issue #2365, opened 2023-01-10
Allowing to set a 'violationState' via API request


interN3rd:
### Current Behavior

When creating a new policy via `PUT` request to the endpoint `/v1/policy`, the value of `violationState` is ignored, because it is [hardcoded ](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/resources/v1/PolicyResource.java#L127C34-L129)to `INFO`.

Example request:

```
curl -X "PUT" "MY_URL" \
-H 'Content-Type: application/json' \
-H 'X-API-Key: MY_API_KEY' \
-d $'{"name": "MY_NAME", "operator": "MY_OPERATOR", "violationState": "FAIL"}
```

A policy does get created, but its `violationState` is still `INFO`.

### Proposed Behavior

I'd like to propose the behavior of getting the value of the field `violationState` and save it accordingly.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #2997, opened 2023-08-30
Fix: allow operator and violationState to be specified when create policy

takumakume:
### Description

Allow `operator` and `violationState` to be specified when create policy.
In addition, to keep backward compatibility, default values are provided.

### Addressed Issue

Closes https://github.com/DependencyTrack/dependency-track/issues/2365

The `violationState` is problematic, but so is the `operator`.

### Additional Details

Currently, sending the `operator` and `violationState` parameters is ignored and the hard-coded values are used.

If these are to be changed to other values, a policy update is required, which is inefficient.

This is a problem specific to create policy since it is a required parameter in the model layer.

In order to maintain backward compatibility of create policy, a default value has been added.

### Checklist


nscuro (approved):
Thanks @takumakume! 🤘

PR #3070, opened 2023-10-02
Fix false negatives in NVD CPE matching

### LINKED ISSUE #1832: Wrong vulnerabilities reports when logical value NA used
### Current Behavior:

A project with a httpd server component identified by the CPE "cpe:2.3:a:apache:http_server:2.4.53:*:*:*:*:*:*:*" leads to be
vulnerable to the CVE 2007-6420.

The CVE CVE 2007-6420 has the following cpe_match :

```json
    "configurations" : {
      "CVE_data_version" : "4.0",
      "nodes" : [ {
        "operator" : "OR",
        "children" : [ ],
        "cpe_match" : [ {
          "vulnerable" : true,
          "cpe23Uri" : "cpe:2.3:a:apache:http_server:2.2.0:*:*:*:*:*:*:*",
          "cpe_name" : [ ]
        }, {
          "vulnerable" : true,
          "cpe23Uri" : "cpe:2.3:a:apache:http_server:2.2.2:*:*:*:*:*:*:*",
          "cpe_name" : [ ]
        }, {
          "vulnerable" : true,
          "cpe23Uri" : "cpe:2.3:a:apache:http_server:2.2.4:*:*:*:*:*:*:*",
          "cpe_name" : [ ]
        }, {
          "vulnerable" : true,
          "cpe23Uri" : "cpe:2.3:a:apache:http_server:2.2.6:*:*:*:*:*:*:*",
          "cpe_name" : [ ]
        }, {
          "vulnerable" : true,
          "cpe23Uri" : "cpe:2.3:a:apache:http_server:2.2.3:*:*:*:*:*:*:*",
          "cpe_name" : [ ]
        }, {
          "vulnerable" : true,
          "cpe23Uri" : "cpe:2.3:a:apache:http_server:-:*:*:*:*:*:*:*",
          "cpe_name" : [ ]
        } ]
      } ]
    },
```

### Expected Behavior:

The NVD search doesn't report the vuln: [NVD Search](https://nvd.nist.gov/vuln/search/results?isCpeNameSearch=true&query=cpe%3A2.3%3Aa%3Aapache%3Ahttp_server%3A2.4.53%3A*%3A*%3A*%3A*%3A*%3A*%3A*&results_type=overview&form_type=Basic&search_type=all&startIndex=0)

Moreover, according to the "Name Matching Specification Version 2.3" I dont't think the actual implementation in AbstractVulnerableSoftwareAnalysisTask.compareVersions is correct as I udnerstand in chap "6.1 from Attribute Comparison Relations" at [https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf](https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf)

### Steps to Reproduce:

I reproduce the issue with a the test case below:

```java
        var project = new Project();
        project.setName("acme-app");
        project = qm.createProject(project, List.of(), false);
        var component = new Component();
        component.setProject(project);
        component.setName("Apache httpd");
        component.setVersion("2.4.53");
        component.setCpe("cpe:2.3:a:apache:http_server:2.4.53:*:*:*:*:*:*:*");
        component = qm.createComponent(component, false);

        var vs1 = new VulnerableSoftware();
        vs1.setCpe23("cpe:2.3:a:apache:http_server:-:*:*:*:*:*:*:*");
        vs1.setPart("a");
        vs1.setVendor("apache");
        vs1.setProduct("http_server");
        vs1.setVersion("-");
        vs1.setVulnerable(true);
        var vs = qm.persist(vs1);

        var vulnerability = new Vulnerability();
        vulnerability.setVulnId("CVE-2007-6420");
        vulnerability.setSource(Vulnerability.Source.NVD);
        vulnerability.setVulnerableSoftware(List.of(vs));
        qm.createVulnerability(vulnerability, false);

        new InternalAnalysisTask().analyze(List.of(component));

        final PaginatedResult vulnerabilities = qm.getVulnerabilities(component);
        assertThat(vulnerabilities.getTotal()).isEqualTo(1);
        assertThat(vulnerabilities.getList(Vulnerability.class).get(0).getVulnId()).isEqualTo("CVE-2007-6420");
```

### LINKED ISSUE #2243: Extensive test suite for CPE matching
### Current Behavior

We maintain a custom CPE matching logic, yet our test coverage is very scarce.

Making changes to the existing logic is a high risk, because we don't quite know if it negatively affects matching on real-world data.

### Proposed Behavior

Setup a test suite with loads of real-world CPE data. 

Note that we also pull versions information from component PURLs if the analyzed component has no CPE assigned to it, so that aspect should be covered as well.

The test subject will be the [`InternalAnalysisTask`](https://github.com/DependencyTrack/dependency-track/blob/4.6.3/src/main/java/org/dependencytrack/tasks/scanners/InternalAnalysisTask.java).

To cover as much data as possible without having to write too much code, setting up a test that just reads from a CSV file or so would be preferable.

```csv
cpe,componentCpe,componentPurl,match
cpe:2.3:a:...,cpe:2.3:a:...,,true
cpe:2.3:a:...,,pkg:maven/...,true
```

### Checklist

### LINKED ISSUE #2894: CPE matching seems to be case sensitive
### Current Behavior

When encountering a component with a cpe that contains uppercase characters, dependencytrack fails to map vulnerabilities.
getVulnerableSoftwareByCpe23 apparently uses a hash map - but this results in a case-sensitive comparison with the CPEs listed in the components.
Two points:

- The CPE of the component shouldn't have been supplied with uppercase characters in the first place:
"Lexical case SHALL NOT distinguish attributes from one another, e.g., the attributes Foo, foo, FOO, etc., SHALL be considered equivalent. By convention, attributes will be written in all lowercase letters, with the underscore (“_”) character used to separate distinct words within an attribute." (https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7695.pdf)
- But also, comparison should be case insensitive - see:
The official matching logic for 2.3 has "to_lowercase" as a first step: https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf (7.3)

### Steps to Reproduce

Change some characters in a cpe string to uppercase and upload bom to dt.
Vulnerabilities will not be mapped.

### Expected Behavior

DT uses case invariant matching as outlined in the cpe specification.

### Dependency-Track Version

4.8.2

Container Image

### Database Server

PostgreSQL

### Database Server Version

_No response_

### Browser

Google Chrome

### Checklist

### LINKED ISSUE #2988: CPE Name Matching: CPE name relation being SUBSET(⊂) should be a match
### Current Behavior

Currently, DT will only consider CPE name matching a "match", if the CPE name relation is SUPERSET (⊃) or when CPE name relation is EQUAL (=). However, to avoid false negatives a SUBSET relation (⊂) MUST also be treated as a match.

To understand this, we need to take a closer look at the [NIST name matching specification](https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7696.pdf), especially at table 6-2.

Here we see that there are exactly 3 cases in which a SPE attribute relation is a SUBSET: 3, 13, 15

image

Keeping in mind that the CVE CPE is the source CPE and the component CPE is the target CPE, let's take a closer look at what these three cases really mean:

case 5 example:

source: `cpe:2.3:o:linux:linux_kernel:5.15.37:*:*:*:*:*:*:NA`
target:  `cpe:2.3:o:linux:linux_kernel:5.15.37:*:*:*:*:*:*:*`

We have a CVE CPE where the "Other" attribute is "Not applicable" and the component attribute for "Other" is "ANY". Why should the CVE not apply to the target component? Since "ANY" is a superset of "NA" (as proven in case 2), this should still match.

case 13 example:

source: `cpe:2.3:o:linux:linux_kernel:5.15.37:*:*:*:*:*:x86:*`
target:  `cpe:2.3:o:linux:linux_kernel:5.15.37:*:*:*:*:*:*:*`

We have a CVE CPE where the "Target_HW" attribute is "x86" and the component attribute for "Target_HW" is "ANY". Why should the CVE not apply to the target component? Sure, it could be a false positive, if the component actually only runs on arm64, but that is up to the component to provide as accurate information as possible in order to prevent false positives. If however the component actually runs on a x86, a no-match would definitely be a false negative.

case 15 example:

source: `cpe:2.3:o:linu*:linux_kernel:5.15.37:*:*:*:*:*:*:*`
target:  `cpe:2.3:o:*:linux_kernel:5.15.37:*:*:*:*:*:*:*`

We have a CVE CPE where the "Vendor" attribute contains a wildcard (the example is a bit contrived, I know) and the component attribute for "Vendor" is "ANY". Again, why should this not be a CVE match? Sure, it could again be a false positive but that is for the auditor to decide. It should not end up as false negative.

As such, cases 2-4 from table 6-4 should be a match:

image

Related to https://github.com/DependencyTrack/dependency-track/issues/2580, https://github.com/DependencyTrack/dependency-track/issues/2994

### Steps to Reproduce

1. Create a component with CPE `cpe:2.3:*:linux:linux_kernel:5.15.37:*:*:*:*:*:*:*` (make sure to disable fuzzy matching for CPE)
2. This will show 0 CVE matches, with [false negatives](https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&isCpeNameSearch=true&seach_type=all&query=cpe:2.3:o:linux:linux_kernel:5.15.37:*:*:*:*:*:*:*), as the "Part" attribute was empty.

### Expected Behavior

The component should show correct matches + X potential false positives, e.g. since there there was no Target_HW for the component defined.

False positives are bad (that's why accurate component CPE data matters), false negatives are way worse.

### Dependency-Track Version

4.8.1

Container Image

### Database Server

PostgreSQL

### Database Server Version

_No response_

### Browser

N/A

### Checklist

nscuro:
### Description

This PR addresses various issues that cause both false negatives, as well as false positives during CPE matching.

### Addressed Issue

Closes #1832
Closes #1929
Closes #2243
Fixes #2988

### Additional Details

* This PR effectively reverts 1d6ce184c8614b8be71baf2979d7732270814932. This commit made it so that CPEs with version `-` (NA) would match *any* component version, which is incorrect and causes lots of false positives.
* CPE matching is supposed to be case-insensitive (https://github.com/DependencyTrack/dependency-track/issues/2894). Thus, querying the database by CPE part, vendor, and product, must be case-insensitive, too. For various reasons related to database indexes, we cannot support this right now:
  * DataNucleus will translate `field.equalsIgnoreCase(<parameter>)` to SQL `UPPER("FIELD") = UPPER(?)`. This will bypass existing indexes, as they do not take casing into account. DataNucleus and JDO do not allow us to specify expressions like `UPPER` for indexes created via `@Index` annotation. 
  * PostreSQL supports functional indexes (e.g. on `UPPER("FIELD")`)
  * Functional indexes are only supported in MySQL >= 8.0.13
  * MSSQL does not support functional indexes at all, and instead requires usage of computed columns

### Checklist

### DIFF
--- a/docs/_docs/FAQ.md
+++ b/docs/_docs/FAQ.md
@@ -33,6 +33,22 @@ dependencies like packages, libraries, and frameworks. In the local mirror there
 So the local mirror is used, but not for dependencies represented by PURL. Dependency Track will use the Analyzer best
 suited to analyze a given dependency.
 
+#### The NVD CPE search is showing significantly more CVEs than Dependency-Track
+
+The NVD allows users to [search CVEs by CPE](https://nvd.nist.gov/products/cpe/search). Naturally, users assume
+that CVEs returned by this search mechanism are CVEs *affecting* the given CPE. When comparing the number of results
+with those reported by Dependency-Track, it is common to find discrepancies. This does not necessarily mean that
+Dependency-Track is missing CVEs that the NVD search is able to identify.
+
+The NVD utilizes the concept of [Known Affected Software Configurations](https://nvd.nist.gov/vuln/vulnerability-detail-pages)
+to express more complex matching criteria. For example, [CVE-2016-8963](https://nvd.nist.gov/vuln/detail/CVE-2016-8963)
+affects `cpe:2.3:a:ibm:license_metric_tool:9.2.0:*:*:*:*:*:*:*` *running on* `cpe:2.3:o:linux:linux_kernel:*:*:*:*:*:*:*:*`.
+The Linux Kernel itself is *not* affected. But when searching the NVD for `cpe:2.3:o:linux:linux_kernel:*:*:*:*:*:*:*:*`,
+CVE-2016-8963 is returned as result.
+
+Dependency-Track will only flag a component as vulnerable when it itself is vulnerable. If it is merely part of a
+configuration like above, Dependency-Track will not report it as vulnerable.
+
 #### Dependency Track crashes when run as a container
 
 Make sure the container is allowed to allocate enough RAM. For memory requirements see
--- a/src/main/java/org/dependencytrack/persistence/QueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/QueryManager.java
@@ -855,6 +855,11 @@ public class QueryManager extends AlpineQueryManager {
         return getVulnerableSoftwareQueryManager().getAllVulnerableSoftware(cpePart, cpeVendor, cpeProduct, purl);
     }
 
+    public List<VulnerableSoftware> getAllVulnerableSoftwareX(final String cpePart, final String cpeVendor,
+                                                              final String cpeProduct, final PackageURL purl) {
+        return getVulnerableSoftwareQueryManager().getAllVulnerableSoftwareX(cpePart, cpeVendor, cpeProduct, purl);
+    }
+
     public Cwe createCweIfNotExist(int id, String name) {
         return getVulnerableSoftwareQueryManager().createCweIfNotExist(id, name);
     }
--- a/src/main/java/org/dependencytrack/persistence/QueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/QueryManager.java
@@ -847,17 +847,8 @@ public class QueryManager extends AlpineQueryManager {
         return getVulnerableSoftwareQueryManager().getAllVulnerableSoftwareByPurl(purl);
     }
 
-    public List<VulnerableSoftware> getAllVulnerableSoftware(final String cpePart, final String cpeVendor, final String cpeProduct, final String cpeVersion, final PackageURL purl) {
-        return getVulnerableSoftwareQueryManager().getAllVulnerableSoftware(cpePart, cpeVendor, cpeProduct, cpeVersion, purl);
-    }
-
     public List<VulnerableSoftware> getAllVulnerableSoftware(final String cpePart, final String cpeVendor, final String cpeProduct, final PackageURL purl) {
         return getVulnerableSoftwareQueryManager().getAllVulnerableSoftware(cpePart, cpeVendor, cpeProduct, purl);
-    }
-
-    public List<VulnerableSoftware> getAllVulnerableSoftwareX(final String cpePart, final String cpeVendor,
-                                                              final String cpeProduct, final PackageURL purl) {
-        return getVulnerableSoftwareQueryManager().getAllVulnerableSoftwareX(cpePart, cpeVendor, cpeProduct, purl);
     }
 
     public Cwe createCweIfNotExist(int id, String name) {
--- a/src/main/java/org/dependencytrack/persistence/VulnerableSoftwareQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/VulnerableSoftwareQueryManager.java
@@ -241,6 +241,9 @@ final class VulnerableSoftwareQueryManager extends QueryManager implements IQuer
 
     /**
      * Fetch all {@link VulnerableSoftware}s matching the given CPE part, vendor, product, or Package URL.
+     * <p>
+     * This method will not query for <em>exact</em> matches of the given CPE attributes,
+     * but instead follow the CPE name matching specification.
      *
      * @param cpePart    The part attribute of the target CPE
      * @param cpeVendor  The vendor attribute of the target CPE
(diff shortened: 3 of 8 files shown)


Jasper-Ben (approved):
With my limited knowledge of the code base in mind: LGTM, other than the already documented short-comings (case-sensitivity).

PR #3196, opened 2023-11-14
<new community usage example added: import audit data from Excel

JoergBruenner:
Signed-off-by: JoergBruenner <joerg.bruenner@gmx.net>


JoergBruenner:
Seen after the fact: the excel.ini file contains a wrong path. It should be: <Path to Audit.xlsx>

PR #3231, opened 2023-11-24
ignore (suppress) findings globally (system-wide)

### LINKED ISSUE #1495: system wide vulnerability management
Hi,
I would congratulate with you because that is a great project but I'm going to describe some "usability problems" that for me needs an improvement to make this project even better:

Our environment:
- dependencytrack version: 4.4
- a lot of projects
- use cases:
    - in a CI environment: CI pipeline will fail if a vulnerability >= high will be found 
    - to analyze stale projects for new vulnerabilities based on latest uploaded bom

Usability problems:

1) A new vulnerability used by almost all projects has been published but not fixed yet (e.g. https://github.com/advisories/GHSA-57j2-w4cx-62h2). If a fix on our projects needs to be released it will be blocked by CI pipeline check. Right now I must suppress and manage that vulnerability on every project. It would be great if we can also manage vulnerabilities "globally": managing vulnerabilities on every project is not sustainable and leads to errors.
2) Right now we can suppress/manage a vulnerability on every project but there is not a view of active/made suppressions (I need to track it somewhere else and if combined with problem 1 that is a huge management problem). It would be great if there was a recap view with all managed/suppressed vulnerabilities so we can figure out what is suppressed and need to be patched when a fix will be released

tomaszn:
### Description

This change implements a way to ignore security advisories (like CVE, GHSA) globally, system-wide. Vulnerabilities caused by advisories listed in the `IGNORED_ADVISORIES` environment variable will not be created at all.

### Addressed Issue

This fixes #1495, which contains rationale behind this feature.

### Additional Details

I solved the issue as simply as I found possible. I treat it more as a workaround for a critical shortage.

### Checklist

### DIFF
--- a/docs/_docs/triage/suppression.md
+++ b/docs/_docs/triage/suppression.md
@@ -33,3 +33,12 @@ finding will have a positive impact not only on Dependency-Track itself, but the
 
 For example, vulnerability aggregation platforms such as Kenna Security or ThreadFix will become aware of the updated
 list of findings and metrics the next time they sync. These systems will assume suppressed findings have been 'fixed'.
+
+### Ignoring Advisories Systemwide
+
+Individual advisories can be ignored by setting the environment variable "IGNORED_ADVISORIES" before starting
+Dependency-Track API server. Identifiers of the advisories need to be separated by a space character, for example:
+
+```
+export IGNORED_ADVISORIES="CVE-2019-123 GHSA-567"
+```
--- a/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
@@ -18,10 +18,12 @@
  */
 package org.dependencytrack.persistence;
 
+import alpine.common.logging.Logger;
 import alpine.event.framework.Event;
 import alpine.persistence.PaginatedResult;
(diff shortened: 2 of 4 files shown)


valentijnscholten:
How practical is this? Seems like it needs a restart after setting or changing the list of ignored advisories (which ignore vulnerabilities).
Also my feeling is that when vulnerabilities are published, DT will quite quickly have these stored and applied to components. Setting the environment variable after that only prevents future assignment of the vulnerability, but will leave existing assignment behind (I think).

Would it work to have an `ignore` flag on a vulnerability in DT which can be set/unset via the UI and API. Additionaly an event (and/or scheduled task) that removes ignored vulnerabilities from components?

tomaszn:
@valentijnscholten, you are right, it has all these downsides. Yet worked for me for several months now.

It is quite practical, because with the deployment configuration versioned, this approach allows documenting and reviewing the changes via merge requests. With a simple switch in the UI widely accessible, beginning users could by mistake ignore the vulnerability globally, instead of suppressing it in the project. It would need a permission.

Maybe the whole suppression feature and the database table (`ANALYSIS`?) could be reused, not only with a switch as you suggest, but also with the rest of inputs. No value in the "project_id" database column would mean a global suppression. Better solution than mine.

And this use case is also covered by @nscuro's work.

Issue #2935, opened 2023-08-04
An inactive Parent cannot be selected as parent


zeenmc:
### Current Behavior

Hello Team :)

I am having issues to create child project, as I am getting error from the tittle. I am feeling this is a bug as I am able to reproduce multiple times. Our DT is 4.8.2 ,and we are using MS SQL Server.

```
2023-08-04T15:41:44.748118564Z: [INFO]  2023-08-04 15:41:44,747 INFO [ProjectResource] Project test-active10 created by TdCK****ST0B
2023-08-04T15:42:08.231296303Z: [INFO]  2023-08-04 15:42:08,231 ERROR [GlobalExceptionHandler] Uncaught internal server error
2023-08-04T15:42:08.231350404Z: [INFO]  java.lang.IllegalArgumentException: An inactive Parent cannot be selected as parent
```
```
#Parent project creation :

curl --verbose -X PUT "https://dt-api.azurewebsites.net/api/v1/project" \
-H 'Content-Type: application/json' \
-H 'X-Api-Key: TdCKerfddgdn7zsLfFST0B' \
-d '{"name": "management-mfe"}' \
-d '{"active": "true"}'
```

After I created project via curl command (idea is to have whole this proccess in CICD pipelines in Azure DevOps),
In UI, I ass project is active

```
and If I try to push child project:
curl --verbose -X POST "https://dt-api.azurewebsites.net/api/v1/bom" \
-H 'Content-Type: multipart/form-data' \
-H 'X-Api-Key: TdCKerfddgdn7zsLfFST0B' \
-F "classifier=Application" \
-F "projectName=management-mfe" \
-F "autoCreate=true" \
-F "projectVersion=10" \
-F "parentName=management-mfe" \
-F "bom=@/Users/SCN1BG/Desktop/services/frontend/user-management-mfe/bom-lokal.json"
```
I will get error from title, as I said, but if I go to UI, in **Project Details** which I created, and disable project, save changes, again enable project, and wait for a few seconds, I will be able to create child project without any errors.

Where is log when I disable/enable just created project
```
2023-08-04T15:42:43.062880058Z: [INFO]  2023-08-04 15:42:43,062 INFO [ProjectResource] Project test-active10 updated by admin
2023-08-04T15:42:54.218621500Z: [INFO]  2023-08-04 15:42:54,218 INFO [ProjectResource] Project test-active10 updated by admin
```

Any ideas how to resolve this issue ?
Ned.

[...]

PR #3244, opened 2023-11-28
Added a default value for the Project `active` field if it is null

willienel:
### Description

When creating a project from the API, which deserializes a JSON object into a Project object, if you do not specify the `active` field in your JSON payload it will default to null - this causes problems with projects being marked as inactive when they are supposed to be active.

To fix this issue I have added a default value for a Project's `active` field just before it is persisted to the database.

### Addressed Issue

This Fixes issue #2935 where projects are not assignable as parent projects because they are reported as inactive.

### Additional Details

I tried to set a default value on the persistence layer but was not able to do so for some reason - I am not too familiar with JDO but I tried following the official documentation but still could not get it to behave as expected.

I then opted for my current solution where I assign the default value inside of the ProjectQueryManager just before it persists the object. The reason why I added it here and not in the API layer is because I wanted this fix to be available to any area of the app that tries to create a project using this method.

I added a test to confirm that it works and I also made sure that I didn't break any existing tests.

### Checklist


nscuro (approved):
Thanks @willienel! 🚀

PR #3260, opened 2023-12-03
Return processing token when cloning project #2842

### LINKED ISSUE #2842: Queue tokens for project cloning
### Current Behavior

API request "PUT /v1/project/clone" returns response 200 with new project UUID like if operation is finished
In fact, cloning a project is a continuous process that requires copying information about the components in use from the original repository to the new
That can lead to problems like #2768 when running two continuous operations at the same time leads to unwanted results

### Proposed Behavior

Return status token like `PUT /v1/bom` does.
This allows to track cloning process and sync task order for Dependency Track to aviod timing conflicts

### Checklist

rkg-mm:
### Description
Return a token when cloning projects which can be used to query clone processing status.
Add an endpoint to query clone processing status.

### Addressed Issue
Closes #2842 

### Additional Details
I did not find an obvious way to also implement wrapping cloning in a transaction while using the processing tokens, so this part is not implemented.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
@@ -39,6 +39,7 @@ import org.dependencytrack.model.Project;
 import org.dependencytrack.model.Tag;
 import org.dependencytrack.persistence.QueryManager;
 import org.dependencytrack.resources.v1.vo.CloneProjectRequest;
+import org.dependencytrack.resources.v1.vo.IsTokenBeingProcessedResponse;
 
 import javax.jdo.FetchGroup;
 import javax.validation.Validator;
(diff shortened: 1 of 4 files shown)


src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -669,4 +672,27 @@ public Response getProjectsWithoutDescendantsOf(
             }
         }
     }
+
+    @GET
```
nscuro, line 676: This endpoint should have its own API resource instead. There are multiple other cases where event tokens could be returned, and checking whether a given token is being processed will always use the same logic.

Thinking out loud, we may even want to allow users to manually trigger certain events via REST API in the future. With this in mind, how about we move this to `/api/v1/event/token/{uuid}` instead?

nscuro (changes requested):

nscuro:
> not sure why the unit test fails now, it works locally and I see no reason why It should not

Seems to be flaky, it's failing on other PRs as well. I'll have a look.

nscuro (approved):
Thanks @rkg-mm!

PR #3269, opened 2023-12-04
Expand OIDC docs

acdha:
### Description

This clarifies what you have to configure when setting up OIDC authentication for DT - I had been hoping that the OpenID Connect groups list would automatically populate and missed that the group list was case-sensitive in my first test.

### Addressed Issue

https://github.com/DependencyTrack/dependency-track/discussions/3188

### Additional Details

As an alternative, would it make sense to make the group to team mapping case-insensitive? Automatically populating the OIDC groups list would be a little more work since that could potentially be both large and some users might reasonably want that behaviour to be optional, but matching the group names case-insensitively should be less controversial. We have close to half a century of evidence with filesystems that distinct names differing only in case is rarely what people want.

### Checklist

- ~[ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective~
- ~[ ] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)~
- ~[ ] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly~

### DIFF
--- a/docs/_docs/getting-started/openidconnect-configuration.md
+++ b/docs/_docs/getting-started/openidconnect-configuration.md
@@ -22,6 +22,14 @@ If configured properly, users will be able to sign in by clicking the _OpenID_ b
 > which means that Dependency-Track will no longer request the `/userinfo` endpoint if all required claims  
 > are present in the ID token's payload already.
 
+## How OpenID Connect claims are mapped
+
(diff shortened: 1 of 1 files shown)


nscuro:
Thanks @acdha! Could you please sign-off your commits so the DCO check passes?

PR #3275, opened 2023-12-05
Webhook alert token and new user alerts

### LINKED ISSUE #3255: Notification Alert for new user added/ingested
### Current Behavior

Upon new user addition/ingestion - e.g. those from incoming OIDC integration - no notifications is issued from Dependency Track.

### Proposed Behavior

Upon new user addition/ingestion, issue notification to configured recipients.
It could be a `scope:SYSTEM` and `group:NEW_USER` type of event.

### Checklist

### LINKED ISSUE #3256: Alerts to Outbound Webhook: allow API token to be informed
### Current Behavior

No token auth option exists for Alerts issued using Outbound Webhook publisher.
Target APIs that require some sort of token to be present are not feasible to be called.

### Proposed Behavior

Allow to register token value and possibly auth header to be used for Outbound Webhook publishers.

### Checklist

fnxpt:
### Description

This PR adds two new notification types (user_created and user_deleted) and the possibility of setting a token on the webhook alert

### Addressed Issue

Closes #3255
Closes #3256

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
+++ b/src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
@@ -77,7 +77,7 @@ public abstract class AbstractWebhookPublisher implements Publisher {
                 request.addHeader("Authorization", "Bearer " + credentials.password);
             }
         } else if (getToken(config) != null) {
-            request.addHeader("X-Api-Key", getToken(config));
+            request.addHeader(getTokenHeader(config), getToken(config));
         }
 
         try {
@@ -111,6 +111,10 @@ public abstract class AbstractWebhookPublisher implements Publisher {
 
     protected String getToken(final JsonObject config) {
         return config.getString(CONFIG_TOKEN, null);
+    }
+
+    protected String getTokenHeader(final JsonObject config) {
+        return config.getString(CONFIG_TOKEN_HEADER, "X-Api-Key");
     }
 
     protected record AuthCredentials(String user, String password) {
--- a/src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
+++ b/src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
@@ -76,6 +76,8 @@ public abstract class AbstractWebhookPublisher implements Publisher {
             } else {
                 request.addHeader("Authorization", "Bearer " + credentials.password);
             }
+        } else if (getToken(config) != null) {
+            request.addHeader("X-Api-Key", getToken(config));
         }
 
         try {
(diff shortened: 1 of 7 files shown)


fnxpt:
@nscuro couldn't find where is the documentation for these alerts, can you point me to the right place

valentijnscholten:
> @nscuro couldn't find where is the documentation for these alerts, can you point me to the right place

probably here: https://docs.dependencytrack.org/integrations/notifications/#outbound-webhooks 
source: https://github.com/DependencyTrack/dependency-track/blob/master/docs/_docs/integrations/notifications.md

src/main/java/org/dependencytrack/util/NotificationUtil.java
```diff
@@ -272,6 +273,25 @@ public static JsonObject toJson(final Project project) {
...
     }
 
+    public static JsonObject toJson(final UserPrincipal user) {
+        final JsonObjectBuilder userBuilder = Json.createObjectBuilder();
+
+        userBuilder.add("username", user.getUsername());
+
+        if (user.getId() != 0) {
+            userBuilder.add("id", user.getId());
+        }
```
nscuro, line 283: The ID is purely internal and should not be part of the notification IMO.

src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
```diff
@@ -76,6 +76,8 @@ public void publish(final PublishContext ctx, final PebbleTemplate template, fin
             } else {
                 request.addHeader("Authorization", "Bearer " + credentials.password);
             }
+        } else if (getToken(config) != null) {
+            request.addHeader("X-Api-Key", getToken(config));
```
nscuro, line 80: Should the header name be configurable? I think the name will vary depending on what system the webhook is sent to.

nscuro (changes requested):
One major caveat here is that users created via auto-provisioning are not captured. As this is happening in Alpine, there's no easy way for DT to detect it:

* Users created for OIDC: https://github.com/stevespringett/Alpine/blob/5727c590c59881d6e46630839e2a8a9dcadcbf4f/alpine-server/src/main/java/alpine/server/auth/OidcAuthenticationService.java#L228-L241
* Users created for LDAP: https://github.com/stevespringett/Alpine/blob/5727c590c59881d6e46630839e2a8a9dcadcbf4f/alpine-server/src/main/java/alpine/server/auth/LdapAuthenticationService.java#L117-L121

src/main/java/org/dependencytrack/util/NotificationUtil.java
fnxpt: the problem is that after the user is deleted all the other fields are null

src/main/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisher.java
fnxpt: Yes, I will change it

nscuro (approved):

Issue #1823, opened 2022-07-25
Notifications subjects - add additional information (e.g. Project/Component/Severity) into Subject string


JN-CSIRT:
### Current Behavior:
The notification email's subject are hard coded right now. 

### Proposed Behavior:
add the notification subject configuration to the notification template and use same params as there exists for content.

JN-CSIRT:
**Current Behavior:**

The notification email's subject are hard coded now and contain the project name (twice) only.
E.g.: "New Vulnerability Identified on Project: [[PROJECTID]: [PROJECTID] : [VERSION]]"

**Proposed Behavior:**
add the notification subject configuration to the notification template in order to add all necessary information
E.g.:
{{ notification.title }}
--------------------------------------------------------------------------------
{% if notification.group == "NEW_VULNERABILITY" %}
_Email Subject: {{alert.name}}: {{ subject.component.project.name }}: {{ subject.component.project.version }}: {{ subject.component.toString }} {{ subject.component.version.toString }} {{ subject.vulnerability.severity }}_
Vulnerability ID:  {{ subject.vulnerability.vulnId }}
Vulnerability URL: {{ baseUrl }}/vulnerability/?source={{ subject.vulnerability.source }}&vulnId={{ subject.vulnerability.vulnId }}
Severity:          {{ subject.vulnerability.severity }}
Source:            {{ subject.vulnerability.source }}
Component:         {{ subject.component.toString }}
Component URL:     {{ baseUrl }}/component/?uuid={{ subject.component.uuid }}
Project:           {{ subject.component.project.name }}
Version:           {{ subject.component.project.version }}
Description:       {{ subject.component.project.description }}
Project URL:       {{ baseUrl }}/projects/{{ subject.component.project.uuid }}
{% if  notification.subject.affectedProjects|length > 1%}
--------------------------------------------------------------------------------

msymons:
This issue was originally raised in discussion #1811

Customizable Notification Templates were introduced in v4.6 (released October 11 2022, since this enhancement was logged) but support only customization of body...  with subject still being hard-coded as of v4.7.1

JN-CSIRT:
According to syalioune:

My previous answer still apply. While v4.6.0 introduced customization features, it was for the notification body content.
The notification title, thus email subject, is unfortunately still hardcoded.
Enhancement request #1823 is still valid.

msymons:
[...]

PR #3422, opened 2024-01-31
Subject prefix

LaVibeX:
### Description

- Email Subject Prefix is Hardcoded rightnow `.subject("[Dependency-Track] " + notification.getTitle())`, 
- This is a solution to allow the User to modify the Subject Prefix to its will  `.subject(smtpPrefix + " " + notification.getTitle() )`

This modification would be available in `Administration>Configuration>Email` 
and ` "[Dependency-Track]"` would be the first default value. 

If the User change the Subject Prefix, it will need to press Update button in order to save the changes.
Subject Prefix is **not a required field** so leaving the input box in blank is also an option.

 **_This new feature helps to distinguish between several instances of Dependency-Track, e.g. prod and qual systems._**

### Default value: _[Dependency-Track]_
image
### Modified Value:  _[Dependency-Track TEST ®]_
image
### Modified Value: " "
image

This PR depends on its corresponding PR in the **frontend** repository
https://github.com/DependencyTrack/frontend/pull/720

### Addressed Issue
This fixes https://github.com/DependencyTrack/dependency-track/issues/1823

### Additional Details 
Check if smtpPrefix is null in order to clean the value to " "
`smtpPrefix = smtpPrefix == null ? " " : smtpPrefix;`

This PR is supervised by @lukas-braune.

### Checklist

Signed-off-by: Andres Tito <andres.tito@rohde-schwarz.com>

### DIFF
--- a/src/main/java/org/dependencytrack/notification/publisher/SendMailPublisher.java
+++ b/src/main/java/org/dependencytrack/notification/publisher/SendMailPublisher.java
@@ -44,6 +44,7 @@ import java.util.stream.Stream;
 
 import static org.dependencytrack.model.ConfigPropertyConstants.EMAIL_SMTP_ENABLED;
 import static org.dependencytrack.model.ConfigPropertyConstants.EMAIL_SMTP_FROM_ADDR;
+import static org.dependencytrack.model.ConfigPropertyConstants.EMAIL_PREFIX;
 import static org.dependencytrack.model.ConfigPropertyConstants.EMAIL_SMTP_PASSWORD;
 import static org.dependencytrack.model.ConfigPropertyConstants.EMAIL_SMTP_SERVER_HOSTNAME;
 import static org.dependencytrack.model.ConfigPropertyConstants.EMAIL_SMTP_SERVER_PORT;
@@ -97,6 +98,7 @@ public class SendMailPublisher implements Publisher {
 
         final boolean smtpEnabled;
         final String smtpFrom;
+        final String smtpPrefix;
         final String smtpHostname;
         final int smtpPort;
         final String smtpUser;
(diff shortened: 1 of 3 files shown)


src/main/java/org/dependencytrack/notification/publisher/SendMailPublisher.java
```diff
@@ -136,7 +140,7 @@ private void sendNotification(final PublishContext ctx, Notification notificatio
             final SendMail sendMail = new SendMail()
                     .from(smtpFrom)
                     .to(destinations)
-                    .subject("[Dependency-Track] " + notification.getTitle())
+                    .subject(smtpPrefix + " " + notification.getTitle() )
```
valentijnscholten, line 143: Might be nitpicking, but I suggest to change the name of the variable and setting to something more generic like `emailSubjectPrefix` or something like that. The prefix is not really tied to the SMTP protocol. If other email delivery methods are added in the future, the prefix might still be valid to use, but should not be called `smtpPrefix`.

valentijnscholten (changes requested):

nscuro (approved):
Thanks @LaVibeX!

nscuro:
Looks like this broke `SendMailPublisherTest`. Do you mind having a look @LaVibeX?

nscuro (approved):

girtsn:
it says that it implements https://github.com/DependencyTrack/dependency-track/issues/1823
from the ticket "add the notification subject configuration to the notification template and use same params as there exists for content"
is it possible to use variables now as in body of the notification (e.g. severity / others)? the description of this ticket seems to be about adding a static prefix only?

nscuro:
You are right @girtsn, I reopened the issue.

PR #3468, opened 2024-02-15
Exclude pre-releases from NuGet latest version check

### LINKED ISSUE #3467: Remove nuget pre-release packages from being returned.
### Current Behavior

The current behavior for Nuget packages is to show a package as out of date if there is currently pre-release version as the last version released.

In the scenario (at time of writing, feb 2024) the current version of [Microsoft.Extensions.DependencyInjection](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection/) is 8.0.0, however there is a re-release version 9.0.0-preview.1.24080.9, that wont become stable until November 2024. For the next 9 months, dependency track will show this version is out of date. 

### Proposed Behavior

According to the [Nuget documentation](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#pre-release-versions) any version number with a '-' is a pre-release. 

Version strings that contain an "-" should be filtered from the list when determinizing the latest version.

In the scenario that a component being compared is actually a pre-release (contains an "-"), then no filtering should occur. This will allowed a newer pre-release to become the latest version.

### Checklist

brentos99:
### Description

This PR addressed the issue of pre-release nuget packages being returned as the latest version.

The [Nuget Documentation](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#pre-release-versions) identifies packages with a "-" in their version string as a pre-release package.

The change is to filter the list of Nuget Packages that contain an "-" when determining the latestVersion. If the component being inspected is a pre-release itself, then pre-release versions are not filtered.

### Addressed Issue

fixes #3467 

### Additional Details

There are 2 tests in this PR (NugetMetaAnalyzerTest.java), but due to the transient nature of the Nuget packages, these tests could fail if there is a non "pre-release" as the latest nuget version of the Microsoft.Extensions.DependencyInjection package.

I would have liked to have created a mock dataset for this scenario, but I the current tests appear tied to connecting directly to the live nuget package urls.  

These 2 tests have their `@Test` commented as to not break future builds.

### Checklist


src/main/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzer.java
```diff
@@ -111,7 +111,10 @@ private boolean performVersionCheck(final MetaModel meta, final Component compon
                     String responseString = EntityUtils.toString(response.getEntity());
                     var jsonObject = new JSONObject(responseString);
                     final JSONArray versions = jsonObject.getJSONArray("versions");
-                    final String latest = findLatestVersion(versions); // get the last version in the array
+
+                    final boolean excludePreRelease = component.getPurl().getVersion() != null && component.getPurl().getVersion().contains("-") ==false;  // if the version is a pre-release version, we should not exclude pre-release versions
```
sebD, line 115: Just a quick remark regarding Java standards. Rather than writing 
`component.getPurl().getVersion().contains("-") == false`
you should directly write 
`!component.getPurl().getVersion().contains("-")`
as it is already a boolean value.
Your code becomes 
`final boolean excludePreRelease = component.getPurl().getVersion() != null && !component.getPurl().getVersion().contains("-");` not much of a difference I admit, but it conforms better to java standards.

brentos99: Thanks @sebD, not worked in Java for a while so thanks for letting me know about the standards. I have refactored that code as per your suggestion.

taladar:
It would be quite useful if this was merged as we have the same issue.

sebastien-lp (approved):
Looks good to me.

brentos99:
Is there something more I need to do on this pull request or some process I have not followed. I'm new to this, and not sure if this is standard timeframe or you are waiting on something.

src/main/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzer.java
```diff
@@ -111,7 +111,11 @@ private boolean performVersionCheck(final MetaModel meta, final Component compon
                     String responseString = EntityUtils.toString(response.getEntity());
                     var jsonObject = new JSONObject(responseString);
                     final JSONArray versions = jsonObject.getJSONArray("versions");
-                    final String latest = findLatestVersion(versions); // get the last version in the array
+
+                     // if the version is a pre-release version, we should not exclude pre-release versions
+                    final boolean excludePreRelease = component.getPurl().getVersion() != null && !component.getPurl().getVersion().contains("-");
```
nscuro, line 116: Latest version information is stored by repository type, namespace, and name of the component:

https://github.com/DependencyTrack/dependency-track/blob/af4461e86ae56eaee193017ec86ee2a690709932/src/main/java/org/dependencytrack/model/RepositoryMetaComponent.java#L53-L74

Thus, the check can't be conditional based on whether the analyzed component's current version is unstable. It would yield different results depending on whether a component with stable or unstable version was analyzed first.

The preference for including pre-release versions must be either global, or on a per-repository basis.

nscuro (changes requested):

src/main/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzer.java
brentos99: Are you saying that the latest version is stored globally in dependency track and not against a project?


So in this scenario, the latest dependency is 1.1-release6 but the latest stable was 1.0

And we had 2 projects
Project A - used 1.0
Project B - used 1.1-release5

Then it would make a difference as to which project was scanned first? (Or last)

brentos99:
Based on the feedback, the pr doesn't work with the way the data is stored in the database. Their can only be one latest version of a package and this pr provides was to dynamically return different versions depending on if the version testing against is a prerelease or not.

Two options could be
1. Always exclude prerelease packages from being the latest version.

2. Provide configuration flag to allow prerelease versions to be a latest version. This would require the ui (different repo) to be updated as well to support the option as well.


Alternatively further investigation on how other repositories store prerelease data is required.

Gepardgame:
Whats the state of this PR? 4.12 is soon to be released and this has the milestone for this and there was long no development?

nscuro:
Will likely un-assign from v4.12. We can still ship this with a bugfix release so that doesn't mean having to wait for it for too long.

brentos99:
Hi @nscuro,

Could I get some direction on this PR?

The simplest approach would be to always ignore packages with a - in the version, as this aligns with NuGet's pre-release convention. According to the NuGet spec, versions with suffixes are treated as pre-release, and versions without a suffix take precedence.

Would this approach be acceptable as a global solution, or do you think a more configurable option is needed?

Thanks!

src/test/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzerTest.java
```diff
@@ -69,6 +69,47 @@ public void testAnalyzer() throws Exception {
...
+
+        analyzer.setRepositoryBaseUrl("https://api.nuget.org");
+        MetaModel metaModel = analyzer.analyze(component);
+
+        Assert.assertTrue(analyzer.isApplicable(component));
+        Assert.assertEquals(RepositoryType.NUGET, analyzer.supportedRepositoryType());
+        Assert.assertNotNull(metaModel.getLatestVersion());
+
+        Assert.assertFalse(metaModel.getLatestVersion().contains("-"));
+    }
```
nscuro, line 111: I think this test case is no longer valid.

nscuro:
> @brentos99 Would this approach be acceptable as a global solution [...]?

Sounds fine to me for now.

src/test/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzerTest.java
nscuro: Nvm, misunderstanding on my side.

nscuro (approved):

Issue #3509, opened 2024-02-28
API Patch Project endpoint silently ignore purl value in object form


Maxouwell:
### Current Behavior

When using the PATCH Project endpoint, if the purl is in object format, it ignores the purl value.
If it is in string format, it is processed.
I suspect the object deserialization of PackageURL is failling and just use null instead.

It may affect other API using the PackageURL object.

### Steps to Reproduce

1. Do a curl to /api/v1/project/$UUID with this payload :
```
{
	"uuid": "$UUID",
	"group": "my.group",
	"purl": {
		"scheme": "pkg",
		"type": "npm",
		"namespace": "my.namespace",
		"name": "my-project",
		"version": "VERSION"
	}
}
```
2. Request is accepted, group is updated, purl is not
3. Do a curl to /api/v1/project/$UUID with this payload :
```
{
	"uuid": "$UUID",
	"group": "my.group",
	"purl": "pkg:npm/my.namespace/my-project@VERSION"
}
```
4. Request is accepted, group and purl are updated

### Expected Behavior

Object purl should be accepted, or the swagger should use a string for PackageURL.

### Dependency-Track Version

4.8.x

### Dependency-Track Distribution

Container Image

### Database Server

N/A

### Browser

N/A

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

sebD:
Other affected API endpoints:
PUT /api/v1/project/
POST /api/v1/project/
PUT /api/v1/component/project/{uuid}
POST /api/v1/component/

valentijnscholten:
I'm interested to know why the json format for purl is useful? It's more common to communicate it in string form?

nscuro:
Yeah, TBH this seems to be more of an issue with Swagger indicating that `purl` is an object when in fact it is not.

I don't think there is value in specifying PURLs as object.

Resolving this merely requires adding a `@ApiModelProperty(dataType = "string")` annotation here:

https://github.com/DependencyTrack/dependency-track/blob/8b23434d34e241a82e95ede34ac5ec14399e786f/src/main/java/org/dependencytrack/model/Project.java#L377-L384

sebD:
Sure no problems.
[...]

PR #3512, opened 2024-02-29
Fix type of `purl` fields in Swagger docs

sebD:
### Description
Adds a custom deserializer for pURL properties of Project and Component objects. This deserializer allows to describe pURL either as text value or object value in JSON payloads to better conform to the swagger api definitions.

### Addressed Issue

Fixes: #3509 

### Checklist


nscuro:
Thanks for the PR @sebD. However, I don't believe we should fix the issue this way, see https://github.com/DependencyTrack/dependency-track/issues/3509#issuecomment-1975249229.

Are you happy to revise this PR to apply the suggested change instead? Otherwise please let me know, and I'll raise a separate PR.

nscuro (approved):
Thanks!

PR #3546, opened 2024-03-12
[feat] add the hackage meta analyzer

MangoIV:
### Description

This adds support for pulling information from hackage necessary to offer using a newer version for vulnerable packages. 

### Addressed Issue

solves task 1 in #3545 

### Checklist

- [x] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)
- [x] This PR implements an enhancement, and I have provided tests to verify that it works as intended

Issue #3545, opened 2024-03-12
provide support for hackage.haskell.org and nixpkgs


MangoIV:
### Current Behavior

no support for either

### Proposed Behavior

similar to cargo et al we can propose newer package versions to users of hackage and nixpkgs 

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested
```[tasklist]
### Tasks
- [ ] add support for hackage
- [ ] add support for nixpkgs
```

MangoIV:
nixpkgs support is different because to my knowledge there's no API to request information on per package basis; all you can do is get *all* of the information as follows: 
- request a channel: https://channels.nixos.org/nixpkgs-unstable/packages.json.br
- this returns a 302 where `location` is the latest release of that channel, say https://releases.nixos.org/nixpkgs/nixpkgs-24.05pre595901.7eeacecff44e/packages.json.br
- from there you can download the entire information about nixpkgs

I'm planning to implement a singleton "pattern" that builds the database once upon the first request to it; however, I don't know whether this is entirely a good idea because the thing would basically take a very long time on this first component with a `nixpkgs` purl...

MangoIV:
the issue with the nixpkgs support is that you have to refresh the singleton every now and then, best asynchronously; I don't know the project well, do you have any recommendations?

nscuro:
Haven't tried it myself, but you may be able to simply use the ElasticSearch instance that [search.nixos.org](https://search.nixos.org/) is using. The credentials are public[^1][^2], and you can observe the requests being made to ES using your browser's dev tools:

![image](https://github.com/DependencyTrack/dependency-track/assets/5693141/3e14791a-6439-4dab-b981-61d87f173910)


[^1]: https://github.com/NixOS/nixos-search/blob/7081483b91ca079db329e8d1309ca65adc3e90fb/frontend/src/index.js#L10-L12
[^2]: https://github.com/NixOS/nixos-search/blob/main/frontend/netlify.toml#L4

MangoIV:
I would like to avoid to use this. It is clearly not made for the purpose and nixpkgd is an open source project with limited resources.

PR #3549, opened 2024-03-13
Implement the hackage and nixpkgs meta analyzers

MangoIV:
### Description

This solves part 2 of #3545 

### Addressed Issue

#3545 

### Additional Details

this depends on #3546 

the apache http client utils version 4 don't implement brotli decompression; I redid some of the common http client code to keep this scoped; a future refactoring could remove the extraneous code. 

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/repositories/NixpkgsMetaAnalyzer.java
+++ b/src/main/java/org/dependencytrack/tasks/repositories/NixpkgsMetaAnalyzer.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  *
  * SPDX-License-Identifier: Apache-2.0
- * Copyright (c) Steve Springett. All Rights Reserved.
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
  */
 package org.dependencytrack.tasks.repositories;
 
--- a/src/main/java/org/dependencytrack/tasks/repositories/NixpkgsMetaAnalyzer.java
+++ b/src/main/java/org/dependencytrack/tasks/repositories/NixpkgsMetaAnalyzer.java
@@ -19,6 +19,8 @@
 package org.dependencytrack.tasks.repositories;
 
 import alpine.common.logging.Logger;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import org.apache.hc.client5.http.classic.methods.HttpGet;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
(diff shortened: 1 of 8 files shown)


src/main/java/org/dependencytrack/tasks/repositories/NixpkgsMetaAnalyzer.java
```diff
@@ -0,0 +1,118 @@
...
+        this.baseUrl = DEFAULT_CHANNEL_URL;
+        HashMap<String, String> newLatestVersion = new HashMap<>();
+
+        try (final CloseableHttpResponse packagesResponse = processHttpRequest5()) {
+            if (packagesResponse != null && packagesResponse.getCode() == HttpStatus.SC_OK) {
+                final var entity = packagesResponse.getEntity();
+                if (entity != null) {
+                    // TODO(mangoiv): is this the fastest way we can do this?
+                    final var entityString = EntityUtils.toString(new DecompressingEntity(entity, new BrotliInputStreamFactory()));
+                    final var packages = new JSONObject(entityString).getJSONObject("packages").toMap().values();
```
MangoIV, line 41: this line errors with `NoClassDefFoundError org/brotli/dec/BrotliInputStream`

MangoIV:
what's missing: 
- [ ] invoke the refreshing job every now and then
- [ ] make Brotli decompression not error 
- [ ] the usual chores.

MangoIV:
so if my theory is true, then this is some truly cursed shit: 
- at runtime, the http client library looks for presence of the brotli dec class 
- if it doesn't exist, it doesn't try to decode brotli encoded files but still removes the decoding header and goes on 
- if it does exist, it actually converts it into a decompressing entity and then goes on decoding it

MangoIV:
Hi, I would like some input on where to best invoke the update of the nixpkgs meta analyzer singleton. Thanks in advance!

emil-wire:
@nscuro I don't want to pester you and understand you have a lot on your mind. We'd be grateful if you could provide some guidance at your earliest convenience to make this PR a success and land this feature in DT. Thank you!

nscuro:
@MangoIV @emil-wire Hey, will have a look at this ASAP and provide feedback. Thank you for your contributions! :)

nscuro:
@MangoIV Instead of using a singleton of `NixpkgsMetaAnalyzer`, and updating it every now and then, you could use a read-through cache with "expire-after-write" eviction. The [Caffeine](https://github.com/ben-manes/caffeine) cache library is already available in DT.

So the first time a task tries to access nix versions, it will fetch and parse `packages.json`, and put it in the cache. Other tasks trying to access it are blocked until the cache population finishes. After that, they can be quickly served from said cache. The cache will expire after a given interval (e.g. 1h), and the whole thing repeats.

The benefit of this is that you won't have to deal with concurrency controls such as locking on your own. Also refreshing happens implicitly and as such doesn't require a dedicated scheduled task. 

Because we'd want to refresh the entire package index at once, the entire index would be a single cache entry:

```java
class NixPkgMetaAnalyzer {

    private static final Cache<String, Map<String, String>> CACHE = Caffeine.newBuilder()
            .expireAfterWrite(60, TimeUnit.MINUTES)
            .maximumSize(1)
            .build();

    MetaModel analyze(Component component) {
        Map<String, String> latestVersions = CACHE.get("nixpkgs", cacheKey -> {
            final var versions = new HashMap<String, String>();
            // TODO: Download and parse package index
            return versions;
        });

        // TODO: Do things with latestVersions
    }
```

More details here: https://github.com/ben-manes/caffeine/wiki/Population#manual

MangoIV:
Thank you for the answer, that looks very nice! 

It looks like a fairly large hammer for the tiny nail in trying to get into the wall but if it’s already used across the codebase, this looks like a feasible option; I’ll adjust accordingly.

MangoIV:
https://github.com/package-url/packageurl-java/commit/73b2f510a3ad09461ee52ada13e67a9ebd804dbf

btw; this got merged, i.e. theoretically I could do this minor cleanup; however upstream didn't do a release yet so idk if this is worth the effort.

MangoIV:
tests pass locally, let's see if CI is merciful today :pleading_face:

MangoIV:
@nscuro is this how you'd imagine it? what can I do to make this PR progress?

nscuro (approved):
Thanks @MangoIV!

As a future enhancement, we'd want to improve the parsing of the nixpkgs catalogue. At the moment, the entire thing is deserialized at once, which causes quite a large amount of memory allocations:

image

A more efficient way would be to leverage a streaming API such as the one provided by Jackson. That allows us to only deserialize fields we actually care about. We make use of that when parsing CVE data from the NVD already:

https://github.com/DependencyTrack/dependency-track/blob/3e8f93fb06465cea80723392c4e761ef94e20b8b/src/main/java/org/dependencytrack/parser/nvd/NvdParser.java#L82-L107

MangoIV:
It’s a tradeoff. I would be reluctant to make a guess on whether it’s wise to use streaming here without thorough benchmarks.

Issue #3470, opened 2024-02-15
API /api/v1/analysis - PUT call does not populate analisisDetail 


ArtWachowski:
### Current Behavior

I noticed that analysisDetails text is delivered as expected and populated in Details box but it is not actually "pushed" to the Audit Trail. 

### Steps to Reproduce

1. Create API call ` # Loop through each Vuln UUID 
while IFS= read -r VUUID; do
    # Use the UUID in the curl request
    curl -k -X PUT "$BASE_URL" \
         -H "Content-Type: application/json" \
         -H "X-Api-Key: $API_KEY" \
         -d "{
               \"project\": \"$PROJECT_ID\",
               \"component\": \"$COMPONENT_ID\",
               \"vulnerability\": \"$VUUID\",
               \"analysisState\": \"EXPLOITABLE\",
               \"analysisResponse\": \"NOT_SET\",
               \"analysisDetails\": \"MY TEXT HERE\",
               \"isSuppressed\": false
             }"
done < "$FILE_PATH"`
2. Access WebUI 
3. Investigate Audit Trial section  // observe Details section 

### Expected Behavior

Audit Trial should have a full record from API transaction, AnalysisState works as expected,  AnalysisDetais does not. 

8 Feb 2024 at 14:55:25
Analysis: NOT_SET → EXPLOITABLE

**8 Feb 2024 at 14:55:26
Details: MY TEXT HERE**

**![image (8)](https://github.com/DependencyTrack/dependency-track/assets/47650079/459d6fcb-7043-4bae-92fe-d2b031f7fa0f)**

### Dependency-Track Version

4.9.x

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Browser

Google Chrome

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

surajbora59:
@nscuro , currently their is details Column in Analysis table

<img width="834" alt="image" src="https://github.com/DependencyTrack/dependency-track/assets/51887332/688852ab-2dd1-4c9d-9f95-5a6c5bef8748">

Response:
[...]

PR #3551, opened 2024-03-14
Adds audit trail comments for all fields of new analysis

sebD:
### Description
Adds audit comments for all properties of a newly created analysis entry.

### Addressed Issue
Fixes: #3470

### Additional Details
This fix proposition changes the current behavior a bit when audits are performed manually via the UI. At the moment, setting the Analysis state only adds a comment about this particular field even though justification and vendor response are also set with default values. The proposed fix would add comments for those too. 

This behavior seem coherent with what I understood from the Vex import feature (cf [CycloneDXVexImporter](https://github.com/DependencyTrack/dependency-track/blob/2182706921bd7d3932db1c6dad4aa1737675fa74/src/main/java/org/dependencytrack/parser/cyclonedx/CycloneDXVexImporter.java#L140).

This is a quick fix solution, a better approach would be to create a dedicated AnalysisService that would be called from the CycloneDXVexImporter and the AnalysisResource. Currently these two classes have too many responsabilities. Moreover, they share the analysis creation responsability and do it in a different manner.

I'll be happy to discuss this refactor further if deemed interesting.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/util/AnalysisCommentUtil.java
+++ b/src/main/java/org/dependencytrack/util/AnalysisCommentUtil.java
@@ -28,32 +28,70 @@ public final class AnalysisCommentUtil {
 
     private AnalysisCommentUtil() { }
 
+
+    public static void makeFirstStateComment(final QueryManager qm, final Analysis analysis, final String commenter) {
+        if (analysis.getAnalysisState() != null) {
+            addAnalysisStateComment(qm, analysis, null, analysis.getAnalysisState(), commenter);
+        }
+    }
+
     public static boolean makeStateComment(final QueryManager qm, final Analysis analysis, final AnalysisState analysisState, final String commenter) {
         boolean analysisStateChange = false;
         if (analysisState != null && analysisState != analysis.getAnalysisState()) {
             analysisStateChange = true;
-            qm.makeAnalysisComment(analysis, String.format("Analysis: %s → %s", analysis.getAnalysisState(), analysisState), commenter);
+            addAnalysisStateComment(qm, analysis, analysis.getAnalysisState(), analysisState, commenter);
         }
         return analysisStateChange;
+    }
+
+    private static void addAnalysisStateComment(QueryManager qm, Analysis analysis, AnalysisState before, AnalysisState after, String commenter) {
+        qm.makeAnalysisComment(analysis, String.format("Analysis: %s → %s", before, after), commenter);
+    }
+
+    public static void makeFirstJustificationComment(QueryManager qm, Analysis analysis, String commenter) {
+        if (analysis.getAnalysisJustification() != null) {
+            addAnalysisJustificationComment(qm, analysis, null, analysis.getAnalysisJustification(), commenter);
+        }
     }
 
     public static void makeJustificationComment(final QueryManager qm, final Analysis analysis, final AnalysisJustification analysisJustification, final String commenter) {
         if (analysisJustification != null) {
             if (analysis.getAnalysisJustification() == null && AnalysisJustification.NOT_SET != analysisJustification) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/dependencytrack/util/AnalysisCommentUtil.java
```diff
@@ -28,42 +28,85 @@ public final class AnalysisCommentUtil {
 
     private AnalysisCommentUtil() { }
 
+
+    public static void makeFirstStateComment(final QueryManager qm, final Analysis analysis, final String commenter) {
+        if (analysis.getAnalysisState() != null) {
+            addAnalysisStateComment(qm, analysis, null, analysis.getAnalysisState(), commenter);
```
nscuro, line 34: We should use `NOT_SET` here instead of `null`. `null` should not be exposed to users.

```suggestion
            addAnalysisStateComment(qm, analysis, AnalysisState.NOT_SET, analysis.getAnalysisState(), commenter);
```

src/main/java/org/dependencytrack/util/AnalysisCommentUtil.java
```diff
@@ -28,42 +28,85 @@ public final class AnalysisCommentUtil {
...
         return analysisStateChange;
     }
 
+    private static void addAnalysisStateComment(QueryManager qm, Analysis analysis, AnalysisState before, AnalysisState after, String commenter) {
+        qm.makeAnalysisComment(analysis, String.format("Analysis: %s → %s", before, after), commenter);
+    }
+
+    public static void makeFirstJustificationComment(QueryManager qm, Analysis analysis, String commenter) {
+        if (analysis.getAnalysisJustification() != null) {
+            addAnalysisJustificationComment(qm, analysis, null, analysis.getAnalysisJustification(), commenter);
```
nscuro, line 53:
```suggestion
            addAnalysisJustificationComment(qm, analysis, AnalysisJustification.NOT_SET, analysis.getAnalysisJustification(), commenter);
```

nscuro (changes requested):
Thanks for the PR @sebD! A few suggestions.

> a better approach would be to create a dedicated AnalysisService that would be called from the CycloneDXVexImporter and the AnalysisResource. Currently these two classes have too many responsabilities. Moreover, they share the analysis creation responsability and do it in a different manner.

Indeed it would be great to have this logic centralized. Would you like to take a shot at that in this PR?

src/test/java/org/dependencytrack/resources/v1/AnalysisResourceTest.java
```diff
@@ -346,17 +349,32 @@ public void updateAnalysisCreateNewTest() throws Exception {
         assertThat(response.getStatus()).isEqualTo(HttpStatus.SC_OK);
         assertThat(response.getHeaderString(TOTAL_COUNT_HEADER)).isNull();
 
+        Condition<String> elfBearer = new Condition<String>() {
+            public boolean matches(String value) {
+                return value.contains("toto");
+            }
+        };
```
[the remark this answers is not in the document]
sebD: Oops. Good catch

sebD:
I'll have a look.

sebD:
@nscuro I am currently working on fixing some tests. Do not hesitate to give me any remarks on the proposed solution. I have taken some initiatives that differ a bit from the rest of the app. As I'm a new contributor, I remain open to make any change in the design.

sebD:
@nscuro PR ready for second review round

nscuro:
Thanks @sebD I'll have a look over the coming days. Trying to wrap up remaining work for v4.11 at the moment.

sebD:
No problem. Take your time.

sebD:
@nscuro I've started a new job recently and spent less time following d-track lately. Don't hesitate to ping me if work needs to be done on this code.

PR #3555, opened 2024-03-16
Perform License Resolution On Name Field During SBOM Import

### LINKED ISSUE #3433: SBOM import does not perform License Resolution like it is done when updates are done via the API
### Current Behavior

When uploading an SBOM containing an SPDX compliant license id in the name field (like):

```
license: [
  name: "GPL-3.0-or-later"
]
```

DT does not recognize and resolve the license. It remains an unrecognized text value.

If I instead perform an update of the license via the API and submit an identical value as the "licenseid:" in the call DT will resolve the license and treat it as a known SPDX Id license value capable of triggering Policy Violations.

### Steps to Reproduce

1.
Upload an SBOM file containing a license section for one of the components like:
```
license: [
  name: "GPL-3.0-or-later"
]
```

2.
Go to the UI and see that the value "GPL-3.0-or-later" is shown in white indicating that it is not recognized.

3.
Perform an API call updating the same component record in DT setting the value to what it already was.

4.
Check the UI and see that DT now recognized the license as being one of the known SPDX licenses.

### Expected Behavior

DT should recognize known SPDX license IDs provided in the name: field of the SBOM component.license[].
Preferably it should even be able to match a set of common synonyms (like "GPLv3+")

  

### Dependency-Track Version

4.9.x

Container Image

### Database Server

PostgreSQL

### Database Server Version

_No response_

### Browser

Google Chrome

### Checklist

aravindparappil46:
### Description
On import of SBOMs with SPDX compliant license ID in the `name` field and with no `id` field, license resolution should first occur against standard licenses. 

If none found, only then should license resolution occur against custom licenses.

### Addressed Issue
Fixes #3433 

### Additional Details
On SBOM import, if the license candidate/license option did not have an `id` field and only a `name` field, we were assuming the license to lookup would be a custom license (whereas in fact, the `name` field could contain a standard license ID and not a custom one)

In this change, during BOM upload, if license does not have an `id`, then we call `qm.getLicense()` with the `name` field first. If it returns no results, only then do we call `qm.getCustomLicense()`

Sample SBOM used for testing is attached to this PR:
[tiny-bom.json](https://github.com/DependencyTrack/dependency-track/files/14622478/tiny-bom.json)

Please do let me know if I am doing anything blatantly wrong :) 

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
+++ b/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
@@ -526,9 +526,14 @@ public class ModelConverter {
                         }
                         else if (StringUtils.isNotBlank(cycloneLicense.getName()))
                         {
-                            final License license = qm.getCustomLicense(StringUtils.trimToNull(cycloneLicense.getName()));
+                            final License license = qm.getLicense(StringUtils.trimToNull(cycloneLicense.getName()));
                             if (license != null) {
                                 component.setResolvedLicense(license);
+                            } else {
+                                final License customLicense = qm.getCustomLicense(StringUtils.trimToNull(cycloneLicense.getName()));
+                                if (customLicense != null) {
+                                    component.setResolvedLicense(customLicense);
+                                }
                             }
                         }
                         component.setLicense(StringUtils.trimToNull(cycloneLicense.getName()));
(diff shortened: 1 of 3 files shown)


nscuro (approved):
Great first PR @aravindparappil46, thank you!

PR #3556, opened 2024-03-17
Update License Of Existing Components On BOM Upload

### LINKED ISSUE #3498: Uploading a BOM doesn't update the license of any existing components
### Current Behavior

When a BOM is uploaded that has different license information for a component than previously, that information is ignored.

A BOM upload can _add_ license information to a component that has no license, but it can't remove a license or otherwise change one that is already present. Component licenses can be manually modified in the UI - it's just uploading a BOM that doesn't change them.

There is no relevant information in the logs that I can see at TRACE level. Reproduced with postgres and H2.

### Steps to Reproduce

1. Upload a BOM to a project, with a component that has a license.
2. Change the license in the BOM to a different license (or remove it)
3. Upload the revised BOM

### Expected Behavior

The license in the component reflects what is in the BOM that has been uploaded, regardless of what license information was originally present.

This is what happens when using BOM_PROCESSING_TASK_V2_ENABLED=true on the #3357 branch, so it looks like this has been incidentally fixed by that refactor.

### Dependency-Track Version

4.10.1, also present in today's (2024-02-26) snapshot image _without_ V2 processing enabled.

We didn't notice this in 4.9, so think this is a regression from around then, but haven't verified that.

Container Image

### Database Server

N/A

### Database Server Version

_No response_

### Browser

N/A

### Checklist

aravindparappil46:
### Description

When an SBOM is re-uploaded with different license information on an already imported component, the component should be updated with the new license.

Similarly, if license information previously existed on a component but not on a subsequent SBOM upload, the resolved license should be deleted from the component

### Addressed Issue

Fixes #3498 

### Additional Details

In the V1 BomUploadProcessingTask, we were looking up licenses _only_ if the component does not have a `resolvedLicense` already.

In such a case, if a new SBOM with an updated license for the component was uploaded, `component.getResolvedLicense() == null` will be false, since the component already had a previously resolved license and we would ignore the new license. Hence, removed this null check for `getResolvedLicense()`.

On a similar note, the logic to lookup license is dependent on whether `licenseChoice != null`, which is correct. However, if `licenseChoice == null` , then we should consider that the component does not have a license and hence nullify the previously resolved license (which came from an earlier BOM upload)

The changes in `ModelConverter.java` are:
 - Remove `if (component.getResolvedLicense() == null)` and fix indentation
 - Add an `else` block to `if (licenseChoice != null)`, which sets license fields to null

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
+++ b/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
@@ -513,34 +513,36 @@ public class ModelConverter {
                 licenseOptions.addAll(licenseChoice.getLicenses());
             }
 
-            // try to find a license in the database among the license options, but only if none has been
-            // selected previously.
-            if (component.getResolvedLicense() == null) {
-                for (final org.cyclonedx.model.License cycloneLicense : licenseOptions) {
-                    if (cycloneLicense != null) {
-                        if (StringUtils.isNotBlank(cycloneLicense.getId())) {
-                            final License license = qm.getLicense(StringUtils.trimToNull(cycloneLicense.getId()));
-                            if (license != null) {
-                                component.setResolvedLicense(license);
-                            }
+            // try to find a license in the database among the license options
+            for (final org.cyclonedx.model.License cycloneLicense : licenseOptions) {
+                if (cycloneLicense != null) {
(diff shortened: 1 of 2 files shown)


nscuro (approved):
Another quality PR, thanks @aravindparappil46!

PR #3561, opened 2024-03-18
Generate SARIF File Of Project Vulnerability Findings

### LINKED ISSUE #909: Add SARIF support
SARIF should be able to describe vulnerable components even though its' original purpose was representing source code vulnerabilities.

Recently, the Dependency-Check project implemented SARIF support with the intended use case that the resulting file would be published to GitHub (which now supports SARIF) and can process and display results directly in GitHub.

This enhancement request is to add support for generating SARIF via an API. This would dynamically generate the SARIF (via pebble), and the resulting file would be downloaded.

aravindparappil46:
### Description
If request header `Accept: application/sarif+json` is provided to the GET Finding By Project UUID API, it will now return a SARIF file with the vulnerability findings in that project.

SARIF file is generated based on a pebble template.

### Addressed Issue
Closes #909 

### Additional Details
In the  GET `/v1/finding/project/{uuid}` API, added a new Media Type `application/sarif+json` to the `@Produces`, to indicate the new response `Content-Type`.

Added a `sarif.peb` template file to a new directory under `templates/findings/sarif.peb`

Sample `bom.json` and it's resulting `.sarif` file is attached below:
[bom-and-sarif.zip](https://github.com/DependencyTrack/dependency-track/files/14631413/bom-and-sarif.zip)

I have tried to include most of the fields from `Finding.java` into a meaningful field in the `SARIF` file based on the `SARIF` schema [here](https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0.json). Those fields for which I could not find an adequate mapping I have kept under the `properties` object (fields such as `epssScore`, `severityRank` etc).

Have tested the generated `SARIF` files against the validator [here](https://sarifweb.azurewebsites.net/Validation) with "GitHub Ingestion Rules" enabled as well. Looks like most of the violations are geared towards static analysis results (like, `region`, `helpUri` etc missing).

### Feedback Needed

1. Would appreciate if the `sarif.peb` template is scrutinized so that I am not making any blatant mistakes with the field mappings (from `Finding.java` to the `SARIF` schema). Perhaps there are more meaningful fields in the sarif schema that we could possibly assign a finding field to. 

A picture showing the fields in `Finding.java` and its mapped field on the `SARIF` file is shown below. The green comments are the target fields in the `SARIF` schema.

`results.properties` mean that the finding field is added as a key/value pair into the `properties` object of the schema.

image

2. I have added logic to calculate the `results.level` field of the `SARIF` schema like below:
```
if finding.vulnerability.severity in ('LOW', 'INFO')
  level = "note"
else if finding.vulnerability.severity == 'MEDIUM':
  level = "warning"
else if finding.vulnerability.severity in ('HIGH', 'CRITICAL'):
  level = "error"
else:
  level = "none"
```
Do let me know in case we need to modify it

3. Even though we have set `content-disposition` header in the response, I was not able to confirm that a file named `findings-{{uuid}}.sarif` was indeed being downloaded locally via cURL or Postman testing. It just displays the sarif json on the terminal now. Please let me know in case there is a way to test the actual file download. (I hope we do not have to change the media type to octet stream instead?)

4. For `rules.name`, I have concatenated `finding.component.name` and `finding.vulnerability.cweName`, hoping that it's more friendly to consume for an end-user. Please do let me know if this needs to be changed.

### Checklist

### DIFF
--- a/src/main/resources/templates/findings/sarif.peb
+++ b/src/main/resources/templates/findings/sarif.peb
@@ -1,0 +1,54 @@
+{
+  "version": "2.1.0",
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0.json",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "OWASP Dependency-Track",
+          "version": "{{ dependencyTrackVersion }}",
+          "informationUri": "https://dependencytrack.org/",
+          "rules": [{% for finding in findings %}
+            {
+              "id": "{{ finding.vulnerability.vulnId }}",
+              "name": "{{ finding.component.name }} - {{ finding.vulnerability.cweName }}",
+              "shortDescription": {
+                "text": "{{ finding.vulnerability.description | split('\n') | join(' ') | trim }}"
+              }
+            }{% if not loop.last %},{% endif %}{% endfor %}
+          ]
+        }
+      },
+      "results": [{% for finding in findings %}
+        {
+          "ruleId": "{{ finding.vulnerability.vulnId }}",
+          "message": {
+            "text": "{{ finding.vulnerability.description | split('\n') | join(' ') | trim }}"
+          },
+          "locations": [
+            {
+              "physicalLocation": {
+                "artifactLocation": {
+                  "uri": "{{ finding.component.purl }}"
+                }
+              }
+            }
+          ],
+          "level": {% if ['LOW', 'INFO'] contains finding.vulnerability.severity %}"note",{% elseif finding.vulnerability.severity == 'MEDIUM' %}"warning",{% elseif ['HIGH', 'CRITICAL'] contains finding.vulnerability.severity %}"error",{% else %}"none",{% endif %}
+          "properties": {
+            "name": "{{ finding.component.name }}",
+            "group": "{{ finding.component.group }}",
+            "version": "{{ finding.component.version }}",
+            "source": "{{ finding.vulnerability.source }}",
+            "cweId": "{{ finding.vulnerability.cweId }}",
+            "cvssV3BaseScore": "{{ finding.vulnerability.cvssV3BaseScore }}",
+            "epssScore": "{{ finding.vulnerability.epssScore }}",
+            "epssPercentile": "{{ finding.vulnerability.epssPercentile }}",
+            "severityRank": "{{ finding.vulnerability.severityRank }}",
+            "recommendation": "{{ finding.vulnerability.recommendation | split('\n') | join(' ') | trim }}"
+          }
+        }{% if not loop.last %},{% endif %}{% endfor %}
+      ]
+    }
+  ]
+}
--- a/src/main/resources/templates/findings/sarif.peb
+++ b/src/main/resources/templates/findings/sarif.peb
@@ -8,12 +8,12 @@
           "name": "OWASP Dependency-Track",
           "version": "{{ dependencyTrackVersion }}",
           "informationUri": "https://dependencytrack.org/",
-          "rules": [{% for finding in findings %}
+          "rules": [{% for vuln in uniqueVulnerabilities %}
             {
-              "id": "{{ finding.vulnerability.vulnId }}",
-              "name": "{{ finding.component.name }} - {{ finding.vulnerability.cweName }}",
+              "id": "{{ vuln.vulnId }}",
+              "name": "{{ vuln.cweName }}",
               "shortDescription": {
-                "text": "{{ finding.vulnerability.description | split('\n') | join(' ') | trim }}"
+                "text": "{{ vuln.description | trim }}"
               }
             }{% if not loop.last %},{% endif %}{% endfor %}
           ]
@@ -23,15 +23,15 @@
         {
           "ruleId": "{{ finding.vulnerability.vulnId }}",
           "message": {
-            "text": "{{ finding.vulnerability.description | split('\n') | join(' ') | trim }}"
+            "text": "{{ finding.vulnerability.description | trim }}"
           },
           "locations": [
             {
-              "physicalLocation": {
-                "artifactLocation": {
-                  "uri": "{{ finding.component.purl }}"
+              "logicalLocations": [
+                {
+                  "fullyQualifiedName": "{{ finding.component.purl }}"
                 }
-              }
+              ]
             }
           ],
           "level": {% if ['LOW', 'INFO'] contains finding.vulnerability.severity %}"note",{% elseif finding.vulnerability.severity == 'MEDIUM' %}"warning",{% elseif ['HIGH', 'CRITICAL'] contains finding.vulnerability.severity %}"error",{% else %}"none",{% endif %}
@@ -45,7 +45,7 @@
             "epssScore": "{{ finding.vulnerability.epssScore }}",
             "epssPercentile": "{{ finding.vulnerability.epssPercentile }}",
             "severityRank": "{{ finding.vulnerability.severityRank }}",
-            "recommendation": "{{ finding.vulnerability.recommendation | split('\n') | join(' ') | trim }}"
+            "recommendation": "{{ finding.vulnerability.recommendation | trim }}"
           }
         }{% if not loop.last %},{% endif %}{% endfor %}
       ]
--- a/src/main/resources/templates/findings/sarif.peb
+++ b/src/main/resources/templates/findings/sarif.peb
@@ -6,6 +6,7 @@
       "tool": {
         "driver": {
           "name": "OWASP Dependency-Track",
+          "fullName": "OWASP Dependency-Track - {{ dependencyTrackVersion }}",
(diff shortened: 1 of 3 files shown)


nscuro:
Great work @aravindparappil46!

Looking at the generated SARIF, I noticed that you chose to key `rule`s by vulnerability ID. This can lead to situation where duplicate `rule`s are created, if a vulnerability affects more than one component. For example:

```json
{
  "id": "CVE-2023-6378",
  "name": "logback-classic - Deserialization of Untrusted Data",
  "shortDescription": {
    "text": "A serialization vulnerability in logback receiver component part of  logback version 1.4.11 allows an attacker to mount a Denial-Of-Service  attack by sending poisoned data."
  }
},
{
  "id": "CVE-2023-6378",
  "name": "logback-core - Deserialization of Untrusted Data",
  "shortDescription": {
    "text": "A serialization vulnerability in logback receiver component part of  logback version 1.4.11 allows an attacker to mount a Denial-Of-Service  attack by sending poisoned data."
  }
},
```

Consider removing the component name from the `name` field, that way you can safely reference the same `rule` multiple times, for multiple components. You will need to add some logic to ensure that you're not creating duplicate `rule` as you iterate over the findings, which in turn might require more pre-computation before rendering the template.

The SARIF validator also appears to prefer `rule.name`s in pascal case, e.g.

> [SARIF2012](http://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html): `runs[0].tool.driver.rules[16].name: 'spring-security-crypto - Use of Insufficiently Random Values'` is not a Pascal-case identifier. For uniformity of experience across all tools that produce SARIF, the friendly name should be a single Pascal-case identifier, for example, `ProvideRuleFriendlyName`.

For the `physicalLocation` of `result`s, it is necessary to provide a `region`:

> [GH1003](http://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html): `runs[0].results[0].locations[0].physicalLocation`: The 'region' property is absent. GitHub Advanced Security code scanning can display the correct location only for results that provide a 'region' object with line and optional column information. At minimum, 'region.startLine' is required. 'region' can also provide 'startColumn', 'endLine', and 'endColumn', although all of those have reasonable defaults.

Looking at how Trivy does it, it just sets all fields to `1`: https://github.com/aquasecurity/trivy-action/blob/062f2592684a31eb3aa050cc61e7ca1451cecd3d/test/data/image-sarif.test#L58-L63

OTOH, I'm wondering if we should use *logical* locations instead? https://github.com/microsoft/sarif-tutorials/blob/main/docs/2-Basics.md#physical-and-logical-locations

src/main/resources/templates/findings/sarif.peb
```diff
@@ -0,0 +1,54 @@
...
+        "driver": {
+          "name": "OWASP Dependency-Track",
+          "version": "{{ dependencyTrackVersion }}",
+          "informationUri": "https://dependencytrack.org/",
+          "rules": [{% for finding in findings %}
+            {
+              "id": "{{ finding.vulnerability.vulnId }}",
+              "name": "{{ finding.component.name }} - {{ finding.vulnerability.cweName }}",
+              "shortDescription": {
+                "text": "{{ finding.vulnerability.description | split('\n') | join(' ') | trim }}"
```
nscuro, line 16: For free-text values, consider using `| escape(strategy="json")` to prevent the SARIF JSON from breaking, for example when the `description` includes double quotes.

You can also set the default escaping strategy when you construct the template engine, e.g.:

https://github.com/DependencyTrack/dependency-track/blob/3b52284f98526f533bd8b7b705e707f6f5847102/src/main/java/org/dependencytrack/notification/publisher/WebhookPublisher.java#L28

src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
```diff
@@ -589,6 +593,45 @@ public void getAllFindingsGroupedByVulnerabilityWithAclEnabled() {
...
+        Assert.assertNotNull(runs);
+        JsonObject runsJson = runs.getJsonObject(0);
+        Assert.assertNotNull(runsJson);
+        Assert.assertEquals("OWASP Dependency-Track", runsJson.getJsonObject("tool").getJsonObject("driver").getString("name"));
+        Assert.assertEquals(new About().getVersion(), runsJson.getJsonObject("tool").getJsonObject("driver").getString("version"));
+        Assert.assertNotNull(runsJson.getJsonObject("tool").getJsonObject("driver").getJsonArray("rules"));
+        Assert.assertEquals("Vuln-1", runsJson.getJsonObject("tool").getJsonObject("driver").getJsonArray("rules").getJsonObject(0).getString("id"));
+        Assert.assertNotNull(runsJson.getJsonArray("results"));
+        Assert.assertEquals("error", runsJson.getJsonArray("results").getJsonObject(0).getString("level"));
+        Assert.assertEquals("Vuln-1", runsJson.getJsonArray("results").getJsonObject(0).getString("ruleId"));
```
nscuro, line 632: It would be great to assert the entire content of the JSON output here. That makes it easier to spot regressions later. You can find an example of how this can be done here:

https://github.com/DependencyTrack/dependency-track/blob/3b52284f98526f533bd8b7b705e707f6f5847102/src/test/java/org/dependencytrack/resources/v1/BomResourceTest.java#L552-L661

src/main/resources/templates/findings/sarif.peb
aravindparappil46: Thank you!

src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
aravindparappil46: Thanks a lot @nscuro ! This is definitely a better way. Will update

aravindparappil46:
Thanks a lot for the detailed feedback @nscuro ! 

Regarding `name` field in the SARIF - makes sense to remove the component name from there. Will do some data wrangling so that only unique vulnerabilities are populated under `rules` (plus can pascal-case the `cweName`)

Regarding `physicalLocation` - Yes, I feel going for `logicalLocations.fullyQualifiedName` is better than hacking around with `region` in `physicalLocation`

I'm done with the above changes in my local branch. Have to work on updating the test and will push soon.

aravindparappil46:
Have pushed the changes. More details in the commit messages :) 

Here's an updated SARIF sample:
[bom-and-sarif-v2.zip](https://github.com/DependencyTrack/dependency-track/files/14681183/bom-and-sarif-v2.zip)

The pascal-casing for some CWE names look weird because of characters like hyphens (e.g., `Out-of-boundsWrite` ) but the SARIF [validator](https://sarifweb.azurewebsites.net/Validation) is not complaining anymore.

One thought I had was whether to conditionally add fields to the `results.properties` object. 
For e.x., Right now, even if a vulnerability does not have a `recommendation`, it is still added as an empty string to its `results.properties`. Could add a null check on the `sarif.peb` for each field under `properties` but wanted to check if it makes sense

src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
```diff
@@ -589,6 +596,225 @@ public void getAllFindingsGroupedByVulnerabilityWithAclEnabled() {
...
+            .withMatcher("vuln3Level", equalTo(getSARIFLevelFromSeverity(v3.getSeverity())))
+            .withMatcher("vuln3Recommendation", equalTo(v3.getRecommendation().trim()))
+            .withMatcher("comp1Purl", equalTo(c1.getPurl().toString()))
+            .withMatcher("comp2Purl", equalTo(c2.getPurl().toString()))
+            .withMatcher("comp1Name", equalTo(c1.getName()))
+            .withMatcher("comp2Name", equalTo(c2.getName()))
+            .withMatcher("comp1Group", equalTo(c1.getGroup()))
+            .withMatcher("comp2Group", equalTo(c2.getGroup()))
+            .withMatcher("comp1Version", equalTo(c1.getVersion()))
+            .withMatcher("comp2Version", equalTo(c2.getVersion()))
```
nscuro, line 651: For these you don't need matchers, you can use literals in the JSON instead. Only use matchers for values that you have no control over (e.g. auto-generated UUIDs, `About` values).

nscuro:
> One thought I had was whether to conditionally add fields to the results.properties object.

Honestly I don't know 😅 

It might make sense to omit fields with empty values, on the other hand it might make sense to have properties consistent across all findings. What's your opinion?

aravindparappil46:
Thanks @nscuro ! Have removed the unwanted matchers.

For the empty `properties` -- I am of the opinion that we should keep the `properties` object consistent across all findings, so that the `SARIF` is more predictable and follows a consistent schema. Might even be useful for downstream tools to ingest and display all properties in a uniform manner.

nscuro (approved):
Thanks!

PR #3571, opened 2024-03-20
perf: Handle Trivy scan results asynchronously

### LINKED ISSUE #3570: Upcoming Trivy integration is awesome, but a bit slow
### Current Behavior

Hey! We're really looking forward to the Trivy support added in 4.11.x so we've been testing out the snapshot. We've noticed some poor performance when scanning sboms with larger numbers of components (~600+), in some cases the scan can take 10 minutes. The trivy scan itself only takes a few seconds to return a result.

It's understood that this functionality is unreleased, but I'm keen to take a run at fixing it 👍

Thanks again everyone, really great project.

### Steps to Reproduce

1. Enable Trivy integration using 4.11.0-SNAPSHOT
2. Start an analysis
3. Make coffee
4. Get results

### Expected Behavior

Repro steps minus step 3 🙂 

It would be nice to reduce the scan time as much as possible to facilitate blocking CICD pipelines.

### Dependency-Track Version

4.11.0-SNAPSHOT

Container Image

### Database Server

PostgreSQL

### Database Server Version

15.6

### Browser

N/A

### Checklist

robert-blackman:
Use CompletableFutures with a FixedThreadPool Executor to asynchronously handle each scan result from Trivy

### Description

This change allows the `TrivyAnalysisTask` to handle scan results asynchronously by borrowing the existing `CompletableFutures` / `CountDownLatch` pattern being used in the `SnykAnalysisTask`. The `FixedThreadPool` can be configured using `TRIVY_THREAD_POOL_SIZE` (default: 10).

### Addressed Issue

fixes #3570

### Additional Details

- The check for `SCANNER_TRIVY_IGNORE_UNFIXED` is no longer done for each vulnerability, as this is expensive and seemingly is not necessary. Instead it is checked once per analysis task.

- I wasn't sure if a ForkJoinPool would do better here for the work stealing capabilities?

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/common/ConfigKey.java
+++ b/src/main/java/org/dependencytrack/common/ConfigKey.java
@@ -32,6 +32,7 @@ public enum ConfigKey implements Config.Key {
     SNYK_RETRY_BACKOFF_MULTIPLIER("snyk.retry.backoff.multiplier", 2),
     SNYK_RETRY_BACKOFF_INITIAL_DURATION_MS("snyk.retry.backoff.initial.duration.ms", 1000),
     SNYK_RETRY_BACKOFF_MAX_DURATION_MS("snyk.retry.backoff.max.duration.ms", 60_000),
+    TRIVY_THREAD_POOL_SIZE("trivy.thread.pool.size", 10),
     TRIVY_RETRY_MAX_ATTEMPTS("trivy.retry.max.attempts", 10),
     TRIVY_RETRY_BACKOFF_MULTIPLIER("trivy.retry.backoff.multiplier", 2),
     TRIVY_RETRY_BACKOFF_INITIAL_DURATION_MS("trivy.retry.backoff.initial.duration.ms", 1000),
--- a/src/main/java/org/dependencytrack/tasks/scanners/TrivyAnalysisTask.java
+++ b/src/main/java/org/dependencytrack/tasks/scanners/TrivyAnalysisTask.java
@@ -23,6 +23,7 @@ import alpine.common.logging.Logger;
 import alpine.common.metrics.Metrics;
 import alpine.common.util.UrlUtil;
 import alpine.event.framework.Event;
+import alpine.event.framework.LoggableUncaughtExceptionHandler;
 import alpine.event.framework.Subscriber;
 import alpine.model.ConfigProperty;
 import alpine.security.crypto.DataEncryption;
@@ -33,6 +34,7 @@ import io.github.resilience4j.retry.Retry;
 import io.github.resilience4j.retry.RetryConfig;
 import io.github.resilience4j.retry.RetryRegistry;
 import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.lang3.concurrent.BasicThreadFactory;
 import org.apache.http.HttpHeaders;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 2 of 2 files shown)


nscuro:
Thanks for the PR @robert-blackman! Very much appreciate folks looking into unreleased code, it's the best opportunity to tweak and fix things!

How is the performance holding up when uploading many BOMs? Processing records in a separate thread pool can potentially have the opposite effect. For example, with a worker pool thread of 20, DT could potentially be executing the `VulnerabilityAnalysisTask` up to 20 times concurrently. But now the Trivy thread pool acts as a bottleneck, as all 20 worker threads shove tasks into it.

robert-blackman:
Good point @nscuro! Maybe a better approach (perhaps a naive one 😂) would be to initialise a new executor with each `TrivyAnalysisTask`, terminating it after each analysis? This simple approach wouldn't allow for finer control over the total number of concurrent executions (not to mention the memory overhead), but would prevent tasks from blocking one-another for the most part. Any thoughts?

nscuro:
> Maybe a better approach (perhaps a naive one 😂) would be to initialise a new executor with each TrivyAnalysisTask, terminating it after each analysis?

In that case we may end up creating more contention as threads compete for database connections. One `TrivyAnalysisTask` alone could hog all database connections in the connection pool, depending on how many threads we're allocating per task. You could test this by [building a container image with your changes locally](https://github.com/DependencyTrack/dependency-track/blob/master/DEVELOPING.md#building-container-images), and using it in [the Docker Compose dev setup with monitoring](https://github.com/DependencyTrack/dependency-track/blob/master/DEVELOPING.md#with-monitoring-stack). The Grafana dashboard includes metrics for the database connection pools.

I believe throwing concurrency at the problem might not solve it without creating new issues. Instead, we may want to look at making the code more efficient. One example would be here:

https://github.com/DependencyTrack/dependency-track/blob/35c7f862b2e30815465dde8cb6f691405304472c/src/main/java/org/dependencytrack/tasks/scanners/TrivyAnalysisTask.java#L447

Where we fetch a `Component` from the database, for each reported vulnerability. If multiple vulnerabilities are reported for the same component, we're doing redundant queries. By grouping the vulnerabilities by affected component, we should be able to minimize those queries.

Similarly, `super.isEnabled` is executed once for each vulnerability:

https://github.com/DependencyTrack/dependency-track/blob/35c7f862b2e30815465dde8cb6f691405304472c/src/main/java/org/dependencytrack/tasks/scanners/TrivyAnalysisTask.java#L319

The method executes a DB query behind the scenes, which again can be expensive. Really it only needs to be executed once when the task is started.

There's also logic here that checks whether a vulnerability exists, and creates it if necessary:

https://github.com/DependencyTrack/dependency-track/blob/35c7f862b2e30815465dde8cb6f691405304472c/src/main/java/org/dependencytrack/tasks/scanners/TrivyAnalysisTask.java#L450-L456

This is executed for every Component<->Vulnerability pair. So if the same vulnerability is reported for multiple components, we're doing redundant work.

So, to summarize, there are multiple opportunities here to structure the code in such a way that:

* Each component is only fetched once
* Each vulnerability is only synchronized once
* Configuration (`isEnabled`) is only fetched once

robert-blackman:
Thanks for the feedback @nscuro 🙂 My original approach really revolved around making the smallest change possible in order to meet the release window, but it did seem that the code in general could do with some optimisation. I'll have a go at implementing some form of local cache to handle the proposed optimisations, not sure if you had an idea for how that would best be implemented? A simple map would do the trick, but I'm not sure if something more robust would be warranted, for example caching vulnerabilities across scans 🤔 

I do think (from our usage at least) that concurrency might still be necessary. For example, we are seeing some scans drop from 10-12 minutes down to 1:30-2:30 minutes. There are likely some threadpools out there that are better suited to this use case where fairness can be taken into account, did you have any thoughts on how concurrency might be best achieved here?

New plan:
1. do less things
2. (only then) do things faster 😉

Issue #3580, opened 2024-03-26
VulnDB vulnerabilities are not assigning CVEs as Aliases


LaVibeX:
### Current Behavior

VulnDB is an important vulnerability database that provides great benefit to Dependency Track. Although we can see the issued CVE ID in the References area, it is not used as an alias. 

We should use this feature to add and pair CVEs as aliases.

By adding those CVEs as aliases, we can considerably minimize duplicate noise in DT while also improving the overall efficiency and accuracy of the vulnerability tracking and reporting functionality.

### Proposed Behavior

Paired CVEs should be assigned as aliases to VulnDB vulnerabilities.

![image](https://github.com/DependencyTrack/dependency-track/assets/52439101/2a8d4248-4284-48a4-8427-052f033c986f)

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

nscuro:
Implemented via #3588.

PR #3588, opened 2024-03-28
New feature: VulnDB Aliases!

LaVibeX:
### Description
VulnDB is an important vulnerability database that provides great benefit to Dependency Track. Although we can see the assigned CVE ID in the References section, it is not used as an alias.

This PR takes paired CVE and added as an VulnDB Alias.
image

I am also introducing a ModelConverterTest, which is important to demonstrate the correct conversion between a VulnDB vulnerability object and a Dependency-Track vulnerability object.

### Addressed Issue
This PR address the Issue #3580

### Additional Details
It is important to verify that the string provided as "CVE" is correct. That is why the getValidCveId function was created and tested. Regex will match any introduced string with the correct CVE pattern.

VulnDB vulnerabilities come with two sources of information: the CVSS and nvdAdditionalInfo fields. The last source will always be preferred over the others. However, in case that field is empty, the cveId from the CVSS fields will be used.

### Checklist


src/main/java/org/dependencytrack/parser/common/resolver/CveResolver.java
```diff
@@ -0,0 +1,64 @@
...
+                cveString = "CVE-" + cveString;
+            }
+            // Match the input against the pattern
+            Matcher matcher = regex.matcher(cveString);
+            if (matcher.matches()) {
+                return cveString;
+            }
+        }
+        return null;
+    }
```
nscuro, line 63: Consider moving this as a `static` method to [`VulnerabilityUtil`](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/util/VulnerabilityUtil.java).

src/main/java/org/dependencytrack/parser/common/resolver/CveResolver.java
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @param cveString the input string, potentially containing a CVE ID
+     * @return the valid CVE ID
+     */
+    public String getValidCveId(String cveString) {
+        if (StringUtils.isNotBlank(cveString)) {
+            // Define the regex pattern
+            String pattern = "^CVE-\\d{4}-\\d+$";
+            // Compile the pattern
+            Pattern regex = Pattern.compile(pattern);
```
nscuro, line 51: Re-compiling the pattern for every invocation of this method can get expensive quite fast. It's usually better to make the `Pattern` a `private static final` field, so it's only compiled once.

src/main/java/org/dependencytrack/parser/vulndb/ModelConverter.java
```diff
@@ -247,4 +259,39 @@ public static CvssV3 toNormalizedMetric(CvssV3Metric metric) {
...
+     * Set corresponding Alias to vulnDbVuln
+     * If the input `cve_idString` represents a valid CVE ID, this function sets
+     * the corresponding aliases for the `vuln` object by calling `computeAliases`.
+     *
+     * @param vuln the `Vulnerability` object for which to set the aliases
+     * @param qm the `QueryManager` object used for synchronization
+     * @param cve_idString the string that may represent a valid CVE ID
+     */
+    private static void setAliasIfValid(Vulnerability vuln,QueryManager qm, String cve_idString) {
+        final String cve_id = CveResolver.getInstance().getValidCveId(cve_idString);
```
nscuro, line 272: Please use camelCase for variables instead of snake_case.

src/main/java/org/dependencytrack/parser/vulndb/ModelConverter.java
```diff
@@ -247,4 +259,39 @@ public static CvssV3 toNormalizedMetric(CvssV3Metric metric) {
...
+     */
+    private static List<VulnerabilityAlias> computeAliases(Vulnerability vulnerability, QueryManager qm, String cve_id) {
+        List<VulnerabilityAlias> vulnerabilityAliasList = new ArrayList<>();
+        final VulnerabilityAlias vulnerabilityAlias = new VulnerabilityAlias();
+        vulnerabilityAlias.setVulnDbId(vulnerability.getVulnId());
+        vulnerabilityAlias.setCveId(cve_id);
+        qm.synchronizeVulnerabilityAlias(vulnerabilityAlias);
+        vulnerabilityAliasList.add(vulnerabilityAlias);
+        return vulnerabilityAliasList;
+    }
```
nscuro, line 296: This only ever returns a list of 0-1 items, should it return a single `null`-able `VulnerabilityAlias` instead?

nscuro (changes requested):

src/main/java/org/dependencytrack/parser/vulndb/ModelConverter.java
LaVibeX: that would be ideal but setAlias expects `//The method setAliases(List<VulnerabilityAlias>)`

One option I can imagine is doing the following

```suggestion
    private static VulnerabilityAlias computeAliases(Vulnerability vulnerability, QueryManager qm, String cve_id) {
        final VulnerabilityAlias vulnerabilityAlias = new VulnerabilityAlias();
        vulnerabilityAlias.setVulnDbId(vulnerability.getVulnId());
        vulnerabilityAlias.setCveId(cve_id);
        qm.synchronizeVulnerabilityAlias(vulnerabilityAlias);
        return vulnerabilityAlias ;
    }
```
and modify the method setAliasIfValid as
```java
    private static void setAliasIfValid(Vulnerability vuln,QueryManager qm, String cveString) {
        final String cveId = VulnerabilityUtil.getValidCveId(cveString);
        final List<VulnerabilityAlias> aliases = new ArrayList<>();
        if (cveId != null) {
            aliases.add(computeAlias(vuln,qm,cveId));
            vuln.setAliases(aliases);
        }
    }
```

nscuro (approved):
Thanks @LaVibeX!

Issue #3585, opened 2024-03-27
Add an API endpoint that returns the UUID of the last uploaded BoM for a project


ddresser:
### Current Behavior

There is no API endpoint that, given a project name, returns the UUID of only the latest uploaded BoM for that project.

### Proposed Behavior

A new API endpoint that takes a project name and will return only the latest uploaded project version, including UUID.

I have a use case that I believe is quite common where I want to analyze the most recently uploaded BoM for a project.  For me, this represents the latest version of the BoM, but I understand that sometimes an older BoM may be re-uploaded.  Even in that case, I believe it is common to want to analyze the most recently uploaded version.  Being able to easily retrieve the UUID of the most recently uploaded version would eliminate the need to parse all the versions and compare dates and would make this type of analysis simpler.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

sebD:
I'll have a look

nscuro:
@ddresser I just realized this is already possible using the `/api/v1/project` endpoint:

```shell
curl -H 'X-Api-Key: <API_KEY>' \
    'http://localhost:8080/api/v1/project?name=foo&sortName=lastBomImport&sortOrder=desc&offset=0&limit=1'
```

Essentially you limit the list of returned items to projects with name `foo`, then sort by `lastBomImport` in descending order, and ask for at most one item to be returned (`offset=0`, `limit=1`).

Still, thanks @sebD for working on this!

PR #3616, opened 2024-04-13
3585: Adds new rest end point to fetch the version of a project with the most recent bom upload

sebD:
### Description

Adds new endpoint : /latestUpdated/{name} to fetch the version of a project with the most recent bom update.

### Addressed Issue

#3585 

### Additional Details

I tried to keep the endpoint's name somewhat short, but I remain open for a renaming if anyone has a better idea.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
@@ -61,6 +61,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 
 final class ProjectQueryManager extends QueryManager implements IQueryManager {
@@ -373,6 +374,23 @@ final class ProjectQueryManager extends QueryManager implements IQueryManager {
     @Override
     public PaginatedResult getProjects(final Tag tag) {
         return getProjects(tag, false, false, false);
+    }
+
+    /**
+     * Returns an optional containing the project version with the latest BOM import date.
+     * @param name the name of the project
+     * @return the version of the project with the latest BOM update date
+     */
+    @Override
+    public Optional getLastImportedVersionProject(final String name) {
+        final PaginatedResult result = getProjects(name, false, false, null);
+        return result.getObjects()
+                .stream()
+                .min((o1, o2) -> {
+                    Project p1 = (Project) o1;
+                    Project p2 = (Project) o2;
+                    return p2.getLastBomImport().compareTo(p1.getLastBomImport());
+                });
     }
 
     /**
--- a/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
@@ -383,7 +383,6 @@ final class ProjectQueryManager extends QueryManager implements IQueryManager {
      */
     @Override
     public Optional<Project> getLastImportedVersionProject(final String name) {
-        final PaginatedResult result = getProjects(name, false, false, null);
         Query<Project> query = pm.newQuery(Project.class);
         query.setFilter("name == :name");
         query.setParameters(name);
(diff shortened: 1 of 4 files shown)


src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -375,6 +376,23 @@ public PaginatedResult getProjects(final Tag tag) {
...
+    @Override
+    public Optional getLastImportedVersionProject(final String name) {
+        final PaginatedResult result = getProjects(name, false, false, null);
+        return result.getObjects()
+                .stream()
+                .min((o1, o2) -> {
+                    Project p1 = (Project) o1;
+                    Project p2 = (Project) o2;
+                    return p2.getLastBomImport().compareTo(p1.getLastBomImport());
+                });
```
nscuro, line 393: The same outcome could be achieved in a single database query instead. In JDOQL, it would be (roughly, haven't tested):

```java
Query<Project> query = pm.newQuery(Project.class);
query.setFilter("name == :name");
query.setParameters(name);
query.setOrdering("lastBomImport DESC");
query.setRange(0, 1);
return Optional.ofNullable(query.executeUnique());
```

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -375,6 +376,23 @@ public PaginatedResult getProjects(final Tag tag) {
         return getProjects(tag, false, false, false);
     }
 
+    /**
+     * Returns an optional containing the project version with the latest BOM import date.
+     * @param name the name of the project
+     * @return the version of the project with the latest BOM update date
+     */
+    @Override
+    public Optional getLastImportedVersionProject(final String name) {
```
nscuro, line 385: `Optional` accepts a type argument, so this should be:

```suggestion
    public Optional<Project> getLastImportedVersionProject(final String name) {
```

sebD: Thanks. I wasn't entirely happy with my solution, especially with the need to cast. Yours is more elegant.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -375,6 +376,22 @@ public PaginatedResult getProjects(final Tag tag) {
...
     }
 
+    /**
+     * Returns an optional containing the project version with the latest BOM import date.
+     * @param name the name of the project
+     * @return the version of the project with the latest BOM update date
+     */
+    @Override
+    public Optional<Project> getLastImportedVersionProject(final String name) {
+        final PaginatedResult result = getProjects(name, false, false, null);
```
nscuro, line 386: The `getProjects` call can be removed now as its results are no longer used.

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -141,6 +142,38 @@ public Response getProject(
         }
     }
 
+    @GET
+    @Path("/latestUpdated/{name}")
```
nscuro, line 146: The path and `ApiOperation` description should not include "updated", because a project can be updated in many different ways that do not involve uploading a BOM. Since this endpoint returns a project based on the latest BOM upload timestamp, both the path and the operations description should reflect that.

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
sebD: What name would you have ? Would **latestImportedBom** suit better ? I'm happy to use any wording proposition.

nscuro:
@sebD, I was just about to merge this, but while testing I realized that the desired functionality is already possible using this endpoint:

https://github.com/DependencyTrack/dependency-track/blob/4e9bab5013a1fac200a5b08004fe2179d46d0588/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L79-L100

For example:

```shell
curl -H 'X-Api-Key: <API_KEY>' \
    'http://localhost:8080/api/v1/project?name=foo&sortName=lastBomImport&sortOrder=desc&offset=0&limit=1'
```

Apologies, I should've checked this before asking the community member to raise a ticket for this... Still, thanks so much for working on this and raising a PR! ❤️

sebD:
No worries. I should have also checked. It's still a good way to practice navigating the code base.

PR #3621, opened 2024-04-15
support for experimental configurations

fnxpt:
### Description
Add support for experimental configurations

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/EventSubsystemInitializer.java
+++ b/src/main/java/org/dependencytrack/event/EventSubsystemInitializer.java
@@ -18,14 +18,14 @@
  */
 package org.dependencytrack.event;
 
-import alpine.Config;
 import alpine.common.logging.Logger;
 import alpine.event.LdapSyncEvent;
 import alpine.event.framework.EventService;
 import alpine.event.framework.SingleThreadedEventService;
 import alpine.server.tasks.LdapSyncTask;
 import org.dependencytrack.RequirementsVerifier;
-import org.dependencytrack.common.ConfigKey;
+import org.dependencytrack.model.ConfigPropertyConstants;
+import org.dependencytrack.persistence.QueryManager;
 import org.dependencytrack.tasks.BomUploadProcessingTask;
 import org.dependencytrack.tasks.BomUploadProcessingTaskV2;
 import org.dependencytrack.tasks.CallbackTask;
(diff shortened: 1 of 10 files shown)


src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -108,7 +108,9 @@ public enum ConfigPropertyConstants {
...
     SEARCH_INDEXES_CONSISTENCY_CHECK_CADENCE("search-indexes", "consistency.check.cadence", "4320", PropertyType.INTEGER, "Lucene indexes consistency check cadence (in minutes)"),
-    SEARCH_INDEXES_CONSISTENCY_CHECK_DELTA_THRESHOLD("search-indexes", "consistency.check.delta.threshold", "20", PropertyType.INTEGER, "Threshold used to trigger an index rebuild when comparing database table and corresponding lucene index (in percentage). It must be an integer between 1 and 100");
+    SEARCH_INDEXES_CONSISTENCY_CHECK_DELTA_THRESHOLD("search-indexes", "consistency.check.delta.threshold", "20", PropertyType.INTEGER, "Threshold used to trigger an index rebuild when comparing database table and corresponding lucene index (in percentage). It must be an integer between 1 and 100"),
+    BOM_PROCESSING_TASK_V2_ENABLED("experimental", "bom.processing.task.v2.enabled", "false", PropertyType.BOOLEAN, "Flag to enable BOM UPLOAD V2"),
+    BOM_VALIDATION_ENABLED("experimental", "bom.validation.enabled", "true", PropertyType.BOOLEAN, "Flag to control bom validation");
```
nscuro, line 113: BOM validation is not really an experimental option, it's more that we provide an "escape latch" for users who are currently uploading invalid BOMs without knowing it.

src/main/java/org/dependencytrack/event/EventSubsystemInitializer.java
```diff
@@ -87,11 +87,18 @@ public void contextInitialized(final ServletContextEvent event) {
...
-        } else {
-            EVENT_SERVICE.subscribe(BomUploadEvent.class, BomUploadProcessingTask.class);
+
+        //EXPERIMENTAL: FUTURE RELEASES SHOULD JUST REMOVE THE FOLLOWING BLOCK AND ENABLE THE COMMENT CODE
+        try (QueryManager qm = new QueryManager()) {
+            if (qm.isEnabled(ConfigPropertyConstants.BOM_PROCESSING_TASK_V2_ENABLED)) {
+                EVENT_SERVICE.subscribe(BomUploadEvent.class, BomUploadProcessingTaskV2.class);
+            } else {
+                EVENT_SERVICE.subscribe(BomUploadEvent.class, BomUploadProcessingTask.class);
+            }
```
nscuro, line 97: I feel like adding this check to `BomUploadProcessingTask`, and conditionally delegating to `BomUploadProcessingTaskV2` would be both less disruptive (no risk of losing events), and simpler (only one code area to check vs. two).

nscuro (changes requested):

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
fnxpt: I got it maybe this should be moved to another panel or maybe we should rename this option, but I think it will still be useful to configure it from UI

src/main/java/org/dependencytrack/event/EventSubsystemInitializer.java
fnxpt: Sorry didn't really get what you are advising

nscuro: What I'm suggesting is something like this:

```java
class BomUploadProcessingTask {
    void inform(Event event) {
        if (qm.isEnabled(BOM_UPLOAD_PROCESSING_TASK_V2_ENABLED)) {
            new BomUploadProcessingTaskV2().inform(event);
            return;
        }
        
        // ...
    }
}
```

fnxpt: ok got it, but maybe the other way around... put this code on BomUploadProcessingTaskV2, and declare the subscription against it... this way if we want to remove v1 we just need to remove the if

nscuro (approved):
Thanks!

PR #3648, opened 2024-04-28
Gracefully handle unique constraint violations

nscuro:
### Description

When uploading many BOMs in bulk, it can happen that vulnerability and repository metadata analysis is performed multiple times for the same component identity in parallel. It is more likely to happen when the BOMs being uploaded have big overlap in terms of components they contain.

This can lead to unique constraint violations when DT updates the `COMPONENTANALYSISCACHE` table. Occurrences of this are non-critical, and kind of expected. Instead of letting the task crash, simply "accept" the exception and consider the job done.

Note that DataNucleus will log a warning even when we handle the exception.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


nscuro:
Purposefully not adding tests since the conditions require race conditions in order to be triggered. Such tests have proven to be incredibly flaky. The changes made are low risk.

Issue #3637, opened 2024-04-22
Enhance MS Teams alerts with project name and URL on BOM_PROCESSING_FAILED events


lukas-braune:
### Current Behavior

Currently, the Microsoft Teams alerts for BOM_PROCESSING_FAILED events do not contain any details about the affected project. For example:

![image](https://github.com/DependencyTrack/dependency-track/assets/9955029/b209195a-e946-4a69-a8ee-7e6709a45ceb)

Without this information, it is challenging to locate and identify the project in question.

### Proposed Behavior

Enhance the Microsoft Teams alert for BOM_PROCESSING_FAILED events to include:
- **Project Name**: The name of the affected project that triggered the BOM_PROCESSING_FAILED event.
- **Project URL**: A link to the specific project in Dependency-Track for quick access and review.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

2000rosser:
I'd like to give this issue a try

PR #3666, opened 2024-05-02
Add the project name and project URL to bom processing notifications

2000rosser:
### Description
Previously, the user was only notified of the result of the BOM processing.

Now the user is told the name of the project for each notificication, and for processing errors, they are also given the project URL

### Addressed Issue
#3637 

### Additional Details

This change modifies all the bom processing notification rather than just MS Teams ones. Adding the URL to BOM_PROCESSED might also make sense.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
+++ b/src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
@@ -19,8 +19,10 @@
 package org.dependencytrack.tasks;
 
 import alpine.common.logging.Logger;
+import alpine.common.util.UrlUtil;
 import alpine.event.framework.Event;
 import alpine.event.framework.Subscriber;
+import alpine.model.ConfigProperty;
 import alpine.notification.Notification;
 import alpine.notification.NotificationLevel;
 import org.cyclonedx.BomParserFactory;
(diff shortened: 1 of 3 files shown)


2000rosser:
When I cloned the repo and ran the App, it was using the BomUploadProcessingTask class. I see you also have BomUploadProcessingTaskV2, which is what the unit tests are checking. Which class is the correct one to make the changes?

nscuro:
@2000rosser I don't think you need to touch the code *dispatching* the notification internally at all. The [notification subject](https://github.com/DependencyTrack/dependency-track/blob/7285d90366859dc81f22c21cc821ddb7189672eb/src/main/java/org/dependencytrack/notification/vo/BomProcessingFailed.java#L26-L30) already includes the `project` object, it's just that when transforming this information to the MS Teams Webhook format, it is not included in the output.

When calling `Notification.dispatch`, what happens is that the `Notification` object is sent to [`NotificationRouter`](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/notification/NotificationRouter.java), which rules (i.e. the *Alerts* configured in the UI) apply.

For each applicable rule, it will [invoke the corresponding publisher](https://github.com/DependencyTrack/dependency-track/blob/7285d90366859dc81f22c21cc821ddb7189672eb/src/main/java/org/dependencytrack/notification/NotificationRouter.java#L79-L92) (i.e. email, MS Teams, Slack).

`Publisher`s basically use Pebble templates to format a `Notification` to whatever the destination is expecting. You can see the variables made available in the template context here: https://github.com/DependencyTrack/dependency-track/blob/7285d90366859dc81f22c21cc821ddb7189672eb/src/main/java/org/dependencytrack/notification/publisher/Publisher.java#L83-L143

The template files live [here](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/resources/templates/notification/publisher). I recently added a suite of tests that assert the current output of all publishers. The relevant test case for this issue can be found here:

https://github.com/DependencyTrack/dependency-track/blob/7285d90366859dc81f22c21cc821ddb7189672eb/src/test/java/org/dependencytrack/notification/publisher/MsTeamsPublisherTest.java#L71-L107

You will notice that the corresponding test case for the email publisher *does* include the desired information: https://github.com/DependencyTrack/dependency-track/blob/7285d90366859dc81f22c21cc821ddb7189672eb/src/test/java/org/dependencytrack/notification/publisher/SendMailPublisherTest.java#L144-L178

So ideally you shouldn't have to touch any code at all, except the test that is asserting the current output for MS Teams once the template has been updated.

2000rosser:
@nscuro Thank you for taking the time to explain, it makes a lot more sense now.

nscuro (approved):
Thanks @2000rosser!

PR #3691, opened 2024-05-10
Add active Field To Project Versions

aravindparappil46:
### Description
In order for the frontend to distinguish between active or inactive project versions, we need to include the `active` boolean field to `ProjectVersion` model.

### Addressed Issue
Required for https://github.com/DependencyTrack/frontend/issues/844

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/ProjectVersion.java
+++ b/src/main/java/org/dependencytrack/model/ProjectVersion.java
@@ -34,15 +34,18 @@ public class ProjectVersion implements Serializable {
 
     private String version;
 
+    private Boolean active;
+
     public ProjectVersion() {
         this.uuid = null;
         this.version = null;
+        this.active = null;
     }
 
-    public ProjectVersion(UUID uuid, String version) {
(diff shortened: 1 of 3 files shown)


nscuro:
@aravindparappil46 The coverage check is failing, primarily because the change itself is so tiny. 

We can bypass it, but another solution could be to remove the `set*` methods from the `ProjectVersion` class. DataNucleus can use the constructor to instantiate the class. It seems the missing coverage here comes from the fact that the `setActive` method is never called.

aravindparappil46:
Thank you @nscuro ! I removed the `setActive` which I added and the coverage check is passing now.
Do let me know if you'd prefer removing all of the setters _(I didn't want to refactor parts I didn't touch without confirming)_

nscuro:
> Do let me know if you'd prefer removing all of the setters [...]

I think that would be good to avoid future confusion. You could also simply convert `ProjectVersion` to a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

nscuro (approved):

PR #3720, opened 2024-05-16
Add missing resource to get all components

### LINKED ISSUE #3719: Make components page default show all components
### Current Behavior

Displays No matching records found

This is a dealbreaker to comply with CIS Control 16.4

### Proposed Behavior

Show a list of all components.

### Checklist

starfishfive:
### Description

Added a new get request to components resource for getting all components.

### Addressed Issue

Fix #3719 

### Checklist

- [ ] ~This PR introduces changes to the database model, and I have added corresponding [update logic]~(https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)


src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
```diff
@@ -73,7 +73,29 @@
 @Path("/v1/component")
 @Api(value = "component", authorizations = @Authorization(value = "X-Api-Key"))
 public class ComponentResource extends AlpineResource {
-
+    
+    @GET
+    @Path("/all")
```
valentijnscholten, line 78: Should this be just "/" (or removed)? Other resources return the list of all entities from "/".

src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
```diff
@@ -73,7 +73,29 @@
...
+            notes = "<p>Requires permission <strong>VIEW_PORTFOLIO</strong></p>"
+    )
+    @PaginatedApi
+    @ApiResponses(value = {
+            @ApiResponse(code = 401, message = "Unauthorized"),
+    })
+    @PermissionRequired(Permissions.Constants.VIEW_PORTFOLIO)
+    public Response getComponents(
+        try (QueryManager qm = new QueryManager(getAlpineRequest())) {
+            final PaginatedResult result = qm.getComponents();
```
valentijnscholten, line 94: Can you add ACL checks / filtering, similar to https://github.com/DependencyTrack/dependency-track/blob/db58e69558845a53e3342c0e749d8abf5376e6ab/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L187

valentijnscholten (changes requested):
Can you elaborate on your use-case for retrieving all components of all projects? Are you exporting/importing them into a BI solution?

src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
starfishfive: Yes, I have removed it.

redbaty:
Hey guys, just to chime in my use case here, this would be useful since we have a lot of microservices (which are their own project), that we group with a parentProject

It would be great if either the parentProject automatically includes components from its children, or that we could list all components so that we can keep track of all packages we are currently using.

PR #3736, opened 2024-05-20
Add date format to support offset in nuget analyser

sahibamittal:
### Description

Nuget Analyzer has list of two supported SimpleDateFormat which are allowed for parsing the dates being received from the repository, based on limited test data.

Now, as seen in logs, below date format is also received from the nugget repo. This date has offset dateTime and needs to be added to the list of supported formats in order to avoid the ParseException.

Date example: 1900-01-01T00:00:00+00:00
Format needed: yyyy-MM-dd'T'HH:mm:ss

### Checklist


nscuro:
Thanks @sahibamittal! Could you please rebase with `master` again?

The test suite failed due to #3737, which has now been resolved via #3738.

nscuro (approved):

PR #3746, opened 2024-05-22
Add EPSS conditions to policies

### LINKED ISSUE #3703: Allow Policies to have rules based on EPSS values
### Current Behavior

We are using Policies to trigger notifications about Issues that need to be urgently adressed.

A Policy can be defined based on the Severity of a Vulnerability, but not based on its EPSS value. 

While this is a good start, a vulnerability with medium severity but a high EPSS value might be more urgent to adress than one with high severity but very low EPSS. If we alert on any medium severity issue, we might run into alert fatigue, rendering our efforts moot.

### Proposed Behavior

I would like a new Condition added to Policies that allows operations on the EPSS. For example, you might configure it to only violate the Policy if EPSS greater than 0.5

I could then combine this with a Severity Condition to, for example, alert me if a new Vulnerability is Severity medium and EPSS greater 0.5

This would allow our team to prioritise updates where they are likely to be an issue, while adressing less urgent applications or components later.

### Checklist

2000rosser:
### Description

This change will allow users to use EPSS scores in their policies. 

### Addressed Issue
closes #3703 

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/policy/EpssPolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/EpssPolicyEvaluator.java
@@ -1,0 +1,88 @@
+/*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.policy;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.dependencytrack.model.Component;
+import org.dependencytrack.model.Policy;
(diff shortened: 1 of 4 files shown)


src/main/java/org/dependencytrack/policy/PolicyEngine.java
```diff
@@ -140,7 +141,7 @@ public PolicyViolation.Type determineViolationType(final PolicyCondition.Subject
         }
         return switch (subject) {
             case CWE, SEVERITY, VULNERABILITY_ID -> PolicyViolation.Type.SECURITY;
-            case AGE, COORDINATES, PACKAGE_URL, CPE, SWID_TAGID, COMPONENT_HASH, VERSION, VERSION_DISTANCE ->
+            case AGE, COORDINATES, PACKAGE_URL, CPE, SWID_TAGID, COMPONENT_HASH, VERSION, VERSION_DISTANCE, EPSS ->
```
nscuro, line 144: As EPSS is a metric related to vulnerabilities, I think it should map to `PolicyViolation.Type.SECURITY`.

src/main/java/org/dependencytrack/policy/EpssPolicyEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+
+import org.dependencytrack.model.Component;
+import org.dependencytrack.model.Policy;
+import org.dependencytrack.model.PolicyCondition;
+import org.dependencytrack.model.Vulnerability;
+
+import alpine.common.logging.Logger;
+
+public class EpssPolicyEvaluator extends AbstractPolicyEvaluator {
+    private static final Logger LOGGER = Logger.getLogger(CpePolicyEvaluator.class);
```
nscuro, line 33: This should refer to `EpssPolicyEvaluator.class`.

nscuro (changes requested):
Thanks @2000rosser! Two minor changes needed, otherwise looks good!

nscuro (approved):
Thanks!

PR #3761, opened 2024-05-27
Search component by group

### LINKED ISSUE #3749: Search on group field
### Current Behavior

When we search on the "Components" tab, it only matches the components by name. It makes the search incomplete, as some components use generic names. For example, if you search "angular" the package "@angular/core" isn't returned because "@angular" is the group and "core" is the name.

### Proposed Behavior

Search by name and group so, in my example, searching "angular" would return "@angular/core" package.

### Checklist

rcsilva83:
### Description

On the "Components" tab, search components also by group. Today it only searches by name but some components use generic names and differentiate themselves by group. For example, @angular/corehas group @angular and name core, so searching for "angular" I don't get all the Angular packages on the results.

### Addressed Issue

Closes #3749

### Checklist

Obs: Couldn't find any documentation about component search


nscuro (approved):
Thanks!

PR #3797, opened 2024-06-02
🚧 WIP: Add Project Tags To Findings

aravindparappil46:
> ⚠️ **NOTE**: This is a Draft PR. Need help figuring out how to filter `Findings` by tags

### Description
In order to display tags in the vulnerability audit page, added tags to the resultant finding object.

### Addressed Issue
Partially addresses frontend issue: https://github.com/DependencyTrack/frontend/issues/849

### Additional Details
Currently, this PR just adds the `tags` to the `/finding` API response (used by the Vulnerability Audit page to display it)

**Help Needed**
Need some help figuring out how to filter `Finding` by `tags` 🤔 

I understand that filtering is currently done in `FindingsSearchQueryManager.processFilters()`, but unsure how to edit the SQL query to be able to filter by project tags.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/FindingsSearchQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/FindingsSearchQueryManager.java
@@ -28,6 +28,7 @@ import org.dependencytrack.model.Component;
 import org.dependencytrack.model.ConfigPropertyConstants;
 import org.dependencytrack.model.Finding;
 import org.dependencytrack.model.GroupedFinding;
+import org.dependencytrack.model.Project;
 import org.dependencytrack.model.RepositoryMetaComponent;
 import org.dependencytrack.model.RepositoryType;
 import org.dependencytrack.model.Vulnerability;
@@ -128,6 +129,10 @@ public class FindingsSearchQueryManager extends QueryManager implements IQueryMa
             final Finding finding = new Finding(UUID.fromString((String) o[29]), o);
             final Component component = getObjectByUuid(Component.class, (String) finding.getComponent().get("uuid"));
             final Vulnerability vulnerability = getObjectByUuid(Vulnerability.class, (String) finding.getVulnerability().get("uuid"));
+            final Project project = component.getProject();
+            if (project != null) {
+                finding.getComponent().put("tags", project.getTags());
+            }
             final Analysis analysis = getAnalysis(component, vulnerability);
             final List<VulnerabilityAlias> aliases = detach(getVulnerabilityAliases(vulnerability));
             aliases.forEach(alias -> alias.setUuid(null));


nscuro:
> Need some help figuring out how to filter Finding by tags 🤔
>
> I understand that filtering is currently done in FindingsSearchQueryManager.processFilters(), but unsure how to edit the SQL query to be able to filter by project tags.
> 
> I see that Tag is a child table of Parent, so guessing we need some LEFT JOIN magic in Finding.QUERY_ALL_FINDINGS 🪄 🧠

I would advise against `LEFT JOIN`s for 1:N relationships like the `Project` <-> `Tag` one, since it will cause duplicate rows for each tag.

This should be solvable with a simple `EXISTS` subquery, for example: 

```sql
EXISTS (
  SELECT 1
    FROM "TAG"
   INNER JOIN "PROJECTS_TAGS"
      ON "PROJECTS_TAGS"."TAG_ID" = "TAG"."ID"
   WHERE "PROJECTS_TAGS"."PROJECT_ID" = "PROJECT"."ID"
     AND "TAG"."NAME" = 'foo'
)
```

Similar to how it's done for the portfolio ACL check:

https://github.com/DependencyTrack/dependency-track/blob/f785fc5998ebc43ec18810ae04c33f11a59967e9/src/main/java/org/dependencytrack/persistence/FindingsSearchQueryManager.java#L351-L378

src/main/java/org/dependencytrack/persistence/FindingsSearchQueryManager.java
```diff
@@ -128,6 +129,10 @@ public PaginatedResult getAllFindings(final Map<String, String> filters, final b
             final Finding finding = new Finding(UUID.fromString((String) o[29]), o);
             final Component component = getObjectByUuid(Component.class, (String) finding.getComponent().get("uuid"));
             final Vulnerability vulnerability = getObjectByUuid(Vulnerability.class, (String) finding.getVulnerability().get("uuid"));
+            final Project project = component.getProject();
+            if (project != null) {
+                finding.getComponent().put("tags", project.getTags());
+            }
             final Analysis analysis = getAnalysis(component, vulnerability);
```
valentijnscholten, line 136: Should it be made more clear that these are **project** tags and not finding tags? A future DT version might have finding level tags. Name the field `project-tags`?

aravindparappil46: @valentijnscholten Totally agree. Will make the change!

valentijnscholten:
There's also users that use the `parent` construct to create a hierarchy of projects. How would that be handled? Might be a lot easier if tags can be set on findings and filter on those. The risk is that you get a LOT of tag-finding relationship entries if you have lots of findings with tag `java` for example.

nscuro:
> @valentijnscholten There's also users that use the parent construct to create a hierarchy of projects. How would that be handled?

Good question. How would you *expect* it to be handled? At the moment there's not a lot of "inheritance" logic for the parent-child construct.

Issue #1674, opened 2022-05-30
Add ability to tag projects on BOM upload


UgniusV:
It would be useful for automation if projects could be tagged on BOM upload.

### Current Behavior:
Currently, when uploading BOMs via `PUT /v1/bom`, we can specify the following properties:
- projectName
- projectVersion
- autoCreate
- bom

This works fine, but in order to tag a project we need to make one more request to `POST /api/v1/project` with the following payload:

```json
{
   "uuid":"ad143ba-1c1d-40ca-8054-2cf57f101206",
   "name":"SampleProject",
   "version":"2022-05-30 11:18:08",
   "classifier":"APPLICATION",
   "tags":[
      {
         "name":"production"
      }
   ],
   "active":true
}
```
This is cumbersome for automation processes, it would be nice if we could upload BOM & tag the project with a single request.

### Proposed Behavior:
Add `projectTag` as an accepted field to `PUT /v1/bom`.

Inside `BomResource`  class `PUT /v1/bom` handler - accept a tag field & use it when creating a project.
Currently, we have the following code: `null` is being passed in as a `tag` parameter to `qm.createProject`

```java
if (hasPermission(Permissions.Constants.PORTFOLIO_MANAGEMENT) || hasPermission(Permissions.Constants.PROJECT_CREATION_UPLOAD)) {
    project = qm.createProject(StringUtils.trimToNull(request.getProjectName()), null, StringUtils.trimToNull(request.getProjectVersion()), null, null, null, true, true);
    //TODO - If portfolio access control is enabled, retrieve the principal (ApiKey only) and automatically grant access to the project to the team the key belongs to.
}
``` 

It would be better to use `null` for a tag when no tag is provided by the user, but if the tag is provided - use it.
E.g. pseudocode
```java
project = qm.createProject(StringUtils.trimToNull(request.getProjectName()), null, StringUtils.trimToNull(request.getProjectVersion()), request.getProjectTags(), null, null, true, true);
```

Cheers

valentijnscholten:
I like this proposal, but why not support `projectTags` (so plural)?

UgniusV:
I agree - providing multiple project tags would be even better

UgniusV:
I personally think that option number three is the best one. This way the behavior is consistent with the `/api/v1/project` endpoint. E.g if we post the following payload to `POST /api/v1/project`:
```json
{
   "uuid":"ad143ba-1c1d-40ca-8054-2cf57f101206",
[...]

PR #3843, opened 2024-06-13
feat: autocreate project with tags

JCHacking:
### Description
To be able to upload a bom and auto-create the project with specific tags

### Addressed Issue
https://github.com/DependencyTrack/dependency-track/issues/1674

### Additional Details
I have added the optional property to the PUT method to upload BOM files, I have not programmed much in Java so any suggestion to improve this PR is welcome.
Thanks!!!

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/vo/BomSubmitRequest.java
+++ b/src/main/java/org/dependencytrack/resources/v1/vo/BomSubmitRequest.java
@@ -18,6 +18,7 @@
  */
 package org.dependencytrack.resources.v1.vo;
 
+import org.dependencytrack.model.Tag;
 import alpine.common.validation.RegexSequence;
 import alpine.server.json.TrimmedStringDeserializer;
 import com.fasterxml.jackson.annotation.JsonCreator;
@@ -28,6 +29,7 @@ import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Pattern;
+import java.util.List;
 
 /**
  * Defines a custom request object used when uploading bill-of-material (bom) documents.
@@ -51,6 +53,8 @@ public final class BomSubmitRequest {
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The project version may only contain printable characters")
     private final String projectVersion;
 
+    private final List<Tag> projectTags;
+
     @Pattern(regexp = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", message = "The parent UUID must be a valid 36 character UUID")
     private final String parentUUID;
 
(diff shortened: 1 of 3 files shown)


valentijnscholten:
I like it. Can you add a test with tags populated?

JCHacking:
Sure! Tests added

src/main/java/org/dependencytrack/resources/v1/vo/BomSubmitRequest.java
```diff
@@ -51,6 +53,8 @@ public final class BomSubmitRequest {
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The project version may only contain printable characters")
     private final String projectVersion;
 
+    private final List<Tag> projectTags;
```
nscuro, line 56: We should probably accept a list of `String` here instead. From the client's perspective, passing this:

```json
{
  "projectTags": ["foo", "bar"]
}
```

is more straightforward than:

```json
{
  "projectTags": [
    {
      "name": "foo"
    },
    {
      "name": "bar"
    }
  ]
}
```

src/main/java/org/dependencytrack/resources/v1/vo/BomSubmitRequest.java
```diff
@@ -110,6 +117,10 @@ public String getProjectVersion() {
         return projectVersion;
     }
 
+    public List<Tag> getProjectTags() {
```
nscuro, line 120: Can you add a `@Schema(description = "")` here, where `description` explains that the tags defined here will **overwrite** any tags on existing projects that were manually applied?

It was a discussion point in #1674 so I believe the API docs should call out the behavior.

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/vo/BomSubmitRequest.java
JCHacking: I agree, but I also think that using Tag keeps the API consistent, as in other methods this happens.

For example:
```json
PUT /v1/project

{
  "tags": [
    {
      "name": "LNsN}MPtIehaapsLeihSsMtMctLN{_aNcipcseIaLSWaNs}t_IesSSSsicpe_hIWtL_McIhs{L}P_SPc}S_M}{S}peWhcP_PeWhs"
    }
  ],
  ...
}
```


Do I make the change to make it easier or do I leave it like this to make the api more consistent?

JCHacking:
I see that the test has failed, I imagine that it is because it is not the same object since one is the tags only with name and the others are the tags obtained from the persistence that will have more properties.

I should only make an assertEquals of the list of strings with the name of the tags, right?

src/main/java/org/dependencytrack/resources/v1/vo/BomSubmitRequest.java
nscuro: Good point. Let's leave it consistent then.

nscuro:
> I should only make an assertEquals of the list of strings with the name of the tags, right?

Yes. Consider using AssertJ's fluent assertions, like so:

```java
assertThat(project.getTags())
    .extracting(Tag::getName)
    .containsExactlyInAnyOrder("foo", "bar");
```

JCHacking:
Fixed it with your suggestion, thanks!!

nscuro (approved):
Thanks!

Issue #3897, opened 2024-06-30
Schema validation error when importing SBOMs that were exported by DT itself


malice00:
### Current Behavior

When exporting an SBOM for a project and then trying to import it again, DT gives an exception that the Schema validation failed.

### Steps to Reproduce

1. Select a project
2. Go to the components tab
3. Click 'Download BOM' -- both 'Inventory' and 'Inventory with Vulnerabilities' trigger the problem
4. Click 'Upload BOM' and select the just saved JSON

### Expected Behavior

The BOM should be imported without problems.

### Dependency-Track Version

4.11.4

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Database Server Version

14.9

### Browser

Mozilla Firefox

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

nscuro:
Can you share the error messages you're getting for the upload requests? The response body will list the things it found to be wrong.

Our tests of the export functionality do validate the generated BOMs, but of course it all comes down to the data being exported. Knowing what failed would help to reproduce and fix.

malice00:
There is nothing in the logs (at least not the ones I have access to), but I was able to get the response by manually executing a POST. Hope this helps...

[response_1719748385096.json](https://github.com/user-attachments/files/16044217/response_1719748385096.json)

nscuro:
That does help, it all seems to be license related. Thanks for providing the sample!

malice00:
I figured it out! We have added some licenses and set those on components. The export does add these in the SBOM, but the import apparently only validates against the official licenses, which causes this exception!

nscuro:
Yeah I think we need to populate the `license.name` field rather than `license.id` for custom licenses. Since the CycloneDX schema strictly requires valid SPDX license IDs in the `license.id` field.

We already match `license.name` against custom licenses when *importing* BOMs, so everything should just continue to work if we make this small change.

PR #3915, opened 2024-07-03
Set license name instead of ID when using custom license

2000rosser:
### Description
A valid SPDX is required when using the ID field for licenses. If the license is a custom license, use the name field instead.

### Addressed Issue
#3897 

### Checklist


2000rosser:
I actually don't think the unit test I added is necessary, will remove it.

nscuro:
Thanks @2000rosser!

Do you think you could add a test for this scenario here? https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/resources/v1/BomResourceTest.java

Have a look at the existing export tests, and either extend them or add a smaller test case that's scoped to only this problem.

We want to ensure that, given a component with custom license, the exported BOM looks like expected (`assertThatJson`...), and passes schema validation (`assertThatNoException().isThrownBy(() -> CycloneDxValidator.getInstance().validate(jsonResponse.getBytes()));`).

2000rosser:
@nscuro Sure, I'll take a look and add a test for it

Issue #322, opened 2019-04-16
Portfolio Notification: Support for "Summary" Group


msymons:
### Issue Type:
  - [ ] defect report
  - [X] enhancement request
  
In Dependency Track v3.4.0 there are two types of Portfolio Notification Groups that are related to alerting on new vulnerabilities (ie, ignoring AUDIT Groups):

* NEW_VULNERABILITY
   * 1 component
   * 1 vulnerability
   * 1 or more affected projects
* NEW_VULNERABLE_DEPENDENCY
   * 1 project
   * 1 component
   * 1 or more vulnerabilities

This can result in a LOT of verbosity.  To illustrate,  creating a new project that uses dropwizard 1.25 results in over 100 notification just relating to vulnerabilities in Jetty and Jackson.  

Consequence?  A big risk that everyone receiving notifications immediately creates a filter (eg for email) to delete anything sent by Dependency-Track, or a rule to move it to a folder (where stuff still risks getting overlooked because there is so much to wade through).

I think that it might help if DT were to implement a new Notification Group that would allow for sending out a summary email.  eg, something like

```
Project:   Name: Version
New Vulnerable Components:  26
New Vulnerabilities: 108
Increase in Inherited Risk Score:  748

```
...ideally with a link back to the project.

I use email to illustrate, but I think that this enhancement would also be useful in Slack

stevespringett:
In its current form, notification are a stream of in-memory events. Once the events have been consumed by any event listeners that care, they are quickly GCed. They are not persistent and reside only in memory.

This enhancement would require storing notification preferences, storage of event summary and details in a time-series data structure so the system could quickly determine what events have occurred since the last time a summary notification occurred for a given user - accounting for permissions and subscription preferences.

There's a lot of complexities here and a lot of overhead required to to fit into the existing architectural pattern. This might make sense to implement post v4.0. As a workaround, it would likely be elementary to consume webhooks external to the system, store then for 24 hours, and send out summaries by reading the the payloads received since the last purge.

msymons:
@stevespringett,  thanks for the feedback.  I have created a task to get webhooks tested out per your suggestion.

Would it be just as much work for DT to have a totally new notification which I might call:

VULNERABILITY_FIRST_AFFECT

...which would fire once and once only, on the first time that the portfolio is impacted by a vulnerability?  The usage would be to tell the receiver "Here's something that needs triaging that you have never seen before in any project".

[...]

PR #3925, opened 2024-07-07
Support for summarized and scheduled notifications

MM-msr:
### Description

Currently, Dependency-Track Alerts send a single notification for every single event happening that an alert is subscribed to (e.g. NEW_VULNERABILITY). This can lead to a lot of emails for users and creates the risk of overlooking important notifications.

To improve the user experience of the Alerts, scheduled notifications are implemented in Dependency-Track to send summarized notifications of new events between the last and current scheduled notification. The schedule is defined as cron expression.

Scheduled notification management is available via API and Frontend under Administration -> Notifications -> Scheduled Alerts, which is introduced in the Frontend Pull Request [reference will be added after creation].

The PR includes default templates for console and email publishing. The needed UI-Changes are available in the Frontend-PR at DependencyTrack/frontend#903.

### Addressed Issue

#322

### Additional Details

- DEFAULT_SCHEDULED_CRON_EXPRESSION as environment variable for the default cron expression of new rules
- Error: "The type javax.validation.Payload cannot be resolved. It is indirectly referenced from required type com.github.packageurl.validator.PackageURL" was introduced in my local development environment after syncing the main project and the transition of some javax packages to jakarta
  - was fixed by manually adding javax.validation-api to pom.xml as depencency
- While evaluating and debugging the consideration of child projects in the notification process, it has turned out that projects delivered by QueryManager.getAllProjects() and some other methods didn't consistently include the correct child references. To workaround this behavior, the notification task "SendScheduledNotificationTask" uses a project-by-project-retrieval approach in "evaluateAffectedProjects(qm, rule)". This worked out empirically to deliver consistent child references.
- DefaultPublisher-Retrieval was changed from by-class to a combination of name + class (or by-enum on a higher layer) to support multiple default publishers with same publisher class. This change is needed to use the existing publisher classes without defining new ones for scheduled default publishers.
---
- There might be an open bug currently in the `toJson`-Process for PolicyViolation, where the policy is not fully included in the PolicyCondition and thus the `Policy.getUUID().toString()` call fails with NullRefException. Often it worked for me just fine, sometimes not. Sometimes fixed with creating a new policy, sometimes with reboot of the application. With some additional random exceptions with the H2 database on ORM-Queries, i don't know for sure, if my code is the reason for this behavior or the instability of H2 with the ORM-mapper in my development environment.

Getting violations for scheduled mail:
> `PolicyQueryManager, L. 340, getPolicyViolations(Project, boolean, ZonedDateTime)`

JSONify the data:
> `NotificationUtil, L. 550ff., toJson(PolicyViolation) -> toJson(pv.getPolicyCondition) -> toJson(pc.getPolicy) -> policy.getUUID().toString() -> [EXCEPTION]`

Please test the expected behavior in a more robust environment (docker, other infrastructures, etc.) to validate

#### Screenshots

Administration_CreateScheduledAlert_Popup
Administration_CreateScheduledPublisher_Popup
Administration_ScheduledAlerts
ScheduledMail_Vulnerabilities

### Checklist

- ~This PR introduces changes to the database model, and I have added corresponding [update logic]~(https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)


MM-msr:
@nscuro After i finally found some time this week to restart the PR process, i cherry-picked all necessary commits from my previous branch, corrected all sign-offs and so on. From the coding side it seems to be fine for now, although additional testing from different perspectives should be done.

Please note in particular the missing javax.validation reference in "Additional details", that seems to be ignored in the current master and seems to be a personal problem of my workspace or an unseen error until now from the transition to jakarta. There may be some more topics to discuss, but the main implementation should be completed by now.

The branch for the PR is based on the same commit my other branches started, but merged with a master newer than this feature branch was created on. Because of this and the recent cherry-picks, the commit history is a bit mixed up with other commits, but that may be gone when the feature branch gets updated with the missing commits from the current master branch.

Please provide me with additional information for the errors of the Static Code Analysis as i cannot see any useful information in the details link. If there is anything else i need to do or you need/want to know, please get in touch with me. I will assist to the best of my abilities.

valentijnscholten:
This looks very interesting. But so does the commit history ;-) Could you rebase it onto current master to make the code changes in the PR "reviewable"?

MM-msr:
@valentijnscholten Yeah, i know :D In my git graph, all looked fine, after the PR i had this 😅 I'll try to rebase it and see, if it's possible without making it worse. After the first PR-attempt went terribly wrong, i'm a bit scared tbh 😆 But i will give it a try and see where i land. Maybe it helps already in the mean time to update the feature branch with the commits it is behind to resolve some "false-included" commits in this PR?

nscuro:
> Because of this and the recent cherry-picks, the commit history is a bit mixed up with other commits, but that may be gone when the feature branch gets updated with the missing commits from the current master branch.

Let me push the latest changes from `master` to that branch.

> Please provide me with additional information for the errors of the Static Code Analysis as i cannot see any useful information in the details link.

Unfortunately Codacy requires us to manually enable each "base" branch for analysis. Unless that is done, analysis will fail immediately. I have now enabled the feature branch, so the check should work for the next push.

nscuro:
Updated the branches for `dependency-track` and `frontend` with the latest `master`. You may need to close and reopen the PR since GitHub doesn't update PRs automatically when the base branch gets updated.

MM-msr:
@valentijnscholten Should be "fixed" by now, thanks to @nscuro :) The update did thankfully what i hoped for.

The current CI test exceptions in `BomUploadProcessingTaskTest.java` seems to be caused by other stuff, that this PR does not affect, but if i can provide any support, please let me know.

nscuro:
> The current CI test exceptions in BomUploadProcessingTaskTest.java seems to be caused by other stuff, that this PR does not affect, but if i can provide any support, please let me know.

You got lucky and ran into a flaky test 😅 Re-ran the workflow and it completed successfully this time.

src/main/java/org/dependencytrack/notification/ScheduledNotificationTaskManager.java
```diff
@@ -0,0 +1,75 @@
...
+        scheduleNextRuleTask(ruleUuid, schedule, schedule.nextDuration(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
+    }
+
+    public static void scheduleNextRuleTaskOnce(UUID ruleUuid, long customDelay, TimeUnit delayUnit){
+        scheduleNextRuleTask(ruleUuid, null, customDelay, delayUnit, () -> cancelActiveRuleTask(ruleUuid));
+    }
+
+    private static void scheduleNextRuleTask(UUID ruleUuid, Schedule schedule, long customDelay, TimeUnit delayUnit, Runnable actionAfterTaskCompletion){
+        var scheduledExecutor = Executors.newSingleThreadScheduledExecutor();
+        var futureTask = new ActionOnDoneFutureTask(new SendScheduledNotificationTask(ruleUuid), actionAfterTaskCompletion);
```
nscuro, line 53: You can use `CallbackEvent`s to execute arbitrary code when processing of a given `Event` completed successfully. You can see it being used here to decrement a `CountdownLatch`:

https://github.com/DependencyTrack/dependency-track/blob/af4461e86ae56eaee193017ec86ee2a690709932/src/main/java/org/dependencytrack/tasks/metrics/PortfolioMetricsUpdateTask.java#L78-L86

src/main/java/org/dependencytrack/model/scheduled/vulnerabilities/VulnerabilityDetailsInfo.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Date;
+import org.dependencytrack.model.Finding;
+import org.dependencytrack.util.ScheduledUtil;
+
+/* 
+ * Part of the ScheduledPolicyViolationsIdentified Template Models.
+ * Contains detailed information about the vulnerability/finding it represents.
+ * This class helps to extract the relevant data from the Finding object, to make it easier to use in the template.
+ */
+public class VulnerabilityDetailsInfo {
```
nscuro, line 30: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/scheduled/vulnerabilities/VulnerabilityOverview.java
```diff
@@ -0,0 +1,98 @@
...
+import org.dependencytrack.model.Project;
+import org.dependencytrack.model.Severity;
+import org.dependencytrack.model.Vulnerability;
+import org.dependencytrack.persistence.QueryManager;
+
+/* 
+ * Part of the ScheduledNewVulnerabilitiesIdentified Template Models.
+ * Contains several key metrics to provide an overview of the new vulnerabilities identified.
+ */
+public final class VulnerabilityOverview {
```
nscuro, line 36: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/scheduled/vulnerabilities/VulnerabilitySummary.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.dependencytrack.model.Finding;
+import org.dependencytrack.model.Project;
+import org.dependencytrack.persistence.QueryManager;
+
+/* 
+ * Part of the ScheduledNewVulnerabilitiesIdentified Template Models.
+ * Contains packed summarized informations about the identified new vulnerabilities, grouped by the affected projects.
+ */
+public final class VulnerabilitySummary {
```
nscuro, line 33: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/scheduled/vulnerabilities/VulnerabilitySummaryInfo.java
```diff
@@ -0,0 +1,69 @@
...
+import org.dependencytrack.model.Finding;
+import org.dependencytrack.model.Severity;
+import org.dependencytrack.model.Vulnerability;
+import org.dependencytrack.persistence.QueryManager;
+
+/* 
+ * Part of the ScheduledNewVulnerabilitiesIdentified Template Models.
+ * Contains detailed information about the amount of the identified new vulnerabilities grouped by their severity.
+ */
+public final class VulnerabilitySummaryInfo {
```
nscuro, line 34: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/scheduled/vulnerabilities/VulnerabilityDetails.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.dependencytrack.model.Finding;
+import org.dependencytrack.model.Project;
+import org.dependencytrack.persistence.QueryManager;
+
+/* 
+ * Part of the ScheduledNewVulnerabilitiesIdentified Template Models.
+ * Contains packed summarized informations about the new vulnerabilities identified, grouped by the affected projects.
+ */
+public final class VulnerabilityDetails {
```
nscuro, line 33: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/scheduled/policyviolations/PolicyViolationDetails.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Map;
+
+import org.dependencytrack.model.PolicyViolation;
+import org.dependencytrack.model.Project;
+
+/* 
+ * Part of the ScheduledPolicyViolationsIdentified Template Models.
+ * Contains all detailed policy violations, grouped by the affected projects.
+ */
+public final class PolicyViolationDetails {
```
nscuro, line 32: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/scheduled/policyviolations/PolicyViolationOverview.java
```diff
@@ -0,0 +1,95 @@
...
+import org.dependencytrack.model.Component;
+import org.dependencytrack.model.PolicyViolation;
+import org.dependencytrack.model.Project;
+import org.dependencytrack.persistence.QueryManager;
+
+/* 
+ * Part of the ScheduledPolicyViolationsIdentified Template Models.
+ * Contains several key metrics to provide an overview of the policy violations identified.
+ */
+public final class PolicyViolationOverview {
```
nscuro, line 35: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/scheduled/policyviolations/PolicyViolationSummary.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Map;
+
+import org.dependencytrack.model.PolicyViolation;
+import org.dependencytrack.model.Project;
+
+/* 
+ * Part of the ScheduledPolicyViolationsIdentified Template Models.
+ * Contains packed summarized informations about the identified policy violations, grouped by the affected projects.
+ */
+public final class PolicyViolationSummary {
```
nscuro, line 32: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/scheduled/policyviolations/PolicyViolationSummaryInfo.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.dependencytrack.model.PolicyViolation;
+
+/* 
+ * Part of the ScheduledPolicyViolationsIdentified Template Models.
+ * Contains detailed information about the amount of the identified policy violations grouped by their type.
+ */
+public final class PolicyViolationSummaryInfo {
```
nscuro, line 31: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

nscuro (commented):
Thanks for the PR @MM-msr! Some initial feedback, I figured it would be easier to give it before merging the PR. I can still merge it now if you'd like to get feedback from users sooner.

Please have a look at my comments and let me know how you'd like to proceed.

MM-msr:
Thanks a lot @nscuro for the initial feedback! I'm currently in the process of writing additional test cases to meet the Code Coverage Quota. I may consider to add some of your suggested changes after that as well or reply to them later this week to discuss them. I think by the end of the week, we know if it's a good time to merge it :) I'll try to get the most things done until then.

src/main/java/org/dependencytrack/model/ScheduledNotificationRule.java
```diff
@@ -0,0 +1,349 @@
...
+import jakarta.validation.constraints.Pattern;
+import jakarta.validation.constraints.Size;
+
+/**
+ * Defines a Model class for scheduled notification configurations.
+ */
+@PersistenceCapable
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ScheduledNotificationRule implements Rule, Serializable {
```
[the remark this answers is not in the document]
MM-msr: Actually, that was my first approach. I changed it to an extra table, because i don't know ORMs very well and it didn't work as i was expecting it with adding/editing the ORM-related attributes. Especially when loading all items depending on their class-type (such as fetching all ScheduledNotificationRule items from the database), i had too many problems to invest more time to get it to work.

I may investigate it again, maybe we can do sth. like filters to fetch differently when requesting NotificationRule vs ScheduledNotificationRule (like "isScheduled == true"). But if you can give me an advice how to maybe do it automatically with ORM-magic, i appreciate it. Depending on how much it affects the existing lines of code for NotificationRule, we can definitely merge it.

src/main/java/org/dependencytrack/notification/ScheduledNotificationTaskInitializer.java
```diff
@@ -0,0 +1,62 @@
...
+
+import alpine.common.logging.Logger;
+import jakarta.servlet.ServletContextEvent;
+import jakarta.servlet.ServletContextListener;
+
+/* 
+ * Initializes the scheduled notification task service.
+ * It schedules the task executions for each enabled scheduled notification rule on startup, based on their cron configuration.
+ */
+public class ScheduledNotificationTaskInitializer implements ServletContextListener {
```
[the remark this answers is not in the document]
MM-msr: The TaskInitializer class is only existing to register the existing cron expressions (CE) of enabled rules at startup. Every change of a CE will be handled by the TaskManager class, such as cancelling the current cron countdown and initiate the new one. So if you change a CE in the ui, the manager will cancel the old notification task and reschedule a new one. If a notification was performed, it calculates the new time until the next execution and schedules the next notification. And so on. So there is no need to restart to get new expressions to work.

If you mean sth different here that i don't get, please tell me, else mark it as resolved.

src/main/java/org/dependencytrack/tasks/SendScheduledNotificationTask.java
```diff
@@ -0,0 +1,224 @@
...
+         * by QueryManager.getAllProjects() (and some other multi-project methods), we
+         * need to retrieve them one by one by their UUIDs. This way it was empirically
+         * proven that the parent-child relationship is (more) consistent.
+         */
+        if(rule.getProjects().isEmpty()){
+            // if rule does not limit to specific projects, get all projects and their children, if configured
+            affectedProjects = qm.detach(qm.getAllProjects())
+                    .stream()
+                    .filter(p -> rule.isNotifyChildren() ? true : p.getParent() == null)
+                    .collect(Collectors.toList());
```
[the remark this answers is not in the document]
MM-msr: We might avoid that, if you have a good idea how. What we really need in the end is the project `id`, `uuid` and the basic information for `toJson(Project)` for the template. In the basic mail template, we need the project `name` and `version` (if i didn't forget anything). The last thing we need to have, is the proper children-connection to find all projects with their children, if the setting in the scheduled notification rule is active.

A different approach here might be the following: instead of getting new findings since date x for each of the affected projects, we could fetch all findings since date x and then filter them similar to `NotificationRouter.resolveRules(...)`:
if(rule has projects)
-> get only findings, which have a project id of the rule projects
else
-> use all findings to get the affected project ids, load them and use this project list for the scheduled notification.

This way, we only load what is needed. But what is important nonetheless: we need a proper children-connection in `rule.getProjects()`. If that's working fine, this approach should work well.

What do you think about that?

MM-msr:
Short update from my side: i added some tests and applied your suggested changes for records. Therefore, i created a factory to create these notification subjects with the logic which was previously located in the specific classes.
I want to test it a bit more and implement some additional suggested changes before i push it, but i will try to finish my local changes in the next days.

MM-msr:
@nscuro Hey there :) Could you trigger a test execution? As i can't really test most of the net-related test cases i want to know if i have to fix sth prior to releasing the PRs from draft state. Thanks in advance!

MM-msr:
Is this by any chance another flaky test? :D

nscuro:
> Is this by any chance another flaky test? :D

Looks like a hiccup in fetching dependencies. I re-started the workflow.

MM-msr:
I will release the PR for now from the draft state to allow a broader review. I fixed some minor test errors. Sadly, i can't really get a grip on the other major errors:

- the 500 error codes seem like a server error and may not be related to the code itself...but that's only a guessing
- the publisher tests were created like the others, but i can't really interpret the test error message

I can look into it with assistance, but if a developer with more experience in java testing and especially this project might take a look, i would really appreciate it. Maybe it's a hiccup from the test env, as e.g. the invalid cron test (ScheduledNotificationRuleResourceTest.java, L. 182ff.) returns the expected string in my browser, but as it seems not in the github tests.

src/main/java/org/dependencytrack/tasks/SendScheduledNotificationTask.java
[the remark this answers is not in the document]
MM-msr: This should be "fixed" by now, as i introduced the ScheduledNotificationFactory, which will only fetch the projects that are included in the findings and uses the data from this database call to extract all information for the sub-parts needed for the email-template. This way, the performance impact should be small.

rkg-mm:
@nscuro can you give feedback if you need further changes and/or by when you will have time to look at this?

src/main/java/org/dependencytrack/notification/ScheduledNotificationTaskManager.java
MM-msr: I looked a bit into that and i think using the Callback Event is not benifitial here. We need the cron-scheduled execution, so the runnable task for the scheduled executor with be needed nonetheless. I could rewrite it in a way, that the runnable will just dispatch the event and the subscriber will react to it by executing the SendScheduleNotificationTask (or scheduling a new one), but it's just another layer on the same logic. Please explain the benefit your seeing in this shortly, if you would like to change it that way.

nscuro:
@rkg-mm I'll try to complete a review this week.

rkg-mm:
@nscuro  could you please update the feature branch to the latest master? our branch is updated to the master branch which results in a lot of changes now :D

nscuro:
@rkg-mm Done. The GitHub diff of both PRs won't update though until you push a new commit.

rbt-mm:
The previously failing unit tests for this PR should now all be fixed.
@nscuro could you please re-approve the pending GitHub workflows for the latest commit?

rbt-mm:
Fix timestamps in scheduled notification tests

Asserting the timestamp for the test `testPublishWithScheduledNewVulnerabilitiesNotification` is fixed by setting it to a static value.
Asserting the timestamp for the test `testPublishWithScheduledNewPolicyViolationsNotification` is fixed by changing the assertion to only include relevant and deterministic information.
```diff
diff --git a/src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java b/src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java
index 09874d6a..6712914a 100644
--- a/src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java
+++ b/src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java
@@ -69,6 +69,7 @@ import jakarta.json.JsonObjectBuilder;
 import java.math.BigDecimal;
 import java.time.Instant;
 import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.util.Date;
 import java.util.EnumMap;
@@ -283,7 +284,7 @@ public abstract class AbstractPublisherTest<T extends Publisher> extends Persist
                 vuln.getSeverity().name(),
                 "analyzer",
                 "http://example.com",
-                Date.from(Instant.ofEpochSecond(66666, 666)).toString(), // Thu Jan 01 18:31:06 GMT 1970
+                "Thu Jan 01 18:31:06 GMT 1970", // Thu Jan 01 18:31:06 GMT 1970
                 AnalysisState.EXPLOITABLE.name(),
                 false)));
 
diff --git a/src/test/java/org/dependencytrack/notification/publisher/SendMailPublisherTest.java b/src/test/java/org/dependencytrack/notification/publisher/SendMailPublisherTest.java
index 84bf7ad7..fa743dbd 100644
--- a/src/test/java/org/dependencytrack/notification/publisher/SendMailPublisherTest.java
+++ b/src/test/java/org/dependencytrack/notification/publisher/SendMailPublisherTest.java
@@ -466,7 +466,7 @@ public class SendMailPublisherTest extends AbstractPublisherTest<SendMailPublish
                                         </a>
                                         </td>
                                         <td>
-                                        Thu Jan 01 19:31:06 CET 1970
+                                        Thu Jan 01 18:31:06 GMT 1970
                                         </td>
                                         <td>
                                         EXPLOITABLE
@@ -497,177 +497,46 @@ public class SendMailPublisherTest extends AbstractPublisherTest<SendMailPublish
             final MimeMultipart content = (MimeMultipart) message.getContent();
             assertThat(content.getCount()).isEqualTo(1);
             assertThat(content.getBodyPart(0)).isInstanceOf(MimeBodyPart.class);
-            assertThat((String) content.getBodyPart(0).getContent()).isEqualToIgnoringWhitespace("""
-            <html>
-            <head>
-                <style>
-                table {
-                    border-collapse: collapse;
-                    width: 100%;
-                }
-
-                th,
-                td {
-                    border: 1px solid #2174c7;
-                    padding: 5px;
-                    text-align: left;
-                }
-
-                th {
-                    background-color: #2174c7;
-                    color: white;
-                }
-
-                tr:nth-child(even) {
-                    background-color: #afafaf;
-                }
-
-                tr:hover {
-                    background-color: #afafaf;
-                }
-                </style>
-            </head>
-            <body>
-                <p>New Policy Violation Identified</p>
-                <p>-------------</p>
-                <p></p>
-
-                
-                <h2>Overview</h2>
-                <div style="overflow-x: auto">
-                <table style="width: 50%;">
-                    <tbody>
-                    <tr>
-                        <td>Projects included in this rule</td>
-                        <td>1</td>
-                    </tr>
-                    <tr>
-                        <td>Total new policy violations</td>
-                        <td>1</td>
-                    </tr>
-                    
-                    <tr>
-                        <td>New policy violations (LICENSE)</td>
-                        <td>1</td>
-                    </tr>
-                    
-                    <tr>
-                        <td>Components affected by new policy violations</td>
-                        <td>1</td>
-                    </tr>
-                    <tr>
-                        <td>Suppressed new policy violations (not included above)</td>
-                        <td>0</td>
-                    </tr>
-                    </tbody>
-                </table>
-                </div>
-                
-                <h2>Summary per project in this rule</h2>
-                <div style="overflow-x: auto">
-                <table style="width: 70%;">
-                    <thead>
-                    <tr>
-                        <th>Project Name</th>
-                        <th>Version</th>
-                        <th>New Policy Violation</th>
-                        <th>All Policy Violations</th>
-                        <th>Suppressed Policy Violations</th>
-                    </tr>
-                    </thead>
-                    <tbody>
-                    
-                    <tr>
-                        <td>
-                        <a href="/projects/c9c9539a-e381-4b36-ac52-6a7ab83b2c95">
-                            projectName
-                        </a>
-                        </td>
-                        <td>projectVersion</td>
-                        <td>
-                        
-                        LICENSE: 1<br />
-                        
-                        </td>
-                        <td>
-                        
-                        LICENSE: 1<br />
-                        
-                        </td>
-                        <td>
-                        
-                        </td>
-                    </tr>
-                    
-                    </tbody>
-                </table>
-                </div>
-
-                <h2>New policy violations per project</h2>
-
-                
-                <h3>
-                Project "projectName" [Version:
-                projectVersion]
-                </h3>
-                <div style="overflow-x: auto">
-                <table>
-                    <thead>
-                    <tr>
-                        <th>State</th>
-                        <th>Risk Type</th>
-                        <th>Policy Name</th>
-                        <th>Component</th>
-                        <th>Component Version</th>
-                        <th>Occurred on</th>
-                        <th>Analysis</th>
-                        <th>Suppressed</th>
-                    </tr>
-                    </thead>
-                    <tbody>
-                    
-                    <tr>
-                        <td>
-                        INFO
-                        </td>
-                        <td>
-                        LICENSE
-                        </td>
-                        <td>
-                        policyName
-                        </td>
-                        <td>
-                        
-                            <a href="/components/94f87321-a5d1-4c2f-b2fe-95165debebc6">
-                            componentName
-                            </a>
-                        
-                        </td>
-                        <td>
-                        componentVersion
-                        </td>
-                        <td>
-                        Thu Jan 01 19:31:06 CET 1970
-                        </td>
-                        <td>
-                        APPROVED
-                        </td>
-                        <td>
-                        No
-                        </td>
-                    </tr>
-                    
-                    </tbody>
-                </table>
-                </div>
-                
-                
-            </body>
-            <footer>
-                <p>Executed: 1970-01-01T18:31:06.000000666</p>
-            </footer>
-            </html>
-            """);
+            
+            // Check if relevant information is included in the HTML content of the notification.
+            // HTML content is not directly compared to static value because deprecated Date 
+            // class in Policy Violation Analysis provides different timestamps for different systems.
+            assertThat((String) content.getBodyPart(0).getContent()).containsIgnoringWhitespaces(
+                "New Policy Violation Identified",
+                "Overview",
+                "<td>Projects included in this rule</td><td>1</td>",
+                "<td>Total new policy violations</td><td>1</td>",
+                "<td>New policy violations (LICENSE)</td><td>1</td>",
+                "<td>Components affected by new policy violations</td><td>1</td>",
+                "<td>Suppressed new policy violations (not included above)</td><td>0</td>",
+                "Summary per project in this rule",
+                "<th>Project Name</th>",
+                "<th>Version</th>",
+                "<th>New Policy Violation</th>",
+                "<th>All Policy Violations</th>",
+                "<th>Suppressed Policy Violations</th>",
+                "<a href=\"/projects/c9c9539a-e381-4b36-ac52-6a7ab83b2c95\">projectName</a>",
+                "projectVersion",
+                "LICENSE: 1",
+                "New policy violations per project",
+                "Project \"projectName\" [Version:projectVersion]",
+                "<th>State</th>",
+                "<th>Risk Type</th>",
+                "<th>Policy Name</th>",
+                "<th>Component</th>",
+                "<th>Component Version</th>",
+                "<th>Occurred on</th>",
+                "<th>Analysis</th>",
+                "<th>Suppressed</th>",
+                "INFO",
+                "LICENSE",
+                "policyName",
+                "<a href=\"/components/94f87321-a5d1-4c2f-b2fe-95165debebc6\">componentName</a>",
+                "componentVersion",
+                "APPROVED",
+                "No",
+                "Executed: 1970-01-01T18:31:06.000000666"
+                );
         });
     }
 

```

nscuro:
@rkg-mm @rbt-mm @MM-msr First let me say sorry for dragging this for so long. All your work is truly appreciated and I hate that this has been open for so long.

I have updated the target branch with the latest `master`, and also raised a PR to bringt your branch up to date: https://github.com/MM-msr/dependency-track/pull/2

There are a few changes I'd like to see here and with respect to everyone's time and resources I believe it's best if I simply propose the changes in a PR into your branch, which you can then review whether it still meets your needs.

Things I'd like to change:

1. Reduce the duplication in code and database tables that is introduced by treating `ScheduledNotificationRule`s and `NotificationRule`s as separate types (see also my previous comment [here](https://github.com/DependencyTrack/dependency-track/pull/3925#discussion_r1685752248)).
2. Generation of notification subjects must work for large(r) portfolios. At the moment we're loading all findings since the last submission at once, we do repetitive queries to load projects for every single finding, and so on. I just spent a lot of time to nuke such patterns from legacy code and really don't want to introduce them again in another place.
3. I'm trying to see if we can avoid introducing a separate scheduling mechanism that bypasses the existing task/eventing mechanism. Of course while still allowing users to change Cron expressions at runtime.

Separately, I was wondering whether we should even allow users to configure scheduled notifications *without* limiting them to projects. At the moment, limiting to projects is optional. If users were to create such notifications for their entire portfolio, I'm seriously struggling to see how all involved systems could possibly handle the size of the notification content. It feels like a footgun to me. WDYT?

nscuro:
> we will check the PR tomorrow asap.

Thanks. To be clear, this initial PR is solely to bring the code up-to-date. I will raise follow-up PRs after that so the changes are easier to see.

> Enforce limiting probably makes sense.

Cool. Will look into adding that.

nscuro:
Working on my changes here: https://github.com/MM-msr/dependency-track/pull/3

nscuro:
Admittedly I struggled with finding the time to complete this but it's finally done, you can find the final result, including a high-level description of how this PR evolved here: #4783.

nscuro:
Superseded by #4783

Issue #3834, opened 2024-06-11
VEX export returns invalid cyclonedx


muellerst-hg:
### Current Behavior

Given a project with two components which have the same vulnerability
When I click "Export VEX"
Then a VEX file is delivered which contains duplicate items in `vulnerabilities` section
But according to [cyclonedx 1.5 schema](https://github.com/CycloneDX/specification/blob/1.5/schema/bom-1.5.schema.json#L91) this is invalid 

Tested with 4.11.3 and 4.12-snapshot from 2024/06/11

### Steps to Reproduce

1. Create a new project and upload the following BOM file: 
[bom-express.json](https://github.com/user-attachments/files/15791098/bom-express.json)
2. Wait for analysis to be finished
3. Go to "Audit Vulerabilties" tab and click "Download VEX" (which returns the following VEX file: 
[vex-express.json](https://github.com/user-attachments/files/15791116/vex-express.json)
 )
 4. Enable "BOM Validation" in "Configuration/Bom Format" settings
 5. Click "Apply VEX" and try to upload the above VEX file
 6. Upload fails with > The uploaded BOM is invalid. Schema validation failed
 
 The same duplicates can be found when exporting the BOM using "Download BOM -> Inventory with Vulnerabilities"

### Expected Behavior

I expect valid cyclonedx bom returned by "Export VEX".

### Dependency-Track Version

4.11.3

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Database Server Version

14

### Browser

Mozilla Firefox

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

SaberStrat:
I decided to have a go at this one. Fixed it locally, not sure if it's the prettiest solution, but I'd dare a PR with it.

SaberStrat:
Thinking about the deduplications on vulnerabilities I'm doing, and if I'm understanding @muellerst-hg remark correctly:

> The same duplicates can be found when exporting the BOM using "Download BOM -> Inventory with Vulnerabilities"

[...]

PR #3948, opened 2024-07-08
Fix vex export returning invalid CycloneDX

SaberStrat:
### Description

Fixes exported VEX and BOMs with vulnerabilities not conforming to CycloneDX 1.5.

### Addressed Issue

#3834 

### Additional Details

The general idea is to deduplicate vulnerabilities based on all `Vulnerability` fields.

I settled on an implementation with one HashSet that tracks all `Vulnerability` objects seen when converting `Finding`s into `Vulnerability`s.

It requires an enhancement of the cyclonedx-core-java library, namely adding custom `equals` and `hashCode` methods to the `Vulnerability` class for fieldwise comparison: https://github.com/CycloneDX/cyclonedx-core-java/issues/463.

Other possible implementations I considered: 
1. Comparison of `Vulnerability`s based on all of its fields _except_ for `affects` and accumulate affects objects in one vulnerability if all other `Vulnerability` fields have otherwise equal values. This came from an interpretation of a remark from the original issue reporter in the issue report. But the CycloneDX schema doesn't really require an accumulation, if I'm not mistaken. 
I considered a few different implementations of this (in case an accumulation was not wrong after all).
    1. a HashMap of `Vulnerability`s to HashSets of `Affect`s. This was the way I'd have gone with
    2. generating the affects list in its own convert method. I think this would've meant introducing two more methods. Could do this, if an accumulation of affects is a possibility after all
    3. appending to an `affects` list defined outside of and passed to `Vulnerability`'s convert() calls. Was not sure about having `convert()` convert vulnerability objects AND modify a state of a list passed to it.

### Checklist


src/test/java/org/dependencytrack/parser/cyclonedx/CycloneDXExporterTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.dependencytrack.tasks.scanners.AnalyzerIdentity;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThatNoException;
+
+public class CycloneDXExporterTest extends PersistenceCapableTest {
+
+    @Test // https://github.com/DependencyTrack/dependency-track/issues/3834
+    public void testSchemaValidityOfExportedVexWithMultipleComponentsWithSameVulnerability() throws GeneratorException {
```
nscuro, line 17: Consider moving this to [`VexResourceTest`](https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/resources/v1/VexResourceTest.java), and asserting the entire response JSON.

Also, a test case that demonstrates what happens when `componentA` and `componentB` have different analyses (i.e. one is suppressed the other is not, or one has sate `IN_TRIAGE` the other `EXPLOITABLE`) would be good.

If we de-duplicate vulnerabilities without the `Analysis` in mind, we might drop important data.

nscuro (changes requested):

src/test/java/org/dependencytrack/parser/cyclonedx/CycloneDXExporterTest.java
SaberStrat: Yeah I'm re-implementing it such that the whole Vulnerability object is used for comparison. Still in the process of making it work.

Will do what you wrote with the tests.

SaberStrat: Done

nscuro:
@SaberStrat cyclonedx-core-java 9.0.5 was [released earlier today](https://github.com/CycloneDX/cyclonedx-core-java/releases/tag/cyclonedx-core-java-9.0.5), which includes your changes. You should be able to bump the library version now to make your implementation work.

SaberStrat:
cyclonedx-core-java bumped and rebased to master.

SaberStrat:
@nscuro I take it adding escaped quotes to the expected error messages in the test methods is not a proper solution for this?

nscuro:
Yup! Nothing special to do on our side since we merely forward the error messages. But still good to have such tests to be made aware of *when* these things change.

SaberStrat:
Rinse and repeat. This time, I ran the full test suite locally and it passed.

Should I cleanup/squash the commits down logically?

nscuro:
> Should I cleanup/squash the commits down logically?

Not necessary.

nscuro (approved):
Thanks @SaberStrat! I'll also backport this to v4.11.6.

Issue #2094, opened 2022-10-25
Documentation:  Incorrect Anchors in Changelog 


msymons:
### Current Behavior:

The Change Log for Dependency-Track can be found at: https://docs.dependencytrack.org/changelog/
The page contains multiple anchors.
The anchors are not unique and do not reference (say) version numbers.  Thus, the anchor for 4.5.0 release does not take one to 4.5.0 release.

### Steps to Reproduce:

1. Navigate to changelog URL (link above0
2. Scroll down to a specific release such as 4.5.0
3. Mouse-over the "#" symbol and right-click and select "open link in new tab"
4. Observe where you have been taken to.

### Expected Behavior:
* The anchor for 4.5.0 should take one to 4.5.0 release notes
* The anchor for 4.4.0 should take one to 4.4.0 release notes
etc

Instead, such anchors all point to: https://docs.dependencytrack.org/changelog/#software-bill-of-materials-sbom

Anchors for frontend-dist.zip (etc) are also not unique.

The reason why this is important is to avoid problems where the links are used in emails, blogposts, etc.  One wants to tell people "this was fixed in version X" and provide a URL that links to the correct changelog for X...  and have the link be valid in 6 months or 12 months time.

### Environment:
  
  - Dependency-Track Version: 4.6.1
  - Browser:  Firefox v106.0.1

nscuro:
Support for Anchors was added in #1845, before that the docs had no anchors at all. IIRC, Jekyll does not come with proper support, and it's hard to retro-fit. So I am not sure how and if this can be fixed at all.

For Hyades we switched to MkDocs, which does a better job of generating working anchors. For example: https://dependencytrack.github.io/hyades/0.5.0-SNAPSHOT/operations/database/#bare-metal-docker

Assigning `size/S` nevertheless since changes to the docs shouldn't be too complex.

PR #3965, opened 2024-07-10
fix: fix anchors in changelog documentation

JCHacking:
### Description
Fixed the problem where the links for the changelog in the web documentation were not being generated correctly

### Addressed Issue
#2094 

### Checklist


JCHacking:
As far as I can see it was failing because it always caught the id of the last h6 (Software Bill of Materials).

If we set the ids to be generated, it works in the versions, but in the files it still fails.

nscuro:
Thanks for looking into this!

> Where are these links generated?

They are generated by https://github.com/allejo/jekyll-anchor-headings which is located here in the repo: https://github.com/DependencyTrack/dependency-track/blob/master/docs/_includes/anchor_headings.html

To be honest I added it a while ago due to a lack of better options - before adding it there were no anchors at all. I'm saying this because there *may* be better options out there that do the job more reliably, so please feel free to experiment with other options as you see fit.

JCHacking:
I have checked the tool and it seems that when in a markdown there is a h1 - h6 it automatically generates an id, which is translated into an anchor. So with archives there is no automatic solution, the only one would be to generate the markdown from _post to add the link.

JCHacking:
Hi @nscuro 
After a thorough review, I don't see a better way to do all this, this is the best that can be achieved with Jekyll.

Do you want to put links to the files of each release? For this we would have to modify the _post, they are generated automatically? 

If so, I can edit the existing ones and make the new ones incorporate it, as long as you are ok with it.

nscuro:
Thanks for investigating!

> Do you want to put links to the files of each release? For this we would have to modify the _post, they are generated automatically?

If specifying the links manually fixes the issue, then I'd say let's go with that. It may be tedious, but IMO the important part is that they *work* :D

Would be much appreciated if you could do that!

JCHacking:
Hi @nscuro 
Modified all _posts with a script (Sorry so many commits, you can see better from github the differences than from git by console)

docs/_posts/2024-07-08-v4.11.5.md
```diff
@@ -30,21 +30,21 @@ We thank all organizations and individuals who contributed to this release, from
...
-###### dependency-track-bundled.jar
+###### [dependency-track-bundled.jar](https://github.com/DependencyTrack/dependency-track/releases/download/4.11.5/dependency-track-bundled.jar)
 
 | Algorithm | Checksum                                                         |
 |:----------|:-----------------------------------------------------------------|
 | SHA-1     | eba6cbaa6c2da9ffb295da83ed39af68ff4130a8                         |
 | SHA-256   | 7ebb11573b2a59084ed98fe92d363240c910dc7b5aa7ebeda64bee7d47089d9a |
 
-###### frontend-dist.zip
+###### [frontend-dist.zip](https://github.com/DependencyTrack/dependency-track/releases/download/4.11.5/frontend-dist.zip)
```
nscuro, line 47: Frontend links should point to the `DependencyTrack/frontend` repository. The current links yield a 404.

```suggestion
###### [frontend-dist.zip](https://github.com/DependencyTrack/frontend/releases/download/4.11.5/frontend-dist.zip)
```

nscuro (changes requested):

JCHacking:
I didn't notice that case, fixed in all files

nscuro (approved):
Thanks!

Issue #3936, opened 2024-07-08
Switch to use authors instead of author


thompson-tomo:
### Current Behavior

The author property is populated under components however this field is deprecated in at least 1.6 possibly prior.

Deprecation is mentioned in: https://github.com/CycloneDX/specification/blob/62a669075f1897193a14060e0784e6a7576b693d/schema/bom-1.6.schema.json#L899 and removal tracked via: https://github.com/CycloneDX/specification/issues/387

### Proposed Behavior

The Authors property is populated rather than author

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

nscuro:
Implemented in https://github.com/DependencyTrack/dependency-track/pull/3969

PR #3969, opened 2024-07-11
Add support for authors field

2000rosser:
### Description
In CyclonDX 1.6 the author field is deprecated and replaced with authors

### Addressed Issue
#3936 

### Additional Details
Please provide any suggestions on extra unit tests I could add.

### Checklist


leec94:
looks good to me! maybe you can also implement this changes in hyades once its merged in?

2000rosser:
@leec94 I'd be happy to give it a go!

src/main/java/org/dependencytrack/model/Component.java
```diff
@@ -114,6 +115,11 @@ public enum FetchGroup {
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The author may only contain printable characters")
     private String author;
 
+    @Persistent(defaultFetchGroup = "true")
+    @Convert(OrganizationalContactsJsonConverter.class)
+    @Column(name = "AUTHORS", jdbcType = "CLOB", allowsNull = "true")
+    private List<OrganizationalContact> authors;
```
nscuro, line 121: Keeping `author` and `authors` separate will introduce ambiguity to the data model. The CycloneDX specification had to do it this way to retain backward compatibility, but we should make an effort to keep our data model clean of such redundancies.

What I think we should do:

* When converting from CycloneDX to DT's data model, transform `author` to `authors`
* When converting from DT's data model CycloneDX, populate `author` with values from `authors`, **if targeting a CycloneDX version before 1.6** ([which we currently do](https://github.com/DependencyTrack/dependency-track/blob/af4461e86ae56eaee193017ec86ee2a690709932/src/main/java/org/dependencytrack/parser/cyclonedx/CycloneDXExporter.java#L97-L105))

nscuro (changes requested):

2000rosser:
@nscuro I was wondering, should I be including author for the component and project resources classes? https://github.com/2000rosser/dependency-track/blob/385bf6eacc1b79446f0032f842e1c262262e352f/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java#L264-L326
Currently I have it as just authors, but that would mean users have to follow the authors pattern when creating a new component/project through the API, where they might be used to just using author.

nscuro:
@2000rosser You made a comment somewhere about `author` now missing from the REST API, but I can't seem to find it anymore. But you are definitely right about that.

A possible way to address this would be:

* Add `getAuthor` and `setAuthor` methods back to the `Component` and `Project` classes
* Mark all of them as `@Deprecated`
* Have `getAuthor` return `ModelConverter.convertContactsToString(authors)`
* Have `setAuthor` call a method that implements [this](https://github.com/DependencyTrack/dependency-track/pull/3969/files#diff-1b2176c8578f62555841ee8976ac2e8c45300b52852f104c8ab3e27738a42d2fR128-R137) logic

The methods would be called by Jackson when serializing to and from JSON.

We will remove that logic in Hyades, since we can make breaking changes there.

2000rosser:
@nscuro Gotcha, I like that approach, makes a lot more sense. Thanks

src/test/java/org/dependencytrack/resources/v1/BomResourceTest.java
```diff
@@ -1006,6 +1013,61 @@ public void uploadBomInvalidParentTest() throws Exception {
...
+                .put(Entity.entity("""
+                        {
+                        "projectName": "acme-app",
+                        "projectVersion": "1.0.0",
+                        "bom": "%s"
+                        }
+                        """.formatted(encodedBom), MediaType.APPLICATION_JSON));
+
+        assertThat(response.getStatus()).isEqualTo(200);
+    }
```
nscuro, line 1069: This test seems to be testing the CycloneDX schema validation, both directly in line `1056` and indirectly through the upload in line `1058`.

Do we need this?

src/test/java/org/dependencytrack/resources/v1/ComponentResourceTest.java
```diff
@@ -506,13 +506,15 @@ public void createComponentTest() {
...
         component.setVersion("1.0");
+        component.setAuthor("SampleAuthor");
         Response response = jersey.target(V1_COMPONENT + "/project/" + project.getUuid().toString()).request()
                 .header(X_API_KEY, apiKey)
                 .put(Entity.entity(component, MediaType.APPLICATION_JSON));
         Assert.assertEquals(201, response.getStatus(), 0);
         JsonObject json = parseJsonObject(response);
         Assert.assertNotNull(json);
         Assert.assertEquals("My Component", json.getString("name"));
+        Assert.assertEquals("SampleAuthor" ,json.getJsonArray("authors").getJsonObject(0).getString("name"));
```
nscuro, line 517: Can you add an assertion here that `author` is also set, according to the `Component#getAuthor` logic?

src/test/resources/unit/cyclonedx/valid-bom-1.6.json
```diff
@@ -32,7 +32,11 @@
     ],
     "component": {
       "type": "application",
-      "author": "Acme Super Heros",
```
nscuro, line 35: Consider creating a separate BOM file instead, and using that in a new test in `BomUploadProcessingTaskTest`. Ideally we would want to verify that `Component#authors` is populated correctly when either `authors` or `author` is provided in the BOM. Even better, we should have a test case that demonstrates what happens when both `author` *and* `authors` are set in the BOM.

nscuro (changes requested):

src/test/java/org/dependencytrack/resources/v1/ComponentResourceTest.java
2000rosser: Since that endpoint just returns the `Component` object  as json, shouldn't the `author` value not be present, since it is not a value in the `Component` class?

nscuro: As you said earlier, removing that field would be a breaking change from the REST API's perspective. Internally, we don't need the legacy `getAuthor`, but users who currently rely on that field will still need it.

So I think it should be included in the response.

2000rosser: I see what you mean. I'll work on this change soon

nscuro (approved):

Issue #2924, opened 2023-07-31
Add a way to send test alerts


rdicroce:
### Current Behavior

There is no way to test an alert. So if you set up an alert using Slack/Teams/Jira/etc, you have no idea if you've done it correctly. And if some time later, you suspect that the alert is somehow broken, you have no way to check that it is still working.

### Proposed Behavior

The Alerts screen should have a button that lets you send a test alert, so users can verify that the alert is configured correctly.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

nscuro:
Implemented in https://github.com/DependencyTrack/dependency-track/pull/3983

PR #3983, opened 2024-07-16
Add option to test notification publisher

2000rosser:
### Description
Replaced the /test/smtp endpoint to work for any of the publishers.

Front end implementation: https://github.com/DependencyTrack/frontend/pull/949

### Addressed Issue
#2924 

### Checklist


nscuro (changes requested):
I'm not sure this addresses the original feature request.

Having a test button for the email *publisher* makes sense, because DT only ever connects to a single email server, and those connection details are a global setting. 

However, for pretty much all other publishers, there is no global configuration. Instead, destination URLs are configured for each alert (see for example MS Teams, Slack, Webhook). Testing the publisher without an alert (`NotificationRule`) will not work for those.

We'd need a test functionality on the alert / `NotificationRule` level. This also means we can't use the same hardcoded `Notification`, since alerts are configured for specific levels and groups. Either we dynamically build a `Notification` object that will satisfy the alert under test, or we let users provide an entirely custom one.

2000rosser:
The approach I went for was to test only the destination for the selected alert rule in the setup page, which worked for the publishers I tested (Slack, MS Teams). If we want to test on the `NotificationRule` level, which approach would be preferred, letting users provide a custom `Notification` or build one that satisfies the `NotificationRule`. 

Could possibly POST the `NotificationRule` similar to [here](https://github.com/DependencyTrack/frontend/blob/0a4bdcb2e99380f28b2d9b0d178183c70558ab07/src/views/administration/notifications/Alerts.vue#L333-L363) and then either build the `Notification` that satisfies it or send a user provided one.

nscuro:
> The approach I went for was to test only the destination for the selected alert rule in the setup page, which worked for the publishers I tested (Slack, MS Teams).

That approach is not necessarily wrong, either. But it limits the usefulness of the check a bit, since individual rules can be customized further (i.e. auth headers for *Outbound Webhook* alerts). Ideally you'd want to know whether a *specific* rule works.

> Could possibly POST the NotificationRule similar to [here](https://github.com/DependencyTrack/frontend/blob/0a4bdcb2e99380f28b2d9b0d178183c70558ab07/src/views/administration/notifications/Alerts.vue#L333-L363) (or just the UUID) and then either build the Notification that satisfies it or send a user provided one.

I think if we can assemble one or more notifications (i.e. one for each enabled `NotificationGroup`) automatically, based on the `NotificationRule`, that would be best.

2000rosser:
@nscuro I opted for adding `_TEST` to the group in order to get the default pebble format for the notification group, as when using the provided one, it gives errors due to having empty values for the text fields in the json.

Any feedback/ideas on this implementation?

nscuro:
> I opted for adding _TEST to the group in order to get the default pebble format for the notification group, as when using the provided one, it gives errors due to having empty values for the text fields in the json.
>
> Any feedback/ideas on this implementation?

We could alternatively assemble "correct" dummy subjects depending on the group, similar to how we do it in [`AbstractPublisherTest`](https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java).

This would avoid the errors you were seeing, and also help with testing custom notification templates.

nscuro:
Looks great @2000rosser! Do you think you can increase the test coverage a bit?

2000rosser:
@nscuro Thanks! Yeah np

nscuro (approved):
Thanks!

PR #4059, opened 2024-08-11
Enhance badge API to require authorization

### LINKED ISSUE #3596: Enhance Badge API to support authentication and authorisation and deprecate unauthenticated access to badges
### Current Behavior

As of Dependency-Track v4.11, Badges can only be activated globally for all projects and versions and its API requests do not require authentication or authorisation.

### Proposed Behavior

I'm proposing basically what @stevespringett suggested here as a future enhancement of the current badges implementation: https://github.com/DependencyTrack/dependency-track/issues/252#issuecomment-515848450:

implement a new permission to control access to the badge API. The authentication and authorisation via API key should be handled via URL parameter rather than a header, as there is no pretty/non-hacky-looking way to pass headers to badge URLs in their canonical use cases. For this, the team associated with the API key should have minimal permissions. Together with Portfolio Access Control, this would allow for a convenient way to control access on a project basis and make DT badges more secure.

While convenient as a feature, and allowing any downstream stakeholder to display the state of vulnerabilities and violations about a tracked project, activating badges in the current implementation opens up a hole in the security for any attacker to use with knowledge of project names or versions. They can fetch quite a lot of data about a project from that API that way that would otherwise require authentication and potentially authorisation.

To not break users' key-less badges displays when they update to the version with this change, the switch from the currently optional and unauthenticated badges access to authenticated+authorised-only will be done with a grace period in two steps over the course of at least two minor version updates.

#### Step 1 (upcoming minor version update):
The current unauthenticated and unauthorised access to badges will be deprecated, but will continue to work for users who have had that enabled. 
The existing setting for badges under DT's general configurations 
```
disable <-> enable (unauthenticated)
``` 
will be turned into 
```
disable unauthenticated <-> enable unauthenticated
```
while authenticated+authorised access will remain a working option in both cases. This way, users' existing unauthenticated badge displays still work, while they replace their badges requests with new ones, supplying them with an API key one after after another, which will also work.
#### Step 2 (a future minor version update):
The option of unauthenticated and unauthorised access to badges will be removed entirely and only authenticated and authorised access will be possible.

### Checklist

SaberStrat:
### Description

Enables Dependency-Track to offer badges in a secure manner and change the `badge` API from an opt-in-able, unauthenticated one into one requiring authentication with the new permission `VIEW_BADGES`, either via HTTP header or URI query parameter.

In the currently implementation, this as a breaking change (see Additional Details.)

### Addressed Issue

Closes #3596 

### Additional Details

As a first for DT's API, badges will allow authentication via URI query parameter, as badges are often retrieved from/embedded in clients that do not easily support an injection of headers.

The securing of badges behind API authentication and ACL can be done in two ways, either as a breaking change or gracefully with backwards compatibility: implement authentication with a new permission and
1. remove current checkbox enabling unauthenticated badges support entirely _(breaking change)_
2. keep a checkbox, but redo it to switch between the settings _(backwards compatible change)_:
   1. disable unauthenticated access (allows authenticated access only, corresponds to the previous "off" setting right after updating to the new DT version containing this PR)
   2. enable unauthenticated access (corresponds to the previous "on" setting)

A grace period is nice. But because the previous way of offering badges represents a security flaw, I would've preferred to make users switch to authenticated badges requests right away. 

#### Downstream changes
https://github.com/DependencyTrack/frontend/issues/967

#### Upstream requirements
https://github.com/stevespringett/Alpine/issues/641

### Checklist

### DIFF
--- a/docs/_docs/integrations/badges.md
+++ b/docs/_docs/integrations/badges.md
@@ -11,6 +11,9 @@ basis via permission.
 To enable badges for a team, activate the permission `VIEW_BADGES`. To deactivate badges, remove the permission. To 
 retrieve a badge, use a team's API key either in the badge API header `X-API-Key` or in the URI parameter `apiKey`.
 
+Dependency-Track ships with a default team "_Badge Viewers_" dedicated to badges that already has the necessary
+permission and an API key.
+
 > As badges are typically embedded in places that more people have access to than to Dependency-Track, the API key used
 > for the badge request should have minimal scope to prevent unintended access beyond that badge. Ideally, the API
 > key belongs to a single-purpose team, having just the `VIEW_BADGES` permission, with only one API key and access to 
--- a/docs/_docs/integrations/badges.md
+++ b/docs/_docs/integrations/badges.md
@@ -5,14 +5,21 @@ chapter: 6
 order: 10
 ---
 
-Dependency-Track supports badges in Scalable Vector Graphics (SVG) format. Support for badges is a globally configurable
-option and is disabled by default.
+Dependency-Track supports badges in Scalable Vector Graphics (SVG) format. Support for badges is configurable on a team
+basis via permission. 
 
-> Enabling badge support will provide vulnerability and policy violation metric information to unauthenticated users.
-> Any anonymous user with network access to Dependency-Track and knowledge of a projects information will be able
-> to view the SVG badge.
+To enable badges for a team, activate the permission `VIEW_BADGES`. To deactivate badges, remove the permission. To 
+retrieve a badge, use a team's API key either in the badge API header `X-API-Key` or in the URI parameter `apiKey`.
 
-In all following examples, replace `{name}`, `{version}`, and `{uuid}` with their respective values.
+> As badges are typically embedded in places that more people have access to than to Dependency-Track, the API key used
+> for the badge request should have minimal scope to prevent unintended access beyond that badge. Ideally, the API
+> key belongs to a single-purpose team, having just the `VIEW_BADGES` permission, with only one API key and access to 
+> only the projects/project versions whose badges are displayed at one site--the latter requiring _Portfolio Access 
+> Control_. 
+
+In all following examples, replace `{name}`, `{version}`, `{uuid}`, and `{apiKey}` with their respective values. For
+brevity, the examples use the URI query parameter as the method of authentication, however, they also work with
+authentication by header.
 
 ### Vulnerable components
 Create a badge for vulnerable components of the project. It either shows:
@@ -33,8 +40,8 @@ name and version.
 
 #### Examples
 ```
-https://dtrack.example.com/api/v1/badge/vulns/project/{name}/{version}
-https://dtrack.example.com/api/v1/badge/vulns/project/{uuid}
+https://dtrack.example.com/api/v1/badge/vulns/project/{name}/{version}?apiKey={apiKey}
+https://dtrack.example.com/api/v1/badge/vulns/project/{uuid}?apiKey={apiKey}
 ```
 
 ### Policy violations
@@ -57,8 +64,8 @@ projects name and version.
 #### Examples
 
 ```
-https://dtrack.example.com/api/v1/badge/violations/project/{name}/{version}
-https://dtrack.example.com/api/v1/badge/violations/project/{uuid}
+https://dtrack.example.com/api/v1/badge/violations/project/{name}/{version}?apiKey={apiKey}
+https://dtrack.example.com/api/v1/badge/violations/project/{uuid}?apiKey={apiKey}
 ```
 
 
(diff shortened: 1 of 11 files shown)


SaberStrat:
Forgot readying the PR for the frontend, will remain a draft until then

SaberStrat:
Not sure how to repro this locally. The unit tests are successful.

nscuro:
Just a flaky test. I restarted CI workflow.

nscuro (commented):
Looks good, and I agree with your assessment of intentionally making this a breaking change due to the security implications of the current implementation.

In the original issue, Steve suggested to create a default team with the `VIEW_BADGE` permission. I think we should follow that. You can add a new default team here: https://github.com/DependencyTrack/dependency-track/blob/edb95ce4f7b70e9367433e2bd227bbec4db8b9ba/src/main/java/org/dependencytrack/persistence/DefaultObjectGenerator.java#L138-L173

nscuro:
Oh, and consider adding a [migration](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/upgrade/v4120/v4120Updater.java) to remove the `badge.enabled` column from the `CONFIGPROPERTY` table. Note that #4143 also edits this file so there might be merge conflicts.

savek-cc:
As this is a breaking change:
Multiple projects are using the current badge feature to create overview tables in confluence, Once we update to 4.12, this would be broken directly as I understand. Do you plan on enabling a smooth transition by making it possible to still use unauthenticated access for at least this one version for migration - so that both authenticated and the old way without authkey can co-exist for a grace period?
E.g. having an env_var like REALLY_BAD_IDEA_ENABLE_UNAUTHENTICATED_BADGES_FOR_THIS_ONE_VERSION?

nscuro:
@SaberStrat Thoughts on @savek-cc's remarks? Would you be willing to refactor your work to enable backward-compatibility?

savek-cc:
Maybe a different approach: Could we (in order to prepare for the preparation) create an AuthKey in DT 4.11 and include that in todays API requests (they would be ignored for now) and then give that team/API-key the new permission with DT 4.12?
So we could prepare (but not test) the transition with DT 4.11 and should be good once we update?

nscuro:
I think that might work. You can create a new team without permission, generate an API key for it, and then append `?apiKey=<YOUR_NEW_API_KEY>` to all your existing URLs. I don't think the API will fail a request if it contains parameters it doesn't recognize.

After upgrading to v4.12, you assign the `VIEW_BADGE` permission to your team and everything should work as expected.

SaberStrat:
The clean way would be of course a grace period for 4.12 (for the entirety of the minor version?), for example implemented as a modification of the current badge enabling config, where the setting
```
disabled <-> enabled (unauthenticated)
```
is turned into a

```
enabled (authenticated) <-> enabled (unauthenticated)
```.

The workaround suggested by you two, while meaning less effort for me, is a hacky one after all. We'd need to document it, and I don't know how this solution in documentation would feel like for users.

I'll let the maintainer @nscuro  decide which of the two solutions is sufficient and appropriate enough 🙂 for DT.

**Comment by nscuro** [comment]
Of course the clean way would be best. If you'd be willing to implement that, that would be awesome!

**Comment by SaberStrat** [comment]
Alright then.

Now I'll need help with organising the change in direction, i.e. project management decisions. 
What would be the best way to turn this PR and its issue about a breaking change into a deprecation for (probably) 4.12 and a breaking change in a later version?
1. Repurpose it--i.e. edit their descriptions and continue on the branch--into a deprecation for 4.12. For the feature removal, create a new issue and PR that'll be for later (e.g. 4.13)
2. Keep it for the feature removal later, finish it and mark it for a future release (e.g. 4.13) and fix future conflicts for when we get closer to that release. For the deprecation for 4.12, create a new issue and PR.

**Comment by savek-cc** [comment]
I don't think it should be an either or (authenticated OR unauthenticated) - but rather an override to allow access without an API key if none is provided. If one is provided, it should be checked as currently implemented.
The idea is to enable the authentication an onboard the projects one after another - but keep the old links working (those without key) for the transition period. Once all status pages/projects did add their (in our key probably the same for all) authKey to the request, this override can be disabled.
The override can then be completely removed in e.g. 4.13.

So the change is as already implemented - but there is an override (e.g. as env-var) that makes it not only accept valid authKeys - but also the empty one (as in none supplied).

What are your thoughts?

**Comment by savek-cc** [comment]
Correct - but even if that override is enabled, your code should still check the AuthKey - just on the "no auth key present" path there is an override to gran access anyway IF that config var is set.
This way, projects can already start using (and testing - as an invalid key would still be rejected) the new authentication.

@nscuro your thoughts?

**Comment by savek-cc** [comment]
> Wouldn't it be easier to just leave the existing `badge.enabled` toggle in place then?

Looking at the current documentation https://docs.dependencytrack.org/integrations/badges/ I too think that this is the best idea. The current options is especially about enabling unauthenticated access (pre 4.12)- and there is no need to disable badges if the only access is authenticated (4.13+).
So we add authenticated access independent of config - and make clear that the "unauthenticated access" toggle feature will go away with 4.13 in the release notes?

**Comment by SaberStrat** [comment]
Then I’m modifying this PR and issue and its frontend counterpart and create a separate pair for the removal for 4.13

**Comment by nscuro** [comment]
@SaberStrat Yeah I think that would be the easiest.

**Comment by nscuro** [comment]
@SaberStrat Do you have a rough idea as to how soon you'd be able to complete this work? No pressure, just curious WRT planning the v4.12 release.

**Comment by SaberStrat** [comment]
@nscuro Estimate by feeling: good case this week, bad case by end of next week. What’s the deadline for v4.12?

**Comment by nscuro** [comment]
We were hoping for early next week, since the next community meeting is on Wednesday.

**Comment by SaberStrat** [comment]
I'm actually struggling to come up with a way to toggle authentication on and off for a resource at runtime. 

DT is using alpine-server to apply an authentication filter for incoming requests, which has the method annotation `@AuthenticationNotRequired` to deactivate that filter for requests of that method. 

Previously, the 'Badges-way' was implemented like this: deactivate authentication in general for all methods of the resource with that annotation (static), and the ConfigProperty/general setting toggle made the resource methods return an empty body (dynamic)

Now what we want is for the toggle to basically do what the annotation did, but let the user control it, at runtime, i.e. dynamically. I'd either need to inject and query state of a ConfigProperty either in alpine-server or somewhere in DT where Alpine classes plug into DT. Or maybe work with Reflections and modify the annotation at runtime?

Either way, my thoughts lead down complicated paths with lots of effort, that I'm not sure is justifiable for a temporary deprecation. I'd be thankful for suggestions of justifiable implementations. Otherwise, I'm at the point where @savek-cc's [idea](https://github.com/DependencyTrack/dependency-track/pull/4059#issuecomment-2354983842) of adding a `apiKey` query parameter to badge requests in 4.11 sounds more reasonable 😅 .

**Comment by nscuro** [comment]
I think you could simply implement the permission check "manually" in the method body, instead of relying on the annotation.

There is a `qm.hasPermission` method that you can use, and `getPrincipal` gets you access to the user or API key if you need it (note: will be `null` for unauthenticated access).

You can access the `apiKey` parameter using the `@QueryParam` annotation as usual.

Yes it's a bit of duplicated code, but ultimately it's not a lot, and it's only in this specific endpoint, and only for v4.12. So I believe this is a reasonable option.

**Review by nscuro — COMMENTED** [review]
Looks good @SaberStrat! Could you please look into fixing the failing tests? Happy to merge once those are resolved.

```
 Error:  Failures: 
Error:    DefaultObjectGeneratorTest.testContextInitialized:38->testLoadDefaultPersonas:96 expected:<3> but was:<4>
Error:    DefaultObjectGeneratorTest.testLoadDefaultPersonas:96 expected:<3> but was:<4>
```

SaberStrat:
> I think you could simply implement the permission check "manually" in the method body, instead of relying on the annotation.
> 
> There is a `qm.hasPermission` method that you can use, and `getPrincipal` gets you access to the user or API key if you need it (note: will be `null` for unauthenticated access).
> 
> You can access the `apiKey` parameter using the `@QueryParam` annotation as usual.
> 
> Yes it's a bit of duplicated code, but ultimately it's not a lot, and it's only in this specific endpoint, and only for v4.12. So I believe this is a reasonable option.

Thank you for the help! It makes sense and goes in a similar direction that ACL is implemented in the badge resource.

SaberStrat:
Can't repro the failure locally.

nscuro:
Likely flaky. Triggered a re-run of the workflow.

nscuro (approved):

nscuro:
Seems like the CPAN registry is slow to respond, causing the test failure. Will merge, since the failing test is unrelated to this change.

PR #4067, opened 2024-08-12
[Feature] Vulnerability update notifications

4naesthetic:
### Description

Adds support for sending notifications when the severity of an existing vulnerability associated with a project is updated.

### Addressed Issue

Addresses: #2361
Supersedes: #2730

### Additional Details

This PR builds on the previous work done in #2730 by @Ehoky (thank you for your work!).

The main difference is that the quantity of notifications has been reduced. Instead of generating a separate notification for every affected component associated with an updated vulnerability we only generate one. The previous implementation generated a notification for every instance of a component, even if they represented the same logical software package, which means that if the same vulnerable component was used across 200 projects then 200 notifications would be generated instead of one.

Other differences include:
* No longer reporting on affected projects to prevent huge notification bodies, and only including a single component entry in the notification
* Adding vuln source, vuln ID and aliases to the vulnerability notification subject JSON payload
* Correctly emitting events from the `AbstractNistMirrorTask` that doesn't use the `VulnerabilityQueryManager::updateVulnerability` method

To support the 'Limit To Projects' functionality we still need to read all affected components (and their associated project IDs) before dispatching the notification even though they don't appear in the payload, so there is still some potential for scalability issues.

### Checklist


4naesthetic:
Hmm... one of the new tests is failing in CI but passing locally which I'm having trouble replicating (`NotificationUtilTest.testVulnerabilityUpdateMultipleComponents`). Any ideas?

nscuro:
Notifications are dispatched and processed asynchronously, so tests asserting on them can suffer from race conditions. I would recommend to use [Awaitility](http://www.awaitility.org/) to account for that, see here for an example: https://github.com/DependencyTrack/dependency-track/blob/b4ac9cd50a71db76128977de858c09af4091f282/src/test/java/org/dependencytrack/tasks/BomUploadProcessingTaskTest.java#L1298-L1302

4naesthetic:
> Notifications are dispatched and processed asynchronously, so tests asserting on them can suffer from race conditions. I would recommend to use [Awaitility](http://www.awaitility.org/) to account for that, see here for an example:
> 
> https://github.com/DependencyTrack/dependency-track/blob/b4ac9cd50a71db76128977de858c09af4091f282/src/test/java/org/dependencytrack/tasks/BomUploadProcessingTaskTest.java#L1298-L1302

Thanks for that. I've updated the asynchronous tests to use Awaitility so hopefully that should eliminate any race conditions.

4naesthetic:
Frontend PR: https://github.com/DependencyTrack/frontend/pull/974

I think this is ready for review. Any chance of this making it into the `4.12` release?

nscuro:
> I think this is ready for review. Any chance of this making it into the 4.12 release?

I'll try to get it reviewed this week. Unless there are some major issues I don't see why it wouldn't make it to v4.12.

src/main/java/org/dependencytrack/tasks/AbstractNistMirrorTask.java
```diff
@@ -55,6 +59,15 @@ Vulnerability synchronizeVulnerability(final QueryManager qm, final Vulnerabilit
...
                 if (!diffs.isEmpty()) {
                     logger.debug("%s has changed: %s".formatted(vuln.getVulnId(), diffs));
+
+                    // Check if we need to emit a vulnerability changed event
+                    if (diffs.containsKey("severity") && diffs.get("severity") != null) {
+                        final PersistenceUtil.Diff severityDiff = diffs.get("severity");
+                        if (severityDiff.before() instanceof final Severity oldSeverity && severityDiff.after() instanceof final Severity newSeverity) {
+                            Event.dispatch(new ProjectVulnerabilityUpdateEvent(persistentVuln, new VulnerabilityUpdateDiff(oldSeverity, newSeverity)));
+                        }
+                    }
```
nscuro, line 69: Since this code executes within a transaction, it *can* happen that we emit an event, but committing the transaction fails, leaving us in an inconsistent state. Consider moving the event dispatch outside of the transaction block.

src/main/java/org/dependencytrack/util/NotificationUtil.java
```diff
@@ -108,6 +110,29 @@ public static void analyzeNotificationCriteria(QueryManager qm, Vulnerability vu
...
     }
 
+    public static void analyzeNotificationCriteria(QueryManager qm, Vulnerability vulnerability, VulnerabilityUpdateDiff vulnerabilityUpdateDiff) {
+        // Vulnerabilities are not guaranteed to include component relationships depending on where the event was generated
+        final Vulnerability completeVulnerability = qm.getVulnerabilityByVulnId(vulnerability.getSource(), vulnerability.getVulnId());
+        final List<Component> components = completeVulnerability.getComponents();
+        if (components != null && !components.isEmpty()) {
+            // To reduce noise we only emit a single notification for each updated vulnerability if it affects one
+            // of our components. The component details are still useful for event consumers, so we pick the first one.
+            final Component detachedComponent = qm.detach(Component.class, components.get(0).getId());
```
nscuro, line 120: I feel like this could be a bit misleading. A vulnerability can potentially affect multiple libraries, operating systems, or hardware components. What is the benefit of including only one component?

src/main/java/org/dependencytrack/util/NotificationUtil.java
```diff
@@ -108,6 +110,29 @@ public static void analyzeNotificationCriteria(QueryManager qm, Vulnerability vu
         }
     }
 
+    public static void analyzeNotificationCriteria(QueryManager qm, Vulnerability vulnerability, VulnerabilityUpdateDiff vulnerabilityUpdateDiff) {
+        // Vulnerabilities are not guaranteed to include component relationships depending on where the event was generated
+        final Vulnerability completeVulnerability = qm.getVulnerabilityByVulnId(vulnerability.getSource(), vulnerability.getVulnId());
+        final List<Component> components = completeVulnerability.getComponents();
```
nscuro, line 116: Fetching all associated components at once can be very costly in large portfolios, especially since only one of them is actually used. Do we need to include this info (see comment below as well)?

src/main/java/org/dependencytrack/notification/vo/ProjectVulnerabilityUpdate.java
```diff
@@ -0,0 +1,43 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.notification.vo;
+
+import org.dependencytrack.model.Component;
+import org.dependencytrack.model.Vulnerability;
+import org.dependencytrack.model.VulnerabilityUpdateDiff;
+
+public class ProjectVulnerabilityUpdate {
```
nscuro, line 25: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

I'd further recommend to rename this class to `VulnerabilityUpdated`, since it's not specific to a project, as the current name would suggest.

src/main/java/org/dependencytrack/model/VulnerabilityUpdateDiff.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.model;
+
+public class VulnerabilityUpdateDiff {
```
nscuro, line 21: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/tasks/ProjectVulnerabilityUpdateTask.java
nscuro, line 1: If we can get rid of the fetching of component data, this task and its corresponding event could be removed, and you could simply call `NotificationUtil#analyzeNotificationCriteria` in-line.

Using a task like this is really only required if generating the notification involves heavyweight I/O or computation. Otherwise, notification dispatch via `Notification#dispatch` is already asynchronous.

src/main/java/org/dependencytrack/notification/publisher/PublishContext.java
```diff
@@ -180,4 +190,21 @@ private static Vulnerability convert(final org.dependencytrack.model.Vulnerabili
...
+                                                       final org.dependencytrack.model.VulnerabilityUpdateDiff notificationVulnUpdateDiff) {
+            if (notificationVuln == null || notificationVulnUpdateDiff == null) {
+                return null;
+            }
+            return new VulnerabilityUpdateDiff(
+                    notificationVuln.getVulnId(),
+                    notificationVuln.getSource(),
+                    notificationVulnUpdateDiff.getOldSeverity(),
+                    notificationVulnUpdateDiff.getNewSeverity()
+            );
```
nscuro, line 205: Do not include the actual diff values in here. Subjects are intended to give only surface-level information about the notification being dispatched, and they all appear in the logs. 

Consider just mapping to `SUBJECT_VULNERABILITY`, and re-using `Vulnerability#convert` instead.

nscuro (changes requested):

src/main/java/org/dependencytrack/notification/vo/ProjectVulnerabilityUpdate.java
4naesthetic: By using `ProjectVulnerabilityUpdate` instead of `VulnerabilityUpdate` I was trying to make it clear that the notification only represents an update to a vulnerability that affects a project in a portfolio. I thought it might be confusing to call it `VulnerabilityUpdate`, as that might lead someone to think that the notification would fire on any updates to a vulnerability in the vuln DB. Perhaps that's already clear enough though from the notification being scoped to the `PORTFOLIO`?

nscuro: Ah, I see, that reasoning makes sense. IMO, `VulnerabilityUpdated` would still be less ambiguous than `ProjectVulnerabilityUpdated`. I guess there are still better ways to word the intention though, but my non-native-speaker skills are not helping.

I was thinking something like these, although they still sound a bit awkward:

* `AffectingVulnerabilityUpdated` - i.e. a vuln *affecting* something was updated
* `FindingVulnerabilityUpdated` - i.e. a vuln that is part of a finding was updated

src/main/java/org/dependencytrack/util/NotificationUtil.java
4naesthetic: There's definitely some nuance that I missed here. 

I considered the earlier implementation where the event notification would include 1 vulnerability, 1 component and a list of affected projects (similar to a `NEW_VULNERABILITY` notification), which would mean generating a notification for every affected component entry in the database. As each logical component has a unique entry in the database for every project that uses it, this could mean emitting a large number of notifications where the content is identical except for the `component.uuid` field. So to try and reduce this redundancy and (in theory) reduce the database load, my logic was to choose just one of the "identical" component entries and emit a single event per updated vulnerability.

What I missed was that there could be separate logical components affected by the same vulnerability and picking only one of them doesn't make sense as you said.

I think what this comes down to is what useful contextual information about a vulnerability do we want to include in the notification? For my use case I actually only care about the `vulnerability` object itself, but others may have use cases where they need the affected component information (and potentially affected projects) as well.

I may be missing something, but at the moment there doesn't seem to be a way to fetch a list of unique logical component identities without doing a full table scan and de-duplicating. If this can't be done efficiently, perhaps it makes sense not to include component information in the notification. What do you think?

src/main/java/org/dependencytrack/util/NotificationUtil.java
4naesthetic: This can absolutely be optimised if we don't need component information in the notification. It think it would be more efficient to include the affected projects instead, and use `qm.getAffectedProjects(vuln)` much like the REST API does in `VulnerabilityResource.getAffectedProject()` as we can then reuse this for filtering in `NotificationRouter`.

src/main/java/org/dependencytrack/util/NotificationUtil.java
nscuro: IMO we should keep the information in this notification as minimal as possible. All data we add now, users will start depending on it, and we can't easily remove it again. I think it's better to wait for users to come forward with their use cases, and then evaluating if / how that can be achieved with more data.

Transmitting affected projects is not feasible (see for example https://github.com/DependencyTrack/hyades/issues/467 where we ran into this in another context).

Communicating logical components is complicated as you said, and I am hesitant to add this complexity without us knowing that users really need it.

If we only include the vulnerability itself now, receivers will have enough information to query the `/api/v1/source/{source}/vuln/{vuln}/projects` endpoint to fetch affected projects.

src/main/java/org/dependencytrack/notification/vo/ProjectVulnerabilityUpdate.java
4naesthetic: Out of those two suggestions I think `FindingVulnerabilityUpdated` is better, but still might be confusing as someone might expect to receive a `Finding` in the event payload (which has a specific meaning and schema as an API resource).

Is `ActiveVulnerabilityUpdated` any better? It mirrors the idea of 'Active' and 'Inactive' projects.

src/main/java/org/dependencytrack/util/NotificationUtil.java
4naesthetic: `qm.getAffectedProjects(vuln)` actually calls `vuln.getComponents()` in the query manager anyway (which makes sense now that I've thought about it properly). Even if we don't include affected projects or components in the notification subject, we still need the affected project list to be able to limit notifications by project (or tags in the future). On the other hand, if we don't filter by affected projects then we might emit a large number of irrelevant notifications whenever there is a vulnerability database update, which also has performance implications.

Issue #4092, opened 2024-08-23
Enhance "Create Project" dialog to include team selection


Gepardgame:
### Current Behavior

With Portfolio Access Control enabled, a user who can create a Project (needs PORTFOLIO_MANAGEMENT), but hasn't the right to assign a team to this project in the settings (needs ACCESS_MANAGEMENT), can't see the project afterwards, cause of the PAC. An Admin needs to assign it manually.

### Proposed Behavior

The user who wants to create a new project has the option to select a team to which the project will be assigned
Mandatory only if PAC feature is enabled.
User with ACCESS_MANAGEMENT Permission should be allowed to add any available team to the project. Normal user should only be allowed to add a team he or she belongs to.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #4093, opened 2024-08-23
Feat/add team selection in create project button

Gepardgame:
### Description
Adds a option to the Create Project dialog to add a team

Added API endpoint to get available Teams. Also needed to create two new classes, for returning.

Altered create Project API Point, to be able to also add a initial Team to the project.

### Addressed Issue

Fixes [#4092](https://github.com/DependencyTrack/dependency-track/issues/4092#issue-2482976965)

### Checklist

~~- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective~~


src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -220,26 +192,77 @@ public Response deleteTeam(Team jsonTeam) {
         }
     }
 
+    @GET
+    @Path("/available-teams")
```
nscuro, line 196: Consider renaming to `/visible` instead, since this endpoint returns teams *visible* to the authenticated user.

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -220,26 +192,77 @@ public Response deleteTeam(Team jsonTeam) {
...
+            }
+            qm.getPersistenceManager().detachCopyAll(configProperties);
+            qm.close();
+            List<Permission> permissions = user.getPermissions();
+            for (Permission permission : permissions) {
+                if (permission.getName().equals("ACCESS_MANAGEMENT")) {
+                    isAllTeams = true;
+                    break;
+                }
+            }
```
nscuro, line 237: Consider using `qm.hasAccessManagementPermission(getPrincipal())` instead:

https://github.com/DependencyTrack/dependency-track/blob/355598157f163241f98bf393f2124a4f2a6ac95f/src/main/java/org/dependencytrack/persistence/QueryManager.java#L1329-L1334

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -220,26 +192,77 @@ public Response deleteTeam(Team jsonTeam) {
...
+            for (final ConfigProperty configProperty : configProperties) {
+                // Replace the value of encrypted strings with the pre-defined placeholder
+                if (configProperty.getGroupName().equals("access-management")
+                        && configProperty.getPropertyName().equals("acl.enabled")) {
+                    required = configProperty.getPropertyValue().equals("true");
+                    break;
+                }
+            }
+            qm.getPersistenceManager().detachCopyAll(configProperties);
+            qm.close();
```
nscuro, line 230: Consider using `qm.isEnabled(ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED)` instead:

https://github.com/DependencyTrack/dependency-track/blob/355598157f163241f98bf393f2124a4f2a6ac95f/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L869

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -220,26 +192,77 @@ public Response deleteTeam(Team jsonTeam) {
...
+    @GET
+    @Path("/available-teams")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Returns a list of Teams what are available as selection", description = "Requires permission <strong>PORTFOLIO_MANAGEMENT</strong></p>")
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", description = "The Available Teams", content = @Content(schema = @Schema(implementation = AvailableTeams.class))),
+            @ApiResponse(responseCode = "401", description = "Unauthorized"),
+            @ApiResponse(responseCode = "404", description = "Teams could not be found")
+    })
+    @PermissionRequired(Permissions.Constants.PORTFOLIO_MANAGEMENT)
```
nscuro, line 204: This endpoint should not require any permissions.

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -220,26 +192,77 @@ public Response deleteTeam(Team jsonTeam) {
...
+            if (super.isLdapUser()) {
+                user = qm.getLdapUser(getPrincipal().getName());
+            } else if (super.isManagedUser()) {
+                user = qm.getManagedUser(getPrincipal().getName());
+            } else if (super.isOidcUser()) {
+                user = qm.getOidcUser(getPrincipal().getName());
+            } else {
+                return Response.status(401).build();
+            }
+            final List<Team> allTeams = qm.getTeams();
```
nscuro, line 219: Try to delay the loading of *all* teams until you know it's *really* required (i.e. ACL is disabled or user has the `ACCESS_MANAGEMENT` permission).

src/main/java/org/dependencytrack/model/LittleTeam.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class LittleTeam implements Serializable {
+
+  private UUID value;
+  private String text;
```
nscuro, line 31: Consider using more descriptive names. While `value` and `text` might make sense for rendering in HTML forms, it doesn't for anyone else calling this endpoint via REST clients.

src/main/java/org/dependencytrack/model/LittleTeam.java
```diff
@@ -0,0 +1,53 @@
...
+package org.dependencytrack.model;
+
+import java.io.Serializable;
+
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class LittleTeam implements Serializable {
```
nscuro, line 28: Consider making this a [`record`](https://docs.oracle.com/en/java/javase/21/language/records.html) instead.

src/main/java/org/dependencytrack/model/LittleTeam.java
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.model;
```
nscuro, line 19: This should be in `org.dependencytrack.resources.v1.vo` instead, since it's exclusively used for REST API responses.

Alternatively, consider making it a nested class in `AvailableTeams`.

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -220,6 +225,33 @@ public Response deleteTeam(Team jsonTeam) {
...
+    })
+    public Response availableTeams() {
+        try (QueryManager qm = new QueryManager()) {
+            Principal user = getPrincipal();
+            List<Team> userTeams = new ArrayList<Team>();
+            if (user instanceof final UserPrincipal userPrincipal) {
+                userTeams = userPrincipal.getTeams();
+            } else if (user instanceof final ApiKey apiKey) {
+                userTeams = apiKey.getTeams();
+            }
```
Gepardgame, line 245: @ nscuro Would it suffice to only do that for the UserPrincipal? As API Keys are only for teams, so there is only one team to return, and mostly this and the /project calls are done by users, aren't their?

Gepardgame:
@nscuro Is there a chance that will get into 4.12, we would love to use it.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -284,6 +309,7 @@ public void getProjectByUuidTest() {
                 .withMatcher("childUuid", equalTo(childProject.getUuid().toString()))
                 .isEqualTo("""
                         {
+                          "accessTeams": [],
```
nscuro, line 312: I'd rather not include this field in responses. You can avoid having `accessTeams` returned in responses by annotating `Project.getAccessTeams` with `@JsonIgnore`, and `Project.setAccessTeams` with `@JsonSetter`.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -478,6 +507,102 @@ public void createProjectEmptyTest() {
...
     }
 
+    @Test
+    public void createProjectWithExistingTeamRequiredTest() {
+        getUserToken(false);
+        Team AllowedTeam = qm.createTeam("AllowedTeam", false);
+        Project project = new Project();
+        project.setName("ProjectWithExistingTeamRequired");
+        qm.addUserToTeam(testUser, AllowedTeam);
+        qm.createConfigProperty("access-management", "acl.enabled", "true", IConfigProperty.PropertyType.BOOLEAN, "");
```
nscuro, line 517: Consider using constants for creating config properties, as seen in other places in this file:

```java
qm.createConfigProperty(
        ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED.getGroupName(),
        ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED.getPropertyName(),
        "true",
        ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED.getPropertyType(),
        null
);
```

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -89,6 +97,23 @@ public void after() throws Exception {
         super.after();
     }
 
+    public void getUserToken(boolean isAdmin) {
```
nscuro, line 100: Nit: This method says `get...` but doesn't return anything. Consider either returning the JWT instead of assigning it to a class field, or renaming the method to something like `setupUserToken`.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -413,6 +439,7 @@ public void getProjectByUnknownTagTest() {
     @Test
     public void createProjectTest(){
         Project project = new Project();
+        project.setAccessTeams(new ArrayList<Team>());
```
nscuro, line 442: Making `accessTeams` a mandatory field would be a breaking change in the API. Initializing it with an empty list should not be necessary. The resource code should expect this field to be missing (`null`), and handle it accordingly.

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -309,6 +315,32 @@ public Response createProject(Project jsonProject) {
...
+            jsonProject.setAccessTeams(new ArrayList<Team>());
+            for (Team choosenTeam : choosenTeams) {
+                if (!visibleUuids.contains(choosenTeam.getUuid())) {
+                    return isAdmin ? Response.status(404).entity("This team does not exist!").build()
+                            : Response.status(403)
+                                    .entity("You don't have the permission to assign this team to a project.").build();
+                }
+                Team ormTeam = qm.getObjectByUuid(Team.class, choosenTeam.getUuid());
+                jsonProject.addAccessTeam(ormTeam);
+            }
```
nscuro, line 343: In the spirit of work avoidance, try to only execute this logic when really necessary. Also, keep in mind that existing clients will not submit `accessTeams`, so you have to expect it to be `null`.

If ACL is not enabled, this code doesn't need to run. Checking permissions and fetching available teams would be wasted work.

If ACL *is* enabled, but `jsonProject.getAccessTeams()` is `null` or empty, you can short-circuit immediately and return a `422` response (a `400` response would be more consistent with other API endpoints).

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
nscuro: The API key's team could have the `ACCESS_MANAGEMENT` permission, in which case we'd need to return all teams here.

Also more generally, for consistency, let's keep handling of API keys in here. Even if it will just return one team most of the time.

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -220,6 +225,32 @@ public Response deleteTeam(Team jsonTeam) {
...
+            boolean required = qm.isEnabled(ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED);
```
nscuro, line 245: It's still a bit odd that the `/api/v1/team/visible` endpoint returns a `required` field that indicates to the frontend, in a specific view, whether a team needs to be submitted to the `PUT /api/v1/project` endpoint. `required` doesn't make sense for any other use case. Do you see a way to push this logic to the frontend instead? Then we could just return an array of `Team` here, which from an API standpoint would make a lot more sense.

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
Gepardgame: > If ACL is not enabled, this code doesn't need to run. Checking permissions and fetching available teams would be wasted work.
I thought that it would be useful, that you can add teams even though its not enabled. So I would change it to only be run if accessTeams is not null/empty

src/main/java/org/dependencytrack/resources/v1/TeamResource.java
Gepardgame: Maybe make the ACCESS_MANAGEMENT_ACL_ENABLED public to view, with the /configProperty/public Endpoint to get from.

nscuro: That sounds reasonable to me. That flag is not confidential so exposing it should be fine.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
Gepardgame: I have found a way to revert these changes but still be able to return accessTeams, but only if its not empty. I think for the future that would be helpful, for some things like seeing inside a project, which teams its corresponds to, so you also can change it easily. This would also allow to sort/search for teams.

nscuro: I think in general it would be best to simply not return the teams here *at all*. [`Team` records may include `apiKeys`](https://github.com/stevespringett/Alpine/blob/953ad483b1b12195250732724475aea96fa23ff1/alpine-model/src/main/java/alpine/model/Team.java#L94-L96), and those should not be leaked, given this endpoint is accessible with lower privileges than the `/api/team` endpoints. 

Also, this endpoint already tends to return way too much data, so I'm being extra cautious with adding more fields to it, in particular if those fields are a potentially unbounded collection.

We can always spend time on making `accessTeams` safe to return here once we have a definitive use-case for it. For the time being, please ensure it's not returned.

Gepardgame: >  [Team records may include apiKeys](https://github.com/stevespringett/Alpine/blob/953ad483b1b12195250732724475aea96fa23ff1/alpine-model/src/main/java/alpine/model/Team.java#L94-L96)

@nscuro I just realized that you get the API keys also over the new /team/visible Endpoint, which is open to all authenticated users.

PR #863, opened 2020-12-02
Alerts list fix

technoo10201:
Fix to remove the limitation of 100 alerts in the alerts list. 
https://github.com/DependencyTrack/frontend/issues/46

PR #4131, opened 2024-09-06
Feat/customizable login page

Gepardgame:
### Description

Adds an API Endpoint, open to anyone, that only can return the custom welcome message given by admins.
Adds  2 ConfigProperties to store the message and if it should be shown.

### Addressed Issue

Fixes [#863](https://github.com/DependencyTrack/frontend/issues/863)

### Checklist

~~- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective~~

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/ConfigPropertyResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ConfigPropertyResource.java
@@ -22,6 +22,7 @@ import alpine.model.ConfigProperty;
 import alpine.server.auth.AuthenticationNotRequired;
 import alpine.server.auth.PermissionRequired;
 import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
 import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.media.Schema;
@@ -31,6 +32,7 @@ import io.swagger.v3.oas.annotations.security.SecurityRequirement;
 import io.swagger.v3.oas.annotations.security.SecurityRequirements;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import org.dependencytrack.auth.Permissions;
+import org.dependencytrack.model.ConfigPropertyConstants;
 import org.dependencytrack.persistence.QueryManager;
 
 import jakarta.validation.Validator;
@@ -38,6 +40,7 @@ import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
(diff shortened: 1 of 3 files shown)


src/main/java/org/dependencytrack/resources/v1/ConfigPropertyResource.java
```diff
@@ -156,5 +157,25 @@ public Response updateConfigProperty(List<ConfigProperty> list) {
...
+        try (QueryManager qm = new QueryManager(getAlpineRequest())) {
+            ConfigProperty greetingMessage = qm.getConfigProperty("Message", "welcomeMessage");
+            ConfigProperty isGreetingMessage = qm.getConfigProperty("Message", "isWelcomeMessage");
+            qm.close();
+            List<ConfigProperty> combined = new ArrayList<ConfigProperty>();
+            combined.add(greetingMessage);
+            combined.add(isGreetingMessage);
+            return Response.ok(combined).build();
+        }
+    }
```
nscuro, line 179: I'm not particularly fond of adding new endpoints for every new property that doesn't require authentication, especially considering the language-related question you asked in Slack. I think it would "scale" better if we added a new `/public` endpoint that would only return `ConfigProperty` records that we explicitly mark as "public".

We did something along those lines in Hyades, to mark properties as read-only or read-writable:

https://github.com/DependencyTrack/hyades-apiserver/blob/eca7ab7c672a2b128c2bb6d729b0fb12d532e60d/src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java#L160-L162 

https://github.com/DependencyTrack/hyades-apiserver/blob/eca7ab7c672a2b128c2bb6d729b0fb12d532e60d/src/main/java/org/dependencytrack/resources/v1/AbstractConfigPropertyResource.java#L59-L66

You could follow a similar approach to filter properties that can safely be returned for in the `/public` endpoint. The frontend could then call this endpoint to retrieve welcome message and language preferences (and potentially more in the future).

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/ConfigPropertyResource.java
Gepardgame: I figured that as well, easiest way would be adding a Property to the constants that mark them optionnely as public, like that:
```
    BOM_VALIDATION_TAGS_EXCLUSIVE("artifact", "bom.validation.tags.exclusive", "[]", PropertyType.STRING, "JSON array of tags for which BOM validation shall NOT be performed"),
    WELCOME_MESSAGE("general", "welcome.message.html", "%20%3Chtml%3E%3Ch1%3EYour%20Welcome%20Message%3C%2Fh1%3E%3C%2Fhtml%3E", PropertyType.STRING, "Custom HTML Code that is displayed before login", true),
    IS_WELCOME_MESSAGE("general", "welcome.message.enabled", "false", PropertyType.BOOLEAN, "Bool that says wheter to show the welcome message or not", true);

    private final String groupName;
    private final String propertyName;
    private final String defaultPropertyValue;
    private final PropertyType propertyType;
    private final String description;
    private final Boolean isPublic;


	ConfigPropertyConstants(String groupName, String propertyName, String defaultPropertyValue, PropertyType propertyType, String description) {
        this.groupName = groupName;
        this.propertyName = propertyName;
        this.defaultPropertyValue = defaultPropertyValue;
        this.propertyType = propertyType;
        this.description = description;
        this.isPublic = false;
    }
    
    ConfigPropertyConstants(String groupName, String propertyName, String defaultPropertyValue, PropertyType propertyType, String description, Boolean isPublic) {
        this.groupName = groupName;
        this.propertyName = propertyName;
        this.defaultPropertyValue = defaultPropertyValue;
        this.propertyType = propertyType;
        this.description = description;
        this.isPublic = isPublic;
    }
```

src/main/java/org/dependencytrack/resources/v1/ConfigPropertyResource.java
```diff
@@ -156,5 +160,26 @@ public Response updateConfigProperty(List<ConfigProperty> list) {
...
+    public Response getGreetingMessage(
+            @Parameter(description = "The group name of the value to retrieve", required = true) @PathParam("groupName") String groupName,
+            @Parameter(description = "The property name of the value to retrieve", required = true) @PathParam("propertyName") String propertyName) {
+        try (QueryManager qm = new QueryManager(getAlpineRequest())) {
+            ConfigProperty property = qm.getConfigProperty(groupName, propertyName);
+            qm.close();
+            ConfigPropertyConstants publicConfigProperty = ConfigPropertyConstants.ofProperty(property);
+            if (!publicConfigProperty.getIsPublic()) {
+                return Response.status(Response.Status.FORBIDDEN).entity("This is not a public visible ConfigProperty").build();
+            }
```
nscuro, line 181: A few nits:

* You don't need to close `qm` manually - that's what the try-with-resources is for
* Consider performing the `isPublic` validation *before* querying the database, potentially even before you open the `QueryManager` - database access is not required for the validation

src/test/java/org/dependencytrack/resources/v1/ConfigPropertyResourceTest.java
```diff
@@ -383,4 +384,17 @@ public void updateConfigPropertyBomValidationTagsInclusiveTest() {
         assertThat(getPlainTextBody(response)).isEqualTo("Value must be a valid JSON array of strings");
     }
 
+    @Test
+    public void getPublicForbiddenTest() {
```
nscuro, line 388: A test for the success-case would be good. You could even loop over all values of `ConfigPropertyConstants` to verify the desired behavior for all of them.

Issue #4140, opened 2024-09-12
Emails don't render certain things right


Gepardgame:
### Current Behavior

Emails that are being sent from the server don't render certain things right, like & -> &amp; and also some things not at all, like §.

### Steps to Reproduce

1. Change the Test notification under resources/v1/NotificationPublisher in the function with Path "test/smtp", to include &, §, etc. 
2. Send the email under administration > email >send test Email
3. Look at your inbox

### Expected Behavior

Tat the email renders this symbols right.

### Dependency-Track Version

4.11.7

### Dependency-Track Distribution

Executable WAR

### Database Server

H2

### Browser

N/A

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

PR #4141, opened 2024-09-12
Feat: Fix that Emails render all symbols right

Gepardgame:
### Description

Fixes that emails can render all symbols right:
image

### Addressed Issue

Fixes [#4140](https://github.com/DependencyTrack/dependency-track/issues/4140)

### Additional Details

Changed MIME Type to include chaset=utf-8, so it uses utf-8 and not just ASCII.
Unescapse the content, so e.g. & does not convert to &amp; and other such things.

### Checklist

~~- [ ] This PR implements an enhancement, and I have provided tests to verify that it works as intended~~


src/main/java/org/dependencytrack/notification/publisher/SendMailPublisher.java
```diff
@@ -141,7 +143,7 @@ private void sendNotification(final PublishContext ctx, Notification notificatio
                     .from(smtpFrom)
                     .to(destinations)
                     .subject(emailSubjectPrefix + " " + notification.getTitle())
-                    .body(content)
+                    .body(StringEscapeUtils.unescapeHtml4(content))
```
nscuro, line 146: I wonder if this will cause issues when the MIME type is `text/html` or similar, which would prompt mail clients to attempt to render the HTML. I'm not sure we can guarantee that the notification content is save to render, or it even makes sense to do so. For example, vulnerability descriptions might include HTML as code snippets.

Gepardgame: So the simplest solution would be to either unescape it if its text/plain and then its text/html leave it.

The thing is I saw is that the content is not perfectly suitable for rendering in hml either, if its left, as it was. & symbol is rightly escaped to &amp;, but the § or ' doesn't get escaped, and therefore can't be rendered in html. For that would the simplet solution be unescape it and escape it again, cause StringEscapeUtils.escapeHtml4, does it the right way. Or do you know a way that the content isn't escaped in the first place all along.

nscuro (commented):
Looks OK to me. Could you add a test to [`SendMailPublisherTest`](https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/notification/publisher/SendMailPublisherTest.java) that includes special characters, to verify the escaping works as expected?

Gepardgame:
I don't know why the test fails, locally it runs completely fine.

nscuro:
Maybe flaky, haven't seen that test fail before but could always be. Restarted the test workflow.

nscuro (approved):

Issue #4176, opened 2024-09-26
Visible Endpoints returns API Keys


Gepardgame:
### Current Behavior

The /team/visible endpoint returns API Keys from the teams you belong to, even so your not allowed to see the Keys.

### Proposed Behavior

Should not return the API Keys, as that would be a security issue.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #4177, opened 2024-09-26
Visible Endpoint returns only Visible Teams(name, uuid)

Gepardgame:
### Description

/team/visible Endpoints no longer returns hole team, but instead only name and uuid.

### Addressed Issue

Fixes [#4176](https://github.com/DependencyTrack/dependency-track/issues/4176)

### Additional Details
N/A
### Checklist


nscuro (approved):
Thanks!

The test failures are unrelated. We're observing more aggressive rate limiting from the GitHub container registry, which is being used to pull Trivy images for testing.

PR #4184, opened 2024-09-27
feat#4148 introduce is latest project flag & allow policies to be limited to latest version

### LINKED ISSUE #4148: Introduce isLatest flag to mark latest version of project
### Current Behavior

Currently, Dependency-Track has no idea which version is the latest. Version numbers follow different formats, so figuring this out automatically is difficult, as no solution will fit everyone.
Some users try to get around this, by labelling latest versions with tags. However, having this information in a standardized format would be much more beneficial:
1. Other features (e.g. collection logic, see https://github.com/DependencyTrack/dependency-track/pull/3258)  could make use of it. In future possibly also rules could be applied only to latest versions (e.g. outdated dependency rules, which make no sense on older versions)
3. The system could automate the update of this flag, not relying on client scripts to update the "old" latest project to remove the tag and add it to the newest version. This causes additional client scripting effort and can fail, leaving it in inconsistent state.

### Proposed Behavior

Based on discussion with @nscuro  in https://github.com/DependencyTrack/dependency-track/pull/3258 there should be a flag "isLatest" introduced, which marks a project as latest version.
When this is set to TRUE on a project, the flag automatically is reset to FALSE on all other versions of this project.
This can be enforced with UNIQUE constraint on (NAME, IS_LATEST).

Hint:
If https://github.com/DependencyTrack/dependency-track/pull/3258 is merged already, the preparation for Collection logic LATEST_VERSION_CHILDREN in ProjectMetricsUpdateTask.updateLatestVersionChildrenCollectionMetrics() should be activated during implementation of this feature.

### Checklist

rkg-mm:
### Description
Frontend PR: https://github.com/DependencyTrack/frontend/pull/1017

Implements #4148 : 
1. Allows projects to be marked as "latest version" via API endpoints allowing creation and updating of projects
2. It is ensured only 1 project version (per project name) is marked as latest, previously marked version is no longer set to latest after a new version is set to latest
3. Allows policies to be limited to project versions marked as latest (e.g. helpful to only apply the version distance policy to the latest development version)
4. Fixed Bug: ProjectQueryManager contains 2 createProject() Methods, but only one triggered a PROJECT_CREATED event. Refactored both methods to use same implementation to avoid such inconsistencies.

### Addressed Issue

closes #4148 

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/Policy.java
+++ b/src/main/java/org/dependencytrack/model/Policy.java
@@ -142,6 +142,10 @@ public class Policy implements Serializable {
     @Column(name = "INCLUDE_CHILDREN", allowsNull = "true") // New column, must allow nulls on existing data bases)
     private boolean includeChildren;
 
+    @Persistent
+    @Column(name = "ONLY_FOR_LATEST_PROJECT_VERSION", allowsNull = "true") // New column, must allow nulls for existing data bases
+    private boolean onlyForLatestProjectVersion;
+
     public long getId() {
         return id;
     }
@@ -223,5 +227,13 @@ public class Policy implements Serializable {
 
     public void setIncludeChildren(boolean includeChildren) {
         this.includeChildren = includeChildren;
+    }
+
+    public boolean isOnlyForLatestProjectVersion() {
+        return onlyForLatestProjectVersion;
+    }
+
+    public void setOnlyForLatestProjectVersion(boolean onlyForLatestProjectVersion) {
+        this.onlyForLatestProjectVersion = onlyForLatestProjectVersion;
     }
 }
--- a/src/main/java/org/dependencytrack/model/Policy.java
+++ b/src/main/java/org/dependencytrack/model/Policy.java
@@ -143,8 +143,8 @@ public class Policy implements Serializable {
     private boolean includeChildren;
 
     @Persistent
-    @Column(name = "ONLY_FOR_LATEST_PROJECT_VERSION", allowsNull = "true") // New column, must allow nulls for existing data bases
-    private boolean onlyForLatestProjectVersion;
+    @Column(name = "ONLY_LATEST_PROJECT_VERSION", defaultValue = "false")
+    private boolean onlyLatestProjectVersion = false;
 
     public long getId() {
         return id;
@@ -229,11 +229,11 @@ public class Policy implements Serializable {
         this.includeChildren = includeChildren;
     }
 
-    public boolean isOnlyForLatestProjectVersion() {
-        return onlyForLatestProjectVersion;
+    public boolean isOnlyLatestProjectVersion() {
+        return onlyLatestProjectVersion;
     }
 
-    public void setOnlyForLatestProjectVersion(boolean onlyForLatestProjectVersion) {
-        this.onlyForLatestProjectVersion = onlyForLatestProjectVersion;
+    public void setOnlyLatestProjectVersion(boolean onlyLatestProjectVersion) {
+        this.onlyLatestProjectVersion = onlyLatestProjectVersion;
     }
 }
--- a/src/main/java/org/dependencytrack/model/Project.java
+++ b/src/main/java/org/dependencytrack/model/Project.java
@@ -24,6 +24,7 @@ import alpine.server.json.TrimmedStringDeserializer;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonIncludeProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSetter;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.JsonSerializer;
@@ -95,7 +96,8 @@ import java.util.UUID;
                 @Persistent(name = "properties"),
                 @Persistent(name = "tags"),
                 @Persistent(name = "accessTeams"),
-                @Persistent(name = "metadata")
+                @Persistent(name = "metadata"),
+                @Persistent(name = "isLatest")
         }),
         @FetchGroup(name = "METADATA", members = {
                 @Persistent(name = "metadata")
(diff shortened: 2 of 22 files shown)


src/main/java/org/dependencytrack/upgrade/v4120/v4120Updater.java
```diff
@@ -286,4 +287,18 @@ private void dropAuthorColumns(final Connection connection) throws SQLException
...
+        LOGGER.info("Setting IS_LATEST flag to false for all Project entries");
+        try (final Statement stmt = connection.createStatement()) {
+            stmt.executeUpdate("""
+                    UPDATE "PROJECT" SET "IS_LATEST" = false
+                    """);
+            stmt.executeUpdate("""
+                    UPDATE "POLICY" SET "ONLY_FOR_LATEST_PROJECT_VERSION" = false
+                    """);
+        }
+    }
```
nscuro, line 302: Analog to my comment in the `Project` class - This seems to be redundant. DataNucleus will set the specified default value in the schema.

src/test/java/org/dependencytrack/ResourceTest.java
```diff
@@ -123,6 +124,16 @@ public void initializeWithPermissions(Permissions... permissions) {
...
 
+    protected void enablePortfolioAccessControl() {
+        qm.createConfigProperty(
+                ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED.getGroupName(),
+                ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED.getPropertyName(),
+                "true",
+                ConfigPropertyConstants.ACCESS_MANAGEMENT_ACL_ENABLED.getPropertyType(),
+                null
+        );
+    }
```
nscuro, line 135: Ha, good idea!

src/main/java/org/dependencytrack/model/Policy.java
```diff
@@ -142,6 +142,10 @@ public enum ViolationState {
     @Column(name = "INCLUDE_CHILDREN", allowsNull = "true") // New column, must allow nulls on existing data bases)
     private boolean includeChildren;
 
+    @Persistent
+    @Column(name = "ONLY_FOR_LATEST_PROJECT_VERSION", allowsNull = "true") // New column, must allow nulls for existing data bases
```
nscuro, line 146: Consider also using `defaultValue = "false"` here, same as for `Project#isLatest`. Also, can we remove the `_FOR_` part? Seems like an unnecessary filler word for a database column.

src/main/java/org/dependencytrack/model/Project.java
```diff
@@ -109,6 +111,7 @@
                 @Persistent(name = "parent")
         })
 })
+@Index(name = "PROJECT_NAME_IS_LATEST_IDX", members = {"name", "isLatest"})
```
nscuro, line 114: Do we need two indexes covering the `IS_LATEST` column?

Actually I believe we might need none. The DB query planner can already use the index on `NAME` to reduce the result to a specific project. Filtering the remaining records (i.e. all versions of that project) by `IS_LATEST` is then unlikely to yield much benefit by being indexed. But we pay a penalty for inserts and updates by having them.

src/main/java/org/dependencytrack/model/Project.java
```diff
@@ -270,6 +273,11 @@ public enum FetchGroup {
     @JsonSerialize(nullsUsing = BooleanDefaultTrueSerializer.class)
     private Boolean active; // Added in v3.6. Existing records need to be nullable on upgrade.
 
+    @Persistent
+    @Index(name = "PROJECT_IS_LATEST_IDX")
+    @Column(name = "IS_LATEST", defaultValue = "false")
+    private Boolean isLatest; // Added in v4.12. Needs to be nullable therefore
```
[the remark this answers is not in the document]
rkg-mm: thanks for the tip, yea I was wondering but decided to follow the other style as I assumed there is a reason. Simplified it now in all occurances.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -518,21 +596,45 @@ public Project clone(
...
+            LOGGER.warn("Project was supposed to be cloned to version %s, but that version already exists".formatted(newVersion));
+            return null;
+        }
 
+        final Project oldLatestProject;
+        if(makeCloneLatest) {
+            oldLatestProject = source.isLatest() ? source : getLatestProjectVersion(source.getName());
+        } else {
+            oldLatestProject = null;
+        }
```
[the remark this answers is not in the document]
rkg-mm: done

src/main/java/org/dependencytrack/model/Policy.java
rkg-mm: done

src/main/java/org/dependencytrack/model/Project.java
rkg-mm: I removed it

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -518,8 +594,25 @@ public Project clone(
...
+            final String newVersion,
+            final boolean includeTags,
+            final boolean includeProperties,
+            final boolean includeComponents,
+            final boolean includeServices,
+            final boolean includeAuditHistory,
+            final boolean includeACL,
+            final boolean includePolicyViolations,
+            final boolean makeCloneLatest
+    ) {
```
nscuro, line 613: `makeCloneLatest = true` currently never reaches the cloning logic, because `CloneProjectTask` calls the old signature of this method: https://github.com/DependencyTrack/dependency-track/blob/7b183654fec6a5a6e07f5ba54b6f8f43496770cd/src/main/java/org/dependencytrack/tasks/CloneProjectTask.java#L49-L59

You should be able to remove the old method signature, since there are no callers outside `CloneProjectTask` anyway.

Would be good to have a regression test for this, you can find an example of project cloning tests here: https://github.com/DependencyTrack/dependency-track/blob/7b183654fec6a5a6e07f5ba54b6f8f43496770cd/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java#L1477-L1522

nscuro (approved):
Great work, thanks!

Issue #4175, opened 2024-09-26
Components get deleted in update


Gepardgame:
### Current Behavior

If you update an internal vulnerability with components, the the components will no longer be showen.

### Steps to Reproduce

1. Create an internal vulnerability with one or more affected components.
2. Open the vulnerability(View details) and click the "Update" Button
3. Reload site and open Vulnerability again and go to Affected Components tab
4. Wonder why there are no components

### Expected Behavior

Components should be shown, even so the vulnerability gets updated.

The problem is there in the code [VulnerabilityQueryManager.java#L798](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java#L798). I don't know there the attribution would be reported again.

### Dependency-Track Version

4.12.0-SNAPSHOT

### Dependency-Track Distribution

Executable WAR

### Database Server

H2

### Browser

N/A

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

Gepardgame:
@sahibamittal You implemented this function. The old source is INTERNAL and the new one is INTERNAL, which concludes to the deletion of the attributes, even so it is still relevant. I can think that has been implemented for to prevent duplications, but I don't know for sure.

nscuro:
I suspect that we don't create [`AffectedVersionAttribution`](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/model/AffectedVersionAttribution.java)s for manually created `VulnerableSoftware` records.

Then when it comes to `reconcileVulnerableSoftware`, manually created `VulnerableSoftware`s end up in this branch: https://github.com/DependencyTrack/dependency-track/blob/d03cb83725b3f44fa47fbc78c3f601b626178bbd/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java#L779-L785

Gepardgame:
[...]

PR #4193, opened 2024-09-30
Fix: Component get deleted in update for Internal Vuln

Gepardgame:
### Description

Now you can update a internal Vulnerability without getting the affected components deleted.

### Addressed Issue

Fixes [#4175](https://github.com/DependencyTrack/dependency-track/issues/4175)

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
@@ -784,9 +784,9 @@ final class VulnerabilityQueryManager extends QueryManager implements IQueryMana
                 continue;
             }
 
-            final boolean previouslyReportedBySource = attributions.stream()
+            final boolean previouslyReportedBySource = source != Vulnerability.Source.INTERNAL && attributions.stream()
                     .anyMatch(attr -> attr.getSource() == source);
-            final boolean previouslyReportedByOthers = attributions.stream()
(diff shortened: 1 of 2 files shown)


Gepardgame:
Again only trivy tests failed.

Is there a way of caching the docker container on the machine which runs the workflow? This would help with not always pulling the image for every test run

nscuro:
Thanks @Gepardgame! Could you please add a test to `VulnerabilityResourceTest` that verifies the desired behavior?

Gepardgame:
I am currently working on it, but have a issue, cause I cannot run the test, cause the Test class starts like this:
`@RunWith(Suite.class)
@SuiteClasses(VulnerabilityQueryManagerTest.SynchronizeVulnerabilityAliasTest.class)
public class VulnerabilityQueryManagerTest extends PersistenceCapableTest {
`
I want to add this test normally to the class, but that doesn't work quit out. @nscuro  Do you know how to fix that?

Gepardgame:
I found a bug in my implementation, but I don't have time anymore today to fix it. I'm sorry, but will do it tomorrow.

nscuro:
No problem @Gepardgame, we can release tomorrow, too.

Worst case I can pick it up, in that case please let me know what bug you found.

Gepardgame:
The bug is that the affectedComponents get duplicated, but without the AffectedVersionAttribution. The original is the same with the AffectedVersionAttribution. In the gui you don't see it, cause it only shows, if there is a AffectedVersionAttribution, but in the spi directly you can see it.

nscuro:
Superseded by #4208

Issue #4200, opened 2024-10-01
Unauthorized access to projects over /vulnerability/{source}/vuln/{vuln}(/projects) when ACL is enabled


Gepardgame:
### Current Behavior

When I make a request to the either /vulnerability/{source}/vuln/{vuln} and /vulnerability/{source}/vuln/{vuln}/projects I get access to projects I have no access to when ACL is enabled. This occurs only in the API Response, not in the GUI, for some reason.

### Steps to Reproduce

1. Have to projects with the same vulnerability
2. Be part of a team which has only access to one project
3. Make a GET request to either /vulnerability/{source}/vuln/{vuln} or /vulnerability/{source}/vuln/{vuln}/projects, you will get information's about the other project too.
4. That you are part of a project is not necessary, you only need to know the vuln and have VIEW_PORTFOLIO Permission

### Expected Behavior

You should only see the projects/components you are allowed to see.

### Dependency-Track Version

4.12.0-SNAPSHOT

### Dependency-Track Distribution

Executable WAR

### Database Server

H2

### Browser

N/A

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

PR #4201, opened 2024-10-01
Fix: Unauthorized access to projects over /vulnerability/{source}/vuln/{vuln}(/projects) when ACL is enabled

Gepardgame:
### Description

Projects are removed from the vulnerability, from which the principal has no access to so.

### Addressed Issue

Fixes [#4200](https://github.com/DependencyTrack/dependency-track/issues/4200)

### Checklist


Gepardgame:
GET /v1/vulnerability is also vulnerable to this, will add fix for that to

nscuro:
As a side-note, the endpoint will no longer return component data at all, as of v5: https://github.com/DependencyTrack/hyades/issues/1539

Also worth pointing out that doing `hasAccess` calls for every single component will have absolutely horrendous performance characteristics. But I don't see a better way given how ACLs were originally built, and without introducing breaking API changes.

nscuro (approved):
Thanks!

PR #4202, opened 2024-10-01
Fix redundant `ConfigProperty` queries in `BadgeResource`

nscuro:
### Description

Fixes redundant `ConfigProperty` queries in `BadgeResource`.

If unauthenticated badge access is enabled, the same `ConfigProperty` was previously queried three times. Since badges are embedded in public places, it'd be beneficial if the respective resources avoided unnecessary work.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


SaberStrat:
Oops, sorry. Didn’t think about that when I put those multiplemethod calls in the `if`s. 😬

nscuro:
No worries at all, I also missed it during the review.

PR #4276, opened 2024-10-18
Update Deploying Docker guide to Compose v2

### LINKED ISSUE #3939: Update the documentation around Docker Compose
### Current Behavior

Please see https://github.com/DependencyTrack/dependency-track/issues/3938

Followed the instructions for starting a DT instance and it failed. This was due to docker-compose (with the '-') no longer being supported with the version of Docker I was using.

### Proposed Behavior

Update the documentation to include this information and take the guidance from the above issue for installing Docker Compose as a plugin. 

### Checklist

danihengeveld:
### Description

- Updated the Deploying Docker Container guide to use Compose v2, so `docker compose` instead of `docker-compose`. Also added guidance for installing compose.
- Added the `ALPINE_OIDC_CLIENT_ID` environment variable to the Docker Compose example.
- Removed the top-level version element from the example and the `docker-compose.yml` file because it is [obsolete](https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete).

### Addressed Issue

Closes #3939.

### Checklist

### DIFF
--- a/docs/_docs/getting-started/deploy-docker.md
+++ b/docs/_docs/getting-started/deploy-docker.md
@@ -18,7 +18,7 @@ other than a modern version of Docker.
 | 4.5GB RAM   | 16GB RAM    |
 | 2 CPU cores | 4 CPU cores |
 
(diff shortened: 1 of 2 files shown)


nscuro (approved):
Thanks!

PR #4287, opened 2024-10-21
Added Dutch tax office as adopter

SudoHenk:
### Description

Added the Dutch Tax office as an Adopter of Dependency-Track.


nscuro (approved):
Thanks so much @SudoHenk! We really appreciate the kind words 🙌

Issue #4247, opened 2024-10-11
The jekyll image specified in the docs is incompatible with


stefandecimelli:
### Current Behavior

The command to run the doc server via docker in DEVELOPING.md uses jekyll:3, which is incompatible with the current dependency: public_suffix. 

### Steps to Reproduce

1.```
$ docker run --rm -it --name jekyll -p "127.0.0.1:4000:4000" -v "$(pwd)/docs:/srv/jekyll:Z" jekyll/jekyll:3 jekyll serve
Unable to find image 'jekyll/jekyll:3' locally
3: Pulling from jekyll/jekyll
Digest: sha256:9521c8aae4739fcbc7137ead19f91841b833d671542f13e91ca40280e88d6e34
Status: Downloaded newer image for jekyll/jekyll:3
Warning: the running version of Bundler (2.0.2) is older than the version that created the lockfile (2.3.25). We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
Your Ruby version is 2.6.3, but your Gemfile specified ~> 3.1.1
```

### Expected Behavior

```
Fetching gem metadata from https://rubygems.org/...........
Using bundler 2.3.25
Fetching public_suffix 6.0.1
Using colorator 1.1.0
Fetching concurrent-ruby 1.3.4
Installing public_suffix 6.0.1
Installing concurrent-ruby 1.3.4
...
```

### Dependency-Track Version

4.7.x

### Dependency-Track Distribution

Container Image

### Database Server

H2

### Database Server Version

_No response_

### Browser

N/A

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

PR #4296, opened 2024-10-23
update to ruby 3

stefandecimelli:
### Description

Updates the ruby version for the doc server in order to fix dependency issue.
(pull request recreated for proper commit signing)

### Addressed Issue

fixes: https://github.com/DependencyTrack/dependency-track/issues/4247


nscuro:
Do we know if this will work with [GitHub Pages](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/about-github-pages-and-jekyll)? They still list Jekyll v3.x here: https://pages.github.com/versions/

It seems like in order to use newer versions of Jekyll, you need to build and publish the site via GitHub Actions: https://www.moncefbelyamani.com/making-github-pages-work-with-latest-jekyll/

stefandecimelli:
> Do we know if this will work with [GitHub Pages](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/about-github-pages-and-jekyll)? They still list Jekyll v3.x here: https://pages.github.com/versions/
> 
> It seems like in order to use newer versions of Jekyll, you need to build and publish the site via GitHub Actions: https://www.moncefbelyamani.com/making-github-pages-work-with-latest-jekyll/

It seems we are at a bit of an impasse here. 
The highest 3.x tag for jekyll is 3.8 / 3.8.6, which still uses Ruby 2. (https://hub.docker.com/r/jekyll/jekyll/tags).

The next subsequent version tag is 4.x, which Github has not added support for according to [their version doc](https://pages.github.com/versions/).

PR #4305, opened 2024-10-24
Use Caffeine as DataNucleus L2 cache

nscuro:
### Description

Uses Caffeine as DataNucleus L2 cache per default.

Users continue to run into situations where the DataNucleus L2 cache grows to excessive sizes, eventually causing the application to become unresponsive, or crash due to OOM.

The default, `Map`-based L2 cache of DataNucleus does not support time- or size-based expiration of entries. Users are thus unable to limit the impact of L2 caching in their environment.

Disabling the L2 cache entirely is possible, but because the default L2 caches don't expose hit/miss metrics, we've been blind to the impact of disabling it so far.

This change switches the default L2 cache implementation to one backed by Caffeine (https://github.com/ben-manes/caffeine). This cache supports time and size expiry policies, and is capable of exposing more detailed metrics.

The example Grafana dashboard was updated to include a widget for the newly available metrics. This should make it easier for users to determine whether they can safely disable L2 caching altogether (i.e. high miss-rates, low hit-rates), and monitor the effectiveness or their configured expiry policies.

It should also help us to determine if we can safely turn off L2 caching per default, to avoid such problematic situations for users entirely.

### Addressed Issue

N/A

### Additional Details

Screenshot 2024-10-24 211038

### Checklist

### DIFF
--- a/dev/monitoring/grafana/dashboards/apiserver.json
+++ b/dev/monitoring/grafana/dashboards/apiserver.json
@@ -2403,6 +2403,197 @@
       "type": "timeseries"
     },
     {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "Prometheus"
+      },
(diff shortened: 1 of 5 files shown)


nscuro:
Superseded by #4310

Issue #3922, opened 2024-07-05
Database could avoid re-writing/updating vulnerability if value is not different


LaVibeX:
### Current Behavior

Currently, the database re-writes all vulnerabilities that are given, without checking if the value has changed or not. This results in an expensive and unnecessary operation.

### Proposed Behavior

To address this issue, we could consider adding a new column or reusing an existing column in the vulnerability table to store the last_modified value. This would allow us to compare the last_modified value of the existing vulnerability with the new vulnerability data.

Another option could be to process a checksum for the stored vulnerability data and compare it with the given value. This would allow us to determine if the vulnerability data has changed, without having to write non-updated values.

https://github.com/DependencyTrack/dependency-track/blob/26185e40335f4abe97a99280e6b02958f896611e/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java#L145-L151

https://github.com/DependencyTrack/dependency-track/blob/26185e40335f4abe97a99280e6b02958f896611e/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java#L90-L135

I have noticed that the CREATED and PUBLISHED values are often set to the same value. We could probably use published to save the "last modified" value.
![image](https://github.com/DependencyTrack/dependency-track/assets/52439101/b96d6adb-a9b2-41da-a4a4-516d9819d77b)

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

valentijnscholten:
Can you be more specific about which part of DT you are referring to? i.e. what are _all vulnerabilities that are given_ ?

LaVibeX:
> Can you be more specific about which part of DT you are referring to? i.e. what are _all vulnerabilities that are given_ ?

Depending on the Task Scheduler, specifically in the Vulnerability database synchronization:

![image](https://github.com/DependencyTrack/dependency-track/assets/52439101/183d27e5-a334-4983-8442-b9481d1d8dd8)

These values might be modified to be updated every hour (worst case scenario), leading to an expensive and exhausting process of rewriting all the "updated vulnerabilities" in the database.

nscuro:
The default flush behavior of the ORM plays into this, see https://github.com/DependencyTrack/dependency-track/pull/4039

[...]

PR #4359, opened 2024-11-07
Optimized synchronizeVulnerability

LaVibeX:
### Description
This modification is intended to avoid multiple updates/writes in the database when there are no changes. In general we want to reduce unnecessary load on the DB.

Part of the logic in updateVulnerability was optimized. The differ.applyIfChanged method was added, similar to what NIST is doing https://github.com/DependencyTrack/dependency-track/blob/8087a9491d0fb8d05fc3ce4859a4a51d1a32cec4/src/main/java/org/dependencytrack/tasks/AbstractNistMirrorTask.java#L222-L257, and callInTransaction was added as seen https://github.com/DependencyTrack/dependency-track/blob/7ce8e0471a32c37eef81994894d7c1b35759b245/src/main/java/org/dependencytrack/persistence/NotificationQueryManager.java#L162-L171.

synchronizeVulnerability now returns null if there is nothing to synchronize.

Some modifications were needed to align with the new logic. For example:

`if (synchronizeVulnerability == null) continue;`
		
Tests are provided.

### Addressed Issue
Reference the issue addressed by this PR #3922 

### Checklist


src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -81,57 +84,80 @@ public Vulnerability createVulnerability(Vulnerability vulnerability, boolean co
...
+                differ.applyIfChanged("owaspRRTechnicalImpactScore", Vulnerability::getOwaspRRTechnicalImpactScore, existingVulnerability::setOwaspRRTechnicalImpactScore);
+                differ.applyIfChanged("owaspRRVector", Vulnerability::getOwaspRRVector, existingVulnerability::setOwaspRRVector);
+                differ.applyIfChanged("cwes", Vulnerability::getCwes, existingVulnerability::setCwes);
+                if (transientVulnerability.getVulnerableSoftware()!= null) {
+                    differ.applyIfChanged("vulnerableSoftware", Vulnerability::getVulnerableSoftware, existingVulnerability::setVulnerableSoftware);
             }
-            final Vulnerability result = persist(vulnerability);
+                // Persist changes only if there are differences
+                if (!differ.getDiffs().isEmpty()) {
+                    final Vulnerability result = persist(existingVulnerability);
```
nscuro, line 147: This is actually not necessary when in a transaction. The ORM already tracks changes made to persistent objects, and they will take effect when the transaction is committed.

Using `applyIfChanged` etc. is mostly useful to avoid excessive DB roundtrips, since calling getters on persistent objects triggers an `UPDATE` query behind the scenes.

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -81,57 +84,80 @@ public Vulnerability createVulnerability(Vulnerability vulnerability, boolean co
...
             return result;
         }
-        return null;
+                // If no changes were made, return the existing vulnerability without modification
+                return existingVulnerability;
+                }
+            else {
+                // Handle cases where no existing vulnerability is found if needed (e.g., log an error)
+                LOGGER.warn("No existing vulnerability found for update operation.");
+                return new Vulnerability(); // return a new object or handle according to requirements
```
nscuro, line 158: Why not return `null`, or throw a `NoSuchElement` exception instead? 

Returning an empty object feels a bit ambiguous. Either this is an expected case and we return a corresponding result, or it is not and we `throw`.

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -142,12 +168,20 @@ public Vulnerability updateVulnerability(Vulnerability transientVulnerability, b
...
+        if (existingVulnerability != null) {
+            if(hasChanges(existingVulnerability, vulnerability)) {
+                // Update only if changes are detected
+                return updateVulnerability(vulnerability, commitIndex);
+            } else {
+                return null;
+            }
+        } else {
+            // Create new vulnerability if it doesnt exist
+            return  createVulnerability(vulnerability, commitIndex);
```
nscuro, line 183: Consider wrapping this in a `callInTransaction` while you're at it. Select and update should share the same transaction.

nscuro (changes requested):
Thanks @LaVibeX! Looks good, just some minor suggestions.

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
LaVibeX: @nscuro thanks for the suggestion, I fix the code.

nscuro (approved):
Thanks!

PR #4374, opened 2024-11-13
Fix policy evaluation not happening upon creation or update of individual components

### LINKED ISSUE #4373: Creating or updating a component should trigger a policy evaluation but does not
### Current Behavior

The implementation of `ComponentResource` shows that a policy evaluation should be triggered when creating or updating a component.
In `createComponent` and `updateComponent` it's implemented like this:

https://github.com/DependencyTrack/dependency-track/blob/3c265ed6e8880fde6c51302f8c54a04273276269/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java#L368-L374

This does actually not happen. The project field is not set in the `PolicyEvaluationEvent`, therefore the `PolicyEvaluationTask` does nothing.

https://github.com/DependencyTrack/dependency-track/blob/3c265ed6e8880fde6c51302f8c54a04273276269/src/main/java/org/dependencytrack/tasks/PolicyEvaluationTask.java#L50-L52

### Steps to Reproduce

1. Create or update a component
2. There are no logs by the `PolicyEngine` since it's not triggered

### Expected Behavior

Creating or updating a component should trigger the policy evaluation for this single component.
The simplest fix is to set the missing project:

`.onSuccess(new PolicyEvaluationEvent(component).project(component.getProject()))`

I provide an PR for this.

### Dependency-Track Version

4.12.1

Container Image

### Database Server

PostgreSQL

### Database Server Version

_No response_

### Browser

Google Chrome

### Checklist

fupgang:
### Description

Fixes #4373 as suggested in the issue.

### Addressed Issue

#4373 

### Additional Details

The test `PolicyEvaluationTaskTest` does not exactly check the fix.
But I found it valuable since there seems to be no test case yet for `PolicyEvaluationTask`. I used it to study the behavior, it asserts another aspect.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
@@ -370,7 +370,7 @@ public class ComponentResource extends AlpineResource {
                 // Wait for RepositoryMetaEvent after VulnerabilityAnalysisEvent,
                 // as both might be needed in policy evaluation
                 .onSuccess(new RepositoryMetaEvent(List.of(component)))
-                .onSuccess(new PolicyEvaluationEvent(component))
+                .onSuccess(new PolicyEvaluationEvent(component).project(component.getProject()))
             );
             return Response.status(Response.Status.CREATED).entity(component).build();
         }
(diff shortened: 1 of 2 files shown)


nscuro (approved):
Good catch, and thanks for the fix!

Issue #3361, opened 2024-01-08
Make project life-cycle management more effective


carniz:
### Current Behavior

When DT is integrated with a larger CI system, the DB tends to fill up quickly which leads to performance and maintainability issues. The API currently only allows to delete 1 single project at a time, and invoking the DELETE `/v1/project/{uuid}` method once per project UUID is time-consuming and many times leads to stability issues, rendering the application (backend + frontend) unresponsive.

Slack discussion: https://owasp.slack.com/archives/C6R3R32H4/p1701445069732229

### Proposed Behavior

My proposal is to add a `/v1/project/batchDelete` method that allows for deleting a list of projects specified by their UUIDs, which would allow you to make less requests and thus reduce the load on the application.

**NB**: I have a patch ready that implements this behavior, and this will be the basis of an upcoming PR.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #4383, opened 2024-11-14
Add /v1/project/batchDelete API method that deletes with SQL

mikael-carneholm-2-wcar:
### Description

This patch adds a /v1/project/batchDelete method that makes project life-cycle management more effective since multiple projects can be deleted per request.

### Addressed Issue

This fixes https://github.com/DependencyTrack/dependency-track/issues/3361

### Additional Details
This is a modified version of https://github.com/DependencyTrack/dependency-track/pull/3407 that instead uses pure SQL to do the deletion. And by doing so, it's much faster:
- The previous implementation was able to delete 61 projects/second (best case, with a batch size of 1000)
- This implementation has been verified to delete 430 projects/second (with a batch size of 1000)

A caveat is that it uses SQL Common Table Expressions (CTEs) which are part of the ANSI standard since SQL:1999, but still not supported by all DB vendors. The H2 database used for the unit tests has experimental support for CTEs, but does not support CTE DELETE statements. For this reason the code checks the name of the JDBC driver to conditionally run certain statements if the driver is org.postgresql.Driver.

Benchmarks using a local sample DB populated with ~11000 projects imported from a production system:
```
batch_size      time
100             0m34.193s
200             0m27.759s
500             0m31.075s
1000            0m25.451s
```
NB: The effect of the batch size is limited by the checkpoint configuration (see https://www.enterprisedb.com/blog/basics-tuning-checkpoints), but in this test the container DB was running with a default configuration.

### Checklist


src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -804,6 +806,56 @@ public Response deleteProject(
...
+    @Path("/batchDelete")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(
+            summary = "Deletes a list of projects specified by their UUIDs",
+            description = "<p>Requires permission <strong>PORTFOLIO_MANAGEMENT</strong></p>"
+    )
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "204", description = "Projects removed successfully"),
+            @ApiResponse(responseCode = "207", description = "Access is forbidden to the projects listed in the response"),
```
nscuro, line 819: Not sold on this status code. Some surface-level research leads me to believe using this outside of WebDAV is not intended: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/207

Wouldn't it be better to atomically fail the request if deletion of one or more projects failed?

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -804,6 +806,56 @@ public Response deleteProject(
...
+            @ApiResponse(responseCode = "401", description = "Unauthorized")
+    })
+    @PermissionRequired(Permissions.Constants.PORTFOLIO_MANAGEMENT)
+    public Response deleteProjects(List<UUID> uuids) {
```
nscuro, line 823: Consider using `@Size(min = 1, max = ...)` here. The limits should then also appear in the OpenAPI spec automatically.

I think there should be *some* upper limit to the number of UUIDs being submitted in one go. We can always extend it later if people need it, but we can't reduce it later if larger batches end up causing problems.

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -804,6 +806,56 @@ public Response deleteProject(
...
+        List<UUID> inaccessibleProjects = new ArrayList<>();
+        try (QueryManager qm = new QueryManager()) {
+            for (Iterator<UUID> it = uuids.iterator(); it.hasNext();) {
+                UUID uuid = it.next();
+                final Project project = qm.getObjectByUuid(Project.class, uuid, Project.FetchGroup.ALL.name());
```
nscuro, line 828: What's the reason for using `FetchGroup.ALL` here?

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -864,6 +868,264 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
...
+                executeAndCloseWithArray(sqlQuery, (Object) uuidsArray);
+
+                sqlQuery = pm.newQuery(JDOQuery.SQL_QUERY_LANGUAGE, """
+                    DELETE FROM "VEX" WHERE "PROJECT_ID" IN (
+                        SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
+                    );
+                    """);
+                executeAndCloseWithArray(sqlQuery, (Object) uuidsArray);
+
+                // The below has only been tested with Postgres, but should work on any RDBMS supporting SQL:1999
```
nscuro, line 1074: For completeness, how many RDBMSes have you tested with this? Having tested at least H2, MSSQL, and PostgreSQL would be good. We have a Docker Compose file for MSSQL here: https://github.com/DependencyTrack/dependency-track/blob/master/dev/docker-compose.mssql.yml

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
mikael-carneholm-2-wcar: Fair point, the 207 indeed seems to be reserved for WebDAV. The reason for implementing it this way was based on the outcome of [this](https://github.com/DependencyTrack/dependency-track/pull/3407#discussion_r1523485369) discussion, where the suggestion by [sebD](https://github.com/sebD) was to return a 204 with a body containing the list of inaccessible projects. The 207 was my suggestion, and since I got a thumbs up I thought I would do it the same way in this PR. I see a value in letting the client know which projects that are inaccessible, so I'm leaning towards going with sebD's suggestion instead of just returning a 401. Would that be an acceptable solution?

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
mikael-carneholm-2-wcar: That's a good idea. I have tested it with up to 1000 UUIDs, so maybe that could be a reasonable max size to start with?

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
mikael-carneholm-2-wcar: Indeed, that doesn't seem to be needed since Project.accessTeams is[ part of the default fetch group](https://github.com/DependencyTrack/dependency-track/blob/4cb2f3a061335561d606dc391e2bb7c67aa95b24/src/main/java/org/dependencytrack/model/Project.java#L280). [ProjectQueryManager.hasAccess](https://github.com/DependencyTrack/dependency-track/blob/4cb2f3a061335561d606dc391e2bb7c67aa95b24/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L960) only needs that property from the project, so I'll change to the getObjectByUuid(Class, UUID) signature instead.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
mikael-carneholm-2-wcar: I've only tested it with H2 and PostgreSQL so far (H2 in unit tests, PostgreSQL in integration tests) but I can try migrating my test data to MSSQL and test that as well. I'll get back to you with the results when done!

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
nscuro: Yeah that sounds reasonable.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
nscuro: No need to do a full load test or anything. We just need assurance that the queries work as expected. Sadly the small differences in SQL syntax between the different RDBMSes turn out to be tripwires more often than not. :D

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
mikael-carneholm-2-wcar: Ok, I'll look into using ProblemDetails since that will make it consistent with how failed tag deletion is handled. Thanks for the tip!

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
mikael-carneholm-2-wcar: True, and true :) But I'll need some data in the schema to pass the condition that the UUID list is not empty, so I'll try to migrate the relatively small exported sample that I created with [pg_sample](https://github.com/mla/pg_sample).

mikael-carneholm-2-wcar:
@nscuro Am I expected to mark the conversations as resolved, or is there anything else blocking this PR from getting merged?

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -970,6 +979,407 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
         }
     }
 
+    /**
+     * Deletes a list of Projects (identified by their UUIDs) and all objects dependent on them.
+     * @param uuids the UUIDs of the Projects to delete
+     *
+     * NB: if ON DELETE rules had been set up, this would be as simple as "delete from project where ..."
```
nscuro, line 986: I asked myself the same. Virtually no record that is associated with a project is of any use when the project is gone (exception perhaps being the project parent-child relationship).

We'll change that in v5 for sure. Changing this in v4 will be a bit challenging since we'd need to do re-create FKs manually for all supported databases. But it's not impossible. 

The DataNucleus ORM has a way to define cascading deletes on FKs, so it would automatically apply for new deployments: https://www.datanucleus.org/products/accessplatform_6_0/jdo/mapping.html#fk 

If you'd like to look into this, I think it would make your current implementation even more efficient.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -970,6 +979,407 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
...
+                );
+                executeAndCloseWithArray(sqlQuery, queryParameter);
+
+                sqlQuery = pm.newQuery(JDOQuery.SQL_QUERY_LANGUAGE, """                        
+                    DELETE FROM "DEPENDENCYMETRICS" WHERE "COMPONENT_ID" IN (
+                        SELECT "ID" FROM "COMPONENT" WHERE "PROJECT_ID" IN (
+                            SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
+                        )
+                    ) OR "PROJECT_ID" IN (
+                        SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
```
nscuro, line 1033: The condition on `PROJECT_ID` should be sufficient, as that is always populated.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -970,6 +979,407 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
...
+                executeAndCloseWithArray(sqlQuery, queryParameter, queryParameter);
+
+                sqlQuery = pm.newQuery(JDOQuery.SQL_QUERY_LANGUAGE, """                        
+                    DELETE FROM "FINDINGATTRIBUTION" WHERE "COMPONENT_ID" IN (
+                        SELECT "ID" FROM "COMPONENT" WHERE "PROJECT_ID" IN (
+                            SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
+                        )
+                    ) AND "PROJECT_ID" IN (
+                        SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
+                    );
```
nscuro, line 1046: The condition on `PROJECT_ID` should be sufficient, as that is always populated.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -970,6 +979,407 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
...
+        var queryParameter = Config.getInstance().getProperty(Config.AlpineKey.DATABASE_DRIVER).equals("com.microsoft.sqlserver.jdbc.SQLServerDriver") ?
+                commaSeparatedUUIDs : uuidsArray;
+        String inExpression = Config.getInstance().getProperty(Config.AlpineKey.DATABASE_DRIVER).equals("com.microsoft.sqlserver.jdbc.SQLServerDriver") ?
+            "IN(SELECT value FROM STRING_SPLIT(?, ','))" : "= ANY(?)";
+
+        runInTransaction(() -> {
+            try (var ignored = new ScopedCustomization(pm).withProperty(PROPERTY_QUERY_SQL_ALLOWALL, "true")) {
+                Query<?> sqlQuery = pm.newQuery(JDOQuery.SQL_QUERY_LANGUAGE, """
+                    DELETE FROM "PROJECTMETRICS" WHERE "PROJECT_ID" IN (
+                        SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
```
nscuro, line 1021: Doing one initial `select id from project where uuid = any(:uuids)` query to retrieve the projects' IDs would allow you to omit a lot of subqueries like this.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -970,6 +979,407 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
...
+                        SELECT "ID" FROM "COMPONENT" WHERE "PROJECT_ID" IN (
+                            SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
+                        )
+                    );
+                    """.replace("= ANY(?)", inExpression)
+                );
+                executeAndCloseWithArray(sqlQuery, queryParameter);
+
+                // Deletion with CTEs does not work with H2, but verified on Postgres and MS SQL Server
+                if ( ! Config.getInstance().getProperty(Config.AlpineKey.DATABASE_DRIVER).equals("org.h2.Driver")) {
```
nscuro, line 1090: You can use `DbUtil.isH2()`, `DbUtil.isMssql()` etc. to make these conditions less verbose.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -970,6 +979,407 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
...
+                    DELETE FROM "PROJECTS_TAGS" WHERE "PROJECT_ID" IN (
+                        SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
+                    );
+                    """.replace("= ANY(?)", inExpression)
+                );
+                executeAndCloseWithArray(sqlQuery, queryParameter);
+
+                sqlQuery = pm.newQuery(JDOQuery.SQL_QUERY_LANGUAGE, """
+                    DELETE FROM "TAG"
+                    WHERE "ID" NOT IN (SELECT "TAG_ID" FROM "PROJECTS_TAGS");
```
nscuro, line 1170: Tags themselves might be used for other purposes and should not be deleted.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -970,6 +979,407 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
...
+                    DELETE FROM "NOTIFICATIONRULE_PROJECTS" WHERE "PROJECT_ID" IN (
+                        SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
+                    );
+                    """.replace("= ANY(?)", inExpression)
+                );
+                executeAndCloseWithArray(sqlQuery, queryParameter);
+
+                sqlQuery = pm.newQuery(JDOQuery.SQL_QUERY_LANGUAGE, """
+                    DELETE FROM "NOTIFICATIONRULE"
+                    WHERE "ID" NOT IN (SELECT "NOTIFICATIONRULE_ID" FROM "NOTIFICATIONRULE_PROJECTS");
```
nscuro, line 1184: Notification rules should not be deleted.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -970,6 +979,407 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
...
+                    DELETE FROM "POLICY_PROJECTS" WHERE "PROJECT_ID" IN (
+                        SELECT "ID" FROM "PROJECT" WHERE "UUID" = ANY(?)
+                    );
+                    """.replace("= ANY(?)", inExpression)
+                );
+                executeAndCloseWithArray(sqlQuery, queryParameter);
+
+                sqlQuery = pm.newQuery(JDOQuery.SQL_QUERY_LANGUAGE, """
+                    DELETE FROM "POLICY"
+                    WHERE "ID" NOT IN (SELECT "POLICY_ID" FROM "POLICY_PROJECTS");
```
nscuro, line 1246: Policies should not be deleted.

nscuro (changes requested):

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
mikael-carneholm-2-wcar: Now implemented basically as suggested, and it seems the performance increased further by doing so: those ~11000 projects now get deleted in ~17 seconds instead of 25 seconds. Probably mostly thanks to loading all projects with one DB round-trip instead of 1 hit per project.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
mikael-carneholm-2-wcar: Right, that was a bit over-thinking from my side. Now removed.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
mikael-carneholm-2-wcar: See above, same here.

nscuro (approved):

Issue #2134, opened 2022-11-07
Composer Meta Analyzer Exceptions


valentijnscholten:
### Current Behavior:
Log is full of Exceptions of the Composer Meta Analyzer:

```
2022-11-03 12:37:32,722 ERROR [LoggableUncaughtExceptionHandler] An unknown error occurred in an asynchronous event or notification thread
kong.unirest.json.JSONException: JSONObject["magento/adobe-ims"] not found.
        at kong.unirest.json.JSONObject.getProperty(JSONObject.java:963)
        at kong.unirest.json.JSONObject.getJSONObject(JSONObject.java:304)
        at org.dependencytrack.tasks.repositories.ComposerMetaAnalyzer.analyze(ComposerMetaAnalyzer.java:101)
        at org.dependencytrack.tasks.repositories.RepositoryMetaAnalyzerTask.analyze(RepositoryMetaAnalyzerTask.java:95)
        at org.dependencytrack.tasks.repositories.RepositoryMetaAnalyzerTask.inform(RepositoryMetaAnalyzerTask.java:60)
        at alpine.event.framework.BaseEventService.lambda$publish$0(BaseEventService.java:101)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
```

### Expected Behavior:
No exceptions or another log level for non critical errors

### Environment:
  
  - Dependency-Track Version: 4.6.1
  - Distribution: Docker
  - BOM Format & Version: n/a
  - Database Server: PostgreSQL
  - Browser: n/a

### Additional Details:
[Source code: https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java#L101](https://github.com/DependencyTrack/dependency-track/blob/9dac458d755bc15e3d5170c80e4afc27fb172bce/src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java#L97-L101)

valentijnscholten:
I see about 40 occurrences of this error on a daily basis. All related to magento modules. I did check a couple of the responses. I attached the responses (with a .txt extension as github won't allow me to upload .json files).
So far I see two cases:

[...]

PR #4435, opened 2024-12-03
Add Support for Composer 2 Metadata API

ch8matt:
### Checklist

- [x] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)
- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective
- [ ] This PR implements an enhancement, and I have provided tests to verify that it works as intended
- [ ] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)
- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly

Issue #2826, opened 2023-06-16
Add support for `vers` version ranges and ecosystem-specific version comparisons


nscuro:
### Current Behavior

The Package URL project has defined [`vers`](https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst), a specification for version ranges that is capable of dealing with ecosystem-specific version notations.

Using an [example from the spec document](https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst#examples), the following `vers` range:

```
vers:tomee/>=1.0.0-beta1|<=1.7.5|>=7.0.0-M1|<=7.0.7|>=7.1.0|<=7.1.2|>=8.0.0-M1|<=8.0.1
```

translates to:

> affects Apache TomEE 8.0.0-M1 - 8.0.1, Apache TomEE 7.1.0 - 7.1.2, Apache TomEE 7.0.0-M1 - 7.0.7, Apache TomEE 1.0.0-beta1 - 1.7.5.

`vers` has been [adopted by the CycloneDX project](https://cyclonedx.org/docs/1.4/json/#vulnerabilities_items_affects_items_versions_items_range), where it is used to communicate version ranges affected by a given vulnerability.

The [version comparison logic DT has today](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/util/ComponentVersion.java) is generic and is struggling to deal with versioning schemes that are specific to certain ecosystems, like Debian (https://github.com/DependencyTrack/dependency-track/issues/1374, https://github.com/DependencyTrack/dependency-track/issues/2062). Our vulnerability matching logic would benefit greatly from being able to understand ecosystem-specific notations.

Further, supporting `vers` would allow DT to ingest vulnerable version ranges from CycloneDX [Bill of Vulnerabilities](https://cyclonedx.org/capabilities/bov/) into its database. But it will be of use for ingesting data from GitHub Advisories, OSV, and other sources as well, by converting between "native" (e.g. NPM range notation) and `vers`.

For policies, the `Coordinates` and `Version` conditions will be a lot more useful, as currently only a single version constraint can be provided, whereas `vers` [supports multiple](https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst#version-range-specifier).

[...]

PR #4470, opened 2024-12-18
Compose Metadata Analyzer: Refactor to support V2 and V1 repositories

### LINKED ISSUE #2337: Switch to Packagist api v2 for Composer package metadata
### Current Behavior

Compose Packagist api v1 is [deprecated](https://packagist.org/apidoc#get-package-data).
API v2 also handles non-existing packages better, see #2134  And it helps packagist to reduce server load on their end.

### Proposed Behavior

Use API v2. It's a very small change as far as I can see in https://packagist.org/apidoc#get-package-data.

Now that we are caching metadata, we could use the [If-Modified-Since header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) to hel preduce load on their server.
In theory also on our end because if the metadata is not modified since we last cached it, we don't have to process the response. But this might be micro optimizing, probably requires a change to the parameters passed into the `analyze` method and we need to persist the new `lastOccurrence` timestamp to the database anyway.

### Checklist

valentijnscholten:
### Description
Use Composer Repository V2 url for metadata as v1 is deprecated. The https://repo.packagist.org and its private counterparts at packagist.com are becoming read only to clients that request V1 metdata. 

This PR adds support for V2 for all repositories. It also keeps and improves V1 support as lots of third party repositories are still in V1 mode. The V1 support in DT was very minimal and this PR ads more constructs such as packages included in the main `packages.json` file and the user of `available-packages` and `available-package-patterns` to only query for applicable packages.

I analyzed a number of real world projects and looked at all the composer repositories and how those were handled by Composer. The implementation is based on the docs found in https://github.com/composer/composer. The docs are a bit ambiguous / unclear sometimes, so I also looked at the source code. I confirmed the behaviour by looking at real worl requests/responses sent by the Composer CLI.

Test cases are added.

### Addressed Issue
Fixes #2337

### Additional Details
The metadata being returned for V2 has the same format as V1, except that the releases for a package are returned as an Array instead of a JSONObject (minified is what Composer calls this).

One thing that changed is how non-existing (or no longer existing) packages are being handled:
- V1: We got a json metadata response with some packages, but it didn't include the requested package  (#2134)
- V2: We get a 404

### Checklist


valentijnscholten:
The current code only works with repositories that use the default url pattern for package metadata:

```
    /**
     * @see <a href="https://packagist.org/apidoc#get-package-data">Packagist's API doc for "Getting package data - Using the Composer v2 metadata"</a>
     * Example: https://repo.packagist.org/p2/monolog/monolog.json
     */
    private static final String PACKAGE_META_DATA_URL = "/p2/%s/%s.json";
```

More extensive changes are needed to support repositories that use a different url pattern, or provide the package metadata inline. Those changes should go into a different PR so we can keep this PR here simple. Hopefully that helpts to get this PR merged/released before Composer V1 stops providing new metadata (February 1st 2025)

nscuro:
> Hopefully that helpts to get this PR merged/released before Composer V1 stops providing new metadata (February 1st 2025)

So this really could (should!) go out in the 4.12.3 bugfix release then. Even if we get 4.13 out before Feb, not everyone will upgrade non-bugfix releases immediately. So better play it safe and get it out sooner.

valentijnscholten:
Even better. Didn't know if you'd want it in a bugfix release, but it's not too big of a change.

valentijnscholten:
Looks like some third party repositories still expose their (meta)data in V1 format. Example https://packagers.shopware.com.
And there's no field indicating which version. So we'll need to support both versions in DT for now by just trying both.

valentijnscholten:
After a good nights sleep I realized we should just look at https://github.com/composer/composer/blob/main/src/Composer/Repository/ComposerRepository.php and mimic what is don there. The approach in #4435 may work for the short term, but it will result in lots of failing (404) requests for V2 meatada to repositories that are still V1.

valentijnscholten:
Turns out that do it right, we need to retrieve the `packages.json` for the repository and check for the `metadata-url`. If present, it's V2. Otherwise it's V1. The approach in #4435 may work for the short term, but I decided to implement the "correct" solution here by mimicking what `Composer` itself does. We need that anyway if we want to do Vulnerability Analysis as well from these repositories.

Some talking points:
- I added a simple in memory cache to cache the `packages.json` response. This may need something persistent/shared Hyades.
- The ComposerMetaAnalyzer doesn't have enough information to make decent cache key. Any thoughts on providing more fields to the Analyzers besides just the Url. We could provide the ID as well, which should be unique? Or can we just provide the full reposiroy entity? For now the code works with the repositories I have seen, but I had to add a `protected` method to clean the cache before the start of each unit test. 
- Some features of the V2 repository are not implemented yet, but there were not implemented for V1 either. These are `includes` , `packages`, `available-packages` and `available-packages-patterns`.

Valentijn

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
```diff
@@ -46,9 +53,32 @@ public class ComposerMetaAnalyzer extends AbstractMetaAnalyzer {
...
 
     /**
-     * @see <a href="https://packagist.org/apidoc#get-package-metadata-v1">Packagist's API doc for "Getting package data - Using the Composer v1 metadata (DEPRECATED)"</a>
+     * @see <a href="https://packagist.org/apidoc#get-package-data">Packagist's API doc for "Getting package data - Using the Composer v1 metadata (DEPRECATED)"</a>
+     * Example: https://repo.packagist.org/p/monolog/monolog.json
+     */
+    private static final String PACKAGE_META_DATA_PATH_PATTERN_V1 = "/p/%package%.json";
+
+    private static final Cache<String, JSONObject> REPO_ROOT_CACHE = Caffeine.newBuilder()
+                                                                    .expireAfterWrite(10, TimeUnit.MINUTES)
```
nscuro, line 62: Are we expecting this data to change very frequently? I'd think a cache duration of ~1h should still be ok, WDYT?

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
```diff
@@ -73,76 +103,286 @@ public RepositoryType supportedRepositoryType() {
...
+                // Some repositories like packages.drupal.org and composer.amasty.com/entprise do not include the name field for a version, so print purl
+                LOGGER.warn("Field 'time' not present in metadata for " + component.getPurl());
+            }
+        });
         return meta;
     }
 
+    /**
+     * Clears the cache of repository root metadata.
+     * Needed in unit tests as we currently only have url as a cache key and all unit tests run against localhost...
```
nscuro, line 372: With WireMock you can launch the mock server on a dynamic port: https://github.com/DependencyTrack/dependency-track/blob/9a470cda5ef9d5df69684a03f5667b8554e893d0/src/test/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTaskTest.java#L41-L42

nscuro:
> Any thoughts on providing more fields to the Analyzers besides just the Url. We could provide the ID as well, which should be unique? Or can we just provide the full reposiroy entity?

I don't see a problem in providing more info, like the repository type and identifier. Don't think we should pass the entire `Repository` object though.

> I had to add a protected method to clean the cache before the start of each unit test as these are all running against localhost. 

Added a comment about that, it should be possible to avoid this conflict in tests by instructing WireMock to use a dynamic port.

> Some of the mock files are huge. [...] But we could trim them a bit and still have decent test quality.

That would be great.

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
```diff
@@ -73,76 +103,286 @@ public RepositoryType supportedRepositoryType() {
...
+        if (repoRoot == null) {
+            // absence of packages.json shouldn't happen, but let's try to get metadata as we did in <=4.12.2
+            return analyzeFromMetadataUrl(meta, component, PACKAGE_META_DATA_PATH_PATTERN_V1);
+        }
+
+        // Available packages is finite, so we can use it to determine if the package exists in this repository
+        if (repoRoot.has("available-packages")) {
+            final JSONArray availablePackages = repoRoot.getJSONArray("available-packages");
+            if (!availablePackages.toList().contains(getComposerPackageName(component))) {
+                return meta;
```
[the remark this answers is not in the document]
valentijnscholten: I don't read it that way:
https://github.com/composer/composer/blob/fb397acaa0648ba2668893e4b786af6465a41696/doc/05-repositories.md?plain=1#L197

> If your repository only has a small number of packages, and you want to avoid
> the 404-requests, you can also specify an `"available-packages"` key in
> `packages.json` which should be **an array with all the package names that your **
> ** repository contains**. _Alternatively_ you can specify an
> `"available-package-patterns"` key which is an array of package name patterns
> (with `*` matching any string, e.g. `vendor/*` would make Composer look up
> every matching package name in this repository).

But the implementation seems to confirm your interpretation: 
https://github.com/composer/composer/blob/main/src/Composer/Repository/ComposerRepository.php#L1772

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
valentijnscholten: That was my feeling as well. For now I just used 10 minutes which is what Composer CLI does.

valentijnscholten:
> I don't see a problem in providing more info, like the repository type and identifier. Don't think we should pass the entire Repository object though.

I suggest we replace the existing `setRepositoryUrl` and `setUsernameAndPasssword` with a new `setup(...)` method where we provide all parameters and setup everything in the right order. Currently order is not (yet) important, but maybe at some point we want to get the repository packages.json file in the setup call or something else that depends on the ordering.

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
valentijnscholten: This would still require multiple `WireMockRule` instances to have different ports for each Test. So might be easier to use multiple `ClientAndServer` instances. But if we add more info like repository ID to the analyzer we can just extend the cache key with that.

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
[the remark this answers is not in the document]
valentijnscholten: Implemented

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
```diff
@@ -73,76 +103,286 @@ public RepositoryType supportedRepositoryType() {
...
+
+        // V1 and no included packages, so we have to retrieve the package specific metadata
+        return analyzeFromMetadataUrl(meta, component, PACKAGE_META_DATA_PATH_PATTERN_V1);
+    }
+
+    private JSONObject getRepoRoot() {
+        // Code mimicksed from https://github.com/composer/composer/blob/main/src/Composer/Repository/ComposerRepository.php
+        // Retrieve packages.json file, which must be present even for V1 repositories
+        String packageJsonUrl = UriBuilder.fromUri(baseUrl).path("packages.json").build().toString();
+        JSONObject repoRoot = REPO_ROOT_CACHE.getIfPresent(packageJsonUrl);
```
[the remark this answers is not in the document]
valentijnscholten: Implemented

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
valentijnscholten: Changed it to 1 hour.

src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
valentijnscholten: Added repositoryID to the cache key.

valentijnscholten:
>>Some of the mock files are huge. [...] But we could trim them a bit and still have decent test quality.

>That would be great.

Trimmed them down.

src/main/java/org/dependencytrack/tasks/repositories/RepositoryMetaAnalyzerTask.java
```diff
@@ -211,6 +213,7 @@ private void analyze(final QueryManager qm, final Component component, final IMe
 
                 if (StringUtils.trimToNull(model.getLatestVersion()) != null) {
                     // Resolution from repository was successful. Update meta model
+                    //FIXME What happens if multiple repositories return a metamodel result with different lastPublishedTimestamps?
```
nscuro, line 216: The repository loop is broken out of upon encounter of the first matching repository, via the `break` in line `243`.

src/main/java/org/dependencytrack/tasks/repositories/AbstractMetaAnalyzer.java
```diff
@@ -33,25 +35,32 @@
...
-
     protected String username;
-
     protected String password;
 
+    /**
+     * Optionally set repositoryId that can be used in a cache key for example.
+     * @param repositoryId
+     */
+    public void setRepositoryID(String repositoryId) {
```
nscuro, line 60: Consistent casing:

```suggestion
    public void setRepositoryId(String repositoryId) {
```

nscuro, line 60: This should be called from `RepositoryMetaAnalyzerTask#analyze` when the analyzer is constructed. At the moment the setter is only called in tests.

nscuro (changes requested):

src/main/java/org/dependencytrack/tasks/repositories/AbstractMetaAnalyzer.java
valentijnscholten: Good catch, added now.

valentijnscholten: added

src/main/java/org/dependencytrack/tasks/repositories/RepositoryMetaAnalyzerTask.java
valentijnscholten: Yeah, that's what made me wonder. I noticed some components are published in multiple repositories. For example `phpunit`. I came across a third part repo that published an older version of `phpunit`. If DT ends up using that release as `latest`, it might draw the wrong conclusion. Might be safer to analyze all repositories and take the global latest release. But that could be wrong as well if some repo publish a release with some changes/fixes with a higher version number. Not sure how DT handles the suffixes like p1, p2, redhat1, etc.

nscuro: There is no bullet-proof way to determine what the "correct" repository or version is, at least none I am aware of. Hence the current strategy that will work *most of the time*, is to pick the first match. The strategy assumes that authoritative repositories (i.e. Maven Central for Maven) are checked first, which ATM is ensured via the [resolution order](https://github.com/DependencyTrack/dependency-track/blob/2316d55dfdf49879312eef9dbaba34f86609cef4/src/main/java/org/dependencytrack/model/Repository.java#L74-L77).

> Not sure how DT handles the suffixes like p1, p2, redhat1, etc.

It's not possible to handle such things generically. #2826 might help, but only as long as whoever adds these suffixes abides to the ecosystem's de-facto standard notation.

nscuro (approved):
Thanks!

PR #4483, opened 2024-12-20
Add Repository Bearer Authentication

valentijnscholten:
### Description
Add `Bearer` authentication support for repositories.

### Addressed Issue
Some repositories require authentication with a Bearer token, for example https://packages.shopware.com and other (private) packagist.com repositories.

### Checklist


valentijnscholten:
@nscure I didn't go the full monty to add another `Enum` to configure authentication type, but just implemented an if statement where a `Bearer` token gets priority over `Basic` auth. I could go even simpler and just instruct users to use the Password field for the bearer token and avoid having to add an extra column and remove the `bearerToken` field from the PR. Let me know what you think.

src/main/java/org/dependencytrack/upgrade/v4130/v4130Updater.java
nscuro, line 1: New columns are added automatically by DataNucleus, a manual migration is not needed.

nscuro (commented):
> I could go even simpler and just instruct users to use the Password field for the bearer token and avoid having to add an extra column and remove the bearerToken field from the PR. 

That would be best for now IMO. I don't think having a column for each credential type makes a lot of sense. If we get more variety of credentials (e.g. API keys, certificates, OIDC) we'll need to re-think this anyway.

src/main/java/org/dependencytrack/upgrade/v4130/v4130Updater.java
valentijnscholten: Ohhh you're making me feel all nice and warm about DataNucleus all of a sudden.

valentijnscholten:
@nscuro changed the PR to do just that: send a Bearer token if username is null but password is populated.
There was test case for that scenario that expected a Basic Auth style header, but I don't think that's a valid scenario, basic auth with null username?

valentijnscholten:
@AbdelHajou I can see a `NuGetMetaAnalyzerTest` that is asserting that a `Basic Auth` header is sent when the username is null but the password is present. Is this a valid use case for NuGet repositories?

valentijnscholten:
@nscuro What's the plan for this? I'm happy to resolve the conflicts, but would like to do it only once when a review/merge is near.

valentijnscholten:
Thanks @nscuro . I think you also need to merge https://github.com/DependencyTrack/frontend/pull/1128 to make it usable from the frontend.

valentijnscholten:
Thanks @nscuro . I think you also need to merge https://github.com/DependencyTrack/frontend/pull/1128 to make it usable from the frontend.

PR #4485, opened 2024-12-20
Backport: Compose Metadata Analyzer: Use v2 URL

### LINKED ISSUE #2337: Switch to Packagist api v2 for Composer package metadata

valentijnscholten:
### Description
Use Composer Repository V2 url for metadata as v1 is deprecated and becoming readonly soon. V1 endpoint will be removed from packagist later this year. This PR removes V1 metadata handling from DT.

### Addressed Issue
Fixes #2337
Backports #4470 

### Additional Details
The metadata being returned for V2 has the same format as V1, except that the releases for a package are returned as an Array instead of a JSONObject. 

One thing that changed is how non-existing (or no longer existing) packages are being handled:
- V1: We got a json metadata response with some packages, but it didn't include the requested package  (#2134)
- V2: We get a 404

I added a new unit test for the 404 scenario. I also removed the V1 unit test because I don't have a real world example of this happening in V2. Or should that unit test remain in place with some dummy package just to make sure we don't regress on this code path in the future?

### Checklist


valentijnscholten:
no longer needed now 4.13.0 is out.

Issue #4227, opened 2024-10-08
Risk of leaking API keys via new "authenticated Badge API" feature in 4.12


m-erhardt:
### Current Behavior

I've just upgraded my DependencyTrack instance to 4.12 and read about the the new "authenticated Badge API" feature (#3596, #4059) in the changelog.

Whilst I am a strong supporter of security I am deeply concerned about the way this feature is implemented (in combination with how Git hosting platforms have implemented badge support).

Most Git repo hosting platforms like GitLab, GitHub, Gitea, Forgejo, etc embed badge-URLs in the HTML-DOM of the page (either via an explicit Badge-feature like GitLab or embedding an image in a Markdown file within the repo) and clients load the badge-svg directly from DependencyTrack.

Given the way the feature is implemented I would have to append an `apiKey`-URL-parameter to the badge URL which would then leak to all users/clients, even with readonly support to the repo. Plus I would need to commit the API key to the repo for platforms which only support badges via embedded images in Markdown.

This might be unproblematic and only require a secret-scanning exception when using a properly scoped API token which only has the `VIEW_BADGES` permission but considering the nature of users it is only a matter of time until a user uses/leaks an API key with way more permissions this way - putting the security of my DependencyTrack instance in a worse place than before the "authenticated Badge API" feature.

Is there anyone else sharing my concerns?

### Proposed Behavior

This could potentially be resolved by 
* Allowing admins to enable/disable the "authenticated Badge API" feature on an instance/project basis (depending on the Security Model of the instance/project)
* Using a project-specific nonce to authenticate requests to the badge API endpoints (instead of regular API keys)

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

valentijnscholten:
The way I read it is that the apiKey only needs VIEW_BADGES permission. If such a key leaks, it's no worse than the current unauthenticated badge?

m-erhardt:
I agree, **IF** a user configures this correctly it does not matter if API keys with only an `VIEW_BADGE` permission are exposed.
That's why I opened this issue as an "enhancement", not an actual security issue.

[...]

PR #4502, opened 2024-12-29
Postpone deprecation of unauthenticated access to Badge API

### LINKED ISSUE #4500: Postpone deprecation of unauthenticated access to Badge API
### Current Behavior

The current implementation of the authenticated Badge API uses API keys as URL parameters for authn and authz. This was done as a natural way to limit access to Dependency-Track badges to specific projects/project-versions, while still allowing a broad and simple use of the badges, e.g. in Readme Markdowns in GitHub.

#4227 brought to attention some dangers that come with that implementation:

* human error: a developer could accidentally commit and push the wrong API key in the DT badge URL in their Readme, thus creating a security incident 
* sets a bad example: contradicts common compliance rules to never publish secrets like API keys, raising the complexity of a technological environment.

So alternative solutions to API keys in URLs should be explored. Until then, API keys should not be enforced for badges and thus the unauthenticated access to badges should remain an alternative.

### Proposed Behavior

To allow the use of badges without having to an publish an API key, keep the option to enable unauthenticated access to badges beyond Dependency-Track 4.12.

For this, we could either remove just the removal notice for 4.13, or remove the deprecation altogether. The latter feels more appropriate though: 
a functionality should become deprecated when there's a newer one that is planned to replace the current one. But since the new functionality (authenticated Badge API access) has been argued to be flawed, and should thus be reconsidered in favor of a better implementation, why continue calling it deprecated when the replacement likely isn't the final form and a better one isn't known yet?

So the proposal is:

* remove the text about the deprecation of unauthenticated access from the docs and frontend, and removal of unauthenticated access in Dependency-Track 4.13 from the docs
* bring the various security implications of either approaches to badges (unauthenticated and authenticated but with API keys) to the users' attention in the docs.

### Checklist

SaberStrat:
Remove deprecation notice for unauthenticated access to the Badge API, as the use of API keys for authenticated access comes with risks too that the maintainer of the DT instance has to weigh against the use of unauthenticated access which does not use API keys.

### Addressed Issue

Closes #4500 

### Checklist

### DIFF
--- a/docs/_docs/integrations/badges.md
+++ b/docs/_docs/integrations/badges.md
@@ -8,15 +8,10 @@ order: 10
 Dependency-Track supports badges in Scalable Vector Graphics (SVG) format. Support for badges is configurable on a team
 basis via permission or globally for unauthenticated access.
 
-> **Deprecation Notice**
-> 
-> Unauthenticated access to badges as a global configuration is deprecated and slated for removal in Dependency-Track
-> v4.13.
-
 To enable badges for a team, activate the permission `VIEW_BADGES`. To deactivate badges, remove the permission. To 
 retrieve a badge, use a team's API key either in the badge API header `X-API-Key` or in the URI parameter `apiKey`.
 
-As a legacy feature, badges can also be accessed without authentication. On new Dependency-Track installations, this is
+As an alternative, badges can also be accessed without authentication. On new Dependency-Track installations, this is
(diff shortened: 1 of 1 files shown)


nscuro (approved):
Makes sense. Once we have a more appropriate solution (#4227) we can remove the API key authentication completely again.

nscuro:
@SaberStrat Did you mean to make more changes? If not, please change the PR to ready for review.

PR #4521, opened 2025-01-06
Added support of deprecation for Nuget and Npm

### LINKED ISSUE #2388: Recognize NuGet `unlisted` packages and do not log 'error parsing upload time for NuGet component'
### Current Behavior

Using Dependency Track 4.7.0, warnings similar to those below are regularly observed in the logs:

```
2023-01-13 19:12:06,807 INFO [RepositoryMetaAnalyzerTask] Performing component repository metadata analysis against 58 components in project: 9731f6d0-12f0-4625-a688-fd8d0be204ea
2023-01-13 19:12:09,046 WARN [NugetMetaAnalyzer] An error occurred while parsing upload time for a NuGet component - Repo returned: 1900-01-01T00:00:00+00:00
2023-01-13 19:12:09,046 WARN [NugetMetaAnalyzer] An error occurred while parsing upload time for a NuGet component - Repo returned: 1900-01-01T00:00:00+00:00
2023-01-13 19:12:10,152 WARN [NugetMetaAnalyzer] An error occurred while parsing upload time for a NuGet component - Repo returned: 1900-01-01T00:00:00+00:00
```

This was reported in https://github.com/DependencyTrack/dependency-track/issues/537 but closed as an assumed defect with NuGet metadata.

The official NuGet docs describe at least one valid use case for the `1900-01-01` timestamp: unlisted packages, which are de-facto deprecated and should no longer be used.

One concrete example to illustrate this: https://www.nuget.org/packages/Unity/5.7.4 -- note the warning:

> The owner has unlisted this package. This could mean that the package is deprecated, has security vulnerabilities or shouldn't be used anymore. 

Use the NuGet API to retrieve the package registration details (https://api.nuget.org/v3/registration5-semver1/unity/5.7.4.json) and note the `listed: false` and `published: "1900-01-01T00:00:00+00:00"` values in the response:

```json
{
    "@id": "https://api.nuget.org/v3/registration5-semver1/unity/5.7.4.json",
    "@type": [
        "Package",
        "http://schema.nuget.org/catalog#Permalink"
    ],
    "catalogEntry": "https://api.nuget.org/v3/catalog0/data/2018.10.15.06.25.35/unity.5.7.4.json",
    "listed": false,
    "packageContent": "https://api.nuget.org/v3-flatcontainer/unity/5.7.4/unity.5.7.4.nupkg",
    "published": "1900-01-01T00:00:00+00:00",
    "registration": "https://api.nuget.org/v3/registration5-semver1/unity/index.json",
    "@context": { /* elided for brevity */ }
}
```

The NuGet docs at https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource include an explicit note for this behavior:

>  Note
> 
> **On nuget.org, the published value is set to year 1900 when the package is unlisted.**

However, Dependency Track does not currently recognize this value and logs a warning. 

### Proposed Behavior

Update `NugetMetaAnalyzer` to accept a NuGet API `published` value of `1900-01-01T00:00:00+00:00` as valid, and meant to represent *unlisted packages* whose use is discouraged.

https://github.com/DependencyTrack/dependency-track/blob/90437bdfee153ed6082e5dfc344a0bb7b5d2e515/src/main/java/org/dependencytrack/tasks/repositories/NugetMetaAnalyzer.java#L49-L52

Additionally, consider enhancing Dependency Track to flag use of unlisted NuGet packages with warnings, e.g. in its APIs or frontend UI. 

It may be helpful to see in the UI, similar to the warning icons for packages which have a newer version available, that the package is unlisted / deprecated and should no longer be used.

### Checklist

pohli81:
Added support of deprecation for Nuget and Npm

### Description

The component meta information is extended by the deprecation string and a boolean  whether the component is deprecated or not. Currently this is implemented for Nuget and Npm. This enhancement can be used to show deprecation information in the component list as in the following example:
image

### Addressed Issue

This PR should fix #2388.

### Additional Details

Currently it is implemented only for Nuget and Npm. It has to be implemented for other package repositories as well.

### Checklist

- [ x] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)
- [ x] This PR implements an enhancement, and I have provided tests to verify that it works as intended
- [x ] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)


valentijnscholten:
I recently thought about doing something like this. But I realized it might be good to align any changes here with the (hopefully) upcoming [OWASP Common Lifecycle Enumeration](https://owasp.org/www-project-common-lifecycle-enumeration/).
It's not done yet, but there are some initial events defined such as `endOfSupport` and `endOfLife`. 

https://docs.google.com/document/d/1sRMS1IX0r7ZkYthDR0VY1bYyvp_6K_xw4sR1vZwla8E/edit?tab=t.0

Would it be wise to pick fieldnames for DT from those already mentioned in the document?
Should it be made clear in the docs which MetaAnalyzers do (not yet) support these fields?

nscuro:
@pohli81 Thanks for the PR!

Are deprecations in the NuGet and NPM ecosystem on a per-package, or a per-version basis?

The limitation we have is that `RepositoryMetaComponent` records are tracked on the package-level. The records are "keyed" by `repositoryType`, `namespace`, and `name`. `version` is not part of the model.

So if only a subset of versions was to be deprecated, we cannot record this in `RepositoryMetaComponent`, as it would then erroneously flag current versions as deprecated, too.

I do agree with @valentijnscholten in that we should look towards CLE and align with it as much as possible. But in any case, we probably need a slight refactoring to support resolution of metadata on the version-level. Or even more granular, on the full PURL, so that qualifiers etc. are also considered.

pohli81:
Thanks for the answers. The deprecation information is on version level. My implementation sets the deprecation only in case the latest version is marked as deprecated. Unfortunately the reason for deprecation is mentioned in the message (EOL, renamed, no longer maintained, ...) so it is currently not easy to inference the reason. At least there is an enumeration in Nuget describing the reason (see https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-deprecation).

I added further tests to make the limitation transparent. There was also a bug I made in the NpmAnalyzer implementation I also fixed.

You are right @nscuro that the implementation based on version level needs a larger refactoring. The current impementation allows to make it visible in the UI, that there will be no newer package and Dependency Track can show the reason of the deprecation in a message.

I read the specification but currently it is not obvious to map the specification to the information available in the repositories and the data model of Dependency Track. So perhaps there could be a versionEvent field of type enumeration with the potential members out of the specification instead of isDeprecated of type boolean.

If we find an acceptable solution I can implement the same for further repositories supporting deprecation.

TobiasJoergensen:
I was looking into whether Dependency Track supported this functionality, but all I could find was this PR.
It could be a great feature to have the information if any NuGet packages have been marked as deprecated.

Bumping this PR for attention.

PR #4542, opened 2025-01-11
Testcase for adding a second tag to project

### LINKED ISSUE #4539: Tag doesn't get appended to project when added over API and project already has a tag
### Current Behavior

When sending a POST '/api/v1/tag/{name}/project' request, I get back a 204 response, but the new tag isn't appended to the project's tags if the project already has one tag added to it. The initial tag stays as it is.
When in the starting situation, the project has no tags, then the POST request adds the tag as intended. But if this is repeated with a project that already has one tag, nothing changes, or rather the new tag isn't appended.

### Steps to Reproduce

1. Find or create a project with one tag.
2. Send a POST request with URL '/api/v1/tag/{name}/project' using an arbitrary API client to the server with a valid tag that has already been created on the server and isn't the same as the one already present in the project.

### Expected Behavior

The new tag should be appended to the tags of the project.

### Dependency-Track Version

4.12.2

Container Image

### Database Server

N/A

### Database Server Version

_No response_

### Browser

N/A

### Checklist

valentijnscholten:
### Description

Should eventually fix #4539 

### Additional Details
Out of curiosity I created a test case to reproduce. Haven't looked at fixing it yet, feel free to advise.

### Checklist
- [x ] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)

### DIFF
--- a/src/test/java/org/dependencytrack/resources/v1/TagResourceTest.java
+++ b/src/test/java/org/dependencytrack/resources/v1/TagResourceTest.java
@@ -814,6 +814,36 @@ public class TagResourceTest extends ResourceTest {
     }
 
     @Test
+    public void tagProjectsExtraTagTest() {
+        initializeWithPermissions(Permissions.PORTFOLIO_MANAGEMENT);
+
+        final var projectA = new Project();
+        projectA.setName("acme-app-a");
+        qm.persist(projectA);
+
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Closing this as while I created this, @nscuro simultaneously submitted #4541 :-)

Issue #3234, opened 2023-11-25
Make use of logback's MDC to add context to log statements


nscuro:
### Current Behavior

Dependency-Track performs many tasks asynchronously. Status updates, warnings, and errors are logged during task execution, however those logs lack context:

* When BOM processing fails, it's unclear for *which* BOM, uploaded to *what* project
* When VEX processing fails, it's unclear for *which* VEX, uploaded to *what* project
* When repository metadata analysis fails, it's unclear for *which* component, and for *what* repository
* etc.

### Proposed Behavior

Logback offers a feature to add additional context to log statements via [Mapped Diagnostic Context](https://logback.qos.ch/manual/mdc.html) (MDC).

We should use that to attach context information, for example:

* Project UUID, name, version
* Component UUID, name, version
* BOM / VEX format, spec version, serial number, version
* etc.

wherever appropriate. When logging to JSON format, those context variables are represented as dedicated JSON fields and can thus be indexed by log aggregators (Elasticsearch, Splunk, ...).

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #4546, opened 2025-01-12
composer meta analyzer: add DEBUG logging

valentijnscholten:
Add some DEBUG log lines to help debugging (possible/future) problems.

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
+++ b/src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
@@ -114,6 +114,7 @@ public class ComposerMetaAnalyzer extends AbstractMetaAnalyzer {
 
         final JSONObject repoRoot = getRepoRoot();
         if (repoRoot == null) {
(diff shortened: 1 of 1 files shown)


valentijnscholten:
As this was a relatively new / last minute feature some extra logging can be helpful. In my test instance with real world data I am not seeing the log lines I was expecting, so I am not if it's correctly using the V2 API now.

nscuro:
Some of the new log statements do not provide any context (i.e. repository, package URL). `RepositoryMetaAnalyzerTask` does currently not set any MDC variables, so logs would not be all-too-helpful I fear.

Two options:

* Go the way of least resistance and add more context to the individual lines via `.formatted`
* Add MDC variables in `RepositoryMetaAnalyzerTask`

The latter would also greatly help with driving #3234 forward.

valentijnscholten:
Hmm was expecting this, but hoped you would forget :-)

In practice I usually just add the Thread-Id to the log pattern and filter on that. But yeah, more context is helpful. I don't have time to do the MDC now, maybe in the other Composer related draft PR later on.

nscuro (approved):

Issue #2683, opened 2023-04-17
Maintain audit trail for any update in an audit vulnerability


sahibamittal:
### Current Behavior

Currently, for any audit vulnerability in a project if there is any update in analysis say, severity level for a vulnerability, there is no update information documented in its audit trail. 

<img width="1325" alt="Screenshot 2023-04-17 at 1 38 20 p m" src="https://user-images.githubusercontent.com/5709882/232486307-2b64bca4-1fde-49d4-8b77-d380699dd578.png">

### Proposed Behavior

The information on any update on the vulnerability in a project should be documented to give more visibility over its changes.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #4750, opened 2025-03-12
Issue 4720

### LINKED ISSUE #4720: Updating multiple analysis fields with one API call only shows one in audit trail
### Current Behavior

All fields are updated to the correct values but only the Analysis field is mentioned in the audit trail when I set analysis, analysis response and analysis details at the same time via a single API call.

### Steps to Reproduce

1. Call the update analysis API call with multiple fields set
2. Observe audit trail only showing a single field

### Expected Behavior

Audit trail should show all modified fields, either as a single combined entry or as separate entries the way it does when changing the settings in the web UI.

### Dependency-Track Version

4.12.6

Container Image

### Database Server

PostgreSQL

### Database Server Version

13.20

### Browser

Mozilla Firefox

### Checklist

stohrendorf:
### Description

Adds complete analysis audit trail.

### Addressed Issue

Closes #4720 #2683 #3470

### Checklist


stohrendorf:
Regarding the `updateAnalysisCreateNewWithEmptyRequestTest`, I don't know what the expected behaviour is here. I can fix the test with this, but this seems wrong:

```java
        assertConditionWithTimeout(() -> NOTIFICATIONS.size() == 1, Duration.ofSeconds(5));
        final Notification projectNotification = NOTIFICATIONS.poll();
        assertThat(projectNotification).isNotNull();
        assertThat(projectNotification.getScope()).isEqualTo(NotificationScope.PORTFOLIO.name());
        assertThat(projectNotification.getGroup()).isEqualTo(NotificationGroup.PROJECT_CREATED.name());
        assertThat(projectNotification.getLevel()).isEqualTo(NotificationLevel.INFORMATIONAL);
        assertThat(projectNotification.getTitle()).isEqualTo(NotificationConstants.Title.PROJECT_CREATED, project);
        assertThat(projectNotification.getContent()).isEqualTo("Acme Example was created");
```

The original test is this: https://github.com/DependencyTrack/dependency-track/blob/f1352acdedda50e6ccd2b87343e6c9ed83ea236d/src/test/java/org/dependencytrack/resources/v1/AnalysisResourceTest.java#L474-L483

src/main/java/org/dependencytrack/persistence/FindingsQueryManager.java
```diff
@@ -194,15 +194,9 @@ public Analysis makeAnalysis(Component component, Vulnerability vulnerability, A
...
-        }
-        if (analysisJustification != null) {
-            analysis.setAnalysisJustification(analysisJustification);
-        }
-        if (analysisResponse != null) {
-            analysis.setAnalysisResponse(analysisResponse);
-        }
+        analysis.setAnalysisState(analysisState);
+        analysis.setAnalysisJustification(analysisJustification);
+        analysis.setAnalysisResponse(analysisResponse);
```
nscuro, line 199: Any reason for making this change? As per comment above, this could unintentionally nuke existing values.

src/main/java/org/dependencytrack/util/AnalysisCommentUtil.java
```diff
@@ -24,53 +24,43 @@
...
-        }
+        makeCommentIfChanged("Vendor Response", qm, analysis, Objects.requireNonNullElse(analysis.getAnalysisResponse(), AnalysisResponse.NOT_SET), analysisResponse, commenter);
     }
 
     public static void makeAnalysisDetailsComment(final QueryManager qm, final Analysis analysis, final String analysisDetails, final String commenter) {
-        if (analysisDetails != null && !analysisDetails.equals(analysis.getAnalysisDetails())) {
-            final String message = "Details: " + analysisDetails.trim();
-            qm.makeAnalysisComment(analysis, message, commenter);
-        }
+        makeCommentIfChanged("Details", qm, analysis, Objects.requireNonNullElse(analysis.getAnalysisDetails(), "None"), analysisDetails, commenter);
```
nscuro, line 46: Note how details comments did not follow the `old -> new` format, but instead simply logged the new value. Reason being that details can be arbitrarily long, such that including both old and new value in the same comment would flood the audit trail.

The previous value will already be present in the audit trail, so it is not lost.

src/test/java/org/dependencytrack/resources/v1/AnalysisResourceTest.java
```diff
@@ -471,16 +497,14 @@ public void updateAnalysisCreateNewWithEmptyRequestTest() throws Exception {
...
 
-        assertConditionWithTimeout(() -> NOTIFICATIONS.size() == 2, Duration.ofSeconds(5));
+        assertConditionWithTimeout(() -> NOTIFICATIONS.size() == 1, Duration.ofSeconds(5));
         final Notification projectNotification = NOTIFICATIONS.poll();
         assertThat(projectNotification).isNotNull();
-        final Notification notification = NOTIFICATIONS.poll();
-        assertThat(notification).isNotNull();
-        assertThat(notification.getScope()).isEqualTo(NotificationScope.PORTFOLIO.name());
-        assertThat(notification.getGroup()).isEqualTo(NotificationGroup.PROJECT_AUDIT_CHANGE.name());
-        assertThat(notification.getLevel()).isEqualTo(NotificationLevel.INFORMATIONAL);
```
nscuro, line 481: No `PROJECT_AUDIT_CHANGE` notification being emitted here seems like a breaking change. Was its emission a bug that is now resolved?

src/main/java/org/dependencytrack/persistence/FindingsQueryManager.java
stohrendorf: This was a change from static analysis from IntelliJ, and I applied it since it was related. It's clearly an unnecessary change and I will be happy to revert it.

src/test/java/org/dependencytrack/resources/v1/AnalysisResourceTest.java
stohrendorf: That's a valid concern, I will investigate.

src/main/java/org/dependencytrack/util/AnalysisCommentUtil.java
stohrendorf: I understand that this would introduce "noise". However, I am personally more leaning into logging every change _explicitly_. This is debatable. I understand that. But I think this is more like an _output_ problem, not an audit problem.

src/test/java/org/dependencytrack/resources/v1/AnalysisResourceTest.java
stohrendorf: Okay, in this particular case, the submitted analysis has no fields set (the test name even says this, `updateAnalysisCreateNewWithEmptyRequestTest`). With the new code, `null` and `NOT_SET` are handled slightly different. Semantically, I don't think this should lead to a `PROJECT_AUDIT_CHANGE` if nothing actually changed.

stohrendorf:
Regarding the failed test, see my other comment: https://github.com/DependencyTrack/dependency-track/pull/4766#issuecomment-2741311644

src/test/java/org/dependencytrack/resources/v1/AnalysisResourceTest.java
nscuro: Thanks for investigating, makes sense.

nscuro (approved):

src/main/java/org/dependencytrack/util/AnalysisCommentUtil.java
nscuro: Hmmm good point. We should probably add some sort of separator between trail entries in the UI, to make it more obvious where one entry ends and the next begins. Do you think that would be sufficient to address your concern?

stohrendorf: That was exactly my idea to be honest - if HTML is escaped properly, adding a simple `<hr>` tag between entries should be enough.

nscuro:
@stohrendorf I have removed #2683 from the "Closes" part of the PR description, since it describes a different enhancement request from what's being done here. It basically asks to populate the audit trail when, say, the NVD adjusts the vulnerability's severity from `MEDIUM` to `HIGH`.

PR #4766, opened 2025-03-19
add endpoint to mass-create tags

### LINKED ISSUE #4463: Use the API to create new tags
### Current Behavior

Currently the only way to create tags is to update an existing project, either through the API or through the UI.

Image

### Proposed Behavior

I wish to be able to create new Tags either from /tags or more importantly with a PUT to api/v1/tag. The only info should be name: "name-of-tag".

I was planning to implement a self-service tool outside dependency track where users would be allowed to set up alerts and then letting them choose a tag they wish to be added to that alert, using the "Limit to tags" field. The alert would be created using the API-call after using the API to create a tag.

Now instead the tool will need to verify that a tag exists, (finding a random project the tag should be added to), and then creating an alert and adding the tag to the alert.

### Checklist

### LINKED ISSUE #4755: Cannot update a project with new tag names (that are not used already on the platform)
### Current Behavior

If a tag was already attached to another project(tag T1 on project X), you can use the API to associate the tag T1 on project Y as well. But if you use the API to put a new tag T2 on project Y, it will fail with: "A tag with name T2 does not exist."

### Steps to Reproduce

1. Use the Rest API to tag a project with a brand new tag name: 
   ```  @POST
    @Path("/{name}/project")
    @Consumes(MediaType.APPLICATION_JSON)
    @Produces(MediaType.APPLICATION_JSON)
    @Operation(
            summary = "Tags one or more projects.",
            description = "<p>Requires permission <strong>PORTFOLIO_MANAGEMENT</strong></p>"
    )````, it will call `org.dependencytrack.resources.v1.TagResource.tagProjects(String, Set<String>)`
2. An eror is obtained: "A tag with name T2 does not exist.""

### Expected Behavior

The tag should be created automatically, persisted, and associated to the project, OR at least a REST API to properly create a tag should be provided. (By the way, please update the API descriptors, it looks like the API documentation is not updated)
org.dependencytrack.persistence.TagQueryManager.tagProjects(String, Collection<String>)
```    public void tagProjects(final String tagName, final Collection<String> projectUuids) {
        runInTransaction(() -> {
            final Tag tag = getTagByName(tagName);
            if (tag == null) {
              // TODO - CREATE HERE THE TAG,
                throw new NoSuchElementException("A tag with name %s does not exist".formatted(tagName));
            }

            final Query<Project> projectsQuery = pm.newQuery(Project.class);
            final var params = new HashMap<String, Object>(Map.of("uuids", projectUuids));
            preprocessACLs(projectsQuery, ":uuids.contains(uuid)", params, /* bypass */ false);
            projectsQuery.setNamedParameters(params);
            final List<Project> projects = executeAndCloseList(projectsQuery);

            for (final Project project : projects) {
                bind(project, List.of(tag), /* keepExisting */ true);
            }
        });
    }
```

### Dependency-Track Version

4.12.6

Container Image

### Database Server

PostgreSQL

### Database Server Version

_No response_

### Browser

Google Chrome

### Checklist

stohrendorf:
### Description

Adds an endpoint to create multiple tags.

### Addressed Issue

Closes #4755
Closes #4463

### Additional Details

Don't know whether a POST is the correct method here, but since this ignores already existing tags, it seems to be better than a PUT.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/TagResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/TagResource.java
@@ -53,6 +53,7 @@ import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.POST;
+import jakarta.ws.rs.PUT;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.Produces;
@@ -144,6 +145,31 @@ public class TagResource extends AlpineResource {
         }
 
         return Response.noContent().build();
+    }
+
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(
+            summary = "Creates one or more tags.",
+            description = "<p>Requires permission <strong>TAG_MANAGEMENT</strong></p>\n"
+    )
+    @ApiResponses(value = {
+            @ApiResponse(
+                    responseCode = "201",
+                    description = "Tags created successfully."
+            )
+    })
+    @PermissionRequired(Permissions.Constants.TAG_MANAGEMENT)
+    public Response createTags(
+            @Parameter(description = "Names of the tags to create")
+            @Size(min = 1, max = 100) final Set<@NotBlank String> tagNames
+    ) {
+        try (final var qm = new QueryManager(getAlpineRequest())) {
+            qm.createTags(List.copyOf(tagNames));
+        }
+
+        return Response.noContent().status(Response.Status.CREATED).build();
     }
 
     @GET
--- a/src/test/java/org/dependencytrack/resources/v1/TagResourceTest.java
+++ b/src/test/java/org/dependencytrack/resources/v1/TagResourceTest.java
@@ -540,6 +540,37 @@ public class TagResourceTest extends ResourceTest {
     }
 
     @Test
+    public void createTagsTest() {
+        initializeWithPermissions(Permissions.TAG_MANAGEMENT);
+
+        final Response response = jersey.target(V1_TAG)
+                .request()
+                .header(X_API_KEY, apiKey)
+                .property(ClientProperties.SUPPRESS_HTTP_COMPLIANCE_VALIDATION, true)
+                .method(HttpMethod.PUT, Entity.json(List.of("foo")));
+        assertThat(response.getStatus()).isEqualTo(201);
+        qm.getPersistenceManager().evictAll();
+        assertThat(qm.getTagByName("foo")).isNotNull();
+    }
+
+    @Test
+    public void createTagsWithExistingTest() {
+        initializeWithPermissions(Permissions.TAG_MANAGEMENT);
+
+        qm.createTag("foo");
+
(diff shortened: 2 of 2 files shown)


src/main/java/org/dependencytrack/resources/v1/TagResource.java
```diff
@@ -146,6 +146,36 @@ public Response deleteTags(
         return Response.noContent().build();
     }
 
+    @POST
```
nscuro, line 149: The API has historically used `PUT` for creating resources, and `POST` for updating them. Whether that is best practice is probably debatable, but for the sake of consistency, let's use `PUT` here as well.

src/main/java/org/dependencytrack/resources/v1/TagResource.java
```diff
@@ -146,6 +146,36 @@ public Response deleteTags(
...
+    @ApiResponses(value = {
+            @ApiResponse(
+                    responseCode = "201",
+                    description = "Tags created successfully."
+            ),
+            @ApiResponse(
+                    responseCode = "400",
+                    description = "Operation failed",
+                    content = @Content(schema = @Schema(implementation = TagOperationProblemDetails.class), mediaType = ProblemDetails.MEDIA_TYPE_JSON)
+            )
```
nscuro, line 165: `qm.createTags` will not throw a `TagOperationFailedException` so this particular response will never be returned. Best to omit this `@ApiResponse` definition to avoid confusion.

nscuro (changes requested):
Thanks! Just two minor adjustments necessary.

src/main/java/org/dependencytrack/resources/v1/TagResource.java
stohrendorf: Thanks for the clarification. Will change this for consistency.

src/main/java/org/dependencytrack/resources/v1/TagResource.java
stohrendorf: I have copied this from another endpoint. I am hesitant to change the other endpoint which specified this as a response, should I do that?

nscuro: Depends on what other endpoint you mean. If they declare this error response, their logic *should* be throwing a `TagOperationFailedException`, which would then be [mapped to `TagOperationProblemDetails`](https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/resources/v1/exception/TagOperationFailedExceptionMapper.java). If that is not the case, then the error declaration is wrong and could safely be removed.

stohrendorf:
Regarding the failed unit test, I can't reproduce it locally. It seems unrelated to these changes.

It's just guesswork now, but I had similar issues in the past with a different persistence framework, and the cause was change queuing and batch change commits; I had to evict the cache at certain times and force a flush to fix the issues I had. Maybe that's the case here, too? If so, this would explain some open issues.

nscuro:
@stohrendorf I have been unable to identify the root cause of these particular failures so far. The ORM generally flushes everything immediately, unless explicitly disabling this behavior. It also forces a flush whenever the `QueryManager` is closed. We destroy and re-create the persistence framework and its associated in-memory H2 database after every test. Definitely possible I missed something in investigating this, of course.

nscuro (approved):

PR #4783, opened 2025-03-26
Add support for scheduled summary notifications

### LINKED ISSUE #322: Portfolio Notification: Support for "Summary" Group

nscuro:
### Description

Builds upon @MM-msr's PR https://github.com/DependencyTrack/dependency-track/pull/3925 in the following ways:

* Adds validator for cron expressions so syntax mistakes can be caught sooner.
* Removes `ScheduledNotificationRule` model, associated persistence methods, and REST endpoints. Add the fields and capabilities required to support scheduled notifications to the existing `NotificationRule` model instead. This avoids a ton of redundant code.
* Removes `ScheduledNotificationTaskManager` in favour of the existing eventing system. An event is emitted every minute to check for due scheduled notifications. To keep track of when a notification was last sent, and when it's next due, we simply use dedicated fields in the `NotificationRule` model.
* Removes `SendScheduledNotificationTask` and leverages the existing `NotificationRouter` instead. Again avoiding a lot of redundant code.
* Improves efficiency slightly by avoiding redundant queries as much as possible while assembling notification subjects. The assembly is still relatively inefficient because it loads too much data into memory. Ideally we would convert all of this to SQL in v5.
* Enforces scheduled notifications to be limited to a set of projects. Emitting them for the entire portfolio is likely to cause OOM scenarios.
* Introduce dedicated `NotificationGroup`s for scheduled notifications, and integrate them into existing templates. There are no longer separate `NotificationPublisher`s for scheduled notifications. Instead, templates may be updated to support them.
* Unfortunately I had to remove the template for HTML emails. It was not possible to integrate into the existing text-based email template due to how MIME-types are handled. Ideally we would have a separate publisher and template for HTML-formatted emails, but I ran out of time to create such a template that covers all (or at least most) notification groups. For the time being, users can simply create their own publisher and template.

### Addressed Issue

Supersedes #3925
Closes #322

### Additional Details

Frontend PR: https://github.com/DependencyTrack/frontend/pull/1210

### Checklist


rkg-mm:
@nscuro Thank you very much, we will check this asap. 
Some questions:
1) Did you also update the frontend already?
2) We had html mails working, whats the difference now which makes it no longer possible in your approach? Adding own publishers isn't easy without modifying the source code, or is it possible to plug them in without rebuilding the project, ideally using the official containers?

nscuro:
> @rkg-mm Did you also update the frontend already?

Yes, see here: https://github.com/DependencyTrack/frontend/pull/1210

> We had html mails working, whats the difference now which makes it no longer possible in your approach? 

Basically, the [default email publisher](https://github.com/MM-msr/dependency-track/blob/e51d2f16125f76c49c9a594506aa73056b4c47c7/src/main/java/org/dependencytrack/notification/publisher/DefaultNotificationPublishers.java#L28) uses the MIME type `text/plain`. If you send an HTML email with this MIME type, mail clients are likely to not render the HTML. So it wasn't feasible to simply add the HTML to the existing email template.

In order to change the MIME type, you need a new publisher. This is what your original PR [did](https://github.com/MM-msr/dependency-track/blob/e51d2f16125f76c49c9a594506aa73056b4c47c7/src/main/java/org/dependencytrack/notification/publisher/DefaultNotificationPublishers.java#L29). The downside is that that publisher and its associated template only catered to the new scheduled notification, but not any other notification type.

In your PR, you worked around this by [differentiating](https://github.com/MM-msr/dependency-track/blob/e51d2f16125f76c49c9a594506aa73056b4c47c7/src/main/java/org/dependencytrack/model/NotificationPublisher.java#L112-L114) between publishers for scheduled, and those for "normal" notifications. But this was really just a workaround, and it was really only necessary for emai due to the MIME type conflict. On a technical level there is nothing that would prevent a normal publisher to be used for scheduled notifications, other than that its template would need updating.

So I had two options at this point:

1. Keep the additional email publisher for HTML, and extend the template such that it also caters to other notification types.
2. Simply extend the existing plaintext email publisher for the new summary types.

For time reasons, I went with (2). Reasoning being that at least the core functionality would be out sooner, and those wishing to use HTML could simply provide a custom template (see below).

We can of course still include the partial HTML email template per default, but my fear is that it'd be too confusing if it only supports a very small subset of notifications.

If you can think of other options I'm open to that of course.

> Adding own publishers isn't easy without modifying the source code, or is it possible to plug them in without rebuilding the project, ideally using the official containers?

You can just clone the default email publisher via the UI, and provide a custom template: https://docs.dependencytrack.org/integrations/notifications/#creation-of-publisher

nscuro:
> Everytime the scheduled email should be dispatched, the following two exceptions occur and no mail is sent: [...]

Yep, good catch. Here I am adding to a list while iterating over it:

https://github.com/nscuro/dependency-track/blob/c9e952b20fe6252afb2d9367f1e59af8f3062208/src/main/java/org/dependencytrack/tasks/ScheduledNotificationDispatchTask.java#L365-L367

Fixing.

nscuro:
@rbt-mm Should be fixed now, please have another go when you find the time.

> I could also provide our newly adjusted HTML template in some way once I'm done with testing.

What are your thoughts on my previous question?

> If you think it's a safe thing to do, we can simply add your original HTML template code to the email.peb file and update the publisher's MIME type.

I def. see the value in having summaries such as these in HTML format, I'm just not too fond of shipping partial templates. So wondering how to best progress on this.

rbt-mm:
> @rbt-mm Should be fixed now, please have another go when you find the time.

Thanks for the quick fix, the mails are now correctly dispatched! Stuff like limiting to projects and not publishing mails when there is no new information also seems to work fine. So the changes look good to me 👍


> I def. see the value in having summaries such as these in HTML format, I'm just not too fond of shipping partial templates. So wondering how to best progress on this.

I also think its better to have such summaries in an easy to digest format like HTML, but I wouldn't add it to the current `email.peb` and change the MIME type of the publisher, because I'm also not sure if changing to `text/html` could potentially break stuff for the current users.

Maybe providing this HTML summary template (and potentially other useful templates from the community which might be contributed in the future?) in a designated documentation/wiki page could be a solution for this?

nscuro:
> Maybe providing this HTML summary template (and potentially other useful templates from the community which might be contributed in the future?) in a designated documentation/wiki page could be a solution for this?

Like the idea. We could introduce a `contrib` directory to the repo where we keep these. Once templates mature enough we can introduce them as defaults to the core application.

Can I ask one last favour? What are your thoughts on the text-based templates in this PR? Any more info you'd like to see there?

rbt-mm:
I think the text-based templates look good and contain all the necessary information 👍
Two things which I think could be helpful when adjusted:

- In the `Project Summary & Vulnerability/Violation Details` adding an empty newline between separate entries could improve readability
- Not sure if it is possible, but mentioning the number of findings directly in the notification title like `New Policy Violations Summary (208 violations in 4 components)` - same for vulnerability summaries of course- could also help receivers to quickly identify how important/urgent a message is and it also helps when differentiating different notifications because they won't all have the same title anymore

nscuro:
> In the Project Summary & Vulnerability/Violation Details adding an empty newline between separate entries could improve readability

Great feedback, thanks. Will add this.

> Not sure if it is possible, but mentioning the number of findings directly in the notification title [...]

Unfortunately it's not, currently. Email is kind of unique in that the title of the message is specified separate from the main body. We'd need separate templating to enable this. But agree on that it'd be good to have.

PR #4796, opened 2025-03-31
Add feature to define the test title for DefectDojo integration

### LINKED ISSUE #2513: Set Testname in DefectDojo by product property and use better default value
### Current Behavior

When the DD integration is enabled the name of the test in DD is always "Dependency Track Finding Packaging Format (FPF) Export".

I also noticed a side effect when I'm using the same engagementId in multiple products (with reimport enabled). In this case each reimport closes the findings of the previous imports.

### Proposed Behavior

I would like to set the name of the test as a product property like "defectdojo.testName".
With this solution I could have different tests bundled in one engagement.

Also I propose to include the DT product name and version into the default DD test name. Like "Dependency Track - PRODUCT_NAME (PRODUCT_VERSION) - Finding Packaging Format (FPF) Export". 
This also will avoid collisions between multiple imports to DD.

### Checklist

AndreVirtimo:
New optional per project property "defectdojo.testTitle".
When property is set, the given test title will be set in DefectDojo.

### Description

With the current behavior we always get the same test title ("Dependency Track Finding Packaging Format (FPF) Export").

The new property allows the user to set his desired test title.

### Addressed Issue

closes #2513

### Additional Details

If you want to use a single engagement with multiple imports from different DT projects, you can't identify the different tests. When using reimport, it gets worse. Each project will override the findings of the previous ones.

### Checklist


nscuro (approved):
Quality PR, thank you!

PR #4832, opened 2025-04-09
Convert tests to JUnit 5

stohrendorf:
### Description

Converts all tests from JUnit 4 to 5. From personal motivation, so that I can execute the tests from within IntelliJ more easily.

### Additional Details
This might break the GitHub workflows. Additionally, the tests run from within IntelliJ run 4x faster, and it seems that the tests sometimes fail because some resource is not properly closed/cleaned up, but I was unable to identify the underlying issue yet - this _might_ also explain the random errors I had with my past PRs, and I'd like to investigate this.

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,7 +105,7 @@
         <lib.jackson-databind.version>2.18.3</lib.jackson-databind.version>
         <lib.json-java.version>20250107</lib.json-java.version>
         <lib.json-unit.version>4.1.0</lib.json-unit.version>
-        <lib.junit.version>4.13.2</lib.junit.version>
+        <lib.junit.version>5.11.4</lib.junit.version>
         <lib.lucene.version>8.11.4</lib.lucene.version>
         <lib.maven-artifact.version>3.9.9</lib.maven-artifact.version>
         <lib.mockserver-netty.version>5.15.0</lib.mockserver-netty.version>
@@ -115,11 +115,10 @@
         <lib.protobuf-java.version>4.30.2</lib.protobuf-java.version>
         <lib.resilience4j.version>2.2.0</lib.resilience4j.version>
         <lib.swagger-parser.version>2.1.25</lib.swagger-parser.version>
-        <lib.system-rules.version>1.19.0</lib.system-rules.version>
+        <lib.junit-pioneer.version>2.3.0</lib.junit-pioneer.version>
         <lib.testcontainers.version>1.21.0</lib.testcontainers.version>
         <lib.wiremock.version>2.35.2</lib.wiremock.version>
         <lib.woodstox.version>7.0.0</lib.woodstox.version>
-        <lib.junit-params.version>1.1.1</lib.junit-params.version>
         <lib.signpost-core.version>2.1.1</lib.signpost-core.version>
         <lib.httpclient.version>4.5.14</lib.httpclient.version>
         <lib.httpclient5.version>5.4.4</lib.httpclient5.version>
@@ -421,15 +420,21 @@
 
         <!-- Test Dependencies -->
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
             <version>${lib.junit.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>pl.pragmatists</groupId>
-            <artifactId>JUnitParams</artifactId>
-            <version>${lib.junit-params.version}</version>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${lib.junit.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>${lib.junit.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -437,6 +442,12 @@
             <artifactId>jersey-test-framework-provider-grizzly2</artifactId>
             <version>${lib.jersey.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>junit</groupId>
+                    <artifactId>junit</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.glassfish.jersey.connectors</groupId>
@@ -458,9 +469,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>com.github.stefanbirkner</groupId>
-            <artifactId>system-rules</artifactId>
-            <version>${lib.system-rules.version}</version>
+            <groupId>org.junit-pioneer</groupId>
+            <artifactId>junit-pioneer</artifactId>
+            <version>${lib.junit-pioneer.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -489,7 +500,7 @@
         </dependency>
         <dependency>
             <groupId>com.icegreen</groupId>
-            <artifactId>greenmail-junit4</artifactId>
+            <artifactId>greenmail-junit5</artifactId>
             <version>${lib.greenmail.version}</version>
             <scope>test</scope>
         </dependency>
@@ -617,13 +628,6 @@
                         </property>
                     </systemProperties>
                 </configuration>
-                <dependencies>
-                    <dependency>
-                        <groupId>org.apache.maven.surefire</groupId>
-                        <artifactId>surefire-junit4</artifactId>
-                        <version>${maven.surefire.plugin.version}</version>
-                    </dependency>
-                </dependencies>
             </plugin>
             <plugin>
                 <groupId>org.jacoco</groupId>
--- a/src/test/java/org/dependencytrack/JerseyTestExtension.java
+++ b/src/test/java/org/dependencytrack/JerseyTestExtension.java
@@ -28,8 +28,8 @@ import org.glassfish.jersey.test.JerseyTest;
 import org.glassfish.jersey.test.ServletDeploymentContext;
 import org.glassfish.jersey.test.spi.TestContainerException;
 import org.glassfish.jersey.test.spi.TestContainerFactory;
-import org.junit.jupiter.api.extension.AfterEachCallback;
-import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.AfterAllCallback;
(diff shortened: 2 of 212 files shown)


src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -965,7 +965,6 @@ public void recursivelyDelete(final Project project, final boolean commitIndex)
         delete(project.getMetadata());
         delete(project.getProperties());
         delete(getAllBoms(project));
-        delete(project.getChildren());
```
nscuro, line 968: Why was this removed?

pom.xml
```diff
@@ -750,6 +755,24 @@
...
+                </configuration>
+                <executions>
+                    <execution>
+                        <phase>process-classes</phase>
+                        <goals>
+                            <goal>enhance</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
nscuro, line 775: What's the reason for including this directly?

pom.xml
```diff
@@ -134,6 +133,7 @@
         <plugin.retirejs.breakOnFailure>false</plugin.retirejs.breakOnFailure>
         <plugin.jetty.version>12.0.19</plugin.jetty.version>
         <plugin.protoc-jar.version>3.11.4</plugin.protoc-jar.version>
+        <plugin.datanucleus.version>5.2.1</plugin.datanucleus.version>
```
nscuro, line 136: This should be `6.0.0-release`, [as used by Alpine](https://github.com/stevespringett/Alpine/blob/28c424604b37a093c754e14db02a4092866cf916/pom.xml#L109).

src/test/java/org/dependencytrack/ResourceTest.java
```diff
@@ -86,37 +83,21 @@ public abstract class ResourceTest {
...
-        // that was neither committed nor rolled back. Unfortunately some areas of the
-        // code base can leave such a broken state behind if they run into unexpected
-        // errors. See: https://github.com/DependencyTrack/dependency-track/issues/2677
-        if (!qm.getPersistenceManager().isClosed()
-                && qm.getPersistenceManager().currentTransaction().isActive()) {
-            qm.getPersistenceManager().currentTransaction().rollback();
-        }
-
-        PersistenceManagerFactory.tearDown();
-    }
```
nscuro, line 118: Why would this no longer be necessary?

src/test/java/org/dependencytrack/notification/publisher/JiraPublisherTest.java
```diff
@@ -34,20 +34,18 @@
 import static org.dependencytrack.model.ConfigPropertyConstants.JIRA_URL;
 import static org.dependencytrack.model.ConfigPropertyConstants.JIRA_USERNAME;
 
-public class JiraPublisherTest extends AbstractWebhookPublisherTest<JiraPublisher> {
+class JiraPublisherTest extends AbstractWebhookPublisherTest<JiraPublisher> {
```
nscuro, line 37: Executing this test does no actually run the methods annotated with `@Override`, but they run their super class' methods instead. You can observe this by setting a breakpoint on the `verify` calls - the breakpoint won't trigger.

IIRC this is a behavior change in JUnit 5, in that it no longer runs `@Test` methods from superclasses. You'll need to remove `@Test` from methods in `AbstractPublisherTest`, and add them to the `@Override` methods instead.

src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java
```diff
@@ -85,7 +88,7 @@ public abstract class AbstractPublisherTest<T extends Publisher> extends Persist
     }
 
     @Test
-    public void testInformWithBomConsumedNotification() {
+    public void testInformWithBomConsumedNotification(WireMockRuntimeInfo wmRuntimeInfo) {
```
nscuro, line 91: Is there a way to have the WireMock instance as a class-level field, instead of "colouring" all methods by requiring a `WireMockRuntimeInfo` parameter?

src/test/java/org/dependencytrack/notification/publisher/AbstractWebhookPublisherTest.java
```diff
@@ -18,30 +18,26 @@
...
+import org.junit.jupiter.api.BeforeEach;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
 import static com.github.tomakehurst.wiremock.client.WireMock.anyUrl;
 import static com.github.tomakehurst.wiremock.client.WireMock.post;
 import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
-import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.options;
 import static org.dependencytrack.model.ConfigPropertyConstants.GENERAL_BASE_URL;
 
+@WireMockTest
```
nscuro, line 32: Are we creating two WireMock instances by repeating this annotation on `AbstractPublisherTest` and `AbstractWebhookPublisherTest`?

nscuro (changes requested):
Thanks @stohrendorf!

Did you use OpenRewrite for this? https://docs.openrewrite.org/running-recipes/popular-recipe-guides/migrate-from-junit-4-to-junit-5

> the tests run from within IntelliJ run 4x faster

Interesting. Any guesses as to why that is the case? The test suite appears to run roughly for the same duration in CI (~20min).

stohrendorf:
> Did you use OpenRewrite for this? https://docs.openrewrite.org/running-recipes/popular-recipe-guides/migrate-from-junit-4-to-junit-5

No, it was a half-automated conversion using IntelliJ over a few days.

pom.xml
stohrendorf: When using IntelliJ to compile everything, it wasn't enhancing the classes before adding this. Some of the changes in this PR are not exactly necessary and are purely for my own convenience, althouth I have tried to keep these changes to a minimum.

src/test/java/org/dependencytrack/ResourceTest.java
stohrendorf: The ResourceTest is now inherited from PersistenceCapableTest, which does the exactly same thing.

src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
stohrendorf: Because, at some point, I got an exception here, since the children are already deleted in lines 941 to 945.

stohrendorf:
> Interesting. Any guesses as to why that is the case? The test suite appears to run roughly for the same duration in CI (~20min).

It seems that the cause for the speedup was that IntelliJ by default does not fork the JVM, but surefire does per class. Once I configured IntelliJ to fork per class, the tests ran as slow as the maven goal, and the DB errors didn't occur anymore. This _seems_ to confirm my suspicion, that some resource is not properly cleaned up between tests, but that's just a guess.

Please note that Java is not my home ground, I'm used to C++ and Python, so I'm still learning.

src/test/java/org/dependencytrack/notification/publisher/JiraPublisherTest.java
stohrendorf: It seems there's now an error within the new `baseTestInformWithTemplateInclude` (I renamed the base class methods), running from within `SendMailPublisherTest`. I can't tell whether this is a legitimate error.

nscuro:
> It seems that the cause for the speedup was that IntelliJ by default does not fork the JVM, but surefire does per class.

Ah, that does make a lot of sense. I just stumbled over the same thing a few days ago [here](https://github.com/DependencyTrack/hyades-apiserver/pull/1130#discussion_r2047784524). The forking behavior is inherited from Alpine and *ideally* should not be needed for DT. 

The speedup of tests *is* quite significant when enabling reuse of forks. So would be great for sure if we can find the culprits that prevent it from being used.

nscuro:
Just ran the tests with `<reuseForks>true</reuseForks>` in the surefire configuration, and I see what you meant with DB errors. 

I have not observed this in the `hyades-apiserver` repository. The difference between it and this repository, is that `hyades-apiserver` runs tests against PostgreSQL containers, whereas this uses an in-memory H2 database.

There is some internal shenanigans going on with H2 where it may delay shutdown after the last connection is closed, likely that is the problem here.

It seems forcing the H2 database to be [shutdown](https://www.h2database.com/html/commands.html#shutdown) solves most test failures, i.e. by adding the followng snippet [here](https://github.com/DependencyTrack/dependency-track/blob/7b81bc24efabd86e25145102483de91f28959b44/src/test/java/org/dependencytrack/PersistenceCapableTest.java#L52):

```java
try {
    // Make sure the in-memory H2 database is closed before the next test is run.
    qm.getPersistenceManager().setProperty(PropertyNames.PROPERTY_QUERY_SQL_ALLOWALL, "true");
    qm.getPersistenceManager().newQuery(Query.SQL, "SHUTDOWN IMMEDIATELY").execute();
} catch (JDODataStoreException | JDOFatalUserException e) {
    // ignored, DB already closed.
}
```

stohrendorf:
Out of curiousity, I tried to implement the fork/shutdown stuff, but it seems a lot of tests now fail because of concurrency issues or non-deterministic behaviour. Additionally, either the DB is now not fully initialized before some few tests, or still not shutdown properly. Fixing this is a lot more work.

nscuro:
> Out of curiousity, I tried to implement the fork/shutdown stuff, but it seems a lot of tests now fail because of concurrency issues or non-deterministic behaviour. Additionally, either the DB is now not fully initialized before some few tests, or still not shutdown properly.

Bummer. I was running the test suite 3 times locally and the only failure I received was due to some unexpected ordering in one resource test. We'll tackle this another time then, thanks for checking!

nscuro (changes requested):

src/test/java/org/dependencytrack/JerseyTestExtension.java
```diff
@@ -27,18 +28,26 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
 
-import jakarta.ws.rs.client.WebTarget;
+import java.util.function.Supplier;
 
 /**
  * @since 4.11.0
  */
-public class JerseyTestRule extends ExternalResource {
+public class JerseyTestExtension implements BeforeEachCallback, AfterEachCallback {
```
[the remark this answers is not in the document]
stohrendorf: Thanks for catching this.

nscuro (approved):
Thanks, great work!

stohrendorf:
Oh, sorry, you merged the fixup commit. I was about to rebase and squash this PR, so this merge introduced a fixup commit, which should never be merged to keep a clean history. If you see a fixup commit by me, and want to merge, please inform be so I can rebase the branch and squash the fixup commits.

nscuro:
@stohrendorf No worries, I did a squash merge so all of this got merged in a single commit.

PR #4879, opened 2025-04-25
Add mail notification severity options

### LINKED ISSUE #3767: Option to trigger notifications based on severity of the CVE
### Current Behavior

I can configure an alert notification for newly detected vulnerabilities.

### Proposed Behavior

Currently, it is possible to configure a notification when a new vulnerability is detected. I would like to be able to trigger an alert notification only when the severity level is critical or high. 

### Checklist

emyhrberg:
### Description

This PR adds support for configuring which CVE severities should trigger alert notifications.

#### Previously:
All NEW_VULNERABILITY_IDENTIFIED events triggered notifications, regardless of severity (e.g., low, medium, high).

#### Now:
Users can select which severities (e.g., only CRITICAL and HIGH) should trigger notifications. This helps reduce noise and makes it easier to focus on the most important vulnerabilities.

If no severity is selected, the system behaves as before (all severities trigger alerts).

### Addressed Issue

Closes #3767 

### Additional Details

As mentioned by users in the issue, teams managing many microservices currently receive too many alerts for low-priority issues. This change allows them to only receive notifications for critical or high-severity vulnerabilities, reducing email overload and improving focus. 

Also, see [frontend PR](https://github.com/DependencyTrack/frontend/pull/1235)

### Checklist

~- [] This PR fixes a defect, and I have provided tests to verify that the fix is effective~
~- [] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)~

### DIFF
--- a/src/main/java/org/dependencytrack/model/NotificationRule.java
+++ b/src/main/java/org/dependencytrack/model/NotificationRule.java
@@ -18,6 +18,7 @@
  */
 package org.dependencytrack.model;
 
+import alpine.common.logging.Logger;
 import alpine.common.validation.RegexSequence;
 import alpine.model.Team;
 import alpine.notification.NotificationLevel;
@@ -38,6 +39,8 @@ import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Pattern;
 import jakarta.validation.constraints.Size;
+import org.dependencytrack.notification.publisher.SendMailPublisher;
+
 import javax.jdo.annotations.Column;
 import javax.jdo.annotations.Element;
 import javax.jdo.annotations.Extension;
@@ -137,6 +140,10 @@ public class NotificationRule implements Serializable {
     @Persistent
     @Column(name = "NOTIFY_ON", length = 1024)
     private String notifyOn;
+
+    @Persistent
+    @Column(name = "NOTIFY_SEVERITIES", length = 1024)
+    private String notifySeverities;
 
     @Persistent
     @Column(name = "MESSAGE", length = 1024)
@@ -323,6 +330,37 @@ public class NotificationRule implements Serializable {
             }
         }
         this.notifyOn = sb.toString();
+    }
+
+    public List<Severity> getNotifySeverities(){
+        Logger LOGGER = Logger.getLogger(NotificationRule.class);
+        LOGGER.debug("S");
+
+        List<Severity> result = new ArrayList<>();
+        if (notifySeverities != null) {
+            String[] severities = notifySeverities.split(",");
+            for (String s: severities) {
+                result.add(Severity.valueOf(s.trim()));
+            }
+        } else {
+            Collections.addAll(result, Severity.values());
+        }
+        return result;
+    }
+
+    public void setNotifySeverities(List<Severity> notifySeverities){
+        if (notifySeverities.isEmpty()){
+            this.notifySeverities = null;
+            return;
+        }
+        StringBuilder sb = new StringBuilder();
+        for (int i=0; i<notifySeverities.size(); i++) {
+            sb.append(notifySeverities.get(i));
+            if (i+1 < notifySeverities.size()) {
+                sb.append(",");
+            }
+        }
+        this.notifySeverities = sb.toString();
     }
 
     public NotificationPublisher getPublisher() {
--- a/src/main/java/org/dependencytrack/notification/NotificationRouter.java
+++ b/src/main/java/org/dependencytrack/notification/NotificationRouter.java
@@ -29,6 +29,7 @@ import org.dependencytrack.model.Project;
 import org.dependencytrack.model.Tag;
 import org.dependencytrack.notification.publisher.PublishContext;
 import org.dependencytrack.notification.publisher.Publisher;
+import org.dependencytrack.notification.publisher.SendMailPublisher;
 import org.dependencytrack.notification.vo.AnalysisDecisionChange;
 import org.dependencytrack.notification.vo.BomConsumedOrProcessed;
 import org.dependencytrack.notification.vo.BomProcessingFailed;
@@ -90,7 +91,12 @@ public class NotificationRouter implements Subscriber {
                             .add(CONFIG_TEMPLATE_KEY, notificationPublisher.getTemplate())
                             .addAll(Json.createObjectBuilder(config))
                             .build();
-                    publisher.inform(ruleCtx, restrictNotificationToRuleProjects(notification, rule), notificationPublisherConfig);
+                    if (publisherClass == SendMailPublisher.class) {
+                        ((SendMailPublisher) publisher).inform(ruleCtx, restrictNotificationToRuleProjects(notification, rule), notificationPublisherConfig, rule.getTeams(), rule.getNotifySeverities());
+                    } else {
+                        publisher.inform(ruleCtx, restrictNotificationToRuleProjects(notification, rule), notificationPublisherConfig);
+                    }
+
                 } else {
                     LOGGER.error("The defined notification publisher is not assignable from " + Publisher.class.getCanonicalName() + " (%s)".formatted(ruleCtx));
                 }
--- a/src/main/java/org/dependencytrack/notification/publisher/SendMailPublisher.java
+++ b/src/main/java/org/dependencytrack/notification/publisher/SendMailPublisher.java
@@ -19,6 +19,7 @@
 package org.dependencytrack.notification.publisher;
 
 import alpine.common.logging.Logger;
+import alpine.model.Team;
 import alpine.notification.Notification;
 import alpine.server.mail.SendMail;
 import alpine.server.mail.SendMailException;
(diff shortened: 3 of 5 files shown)


rkg-mm:
Love this, I need this :)
Not sure if maintaining them as string is the best solution, but will leave this up to maintainers ;)

But another note: I might have overseen something, but what happens if you already have notification rules and then install this update? Did I miss some migration or default that falls back to the original behavior for already existing rules?

emyhrberg:
> Love this, I need this :) Not sure if maintaining them as string is the best solution, but will leave this up to maintainers ;)
> 
> But another note: I might have overseen something, but what happens if you already have notification rules and then install this update? Did I miss some migration or default that falls back to the original behavior for already existing rules?

Hi @rkg-mm 

Thanks alot! 😄 
About your question: existing notification rules will continue work exactly as before. This means if `notifySeverities ` is not set, it will default to all severities. Only Mail notifications `(SendMailPublisher)` are filtered, and they are only filtered for the event types selected (which work exactly as before).

Also, regarding storing severities as a comma-separated string: It is consistent with how notifyOn was already stored, but we can change this if needed.

src/main/java/org/dependencytrack/notification/NotificationRouter.java
```diff
@@ -91,6 +92,11 @@ public void inform(final Notification notification) {
                             .addAll(Json.createObjectBuilder(config))
                             .build();
                     publisher.inform(ruleCtx, restrictNotificationToRuleProjects(notification, rule), notificationPublisherConfig);
+                    if (publisherClass != SendMailPublisher.class || rule.getTeams().isEmpty() || rule.getTeams() == null) {
+                        publisher.inform(ruleCtx, restrictNotificationToRuleProjects(notification, rule), notificationPublisherConfig);
+                    } else {
+                        ((SendMailPublisher) publisher).inform(ruleCtx, restrictNotificationToRuleProjects(notification, rule), notificationPublisherConfig, rule.getNotifySeverities());
+                    }
```
nscuro, line 99: Only doing this for the email publisher seems like an artificial limitation. 
What's the reason for doing it this way?

emyhrberg: > Only doing this for the email publisher seems like an artificial limitation. What's the reason for doing it this way?

Email is the only publisher we tested, we had trouble setting up SMTP server for Slack and the other publishers.

emyhrberg: @nscuro 

> Only doing this for the email publisher seems like an artificial limitation. What's the reason for doing it this way?

Just pushed a commit that includes all publishers, let me know if that works @nscuro

emyhrberg:
I believe this PR is ready to merge @nscuro

src/main/java/org/dependencytrack/model/NotificationRule.java
```diff
@@ -325,6 +329,34 @@ public void setNotifyOn(Set<NotificationGroup> groups) {
...
 
+    public List<Severity> getNotifySeverities(){
+        List<Severity> result = new ArrayList<>();
+        if (notifySeverities != null) {
+            String[] severities = notifySeverities.split(",");
+            for (String s: severities) {
+                result.add(Severity.valueOf(s.trim()));
+            }
+        } else {
+            return List.of(Severity.values());
```
nscuro, line 340: The problem with returning all severities per default is that this also "bleeds" through to the REST API. We'd then have `NotificationRules` returning all severities, even if they are not configured for vulnerability-related `NotificationGroup`s.

Perhaps this kind of fallback should be in `NotificationRouter` instead, when `notifivySeverities` is actually being evaluated?

src/main/java/org/dependencytrack/notification/publisher/Publisher.java
```diff
@@ -61,6 +63,30 @@ public interface Publisher {
...
+
+            final Severity sev = vuln.getVulnerability().getSeverity();
+            if (sev != null && !notifySeverities.contains(sev)) {
+                return;
+            }
+        }
+
+        // Call the default inform method
+        inform(ctx, notification, config);
+    }
```
nscuro, line 88: Consider performing the filtering [here](https://github.com/DependencyTrack/dependency-track/blob/d6f63c78216915418bdaba36ef0f188c82e008d7/src/main/java/org/dependencytrack/notification/NotificationRouter.java#L220) in `NotificationRouter` instead.

Notifications not eligible for publishing should never reach the `Publisher` to begin with.

src/main/java/org/dependencytrack/model/NotificationRule.java
```diff
@@ -325,6 +329,34 @@ public void setNotifyOn(Set<NotificationGroup> groups) {
...
+            return;
+        }
+        StringBuilder sb = new StringBuilder();
+        for (int i=0; i<notifySeverities.size(); i++) {
+            sb.append(notifySeverities.get(i));
+            if (i+1 < notifySeverities.size()) {
+                sb.append(",");
+            }
+        }
+        this.notifySeverities = sb.toString();
```
nscuro, line 357:
```suggestion
        if (notifySeverities == null || notifySeverities.isEmpty()){
            this.notifySeverities = null;
            return;
        }
        this.notifySeverities = notifySeverities.stream().collect(Collectors.joining(",");
```

src/main/java/org/dependencytrack/notification/publisher/Publisher.java
```diff
@@ -61,6 +63,30 @@ public interface Publisher {
...
+                        JsonObject config,
+                        List<Severity> notifySeverities) {
+
+        if (notification == null || config == null) {
+            return;
+        }
+
+        // Filter out unwanted notifications based on the configured severities
+        if (notifySeverities != null && !notifySeverities.isEmpty()
+            && notification.getSubject() instanceof NewVulnerabilityIdentified vuln) {
```
nscuro, line 78: Shouldn't this also apply to `NEW_VULNERABLE_DEPENDENCY` notifications? In that case, a component can have multiple vulnerabilities. How do we want to treat that case? Do we drop the entire notification, or do we just filter out vulnerabilities that don't fit the configured severity criteria?

nscuro (changes requested):

src/main/java/org/dependencytrack/model/NotificationRule.java
emyhrberg: The reason we implemented it by returning all severities per default is to have it visually check all the checkboxes by default. I figured this would clarify the behaviour of the options to the users. Anyway, I moved the empty list check to `NotificationRouter`, as you suggested, and it still works as expected. Only the visual part changed; all the checkboxes are not checked by default. But we are open to change it, if needed.

src/main/java/org/dependencytrack/notification/publisher/Publisher.java
emyhrberg: Fixed, moved the filtering to where you suggested in `NotificationRouter`, where I also added filtering to `NewVulnerableDependency`.

src/main/java/org/dependencytrack/model/NotificationRule.java
emyhrberg: Fixed.

src/main/java/org/dependencytrack/notification/publisher/Publisher.java
emyhrberg: Added filtering for `NEW_VULNERABLE_DEPENDENCY` notifications.

It currently works by dropping the entire notification if it does not fit the configured severity filters.

emyhrberg:
A new commit has been made addressing the comments from @nscuro .

emyhrberg:
@nscuro Any update on this PR? I believe all the changes requested are already done, let me know otherwise please, thanks!

HeavyRainQ:
Hi, can you please tell me when this feature will be available in the release version? I'm using 4.13.5 and it's not available yet. @emyhrberg

src/main/java/org/dependencytrack/notification/NotificationRouter.java
```diff
@@ -218,12 +220,50 @@ List<NotificationRule> resolveRules(final PublishContext ctx, final Notification
             pm.detachCopyAll(result);
             LOGGER.debug("Matched %d notification rules (%s)".formatted(result.size(), ctx));
 
+            final List<NotificationRule> severityFiltered = new ArrayList<>();
+            for (final NotificationRule rule : result) {
```
Hunroll, line 224: I'd move "if (severities == null || severities.isEmpty())" here (before "for" block) and do severityFiltered=result; in this case, not to enumerate through all the results if severities are not set. Just to skip the whole following block

owingruters:
@emyhrberg and @nscuro can you finalize this PR?

emyhrberg:
> @emyhrberg and @nscuro can you finalize this PR? 

Waiting for @nscuro to respond

PR #4880, opened 2025-04-25
Add suppress and unsuppress for all projects

### LINKED ISSUE #1495: system wide vulnerability management

rufhe513:
### Description

This commit introduces functionality to suppress or unsuppress a vulnerability in all projects where the affected component appears. It streamlines bulk vulnerability management across projects, reducing the need for repetitive actions and ensuring consistent analysis state for shared components.

### Addressed Issue

closes #1495 

### Additional Details

This fix identifies the component to be suppressed across all projects in which it appears. It matches components by name and compares their versions. Only components with a version less than or equal to the selected component’s version will be suppressed.
I have also contributed in frontend

### Checklist


rkg-mm:
Did I miss something, or would this circumvent ACL permissions? If User A suppresses this in Project A with this flag active, this would suppress the finding also in Project B and C. However, If user does only have permissions on A+B, not on C, he still would alter this for project C. This is a no-go :(

I also have the need for something like this, but this needs more detailed planning I think. The even more important use case for me would e.g. be, to add a vuln analysis to all versions of the same project, but not to completely different projects.

src/main/java/org/dependencytrack/resources/v1/AnalysisResource.java
```diff
@@ -170,6 +170,21 @@ public Response updateAnalysis(AnalysisRequest request) {
...
+                        if (c.getName().equals(component.getName()) && c.getVersion().compareTo(component.getVersion()) <= 0) {
+                            matchingComponents.add(c);
+                        }
+                    }
+                }
+            } else {
+                matchingComponents.add(component);
+            }
+
             String commenter = null;
```
valentijnscholten, line 188: This looks inefficient is it will retrieve all components one-by-one. There also seems to be an assumption that the suppression needs to happen for all versions older than or equal to the version of the component selected for the suppression?

valentijnscholten:
Not sure what the future plans of DT are regarding bulk suppression, but maybe it would be better implemented in the Audit Vulnerabilities view.

PR #4894, opened 2025-04-28
Add button to remove SBOM

### LINKED ISSUE #4875: Add button to remove SBOM from a project
### Current Behavior

It is only possible to replace an SBOM via the GUI by uploading a new one. There is no option to remove or clear an existing SBOM from a project through the GUI. This can only be done via the API.

### Proposed Behavior

Add a GUI option to remove an SBOM from a project, not just replace it. This would let users reset a project’s SBOM without using the API.

### Checklist

anttihuovila01:
### Description

 

My change allows users to remove an SBOM completely from a project, and clearing all the details that are part of an SBOM in the frontend (components etc). The removal is done by clicking on a remove SBOM button in Projects/Components, and thereafter the user gets to confirm the removal using a pop up modal. The change is necessary as it makes it possible to clear a project from it's SBOM fully, instead of having to overwrite it with a new SBOM as was the only option until this change.

### Addressed Issue

Closes #4875 

### Additional Details

I found it problematic to refresh the frontend and remove all the components in there. The components still showed up after removing the SBOM with the existing API function and refreshing, so I managed to implement one way around it which was to remove components one batch (currently 50) at a time from the frontend, to fully get them removed from the project.

### Checklist

~- [] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)~


src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -242,6 +243,65 @@ public Response exportComponentAsCycloneDx (
...
+    @PermissionRequired(Permissions.Constants.BOM_UPLOAD) //Same permission as upload
+    public Response deleteBom(
+            @Parameter(description = "The UUID of the project to delete BOM for", required = true)
+            @PathParam("uuid") @ValidUuid String uuid) {
+        try (QueryManager qm = new QueryManager()) {
+            Project project = qm.getObjectByUuid(Project.class, uuid);
+            if (project == null) {
+                return Response.status(Response.Status.NOT_FOUND).entity("Project not found").build();
+            }
+            qm.deleteBoms(project);
```
nscuro, line 271: Deleting BOM records does not delete the components in a project. BOM records currently merely loosely track surface-level metadata about BOM that were uploaded to a project, but they're not directly associated with components, services, etc.

I fear this won't do what you want.

anttihuovila01: I noticed that the components and other related information in the frontend wasn't "refreshing" properly after calling that deleteBoms function alone, so therefore I tried around quite a bit and managed to create this function that also deletes components and refreshes metrics making the BOM-deletion progress more including:

image

This "removeBom" function is located in the code for my frontend PR for this same issue (lines 427-472 in ProjectComponents.vue), here is the link to the frontend PR: https://github.com/DependencyTrack/frontend/pull/1239

valentijnscholten: If you want to clean a project I think this should be handled in the API in a `.../project/uuid/clean` call or similar. But what is the use case for having an empty project?

anttihuovila01: @valentijnscholten That's true, /clean sounds more clean to be fair. In my current implementation as it is, it cleans almost everything, but "Last BOM Import" and "BOM Format" in the home/Projects tab still include the old info, but everything else is removed/reset fully, including all components and services. The purpose is to give users more flexibility, allowing them to clean up or update a project without deleting it entirely. For example, if they're waiting to upload a new SBOM, removing the old one helps clearly mark the project as ready, since an empty project is easier to identify.

Would you like me to adjust the implementation, such as making sure everything is fully cleared and renaming the API endpoint to /clean? Or do you feel this feature isn’t important enough to include in DependencyTrack?

PR #4895, opened 2025-04-28
Add field for timestamp

### LINKED ISSUE #2782: BOM Timestamp Display in UI
### Current Behavior

A BOM that is uploaded to Dependency-Track at time X will result in "Last BOM Import" timestamp (displayed on projects page and also on overview tab for individual projects) with value time X.

The "Last BOM Import" column on the projects page is sortable and this is very useful for quickly seeing which projects in the portfolio have not had a recenet upload.  eg, when a nightly build is failing.

However,  knowing when a BOM was uploaded does not tell you anything at all about when the BOM was generated. The timestamp of the BOM might be months older than "Last BOM Import".

### Proposed Behavior

* Add support for recording BOM timestamp in database.  
* Display BOM timestamp in column on projects page.  The column should be sortable (or it will not really be usable) and should not be displayed by default...  even users who value the data might not want to have it cluttering their UI all the time.

#### Caveats
* Does implementation have to factor in impact of future integration of BOM repository server, where a project might have a history of BOMs, each wth its' own unique timestamp.
* What about merged BOMs, where (say) two BOMs that are a month old get merged today, prior to upload to DT?

#### Notes
This enhancement is inspired by #2774.  It is intended as an MVP for dealing with BOM age.  In broad terms this MVP is easy to implement and requires no design thought about exactly how to rate the risk of an old BOM...  but leaves the door open to later implementation of the linked issue.

### Checklist

anttihuovila01:
### Description

Added a field for showing the timestamp for when an SBOM that has been uploaded to a project was originally generated.  This shows the "age" of the SBOM, which is informative and is a fitting field to have in the Project/Overview window under "last bom import". My code fetches the timestamp from the metadata/timestamp section in a standard CycloneDX SBOM. I modified existing functionality to easily be able to do this as soon as a BOM is uploaded to a project.

### Addressed Issue

Closes #2782 

### Additional Details

Link to original issue: https://github.com/DependencyTrack/dependency-track/issues/2782

### Checklist


src/main/java/org/dependencytrack/model/Project.java
```diff
@@ -305,6 +305,16 @@ public enum FetchGroup {
...
 
+    /**
+     * Convenience field which stores the timestamp of when the BOM was generated (from BOM metadata).
+     */
+    @Persistent
+    @Index(name = "PROJECT_BOM_TIMESTAMP_IDX")
+    @Column(name = "BOM_TIMESTAMP", jdbcType = "BIGINT", allowsNull = "true")
+    @Schema(type = "string", format = "date-time", description = "Timestamp when the BOM was generated (from BOM metadata)")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    private Date bomTimestamp;
```
nscuro, line 316: I think conceptually this would belong to the [`Bom`](https://github.com/DependencyTrack/dependency-track/blob/d8036722708bbd48322e8ff7f96b4498ced15e40/src/main/java/org/dependencytrack/model/Bom.java) entity.

`Bom` is not currently surfaced on the UI, but it should be. Ideally, each project would have a list (i.e. timeline) of `Bom`s that were imported to it. The data is there, it just needs exposure.

PR #5017, opened 2025-06-02
Fix: Handle JSONArray package entries in Composer metadata analysis

ch8matt:
### Description

Some Composer repositories such as Satis serve metadata where the "packages" object contains values as JSONArrays instead of JSONObjects.

This patch enhances ComposerMetaAnalyzer to support both formats:

JSONArray: now correctly expanded into version-keyed JSONObject using expandPackageVersions().
JSONObject: handled as before.
Any other type triggers a MetaAnalyzerException for clarity.
This fixes issues like:
org.json.JSONException: JSONObject[""] is not a JSONObject (class org.json.JSONArray)

Allows Dependency-Track to correctly parse metadata from Satis and similar repositories.

### Checklist


ch8matt:
@valentijnscholten 

Hello,

What do you think of this change ?

valentijnscholten:
In the PR that added support for V2 I tried to cater for this scenario using the `isMinified` function. It looks like Satis is not providing this information in its repo json response? Would it be better to enhance the implementation of the `isMinified` check as that is used in other places as well?

ch8matt:
No it doesn't provide information in its repo json, I made a mock of what its syntax in this file for reference: https://github.com/DependencyTrack/dependency-track/blob/bd25f69ba361262fb1f6801f2a2979f6d5ff1639/src/test/resources/unit/tasks/repositories/https---composer.include.com.metadata-galaxy-cow-arrayentry.json .

Yes I agree with you it would be better to ehance the implementation of the `isMinified` :)

ch8matt:
https://github.com/DependencyTrack/dependency-track/pull/5019 I updated appropriately hopefully there, I am closing this one.

PR #5019, opened 2025-06-03
Improve Composer meta analyzer's ability to deal with minified metadata

ch8matt:
### Description

Some Composer repositories (e.g., Satis) do not include the "minified" key in their packages.json,
even though they return packages in minified format (arrays instead of versioned objects).

This commit enhances the isMinified() method to heuristically detect minified metadata by inspecting
the structure of the packages object. This improves compatibility with such repositories and avoids
JSON parsing issues during metadata analysis.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
+++ b/src/main/java/org/dependencytrack/tasks/repositories/ComposerMetaAnalyzer.java
@@ -170,16 +170,7 @@ public class ComposerMetaAnalyzer extends AbstractMetaAnalyzer {
                     LOGGER.debug("%s: package not found in repository %s.".formatted(component.getPurl(), this.repositoryId));
                     return meta;
                 }
-            
-                Object packageEntry = packages.get(getComposerPackageName(component));
-                JSONObject packageVersions;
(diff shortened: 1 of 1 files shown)


ch8matt:
@valentijnscholten Can you confirm this is okay :) ?

valentijnscholten:
Looking at the source code of `composer` itself, it doesn't do this fallback check in this PR:

https://github.com/composer/composer/blob/8d239c7d3d6b7cff416f4ce3546a0b2b7633cb30/src/Composer/Repository/ComposerRepository.php#L1048C92-L1048C96

And the `expand` method is not called anywehere else. But yet `composer` does work with `Satis` repositories it seems. I have acccess to 1 private `Satis` based repository and I can see also that the minified "header" is missing.

Summarizing I don't know why we need this PR but it does solve the problem and would be safe to merge.

ch8matt:
@nscuro Hello, could you tell me what you think about this PR ? Would love to have the feature fixed in the future version of Dependency-Track :D

nscuro (approved):
Thanks @ch8matt! I wasn't sure if you were more clarification from others but the PR looks good to me.

PR #5026, opened 2025-06-05
feat(analyzer): add option to ignore severity reported by Trivy scan

arjavdongaonkar:
### Description

* A new flag in the UI under:
  `Administration → Analyzer → Trivy`
  Labeled as **"Ignore severity reported by Trivy scan"**
* This flag will be enabled by default (i.e., severity is ignored).
* When disabled, the analyzer **overwrites the existing severity of a vulnerability** with the value reported by Trivy

### Checklist


USer78012:
Hello,
This improvement is really interesting for preserving the severities coming from Trivy. 
Have you abandoned its integration into Dependency Track?

arjavdongaonkar:
> Hello, This improvement is really interesting for preserving the severities coming from Trivy. Have you abandoned its integration into Dependency Track?

As duely stated by @stohrendorf , What if trivy reports multiple severities? Let's say it reports a vulnerability in curl. You then let it scan two docker BOMs - one from a debian based image, one from an alpine based image. Since trivy uses distro-specific databases for severities, you now could get two different severities for the same vulnerability. I don't see how that could work with your proposed changes.

To address this, I’ve implemented a workaround: since vulnerability severity can vary by package version and distro, I use the unique package URL (purl) from Trivy’s reported vulnerabilities. In Dependency Track, I create new findings with the finding ID as the MD5 hash of the purl. This allows me to maintain a distinct set of vulnerabilities without overwriting the severity of CVEs reported by Dependency Track or other sources.

I have implemented this as its working as expected for my org. If you’re interested, I can prepare a PR with this approach.

USer78012:
Thank you for your response.
Yes I am interested.

arjavdongaonkar:
@USer78012 
Here is a simple PR that will change the vuln ID to MD5 of pURL
https://github.com/DependencyTrack/dependency-track/pull/5203

PR #5066, opened 2025-06-18
feat: support configurable match mode for internal component regex (AND/OR)

ch8matt:
### Description

Adds support for configuring how internal component regex filters are applied.

- Introduced a new config property:
  `internal-components.match-mode` (`OR` by default, accepts `AND`)
- When set to `AND`, a component must match **both** the group and name regex to be considered internal.
- When `OR` (default), a match on either field suffices.

### Checklist


ch8matt:
For the frontend: https://github.com/DependencyTrack/frontend/pull/1283

src/main/java/org/dependencytrack/util/InternalComponentIdentifier.java
```diff
@@ -76,7 +89,14 @@ public boolean isInternal(final Component component) {
             matchesName = false;
         }
 
-        return matchesGroup || matchesName;
+        final String matchMode = matchModeSupplier.get();
+        if ("AND".equals(matchMode)) {
+            final boolean groupOk = patterns.groupPattern() == null || matchesGroup;
+            final boolean nameOk = patterns.namePattern() == null || matchesName;
+            return groupOk && nameOk;
```
nscuro, line 96: Why not just `matchesGroup && matchesName`?

src/main/java/org/dependencytrack/util/InternalComponentIdentifier.java
```diff
@@ -76,7 +89,14 @@ public boolean isInternal(final Component component) {
             matchesName = false;
         }
 
-        return matchesGroup || matchesName;
+        final String matchMode = matchModeSupplier.get();
+        if ("AND".equals(matchMode)) {
```
nscuro, line 93: Please also add a [test](https://github.com/DependencyTrack/dependency-track/blob/b4fa5ce1b6372055bc7e9b1eb4823e2d57bf2bb9/src/test/java/org/dependencytrack/util/InternalComponentIdentificationUtilTest.java) or two that covers this new match mode.

src/main/java/org/dependencytrack/util/InternalComponentIdentifier.java
ch8matt: So that the absence of one regex doesn’t block the match

nscuro (approved):

ch8matt:
Hello, do you know why it failed ? https://github.com/DependencyTrack/dependency-track/actions/runs/16002835392/job/45141990386

PR #5099, opened 2025-07-03
added Vulert into README.md

ankitdn:
Added Vulert into Readme


nscuro:
Vulert is not a supported scanner.

Issue #5127, opened 2025-07-22
Tags are not updated on subsequent BOM uploads


henriduflot:
### Current Behavior

I've encountered what appears to be a bug with the tag management when uploading BOMs via the API. When using the PUT `/api/v1/bom` endpoint with the `projectTags `parameter, tags are correctly applied on initial project creation but are not updated on subsequent uploads.

Seems the same as https://github.com/DependencyTrack/dependency-track/issues/4539 but on `/api/v1/bom`  API.

### Steps to Reproduce

Make an initial API call to PUT /api/v1/bom with the following parameters:

```
{
  "projectName": "subProject",
  "projectVersion": "prod",
  "parentName": "parentProject",
  "projectTags": ["tag1"],
  "autoCreate": true,
  "bom": "ewogICJib21Gb3JtYXQiOiAiQ3ljbG9uZURYIiwKICAic3BlY1ZlcnNpb24iOiAiMS40IiwKICAiY29tcG9uZW50cyI6IFsKICAgIHsKICAgICAgInR5cGUiOiAibGlicmFyeSIsCiAgICAgICJuYW1lIjogImFjbWUtbGliIiwKICAgICAgInZlcnNpb24iOiAiMS4wLjAiCiAgICB9CiAgXQp9"
}
```

2. Verify that the project is created with the tag "tag1"
3. Make a second API call with modified tags:

```
{
  "projectName": "subProject",
  "projectVersion": "prod",
  "parentName": "parentProject",
  "projectTags": ["tag1", "tag2"],
  "autoCreate": true,
  "bom": "ewogICJib21Gb3JtYXQiOiAiQ3ljbG9uZURYIiwKICAic3BlY1ZlcnNpb24iOiAiMS40IiwKICAiY29tcG9uZW50cyI6IFsKICAgIHsKICAgICAgInR5cGUiOiAibGlicmFyeSIsCiAgICAgICJuYW1lIjogImFjbWUtbGliIiwKICAgICAgInZlcnNpb24iOiAiMS4wLjAiCiAgICB9CiAgXQp9"
}
```

4. Observe that the project tags remain unchanged (still only "tag1")

### Expected Behavior

When uploading a BOM with updated tags, the project's tags should be updated to match the new set provided in the API call.

### Dependency-Track Version

4.13.0

### Dependency-Track Distribution

Executable WAR, Container Image

### Database Server

PostgreSQL

### Database Server Version

14.5

### Browser

Mozilla Firefox

[...]

PR #5140, opened 2025-07-25
Ensure PROJECT_CREATION_UPLOAD same behaviour for tagging when autocr…

snieguu:
### Description
This PR fixes an authorization inconsistency regarding tag handling for users with the `PROJECT_CREATION_UPLOAD` permission during project auto-creation/update for `PUT/POST` `/v1/bom` resource

Users with only the PROJECT_CREATION_UPLOAD permission were previously able to:

Update tags when bom upload, even without PORTFOLIO_MANAGEMENT permission and could not do the same for existing projects.

This led to inconsistent permission boundaries

Tag creation or updates are now entirely blocked for users who only have the `PROJECT_CREATION_UPLOAD` permission - regardless of whether the project is being created or updated.

Proper tag handling (both creation and modification) now consistently requires the PORTFOLIO_MANAGEMENT permission.

### Addressed Issue

#5127

### Checklist


snieguu:
@nscuro Should I add/fix anything else here or this can be merged ?

Issue #5158, opened 2025-07-30
Policy condition PURL IS_NOT will match a substring


jakub-bochenski:
### Current Behavior

Define a policy with condition:
```
{
	"subject": "PACKAGE_URL",
	"operator": "IS_NOT",
	"value": "pkg:maven/com.mysql/mysql-connector-j@8.0.33"
}
```
Upload a bom:
```
{
	"bomFormat": "CycloneDX",
	"specVersion": "1.5",
	"version": 1,
	"components": [
	   {
	      "type" : "library",
	      "group" : "com.mysql",
	      "name" : "mysql-connector-j",
	      "version" : "8.0.33",
	      "purl" : "pkg:maven/com.mysql/mysql-connector-j@8.0.33?type=jar"
	    }
	]
}
```

Observe that the policy is not triggered.

### Steps to Reproduce

1.

### Expected Behavior

I don't think the behaviour is documented anywhere. 
Yet I think it's reasonable to expect `IS`/`IS_NOT` operator requires equality, not being a substring.

### Dependency-Track Version

4.13.2

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Browser

N/A

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

Issue #5142, opened 2025-07-25
PURL NO_MATCH policy condition doesn't match by ?type


jakub-bochenski:
### Current Behavior

Setup a policy with condition
```
{
	"subject": "PACKAGE_URL",
	"operator": "NO_MATCH",
	"value": "pkg:maven/com.mysql/mysql-connector-j@8.0.33?type=jar"
}
```

Upload a BOM:
```
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.5",
  "version": 1,
  "components": [
     {
        "type" : "library",
        "group" : "com.mysql",
        "name" : "mysql-connector-j",
        "version" : "8.0.33",
        "purl" : "pkg:maven/com.mysql/mysql-connector-j@8.0.33?type=jar"
      }
  ]
}
```

Observe that the policy is triggered.

If you define a policy with condition (notice no ?type)
```
{
	"subject": "PACKAGE_URL",
	"operator": "NO_MATCH",
	"value": "pkg:maven/com.mysql/mysql-connector-j@8.0.33"
}
```

Then the policy will not trigger for the above BOM

### Expected Behavior

PURL policy condition should match a component with exactly the same PURL

### Dependency-Track Version

4.13.2

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Browser

N/A

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

stohrendorf:
The policy condition values are treated as regular expressions, try escaping the `?`, so that your expression does not only match `@8.0.33type=jar` or `@8.0.3type=jar` (notice how `?` gets removed).

jakub-bochenski:
thanks, that explains the behavior

it would be nice if this was documented, I don't think it's described anywhere now?

Issue #4444, opened 2024-12-06
Policy Condition for Direct Components


forsakenwalkatnight:
### Current Behavior

In the policy management we can add coordinates as condition. While coordinates can be specified with a group, name, and version, an _exact_ match is not provided.

Example: Component shown as belonging to group "org.keycloak" is actually part of "org.keycloak.bom" group and shows up as a result in the alert. This results in reporting on 70 components instead of 34 direct ones.

![Image](https://github.com/user-attachments/assets/a54c51a6-e0a1-4704-b412-3da8155a720a)

### Proposed Behavior

Add a new condition to the policy management for direct components only or allow for exact matching to avoid more results than intended.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

stohrendorf:
Unfortunately, dependency de-duplication prevents a proper solution as of right now. For example, importing an SBOM created by trivy with a duplicated component (e.g. a JAR found at 2 places) basically leads to random behaviour what component is picked up by DT (the reason for that is a bit complicated, so I will not explain here). Because of that, a dependency could randomly change between considered being a direct and an indirect dependency. This, in turn, would mean that this undefined behaviour would bleed into the policy engine.

Issue #2673, opened 2023-04-15
Support Common Expression Language (CEL) in policy conditions


nscuro:
### Current Behavior

Policy evaluation today is based on conditions that are evaluated one-by-one, for every component.

Evaluations of conditions do not have a shared context. This means that for certain types of conditions, it is not possible to chain them in a conjunctive way. This is most noticeable for policies that target vulnerability details.

For example, given the policy operator `ALL`, and the following policy conditions:

* CWE `CONTAINS_ANY` `666`
* Severity `IS` `HIGH`

Policy violations will be will be raised, if:

* *Any* vulnerability affecting the component has the CWE `666`, and
* *Any* vulnerability affecting the component has the severity `HIGH`

Whereas in reality, users may want to only receive a violation when the component is affected by a vulnerability that both has the CWE `666`, **and** a `HIGH` severity.

### Proposed Behavior

Instead of introducing a shared context for policy conditions, or complex abstractions and UI elements for such cases, I propose to support expressions instead.

And instead of embedding scripting languages and having to deal with all the security implications that come with that, I propose to use languages that are purpose-built for this kind of thing. A popular choice appears to be the [Common Expression Language (CEL)](https://github.com/google/cel-spec). Java bindings [are available](https://github.com/google/cel-java).

The basic idea would be to define an object model that is passed as input to the CEL evaluator. Users would then be able to provide their conditions as CEL expressions. The policy mentioned above may be implemented like so:

```
vulnerabilities.exists(v, 666 in v.cwes && (v.severity == "HIGH" || v.severity == "CRITICAL"))
```
*(See [macros](https://github.com/google/cel-spec/blob/master/doc/langdef.md#macros))*

> **Note**  
> We should *not* reuse the existing model classes (`org.dependencytrack.model`) as inputs, as they hold internal fields and may cause unintended side effects, like database interactions, due to them being bound to the ORM. CEL inputs should be clearly defined and limited to *useful* fields, so that users know what they can expect as input.

Another option we could consider is integration with [Open Policy Agent](https://www.openpolicyagent.org/).
While CEL can be embedded as library, OPA would be an external application that users need to run. Evaluating conditions will thus also involve network calls, causing a noticeable performance hit.

Supporting an expression language will make the policy engine much more flexible, extensible, and more useful. 

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
[...]

PR #5160, opened 2025-07-30
Fix inverted component matching

### LINKED ISSUE #5109: policy violation not working as expected
### Current Behavior

Hello everyone,

I'm currently facing an issue where the policy violation rule set to trigger when "coordinate does not match" in Dependency-Track does not seem to work as intended. Despite configuring the rule with this condition, violations are not being detected or flagged.

Has anyone experienced similar behavior? Could there be specific reasons why the "coordinate does not match" condition isn't functioning properly?
Best Regards, 

### Steps to Reproduce

1.configuring a policy violation rule with this condition " does not match"

### Expected Behavior

policy violations are detected. 

### Dependency-Track Version

4.11.x

Executable WAR

### Database Server

PostgreSQL

### Database Server Version

_No response_

### Browser

Google Chrome

### Checklist

### LINKED ISSUE #5135: Coordinates policy condition doesn't work if component has blank group
### Current Behavior

I expect this policy should trigger for every component that is not foo/bar@baz

Instead the policy never triggers

Image

I've tried different variations ANY/ALL, MATCHES or NO_MATCH -- it seems that a policy containing the Coordinates condition never fires

### Dependency-Track Version

4.13.2

Container Image

### Database Server

PostgreSQL

### Database Server Version

_No response_

### Browser

N/A

### Checklist

### LINKED ISSUE #5136: Coordinates NO_MATCH policy condition will match a component that has either group, name or version
### Current Behavior

Define a policy with condition
```
{
  "subject": "COORDINATES",
  "operator": "NO_MATCH",
  "value": "{\"group\":\"foo\",\"name\":\"component-one\",\"version\":\"baz\"}"
}
```

upload a BOM:
```
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.5",
  "version": 1,
  "components": [
    {
      "type": "library",
      "group": "example-one",
      "name": "component-one",
      "version": "1.0.0",
      "purl": "pkg:maven/example-one/component-one@1.0.0"
    },
    {
      "type": "library",
      "group": "example-two",
      "name": "component-two",
      "version": "2.0.0",
      "purl": "pkg:maven/example-two/component-two@2.0.0"
    }
  ]
}
```

Policy violation is only triggered for `component-two`, even though the coordinates specify a different group and version than `component-one` has.

The same happens if a component has matching group or version

### Expected Behavior

Current behavior is surprising, if I specify coordinates I expect all segments to match.

### Dependency-Track Version

4.13.2

Container Image

### Database Server

PostgreSQL

### Database Server Version

_No response_

### Browser

N/A

### Checklist

stohrendorf:
### Description

Fixes wrong negative matches with coordinate policies. Also adds a FAQ section about common policy value issues.

### Addressed Issue

Related to #5158 #5142 #4444 (I think it's possible that this PR is enough to consider these issues being fixed)
Fixes #5136
Fixes #5135
Fixes #5109

### Additional Details

Introduces some minor breaking changes for the `NO_MATCH` operator, but should work exactly the same otherwise, at least according to the unit tests. I wonder where to note that this changes behaviour and will introduce (probably) unexpected changes in some cases.

### Checklist

### DIFF
--- a/docs/_docs/FAQ.md
+++ b/docs/_docs/FAQ.md
@@ -110,4 +110,16 @@ nginx.ingress.kubernetes.io/proxy-body-size: "100m"
 
 Please consult the [official documentation](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size)
 
+#### Policy conditions do not work for some PURLs
+
+Policy condition values are treated as regular expressions.
+
+1. Policy condition values are implicitly treated as substring matches.
+   They must be explicitly anchored with `^` and `$` to make them an exact match.
+2. Characters with special meaning in regular expressions should be escaped with a `\\`.
+   This is needed if the PURL contains a `?`, since the question mark makes the previous character optional and is not treated literally.
+   Another special character is `.`, which should also be escaped.
+3. Policy condition values support wildcards, so an `*` means that any text is allowed, including missing text.
+   For example, `^vendor/*$` would match `vendor/lib-1`, `vendor/app`, or even only `vendor/`.
+
 [defect report]: https://github.com/DependencyTrack/dependency-track/issues/new?assignees=&labels=defect%2Cin+triage&template=defect-report.yml
--- a/src/main/java/org/dependencytrack/policy/CoordinatesPolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/CoordinatesPolicyEvaluator.java
@@ -60,31 +60,38 @@ public class CoordinatesPolicyEvaluator extends AbstractPolicyEvaluator {
         for (final PolicyCondition condition : super.extractSupportedConditions(policy)) {
             LOGGER.debug("Evaluating component (" + component.getUuid() + ") against policy condition (" + condition.getUuid() + ")");
             final Coordinates coordinates = parseCoordinatesDefinition(condition);
-            if (matches(condition.getOperator(), coordinates.getGroup(), component.getGroup())
-                    && matches(condition.getOperator(), coordinates.getName(), component.getName())
-                    && versionMatches(condition.getOperator(), coordinates.getVersion(), component.getVersion())) {
-                violations.add(new PolicyConditionViolation(condition, component));
+            final boolean positiveMatch = matches(coordinates.getGroup(), component.getGroup())
+                    && matches(coordinates.getName(), component.getName())
+                    && versionMatches(coordinates.getVersion(), component.getVersion());
+            switch (condition.getOperator()) {
+                case MATCHES -> {
+                    if (positiveMatch) {
+                        violations.add(new PolicyConditionViolation(condition, component));
+                    }
+                }
+                case NO_MATCH -> {
+                    if (!positiveMatch) {
+                        violations.add(new PolicyConditionViolation(condition, component));
+                    }
+                }
+                default -> {
+                    // silently swallow this
+                }
             }
         }
         return violations;
     }
 
-    private boolean matches(final PolicyCondition.Operator operator, final String conditionValue, final String part) {
-        if (conditionValue == null && part == null) {
-            return true;
+    private boolean matches(final String conditionValue, final String part) {
+        if (part == null) {
+            return conditionValue == null;
         }
+
         final String p = StringUtils.trimToNull(part);
-        if (p != null) {
-            if (PolicyCondition.Operator.MATCHES == operator) {
-                return org.dependencytrack.policy.Matcher.matches(p, conditionValue);
-            } else if (PolicyCondition.Operator.NO_MATCH == operator) {
-                return !org.dependencytrack.policy.Matcher.matches(p, conditionValue);
-            }
-        }
-        return false;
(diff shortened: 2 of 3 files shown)


docs/_docs/FAQ.md
```diff
@@ -110,4 +110,16 @@ nginx.ingress.kubernetes.io/proxy-body-size: "100m"
...
+#### Policy conditions do not work for some PURLs
+
+Policy condition values are treated as regular expressions.
+
+1. Policy condition values are implicitly treated as substring matches.
+   They must be explicitly anchored with `^` and `$` to make them an exact match.
+2. Characters with special meaning in regular expressions should be escaped with a `\\`.
+   This is needed if the PURL contains a `?`, since the question mark makes the previous character optional and is not treated literally.
+   Another special character is `.`, which should also be escaped.
+3. Policy condition values support wildcards, so an `*` means that any text is allowed, including missing text.
```
jakub-bochenski, line 122: This is confusing - I don't understand how policy conditions would support both regex and `*` wildcard at the same time?

Is it that some policy conditions use regexp and some support `*` wildcard?

Also I'm guessing that `IS`/`IS_NOT` will be simple equality now?

If I'm following things correctly I think it would be best to enumerate driectly for each operator if it's regex, `*` or simple equality etc.

stohrendorf: `*` is silently replaced with `.*`, and then `..*` is silently replaced with `.*`, I didn't want to explain _that_ logic.

jakub-bochenski: well, for me it would be easier to understand if it was explained there :)

also this could cause unexpected behavior `..*` is one or more characters, someone could use that instead of `.+` and would have no idea why the regex isn't working

docs/_docs/FAQ.md
```diff
@@ -110,4 +110,16 @@ nginx.ingress.kubernetes.io/proxy-body-size: "100m"
 
 Please consult the [official documentation](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size)
 
+#### Policy conditions do not work for some PURLs
+
+Policy condition values are treated as regular expressions.
+
+1. Policy condition values are implicitly treated as substring matches.
+   They must be explicitly anchored with `^` and `$` to make them an exact match.
+2. Characters with special meaning in regular expressions should be escaped with a `\\`.
```
jakub-bochenski, line 119: I think in actual regex the escape is just `\` 
It would be `\\` in Json because it has to be escaped there again, or am I misunderstanding something?

jakub-bochenski: This condition works for me:
```
{
  "subject": "PACKAGE_URL",
  "operator": "NO_MATCH",
  "value": "^pkg:maven/com.mysql/mysql-connector-j@8.0.33\\?type=jar$"
}
```
(Literal value in json is `^pkg:maven/com.mysql/mysql-connector-j@8.0.33\?type=jar$`)
This doesn't
```
{
  "subject": "PACKAGE_URL",
  "operator": "NO_MATCH",
  "value": "^pkg:maven/com.mysql/mysql-connector-j@8.0.33\\\\?type=jar$"
}
```
(Literal value in json is `^pkg:maven/com.mysql/mysql-connector-j@8.0.33\\?type=jar$`)

stohrendorf: It was an assumption of me that one backslash would escape the backtick, which is not true for GHFM, but according to [this](https://talk.jekyllrb.com/t/backslash-treatment-inside-backquotes-is-different-between-jekyll-and-markdown/8537), it's actually needed for Jekyll.

docs/_docs/FAQ.md
stohrendorf: I would say that using `..*` instead of `.+` is _very_ unlikely. I totally agree this is confusing, but I really don't know how to describe it in a way that is not too confusing, since I believe these conditions are created by inexperienced _and_ experienced people. Also, this is a description for a FAQ, which should give short answers to short questions. A more detailed description would need to be placed somewhere else IMHO, but I don't know where.

jakub-bochenski: > but I don't know where.

One thing I like to do in my projects is to document behavior with automated tests.
There is https://github.com/DependencyTrack/dependency-track/blob/master/src/test/java/org/dependencytrack/policy/MatcherTest.java but IMO it would need to be written in "test are specs" way to be useful as documentation. 

> I would say that using ..* instead of .+ is very unlikely. 

In Java no, but if you often use basic `grep` you might have a habit of only using `[]*.`

Also wouldn't this break any construct like `A*` (0 or more A characters) -> `A.*` (A and 0 or more of any character).

I'm not arguing this logic should change now, just pointing out some problems.

stohrendorf: I agree there is some stuff to discuss, but this is not the correct place to discuss the issues, but rather improving the docs to address common issues without getting too detailed.

nscuro: CEL expressions (#2673, ship with v5) will give users more control over the type of matching being performed.

nscuro (approved):
Thanks!

PR #5192, opened 2025-08-07
CompressUtil: handle ArchiveException

aborrero:
Fixes this build problem:

[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /home/build/src/main/java/org/dependencytrack/util/CompressUtil.java:[43,96]
  unreported exception org.apache.commons.compress.archivers.ArchiveException; must be caught or declared to be thrown
[INFO] 1 error

Signed-off-by: Arturo Borrero Gonzalez <arturo.borrero@chainguard.dev>

### DIFF
--- a/src/main/java/org/dependencytrack/util/CompressUtil.java
+++ b/src/main/java/org/dependencytrack/util/CompressUtil.java
@@ -45,7 +45,7 @@ public final class CompressUtil {
             if (ais.canReadEntryData(entry)) {
                 return IOUtils.toByteArray(ais);
             }
-        } catch (IOException e) {
+        } catch (IOException | org.apache.commons.compress.archivers.ArchiveException e) {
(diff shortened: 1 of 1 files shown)


nscuro:
`throws ArchiveException` was removed as part of the commons-compress update to 1.28.0: https://github.com/DependencyTrack/dependency-track/pull/5156. Our builds in CI work just fine, did you modify any dependencies in your environment perhaps?

nscuro:
As evident by the CI builds failing with the patch applied, this is not an issue for us. I suspect the fact [you're bumping dependency versions](https://github.com/wolfi-dev/os/blob/8c156dd8005474c8d229b8f8193129785bab8ed0/dependency-track.yaml#L36) in your build is causing problems.

PR #5197, opened 2025-08-10
Update OssIndexAnalysisTask.java

zer0uno:
Removed workaround for OssIndexAnalysis when component had "v" as version prefix

### Description

I've noticed that calling the OSS Index API `https://ossindex.sonatype.org/api/v3/component-report` with `pkg:composer/symfony/validator@v3.3.18` correctly returns the vulnerability CVE-2024-50343, while calling the API with `pkg:composer/symfony/validator@3.3.18` (note the absence of `v`) I get back no vulnerability.
In the `OssIndexAnalysisTask` class I found that the `minimizePurl()` method removes the `v` just before calling the API, apparently because previously there was a bug (as explained in the comment block above the function definition) that now seems solved.

### Addressed Issue

https://github.com/DependencyTrack/dependency-track/pull/1220

### Checklist

- [X] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java
+++ b/src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java
@@ -247,13 +247,6 @@ public class OssIndexAnalysisTask extends BaseComponentAnalyzerTask implements C
      * Therefore, this method will return a String representation of a PackageURL without qualifier
      * or subpath.
      * <p>
-     * Additionally, as of October 2021, versions prefixed with "v" (as commonly done in the Go and PHP ecosystems)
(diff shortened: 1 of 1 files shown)


agnieszka-docplanner:
Hi, any progress on this? There is a lot of components with "v" in version and they are currently not being properly analyzed.

nscuro:
Here's a related Slack discussion: https://owasp.slack.com/archives/C6R3R32H4/p1754692145518059

I am not confident removing this logic for all ecosystems will yield the desired result, as in some cases adding the `v` will cause *all* vulnerabilities *across all versions* to be returned.

PR #4998, opened 2025-05-27
feat: add AttributedOnPolicyEvaluator 

arjavdongaonkar:
add AttributedOnPolicyEvaluator for vulnerability age-based policy evaluation

Implements production-ready evaluator with caching, error handling, and comprehensive logging. Supports ISO-8601 period formats with injectable dependencies for testing.

### Checklist

- [x] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)
- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective
- [x] This PR implements an enhancement, and I have provided tests to verify that it works as intended
- [ ] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)
- [x] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly

PR #5202, opened 2025-08-12
feat(policy): add PatchedVersionPolicyEvaluator and related tests

### LINKED ISSUE #5731: Add Policy Evaluator for Patched Version Availability (Patchable vs Non-Patchable)
### Current Behavior

Dependency-Track policies cannot evaluate vulnerabilities based on whether a patched version exists.

As a result, policies/SLA checks treat all vulnerabilities similarly, even when no upgrade/patch is available.

### Proposed Behavior

Add a new policy evaluator that checks whether vulnerabilities associated with a component have a patched version available.

Policies should be able to trigger when:

1. Patched version is present, or
2. Patched version is not present

based on configured condition (IS / IS_NOT) and expected value (true / false).

### Checklist

arjavdongaonkar:
### Description
Added a **Patched Version Policy Evaluator** that checks if vulnerabilities in a component have a patched version available, and compares this against policy conditions.
This allows policies to trigger when:

* A patched version **is** present (or)
* A patched version **is not** present
  based on the configured condition (`IS` or `IS_NOT`) and expected value (`true` / `false`).

### Addressed Issue

Enables policy checks for vulnerabilities based on the presence or absence of patched versions. Closes https://github.com/DependencyTrack/dependency-track/issues/5731

### Additional Details

* The evaluator looks at all vulnerabilities for a component.
* It treats any non-empty `patchedVersions` field as "patched version available".
* Invalid policy values for this subject (not `true` or `false`) are ignored with a warning in logs.
* Adds flexibility to enforce upgrade-related security policies.

FE PR: https://github.com/DependencyTrack/frontend/pull/1305
### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/PolicyCondition.java
+++ b/src/main/java/org/dependencytrack/model/PolicyCondition.java
@@ -90,7 +90,8 @@ public class PolicyCondition implements Serializable {
         CWE,
         VULNERABILITY_ID,
         VERSION_DISTANCE,
-        EPSS
+        EPSS,
+        PATCH_VERSION
     }
 
     public enum FetchGroup {
--- a/src/main/java/org/dependencytrack/policy/PatchedVersionPolicyEvaluator.java
+++ b/src/main/java/org/dependencytrack/policy/PatchedVersionPolicyEvaluator.java
@@ -1,0 +1,73 @@
+/*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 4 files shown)


nscuro (commented):
The problem with the `patchedVersions` field is that no source populates it (I believe it was originally added for RetireJS, but that analyzer no longer exists).

It's possible that the Trivy analyzer may populate the field, but that also only happens when it *creates* a vulnerability. It won't do that when the vulnerability already existed.

If we merge this in, I fear users will rightfully expect this feature to work across the board, only to realize it doesn't.

arjavdongaonkar:
Same as #4998, the PolicyEvaluator extension point this plugs into is gone on v5, the CEL engine replaced it. Closing this one too.

PR #5224, opened 2025-08-24
Add Support for CycloneDX Scope Data

### LINKED ISSUE #4647: Feature Request: Support for CycloneDX Scope Data
### Current Behavior

**Summary**
CycloneDX provides component scope data ( https://cyclonedx.org/docs/1.6/json/#components_items_scope ), which indicates whether a component is required, optional or excluded. This data can be used to classify development and other optional dependencies, enhancing transparency while allowing them to be segregated from other scopes.

**Problem Statement**
Currently, Dependency-Track does not utilise scope data from CycloneDX SBOMs. This omission may lead users to generate SBOMs that exclude optional dependencies, potentially creating a false sense of security and reducing overall transparency.

### Proposed Behavior

**Proposed Solution**

MVP Implementation
- Store component scope data from CycloneDX SBOMs.
- Display scope information in the UI to help users make informed decisions.

Future Enhancements

- Integrate scope data into policy enforcement and notifications, allowing users to leverage it for advanced use cases.

### Checklist

anantk24:
### Description

I have added column component table as scope column . While publishing bom I updating component to have state from cyclone dx component. Similarly for fetch request mapping it with component identity. 

### Addressed Issue
Addresssing backend part. Updating front end in different pull request.
fixes #4647 

### Additional Details

- How to process existing uploaded BOM which will not have scope data populated?
- Can we improve on fetching finding based on index resulted query? this can introduce bug if someone is missing index. I am refering to Finding .java

### Checklist


src/main/java/org/dependencytrack/model/Component.java
```diff
@@ -183,6 +183,13 @@ public enum FetchGroup {
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The version may only contain printable characters")
     private String version;
 
+    @Persistent
+    @Column(name = "SCOPE", jdbcType = "VARCHAR")
+    @Index(name = "COMPONENT_SCOPE_IDX")
+    @NotNull
+    @Extension(vendorName = "datanucleus", key = "enum-check-constraint", value = "true")
+    private org.cyclonedx.model.Component.Scope scope;
```
nscuro: Let's define our own `Scope` enum so we're not binding our data model to a class controlled by a 3rd party library.

Also please remove the `@Extension` annotation. Unfortunately it causes a lot of issues when new enum values are added later. The ORM only *creates* the database constraint, but never updates them again.

src/main/java/org/dependencytrack/upgrade/v4140/v4140Updater.java
nscuro, line 1: This migration is not necessary - the ORM will add the new column automatically.

anantk24: Thanks. I was curious on that part, due to PR checklist Added those changes. I will revert back those.

src/main/java/org/dependencytrack/model/Component.java
anantk24: Thanks for clarification.  I will work on that.

anantk24:
@nscuro  Thank you for your earlier review and feedback on this PR. I’ve implemented the suggested changes and updated the code accordingly. When you get a chance, could you please take another look?

I really appreciate your time and guidance in helping improve this contribution.

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
```diff
@@ -91,7 +99,7 @@ public class ModelConverter {
     private ModelConverter() {
     }
 
-    public static ProjectMetadata convertToProjectMetadata(final org.cyclonedx.model.Metadata cdxMetadata) {
+    public static ProjectMetadata convertToProjectMetadata(final Metadata cdxMetadata) {
```
nscuro: Please keep the fully qualified class names here. Since some class names are shared between the CycloneDX library and internal models of Dependency-Track, the qualified names make it easier to differentiate the two.

anantk24:
@nscuro I've addressed the previously failing test cases, everything should be successful now :) .  When you get a chance, could you please review the changes and re-run the pending workflow?
Appreciate your time and support! Thanks!

src/main/java/org/dependencytrack/model/Component.java
```diff
@@ -183,6 +183,12 @@ public enum FetchGroup {
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The version may only contain printable characters")
     private String version;
 
+    @Persistent
+    @Column(name = "SCOPE", jdbcType = "VARCHAR")
+    @Index(name = "COMPONENT_SCOPE_IDX")
+    @NotNull
```
nscuro: I think marking this as `@NotNull` would be a breaking change in the REST API. Clients developed against earlier versions will not provide this field. 

Let's remove that annotation.

```suggestion
```

src/main/java/org/dependencytrack/model/Component.java
```diff
@@ -183,6 +183,12 @@ public enum FetchGroup {
     @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS, message = "The version may only contain printable characters")
     private String version;
 
+    @Persistent
+    @Column(name = "SCOPE", jdbcType = "VARCHAR")
```
nscuro: Let's specify a max length here:

```suggestion
@Column(name = "SCOPE", jdbcType = "VARCHAR", length = 255)
@Size(max = 255)
```

src/main/java/org/dependencytrack/model/Scope.java
```diff
@@ -0,0 +1,51 @@
...
+ * Scope would be deriving from different SBOM provider.
+ * * <a href="https://cyclonedx.org/docs/1.6/json/#components_items_scope">Cyclondx Reference</a>
+ * * @author Anant Kurapati
+ * @since 4.14.0
+ */
+public enum Scope {
+    REQUIRED("Required"),
+    OPTIONAL("Optional"),
+    EXCLUDED("Excluded"),
+    EMPTY("Empty");
```
nscuro: Empty is not a valid scope value, I think simply using `null` when no scope data is available would be more straightforward.

src/main/java/org/dependencytrack/model/Scope.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private final String label;
+
+    Scope(String label) {
+        this.label = label;
+    }
+
+    public String getLabel() {
+        return label;
+    }
```
nscuro: Label doesn't appear to be used anywhere?

src/main/java/org/dependencytrack/model/Scope.java
```diff
@@ -0,0 +1,51 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.model;
+import org.cyclonedx.model.Component;
+/**
+ * Enum class for tracking individual components scope.
+ * Scope would be deriving from different SBOM provider.
+ * * <a href="https://cyclonedx.org/docs/1.6/json/#components_items_scope">Cyclondx Reference</a>
+ * * @author Anant Kurapati
```
nscuro: Use `@see` tags for link references and remove duplicate `*`:

```suggestion
 * @author Anant Kurapati
 * @see <a href="https://cyclonedx.org/docs/1.6/json/#components_items_scope">Cyclondx Reference</a>
```

Also can you run a formatter over this file please, so line breaks etc. are added.

src/main/java/org/dependencytrack/model/Scope.java
anantk24: Purpose was out in response for now skipped it.

anantk24:
@nscuro  I have made the requested changes based on your feedback, including code formatting, scope handling adjustments, and updates to annotations. The branch has been rebased and pushed with the latest corrections.

Could you please review the PR once again at your convenience?

Thank you for your time and guidance.

PR #5280, opened 2025-09-05
make CPE matching case-insensitive

### LINKED ISSUE #4663: CPEs are case-sensitive
### Current Behavior

Currently when you're adding component with CPE like:

cpe:2.3:a:7-Zip:7-Zip:18.03:*:*:*:*:*:*:*

The DependencyTrack analyzers will not find any issues. When switching from "Z" to "z" everything works correctly:

cpe:2.3:a:7-zip:7-zip:18.03:*:*:*:*:*:*:*

### Proposed Behavior

The CPE (and PURL I suppose) should be case-insensitive, all combination should find vulnerabilities.

### Checklist

stohrendorf:
### Description

This is an attempt to address issue #4663, to make CPE matching case-insensitive.

### Addressed Issue

Fixes #4663

### Additional Details

Since this touches areas I'm not familiar with, and since the issue is marked with size M, I'm pretty sure I missed quite a lot of things.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/VulnerableSoftware.java
+++ b/src/main/java/org/dependencytrack/model/VulnerableSoftware.java
@@ -262,7 +262,7 @@ public class VulnerableSoftware implements ICpe, Serializable {
     }
 
     public void setPart(String part) {
-        this.part = part;
+        this.part = part == null ? null : part.toLowerCase();
     }
 
     public String getVendor() {
@@ -270,7 +270,7 @@ public class VulnerableSoftware implements ICpe, Serializable {
     }
 
     public void setVendor(String vendor) {
-        this.vendor = vendor;
+        this.vendor = vendor == null ? null : vendor.toLowerCase();
     }
 
     public String getProduct() {
@@ -278,7 +278,7 @@ public class VulnerableSoftware implements ICpe, Serializable {
     }
 
     public void setProduct(String product) {
-        this.product = product;
+        this.product = product == null ? null : product.toLowerCase();
     }
 
     public String getVersion() {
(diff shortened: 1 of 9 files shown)


src/main/java/org/dependencytrack/upgrade/v4140/v4140Updater.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public void executeUpgrade(final AlpineQueryManager qm, final Connection connection) throws Exception {
+        normalizeCpeData(connection);
+    }
+
+    private void normalizeCpeData(final Connection connection) throws SQLException {
+        try (final Statement statement = connection.createStatement()) {
+            for (var column : List.of("PART", "VENDOR", "PRODUCT")) {
```
nscuro: Let's do a single statement execution that modifies all columns at once instead. Multiple passed would produce unnecessary bloat in MVCC databases like Postgres.

src/main/java/org/dependencytrack/upgrade/v4140/v4140Updater.java
```diff
@@ -0,0 +1,54 @@
...
+import java.sql.Statement;
+import java.util.List;
+
+public class v4140Updater extends AbstractUpgradeItem {
+
+    private static final Logger LOGGER = Logger.getLogger(v4140Updater.class);
+
+    @Override
+    public String getSchemaVersion() {
+        return "4.13.1";
```
nscuro: Let's target 4.13.5 for this.

nscuro (changes requested):

src/main/java/org/dependencytrack/upgrade/v4140/v4140Updater.java
stohrendorf: Oh, sorry. Didn't see that.

stohrendorf:
I'm still not sure this covers everything. It feels a bit above my head, and I still got the feeling that there are cases where this will not work properly. For example, with the recent changes, I added more changes for the lucene engine, and I'm not sure how to test this properly.

nscuro (approved):

PR #5314, opened 2025-09-20
Issue 5196

### LINKED ISSUE #5196: GET v1/tag/policy/$policy_uuid returns ALL tags in the system
### Current Behavior

A

### Steps to Reproduce

1. Define 2 tags
2. Add 1 tag to policy
3. Call api/v1/tag/policy/c2ef20f3-2aec-4061-a3b1-2fbd2cd5e9a5
4. Observe both tags are returned

### Expected Behavior

A

### Dependency-Track Version

4.13.2

Container Image

### Database Server

N/A

### Database Server Version

_No response_

### Browser

N/A

### Checklist

stohrendorf:
### Description

Changes /tag/policy/$policy_uuid to return only tags of the policy itself, independent of whether a project is associated.

### Addressed Issue

Fixes #5196.

### Additional Details

Not sure what the intended behaviour is here. The documentation is unclear here. So this might be completely wrong. Additionally, in the second commit, some static analysis issues were addressed not related to the underlying issue, and I'd be happy to remove that commit.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/TagQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/TagQueryManager.java
@@ -176,10 +176,10 @@ public class TagQueryManager extends QueryManager implements IQueryManager {
         ) {
             this(
                     name,
-                    (long) projectCount,
-                    (long) collectionProjectCount,
-                    (long) policyCount,
-                    (long) notificationRuleCount,
+                    projectCount,
+                    collectionProjectCount,
+                    policyCount,
+                    notificationRuleCount,
                     (long) totalCount);
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/dependencytrack/persistence/TagQueryManager.java
```diff
@@ -175,10 +176,10 @@ public TagListRow(
         ) {
             this(
                     name,
-                    (long) projectCount,
-                    (long) collectionProjectCount,
-                    (long) policyCount,
-                    (long) notificationRuleCount,
```
nscuro, line 181: Ah, I think I added those to avoid recursive constructor calls. But I guess only casting one of the parameters (`totalCount`) suffices. Good catch.

nscuro (approved):

nscuro:
> Not sure what the intended behaviour is here. The documentation is unclear here. So this might be completely wrong. 

Looking at the code it was likely an oversight / copy-paste-induced issue. Your fix makes sense. Thanks!

PR #5323, opened 2025-09-23
Check for non-empty timestamp files in doDownload of NistMirrorTask

### LINKED ISSUE #5322: NistMirrorTask - Download failed: Cannot parse null string
### Current Behavior

While playing around with the `NistMirrorTask` on a test instance I noticed the following error in the logs: 
``` log
2025-09-23 16:26:30,718 ERROR [NistMirrorTask] Download failed : Cannot parse null string
```  

After a short investigation I found some empty timestamp (`.ts`) files in the `dependency-track/nist` directory.
I am not really sure what caused these files to be emtpy, probably a crash or other ungraceful shutdown while the task was running. 

Reading and parsing the empty timestamp files causes the error listed above: 

``` Java
File timestampFile = new File(outputDir, filename + ".ts");
if(timestampFile.exists()) { 
    BufferedReader tsBufReader = new BufferedReader(new FileReader(timestampFile));
    String text = tsBufReader.readLine(); 
    modificationTime = Long.parseLong(text); 
}
```
Calling `tsBufReader.readLine()` on an empty `timestampFile` will return `null`. Passing `null` to `Long.parseLong()` will cause an exception to be thrown. 
**This error causes the affected "year files" to be skipped every time the task runs and they will not be overwritten or deleted.** 
A simple fix would be: also check if the file is not empty (`lenght() > 0`). In this case the "corrupted" emtpy files will simply be overwritten. 

### Steps to Reproduce

1. Activate Nvd feed mirroring (NistMirrorTask)
2. Create or modify a `.ts` file in `dependency-track/nist` (example: `nvdcve-2.0-2025.json.gz.ts`) such that it is empty, i.e. 0 bytes
3. Start DT
4. Observe the logs for the mentioned error

### Expected Behavior

The affected file is downloaded if the `.ts` file is empty instead of skipping it every time

### Dependency-Track Version

4.13.4

Container Image

### Database Server

PostgreSQL

### Database Server Version

17.6 alpine

### Browser

Google Chrome

### Checklist

jonbally:
### Description

In the rare case that a timestamp file (.ts) is empty due to a crash or other ungraceful shutdown, the affected file will be read and parsed which causes an exception to be thrown. This would happen every time the task runs and thus not self correct. This fix will make sure the empty/corrupted .ts files are overwritten and the corresponding file is reacquired.

### Addressed Issue

Fixes #5322

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/NistMirrorTask.java
+++ b/src/main/java/org/dependencytrack/tasks/NistMirrorTask.java
@@ -252,7 +252,7 @@ public class NistMirrorTask extends AbstractNistMirrorTask implements LoggableSu
     private long checkHead(final String cveUrl) {
         final HttpUriRequest request = new HttpHead(cveUrl);
         try (final CloseableHttpResponse response = HttpClientPool.getClient().execute(request)) {
-            return Long.valueOf(response.getFirstHeader(HttpHeaders.CONTENT_LENGTH).getValue());
+            return Long.parseLong(response.getFirstHeader(HttpHeaders.CONTENT_LENGTH).getValue());
         } catch (IOException | NumberFormatException | NullPointerException e) {
             LOGGER.error("Failed to determine content length");
         }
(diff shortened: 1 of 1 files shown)


nscuro (approved):
Thanks!

Issue #5348, opened 2025-09-29
Improving vulnerability audit through a prioritization score


cat-passwd:
### Current Behavior

In the `Audit Vulnerabilities` tab of the project view, I identified a lack of properties useful to the prioritization of the analysis of those vulnerabilities.

The most relevant properties currently present are `CWEs` and `Severity`, which I find lacking by themselves.

### Proposed Behavior

As this enhancement request affects both frontend and backend, I will only detail the effects on the backend in the following proposition. Please review it accompanied by [its corresponding frontend issue](https://github.com/DependencyTrack/frontend/issues/1341) for completeness.

I feel the need for faster and more informed prioritization of the analysis of vulnerabilities, in particular for new projects, combining available pre-analysis metrics like CVSS and EPSS.

I propose to add a prioritization score to findings, based on CVSS base score and EPSS. I performed a simulation of various formulas ideas and arbitrated on the resulting ordering to find the formula of `priorityScore = (1 + epss^0.5) * cvss` matching my arbitration.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #5349, opened 2025-09-29
Adds a priority score to findings

cat-passwd:
### Description

Adding a priority score for project findings, following the formula: `priorityScore = (1 + epss^0.5) * cvss`.
CVSSv3 is preferred over CVSSv2 if present.
EPSS defaults to 0 when undefined.
The priority score itself defaults to zero when neither CVSSv3 nor CVSSv2 is available.

### Addressed Issue

Implements #5348

### Additional Details

View of the prioritization using this score (includes the respective frontend pull request):
dependency-track-more-vulnerabilty-properties-prioritization

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/Finding.java
+++ b/src/main/java/org/dependencytrack/model/Finding.java
@@ -28,6 +28,8 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.Serializable;
 import java.math.BigDecimal;
+import java.math.MathContext;
+import java.math.RoundingMode;
 import java.sql.Clob;
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -205,6 +207,14 @@ public class Finding implements Serializable {
         optValue(vulnerability, "severityRank", severity.ordinal());
         optValue(vulnerability, "epssScore", o[20]);
         optValue(vulnerability, "epssPercentile", o[21]);
+        // Compute the priority score relative to known scores (EPSS and CVSS).
+        BigDecimal priorityScore = BigDecimal.ZERO;
+        if (o[15] != null || o[16] != null) {
+            final BigDecimal cvssBaseScore = o[16] != null ? (BigDecimal) o[16] : (BigDecimal) o[15];
+            final BigDecimal epssScore = o[20] != null ? (BigDecimal) o[20] : BigDecimal.ZERO;
+            priorityScore = epssScore.sqrt(MathContext.DECIMAL32).multiply(cvssBaseScore).add(cvssBaseScore).setScale(5, RoundingMode.HALF_UP);
+        }
+        optValue(vulnerability, "priorityScore", priorityScore);
         final List<Cwe> cwes = getCwes(o[22]);
         if (cwes != null && !cwes.isEmpty()) {
             // Ensure backwards-compatibility with DT < 4.5.0. Remove this in v5!


ptdropper:
I work in the medical device manufacturing space so severity and exploitability are considered carefully. My question is on the source of the formula you propose " priorityScore = (1 + epss^0.5) * cvss. CVSSv3 is preferred over CVSSv2 if present."  From the design/develop/manufacture/ship perspective CVSSv2 is more expressive and valuable compared to v3. Can you comment on the formula and comment on the v2 and v3 proposal? From my perspective using multiple data sources as you suggest is excellent. My request would be to have a choice of the priorityScore formula for me the user to select or to implement on my own. The medical device standards groups are and have been struggling with the various ways to assess the scoring with many proposed ways to do so, and all of the formulas have issues so far. This is an on going effort in the standards bodies, where I participate by offering real world experience. Thanks.

cat-passwd:
Thanks for your questions!

## Rationale of the formula

Developed formula for clarity: `priorityScore = cvss + epss^0.5 * cvss`.

As you mentioned, establishing such a formula to prioritize vulnerability analysis is non-trivial, based on already estimated data and ultimately quite subjective, so I will describe the rationale behind it:
- A vulnerability of high CVSS should be prioritized at a significant degree (here 50% of the priority score) over a vulnerability of lower CVSS: `"cvss + "`.
- The likelihood of exploitation in the wild expressed by the EPSS should be weighted in relation to the CVSS of the vulnerability, with a heavy emphasis on vulnerability of high CVSS: `"epss * cvss^2"`, which produces the same priority ranking as `"epss^0.5 * cvss"` but has clearer bounds of [0, 20], or [0, 10] in the absence of EPSS as its weight then represents at maximum 50% of the total priority score.

I also wanted to take KEV and public exploit into account but since this data is not currently available in Dependency Track, I refrained to integrate it into the formula. If this data is available in the future, the formula should be modified to use it. In my experimentations, I found substituting the EPSS by `1` in cases where the vulnerability was actively exploited matched my personal arbitrations.

While this is not perfect, I feel there is a **need to have at least something better than only using severity** (not CVSS) which is the current behavior, this is the main motivation behind this pull request. This was tested on our projects' vulnerabilities and after arbitration we found the output to be satisfying.

Of course, I'm very interested in any suggestion to better this approach and ultimately have a better formula.

## CVSSv3 preferred over CVSSv2

I did not think very long on this matter, which may be an error on my part. I decided on this selection from two main reasons:
- For new vulnerabilities, CVSSv2 tends to be much less often populated than CVSSv3.
- In the vulnerability view of Dependency Track, the CVSSv3 score is already displayed in priority over the CVSSv2 score when both are present. Below, an example of this view with `CVE-2019-3466` which has a CVSSv2 of 7.2 and a CVSSv3 of 7.8:
cvssv3-over-cvssv2-vuln-view

retanoj:
Combining EPSS and CVSS scores as vulnerability priority is a good idea. At the same time, KEV data should be included (KEV can be simply considered as EPSS = 1.0), and vulnerability priority should be used to calculate metrics instead of CVSS ratings.

Here, I propose using ratings instead of scores, which helps to maintain conceptual consistency with CVSS ratings.

cat-passwd:
Hi @retanoj !

Thanks for your interest on this matter, I agree with your assessment regarding the importance of using the KEV, emulating an EPSS of 1.0 when a vulnerability is included in it:
> If this data is available in the future, the formula should be modified to use it. In my experimentations, I found substituting the EPSS by ` 1`  in cases where the vulnerability was actively exploited matched my personal arbitrations.

However, I don't understand your proposition of using ratings instead of scores. In particular, you mention "CVSS ratings" which surprises me since CVSS stands for "Common Vulnerability Scoring System", clearing meant to output a score and not a rating.

Could you please clarify on this last point?

retanoj:
Hi @cat-passwd 
Thx for your reply.

I know that CVSS is a score system and it maps score ranges to different rating.
I think it would be ideal if EPSS could be mapped to a factor in CVSS, thus influencing the CVSS score. But that's really not feasible.
Alternatively, could we find a formula that incorporates EPSS and CVSS without altering the CVSS rating range, thus enabling the application of the CVSS rating?

cat-passwd:
It's interesting, it would indeed be relevant to consider a CVSS influenced by EPSS in order to better represent potential risk.

The current formula was conceived aiming to produce scores and rank the priority of findings. It may be relevant to perform further research and discussions to find a more fitting formula.

I believe the main challenge to be the wide range of values EPSS can take, factoring it with CVSS dilutes CVSS disproportionately to its importance, as you also identified. Using the EPSS percentile may be a good path forward, but I'm seeking other opinions to produce a result matching the expectations from analysts which will ultimately deal with a high volume of findings.

Since further feedback is necessary, I will try to reach out on Slack next week.

cat-passwd:
From Mattias Jidhage on Slack:
> Going "Beyond CVSS Base" is a great step.
>
> Instead of building our own algo; have you considered using CVSS scoring system as your base algo and using EPSS (and possibly other signals) to influence CVSS Temporal (CVSS 3.1) and/or Threat Metric (CVSS4.0)
>
> I realize a mapping is required from EPSS (and other future signals, like KEV) to CVSS Temporal Exploit Code Maturity and that mapping would be custom (afaik no 'standard' exists for this) but the benefit would be that this approach would open up for other modifiers. Like adding the possibility to adding environmental parameters as well.

retanoj:
@cat-passwd  
emmm, I'm considering something else.
Yes, we can map EPSS (or KEV) scores to a factor of the CVSS score. But will dependency-track accept this until it receives official approval from the CVSS SIG?

cat-passwd:
@retanoj 

From the [CVSS version 4.0 user guide](https://www.first.org/cvss/v4.0/user-guide):
> The application of Environmental and Threat metrics is the responsibility of the CVSS consumer. Assessment providers such as product maintainers and other public/private entities such as the National Vulnerability Database (NVD) typically provide only the Base Scores enumerated as CVSS-B.

What we seek to do is essentially a CVSS-BT (Base and Threat metrics) from the nomenclature in the user guide cited above. As such, I do not believe any approval is needed, as this it reflects characteristics that change over time.

In my research, I found the repository [cvss-bt](https://github.com/t0sche/cvss-bt/tree/main) to fit very closely with what I have in mind.

While I'm very enthusiastic about their approach, this would represent significantly more work than my initial proposition. Which also bring more changes on the whole codebase, tasks running daily to enrich our local CVEs (or an external dependency of questionable reliability compared to NVD).

I personally believe this is worth it, as it is valuable information for analysts and security teams (and matches the intent expressed explicitly in the CVSS specification by the FIRST).

retanoj:
@cat-passwd 
Yes, integrating threat indicators into CVSS scores is indeed a highly valuable initiative. I have also taken note of the cvss-bt project, which incorporates certain commercial data sources. However, I do have some reservations regarding the algorithms employed by the cvss-bt project.

Identifying the right approach is crucial. We are conducting similar experiments internally, utilizing KEV and EPSS scores to influence CVSS scores—specifically, in the context of CVSS 4.0, this involves configuring the "Exploit Maturity" field.

PR #5432, opened 2025-10-21
fix: removed unwanted code and minor typo fixes

arjavdongaonkar:
### Description

Remove unwanted code and typo fix

1. instad --> instead
2. .get(0); --> .getFirst();
3. Removed unwanted .sorted() as we are using only count.

### Checklist


countfnx (approved):

arjavdongaonkar:
These files moved in v5, src/main/java is now apiserver/src/main/java and PolicyEngine was replaced by the CEL engine, so the diff no longer applies. Closing.

Issue #4484, opened 2024-12-20
Project name rendered as PURL in notifications and logs


lukas-braune:
### Current Behavior

Projects that have a PURL assigned (SBOM's metadata component) will be rendered with the PURL instead of their project name at least in the following locations:
- Email notifications: subject and body
- Microsoft Teams notifications: message title
- Log messages: `org.dependencytrack.resources.v1.ProjectResource` logger → project creation and deletion events

This seems to be a regression as it should have been resolved by #4108.

### Steps to Reproduce

1. Create a new project and assign a PURL address under _Identity_ tab
2. Configure MS Teams and email notifications
3. Assign a vulnerable component in order to trigger the delivery of notifications (for instance)
4. Remove the project

### Expected Behavior

Render the project name instead of the PURL

### Dependency-Track Version

4.12.2

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Database Server Version

15.8

### Browser

Google Chrome

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

PR #5438, opened 2025-10-22
Changed the toString() method for project objects.

ElenaStroebele:
### Description
Changed the toString() method for project objects, to return the actual name and version instead of the assigned purl in any case. Added a new test to verify this and corrected existing tests to account for this change.

### Addressed Issue
Addressed issue #4484. Previous to my changes, alerts displayed the purl of the project if set, instead of the name and the version. 

### Additional Details
Used ChatGPT and Copilot to understand existing codebase and find the underlying problem in the toString() method; further used to debug and suggest changes in the tests. 

### Checklist

### DIFF
--- a/src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java
+++ b/src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java
@@ -186,7 +186,7 @@ abstract class AbstractPublisherTest<T extends Publisher> extends PersistenceCap
                 .scope(NotificationScope.PORTFOLIO)
                 .group(NotificationGroup.NEW_VULNERABILITY)
                 .level(NotificationLevel.INFORMATIONAL)
-                .title(NotificationConstants.Title.NEW_VULNERABILITY)
+                .title(NotificationUtil.generateNotificationTitle(NotificationConstants.Title.NEW_VULNERABILITY, project))
                 .content("")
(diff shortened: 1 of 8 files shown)


ElenaStroebele:
@nscuro Please consider this PR in release 4.13.6 .

src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java
```diff
@@ -318,6 +318,29 @@ public final void baseTestPublishWithScheduledNewPolicyViolationsNotification()
                 .isThrownBy(() -> publisherInstance.inform(PublishContext.from(notification), notification, createConfig()));
     }
 
+
+    public void baseTestInformWithNewVulnerabilityNotificationAndSubject(){
```
nscuro: Instead of creating a new test case that is effectively the same as `baseTestInformWithNewVulnerabilityNotification`, let's update `baseTestInformWithNewVulnerabilityNotification` directly instead.

nscuro (changes requested):

src/test/java/org/dependencytrack/notification/publisher/AbstractPublisherTest.java
ElenaStroebele: Implemented your suggestion! Thanks for your feedback!

PR #5456, opened 2025-10-28
Add CVSSv4 support to Dependency-Track

### LINKED ISSUE #4707: Cvss4 support in Dependency Track
### Current Behavior

currently we are not able to see Cvss4 scoring in Dependency Track

Image

### Proposed Behavior

As NVD supports CVSS4 scoring system for vulnerabilities it should also be shown in Dependency Track

### Checklist

tobiasgies:
### Description

This adds support for CVSSv4 scores to Dependency-Track. Scores are stored in the database, will be returned as part of the relevant HTTP API resources following existing conventions, can be updated via the REST API, and will be processed by parsers for most vulnerability sources. They will also be preferred over CVSSv3 and CVSSv2 scores when determining the severity of a vulnerability.

### Addressed Issue

fixes #4707

#### CVSSv4 handling not implemented in certain parsers

The following parsers do not have handling for CVSSv4 scores added by this PR, since the APIs they are based on are commercial offerings that I do not have access to:

- Snyk
- VulnDB

#### Only backend work completed for now

This PR obviously only addresses the server / API side of implementing CVSSv4 support. I am planning to work on support in the frontend next, but wanted to put my work on this repository out there first. I hope to get some feedback on this so I don't run in the wrong direction with any frontend changes.

#### Trivy protobufs updated

To be able to process CVSSv4 scores supplied by trivy, I have updated the protobuf files stored in this repo to the state of their `release/v0.67` branch. I have kept the customizations / changes to the headers of the protobuf files the same, though. For the remainder of each protobuf file, I chose easier copy/paste-ability in the future over keeping the diff small. I hope that's okay.

### Checklist

- [X] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)
  - **NOTE:** This is likely an open to-do item and I'd love some feedback on what needs documenting.

### DIFF
--- a/src/test/java/org/dependencytrack/resources/v1/NotificationPublisherResourceTest.java
+++ b/src/test/java/org/dependencytrack/resources/v1/NotificationPublisherResourceTest.java
@@ -51,6 +51,7 @@ import org.glassfish.jersey.server.ResourceConfig;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 import java.time.Duration;
@@ -366,6 +367,7 @@ public class NotificationPublisherResourceTest extends ResourceTest {
     }
 
     @Test
+    @Timeout(20)
     public void testNotificationRuleTest() {
         NotificationPublisher publisher = qm.createNotificationPublisher(
                 "Example Publisher", "Publisher description",


tobiasgies:
`NotificationPublisherResourceTest.testNotificationRuleTest` failed.

Well that's new. I haven't seen that test fail in my local test runs yet. I'll have a look to see if I can reproduce... any info on if it's just a bit brittle?

**EDIT:** From what I can tell it seems we just found a particularly busy Actions runner. The test takes between 5 and 6 seconds to run on my local machine (AMD Ryzen 9 9900X, 64 GB RAM). I could increase the timeout for this test to 15 seconds to make it less likely for this to re-occur. Not sure if that's desirable though.

stohrendorf:
> I could increase the timeout for this test to 15 seconds to make it less likely for this to re-occur. Not sure if that's desirable though.

IMHO you can do that as long as it's not excessive. I have done so myself in other PRs to mitigate some flaky tests that were clearly failing because of low resources in the CI environment.

tobiasgies:
> IMHO you can do that as long as it's not excessive. I have done so myself in other PRs to mitigate some flaky tests that were clearly failing because of low resources in the CI environment.

Alright, thank you. In that case: Extended timeout to 20s to hopefully alleviate the flakiness.

src/main/java/org/dependencytrack/model/Vulnerability.java
```diff
@@ -284,6 +284,24 @@ public static Source resolve(String id) {
...
+    private BigDecimal cvssV4ThreatScore;
+
+    @Persistent
+    @Column(name = "CVSSV4ENVIRONMENTALSCORE", scale = 1)
+    private BigDecimal cvssV4EnvironmentalScore;
+
+    @Persistent
+    @Column(name = "CVSSV4VECTOR")
+    @JsonDeserialize(using = TrimmedStringDeserializer.class)
+    @Pattern(regexp = RegexSequence.Definition.PRINTABLE_CHARS_PLUS, message = "The CVSSv4 Vector may only contain printable characters")
```
snieguu, line 302: Maybe we should do here more strict validation, accept only valid CVSSv4 vectors

tobiasgies: I agree in principle, I just copy/pasted the validation code for the other CVSS vector versions for the moment. However I think this might be duplicate work - the `org.metaeffekt.core:ae-security` package contains stricter validation logic in `CvssVector.parseVector`. Any invalid vector will cause the method to return null instead of a `CvssVector` instance, causing the data to be thrown out.

With that in mind, do you still think additional validation is needed in this place?

snieguu: Yes I know that for every other places related to CVSS we have `PRINTABLE_CHARS_PLUS`.
Regarding to
> Any invalid vector will cause the method to return null instead of a CvssVector instance, causing the data to be thrown out.

validation in `CvssVector.parseVector` is also very simple, regex is:
```
private static final Pattern CVSS_PATTERN = Pattern.compile("CVSS:\\d+\\.?\\d?");
```
For instance, I am able to create such internal Vulnerability in DependencyTrack 4.13.6(note that i was able to put invalid cvss 4.0 vector into cvss 2.0)
Screenshot 2025-10-30 at 11 54 23

From my experience strict validation should be present from the begining of the feature because then data that normaly will be rejected(because validation) are living in db and can cause a problems

Second thing is that OWASP product should be secure as much as possible, because this is OWASP mission at the end,
how can we teach people to write application according to ASVS 4.X when we not do the same.

On the other hand previous version of cvss was written in a way that is used in PR, making difference here can cause problems that i am not aware of, so I agree with that maybe it can be addressed later(I can do that)

src/main/java/org/dependencytrack/upgrade/v4140/v4140Updater.java
```diff
@@ -0,0 +1,54 @@
...
+            LOGGER.info("Adding CVSSv4 columns to \"VULNERABILITY\"");
+            statement.execute(/* language=SQL */ """
+                    ALTER TABLE "VULNERABILITY"
+                        ADD COLUMN "CVSSV4BASESCORE"    numeric,
+                        ADD COLUMN "CVSSV4THREATSCORE" numeric,
+                        ADD COLUMN "CVSSV4ENVIRONMENTALSCORE"  numeric,
+                        ADD COLUMN "CVSSV4VECTOR"       varchar(255);
+                    """);
+        }
+    }
```
stohrendorf, line 53: A migration shouldn't be necessary, the persistence framework should add columns itself.

src/main/java/org/dependencytrack/model/Vulnerability.java
nscuro: It might make sense to look into creating a custom validator, like we have for [cron expressions](https://github.com/DependencyTrack/dependency-track/blob/07316d9b82e4c8e8fa68d2103540154665d2b820/src/main/java/org/dependencytrack/model/validation/CronExpressionValidator.java) for example.

I do agree that adding that for existing CVSS types might cause issues that would need proper data migration first. But for everything we add going forward, it makes sense to be more defensive as to what we accept.

stohrendorf:
Regarding the other analyzers you didn't extend, Snyk docs are [here](https://docs.snyk.io/snyk-api/reference). For VulnDB, [this](https://vuldb.com/de/?kb.datapoints#vulnerabilities) seems to be the official reference.

AndreVirtimo:
Please also add the Vector to the export (Findings class).

Even vor v2 and v3 It's not in the export, but I provided an PR to add them -> #5844

nscuro:
@tobiasgies I'm going to take this up from here to make sure this makes it to the 4.14.0 release. Thanks for your work on this, we'll make sure to give appropriate credit.

nscuro:
Superseded by #5863

Issue #5499, opened 2025-11-05
Display Project UUID in log messages.


ElenaStroebele:
### Current Behavior

When creating/updating/deleting/cloning a project, the log-messages only include group, projectName and projectVersion to reference a project. 
For users it is difficult to uniquely identify a project like this through changes (updating the version/name), because the name and version are not unambiguous anymore. 

### Proposed Behavior

When the projectName/projectVersion/projectGroup display is used within logging for updating/deleting/creating projects, further include the projectUUID in the logging message. 

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

PR #5500, opened 2025-11-05
Include project UUID in log messages.

ElenaStroebele:
### Description
Added MDC within ProjectResource.java to be able to include the project UUID in log messages. UUID is now included in the log messages for creating/cloning/updating/deleting a project, to be able to identify a project unambiguously even through changes of name/version of the project. 

### Addressed Issue
Addressed issue #5499 .

### Additional Details
Used Copilot and ChatGPT to understand existing codebase. 

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
@@ -39,6 +39,7 @@ import io.swagger.v3.oas.annotations.security.SecurityRequirement;
 import io.swagger.v3.oas.annotations.security.SecurityRequirements;
 import jakarta.validation.constraints.Size;
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.MDC;
 import org.dependencytrack.auth.Permissions;
 import org.dependencytrack.event.CloneProjectEvent;
 import org.dependencytrack.model.Classifier;
@@ -77,6 +78,11 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
+
+import static org.dependencytrack.common.MdcKeys.MDC_PROJECT_NAME;
+import static org.dependencytrack.common.MdcKeys.MDC_PROJECT_UUID;
+import static org.dependencytrack.common.MdcKeys.MDC_PROJECT_VERSION;
+import static org.dependencytrack.common.MdcKeys.MDC_EVENT_TOKEN;
 
 import static java.util.Objects.requireNonNullElseGet;
 import static org.dependencytrack.util.PersistenceUtil.isPersistent;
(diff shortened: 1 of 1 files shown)


ElenaStroebele:
@nscuro Please consider this PR in release 4.13.6 .

src/main/java/org/dependencytrack/model/Project.java
```diff
@@ -676,6 +676,19 @@ public String toString() {
         return sb.toString();
     }
 
+    public String toStringForLogs(){
```
stohrendorf: Just an idea for the name: `toPrintableString` - because I think this could also be used for notifications. But also a style issue (missing a space before {).

stohrendorf (commented):
I also wonder if you could add some audit logs (or logging in general) since you're working on this. For example, recursive deletion of projects only logs the root project, but not the children as far as I can tell. But that would be just a bonus.

src/main/java/org/dependencytrack/parser/cyclonedx/CycloneDXVexImporter.java
```diff
@@ -53,7 +53,7 @@ public void applyVex(final QueryManager qm, final Bom bom, final Project project
             return;
         }
         if (qm.getVulnerabilityCount(project, true) == 0) {
-            LOGGER.info("The project %s does not have any vulnerabilities; Skipping VEX import".formatted(project));
+            LOGGER.info("The project %s does not have any vulnerabilities; Skipping VEX import".formatted(project.toStringForLogs()));
```
nscuro: I think it would be better to populate the [MDC](https://github.com/DependencyTrack/dependency-track/issues/3234) with appropriate keys instead.

Relevant MDC keys are defined here: https://github.com/DependencyTrack/dependency-track/blob/72dfb1a55073e4095dd7cc15814b5a38cb29e54a/src/main/java/org/dependencytrack/common/MdcKeys.java#L36-L38

Example usage here: https://github.com/DependencyTrack/dependency-track/blob/72dfb1a55073e4095dd7cc15814b5a38cb29e54a/src/main/java/org/dependencytrack/tasks/VulnerabilityAnalysisTask.java#L111-L113

src/main/java/org/dependencytrack/model/Project.java
ElenaStroebele: Thank you for your review! As I will continue to include the proposed MDC functionality, rather than this simple additional method approach, this will not be relevant for my future implementation. Therefore, I will mark this as resolved!

ElenaStroebele:
> I also wonder if you could add some audit logs (or logging in general) since you're working on this. For example, recursive deletion of projects only logs the root project, but not the children as far as I can tell. But that would be just a bonus.

I think it would be easier to open a new issue for this (if none already exists) to keep a clear task and goal.

src/main/java/org/dependencytrack/parser/cyclonedx/CycloneDXVexImporter.java
ElenaStroebele: That is a cool idea! I will try to include this in the project for these cases.

ElenaStroebele: I have implemented MDC for the requested changes in Issue #5499. Therefore I will resolve this conversation.

PR #5537, opened 2025-11-14
Incremental updates for OsvDownloadTask 

jonbally:
### Description

In it's current implementation the `OsvDownloadTask` will download and process **all** existing advisory files for each selected ecosystem on **every** run. Depending on the selected ecosystems, this can lead to a big daily resource demand (e.g. mirroring the entire collection of OSV Ubuntu advisories takes more than one hour on our instance).

This PR adds changes to the `OsvDownloadTask` to support incremental updates for the selected OSV ecosystems.
I also adapted existing tests and included new ones for full and incremental updates. 

A full mirror for the selected ecosystems will still be performed every 5 days (similar to the `NistMirrorTask`) to stay in sync with the OSV database. In certain cases some changes might be missed in the incremental update, e.g. when the modified_id.csv file contains a lot of new or modified ids, sometimes more than 40k, the incremental update will cut off at 10k.

### Addressed Issue

Enhancement for `OsvDownloadTask`
* Added support for incremental updates
* Added retry mechanism for downloads and requests
* Added notifications 
* Added time tracking

### Additional Details

The incremental updates are utilizing the `storage.googleapis.com/osv-vulnerabilities/<ecosystem>/modified_id.csv` file.
It contains the IDs of recently added and modified advisories as well as the timestamp of the addition/modification.  

The implementation is conceptually similar to the `NistMirrorTask`: 
* When running the task, it first checks if certain files exist in the `osv` directory for the respective ecosystem 
  * an ecosystem `all.zip` (renamed), 
  * a `.zip.ts` timestamp file for the `all.zip` file, and 
  * a `.csv.ts` timestamp file for the `modified_id.csv` (renamed) file
* If these files exist, are non-empty, and the "full update" timestamp is within 5 days before now, then an incremental update is performed using the timestamp in the `.csv.ts` file for selecting relevant entries of the `.csv`
* If any condition fails, then a full update is performed

\
I did not modify the parts of the `OsvDownloadTask` that perform the database update.
At the moment each advisory is still individually going through these steps, causing many DB reads and writes:

Full mirror: 
* Unzip entry -> Parse JSON -> Parse OSV -> map to DT vulnerability model -> Check if new or modified -> Update DB

Incremental mirror: 
* Request JSON -> Parse JSON -> Parse OSV -> map to DT vulnerability model -> Check if new or modified -> Update DB

I think this could be optimized a lot by processing the advisories in batches. 
An alternative to the current implementation, which I saw discussed somewhere recently, would be to implement an OSV analyzer using the OSV API instead of mirroring all of the advisories and performing the matching in DT.

\
Some points about this PR which might be worth discussing:
* The size limit of 1 GiB which I added for the acquired `all.zip` files 
* The hard-coded cadence of 5 days for the full mirror 
* The cutoff after 10k modified advisories for incremental updates

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
+++ b/src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
@@ -101,6 +101,7 @@ import static org.dependencytrack.model.ConfigPropertyConstants.VULNERABILITY_SO
 import static org.dependencytrack.model.ConfigPropertyConstants.VULNERABILITY_SOURCE_GOOGLE_OSV_BASE_URL;
 import static org.dependencytrack.model.ConfigPropertyConstants.VULNERABILITY_SOURCE_GOOGLE_OSV_ENABLED;
 import static org.dependencytrack.model.Severity.getSeverityByLevel;
+import static org.dependencytrack.util.RetryUtil.maybeClosePreviousResult;
 import static org.dependencytrack.util.VulnerabilityUtil.normalizedCvssV2Score;
 import static org.dependencytrack.util.VulnerabilityUtil.normalizedCvssV3Score;
 
@@ -113,7 +114,7 @@ public class OsvDownloadTask implements LoggableSubscriber {
     private static final long MAX_ZIP_BYTES = 1024 * 1024 * 1024;
     private static final String OSV_FILENAME_PREFIX = "google-osv-";
     private static final Logger LOGGER = Logger.getLogger(OsvDownloadTask.class);
-    private static final Retry DOWNLOAD_RETRY;
+    private static final Pattern JSON_PATTERN = Pattern.compile("\\.json$", Pattern.CASE_INSENSITIVE);
     private static final Retry REQUEST_RETRY;
     private Set<String> ecosystems;
     private String osvBaseUrl;
@@ -126,27 +127,7 @@ public class OsvDownloadTask implements LoggableSubscriber {
     private boolean mirroredWithoutErrors = true;
 
     static {
-        RetryRegistry registry = RetryRegistry.ofDefaults();
-        final RetryConfig downloadRetryConfig = RetryConfig.custom()
-                .intervalFunction(ofExponentialBackoff(
-                        Duration.ofSeconds(2),
-                        2,
-                        Duration.ofSeconds(32)
-                ))
-                .maxAttempts(6)
-                .retryExceptions(
-                        SocketException.class, SocketTimeoutException.class,
-                        NoRouteToHostException.class, ConnectTimeoutException.class
-                )
-                .build();
-        DOWNLOAD_RETRY = registry.retry("osv-mirror-download", downloadRetryConfig);
-        DOWNLOAD_RETRY.getEventPublisher()
-                .onRetry(event -> LOGGER.warn("Encountered retryable exception; Retries: "
-                        + event.getNumberOfRetryAttempts() + "; Next retry in: "
-                        + event.getWaitInterval().toSeconds() + " s", event.getLastThrowable()))
-                .onError(event -> LOGGER.error("Failed after "
-                        + event.getNumberOfRetryAttempts() + " attempts"));
-
+        final RetryRegistry registry = RetryRegistry.ofDefaults();
         final RetryConfig requestRetryConfig = RetryConfig.<CloseableHttpResponse>custom()
                 .intervalFunction(ofExponentialBackoff(
                         Duration.ofSeconds(2),
@@ -154,6 +135,7 @@ public class OsvDownloadTask implements LoggableSubscriber {
                         Duration.ofSeconds(32)
                 ))
                 .maxAttempts(6)
+                .consumeResultBeforeRetryAttempt(maybeClosePreviousResult())
                 .retryExceptions(
                         SocketException.class, SocketTimeoutException.class,
                         NoRouteToHostException.class, ConnectTimeoutException.class
@@ -221,7 +203,7 @@ public class OsvDownloadTask implements LoggableSubscriber {
             final long start = System.currentTimeMillis();
             setOutputDir(mirrorDirPath.toAbsolutePath().toString());
             LOGGER.info("Starting Google OSV mirroring for: " + ecosystems);
-            ecosystems.forEach(this::processOsvEcosystem);
+            ecosystems.forEach(ecosystem -> processOsvEcosystem(ecosystem, false));
             final long end = System.currentTimeMillis();
             if (mirroredWithoutErrors) {
                 LOGGER.info("Google OSV mirroring complete");
@@ -248,19 +230,15 @@ public class OsvDownloadTask implements LoggableSubscriber {
         }
     }
 
-    private void processOsvEcosystem(String ecosystem) {
+    private void processOsvEcosystem(String ecosystem, boolean forceFullMirror) {
         try (var ignoredMdcOsvEcosystem = MDC.putCloseable("osvEcosystem", ecosystem)) {
             final Instant currentTime = Instant.now(this.clock);
-            if (shouldDoIncrementalUpdate(ecosystem, currentTime)) {
-                String url = this.osvBaseUrl
-                        + URLEncoder.encode(ecosystem, StandardCharsets.UTF_8).replace("+", "%20")
-                        + "/modified_id.csv";
+            if (shouldDoIncrementalUpdate(ecosystem, currentTime) && !forceFullMirror) {
+                String url = this.osvBaseUrl + encodeURL(ecosystem) + "/modified_id.csv";
                 LOGGER.info("Incremental update - Initiating download of " + url);
                 doUpdate(url, ecosystem, true, currentTime);
             } else {
-                String url = this.osvBaseUrl
-                        + URLEncoder.encode(ecosystem, StandardCharsets.UTF_8).replace("+", "%20")
-                        + "/all.zip";
+                String url = this.osvBaseUrl + encodeURL(ecosystem) + "/all.zip";
                 LOGGER.info("Full mirror - Initiating download of " + url);
                 doUpdate(url, ecosystem, false, currentTime);
             }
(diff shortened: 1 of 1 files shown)


src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+        }
+    }
+
+    private void doUpdate(String url, String ecosystem, boolean incremental, Instant startTime) throws Throwable {
+        File incrementalTimestampFile = new File(outputDir, OSV_FILENAME_PREFIX + ecosystem
+                + "-modified.csv.ts").getAbsoluteFile();
+        File fullTimestampFile = new File(outputDir, OSV_FILENAME_PREFIX + ecosystem
+                + ".zip.ts").getAbsoluteFile();
+        if (incremental) {
+            Path modifiedCsv = DOWNLOAD_RETRY.executeCheckedSupplier(() -> downloadModifiedCsvFile(url, ecosystem));
```
nscuro: Do we really need the nested retries? `downloadModifiedCsvFile` already retries failing requests. If anything else fails in that method it's very unlikely to be retryable.

Same applies for `downloadOsvZipFile` below.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+            final HttpEntity entity = response.getEntity();
+            try {
+                LOGGER.info("Downloading...");
                 if (status.getStatusCode() != HttpStatus.SC_OK) {
-                    LOGGER.error("Download failed : " + status.getStatusCode() + ": " + status.getReasonPhrase());
-                    continue;
+                    LOGGER.error("Download of modified_id.csv failed for: " + ecosystem + " - " +
+                            status.getStatusCode() + " " + status.getReasonPhrase());
+                    mirroredWithoutErrors = false;
+                    throw new Exception("Download failed: " + status);
```
nscuro: Don't use generic `Exception` or `RuntimeException`:

```suggestion
                    throw new IllegalStateException("Download failed: " + status);
```

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+        }
+        LOGGER.info("Downloading and processing new or modified advisories");
+        final OsvAdvisoryParser parser = new OsvAdvisoryParser();
+        final ArrayList<String> unsuccessfulIds = new ArrayList<>();
+        int count = 0;
+        int lastLoggedPercent = 0;
+        for (String id : modifiedIds) {
+            final long downloadStartTime = System.currentTimeMillis();
+            String url = this.osvBaseUrl
+                    + URLEncoder.encode(ecosystem, StandardCharsets.UTF_8).replace("+", "%20")
```
nscuro: Perhaps the URL encoding could be moved to a small helper method?

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+            // Should never be reached as the .ts file is already checked in shouldDoIncrementalUpdate()
+            LOGGER.error("Could not obtain last update time from timestamp file due to previous error,"
+                    + " using fallback timestamp");
+            lastUpdate = Instant.EPOCH;
+        }
+        final ArrayList<String> modifiedIds = parseModifiedOsvAdvisoryCsv(modifiedCsvFilePath, lastUpdate);
+        if (modifiedIds.isEmpty()) {
+            LOGGER.info("No new or modified advisories since the last update, skipping");
+            return true;
+        }
```
nscuro, line 350: Would it make sense to have an escape latch here, where if we detect more than N (say, >100) modified IDs, we fall back to the full ZIP download? If there are many modified IDs, we'll be spending *a lot* of time doing individual download requests.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+                    }
+                    if (unsuccessfulIds.size() > 5) {
+                        LOGGER.error("Failed to acquire more than 5 out of " + modifiedIds.size() + " advisories, aborting."
+                                + " IDs that could not be acquired: " + unsuccessfulIds);
+                        return false;
+                    }
+                    try (final InputStream in = response.getEntity().getContent()) {
+                        final BufferedReader reader = new BufferedReader(
+                                new InputStreamReader(in, StandardCharsets.UTF_8), 8192
+                        );
```
nscuro: Can we move the readers to the `try` statement so they're guaranteed to be closed?

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+                        break;
+                    }
+                    // For some ecosystems more than 10k advisories (npm sometimes has 40k+) might be modified within a day.
+                    // Those mass updates are usually non-critical i.e. versions being added after a new release, but
+                    // they might also contain many new "malicious package" notices for npm
+                    if (modifiedIds.size() >= 10000) {
+                        LOGGER.warn("Cutting off after 10k new or modified advisories, "
+                                + "remaining updates will be retrieved in full mirror every 5 days");
+                        break;
+                    }
```
nscuro: Somewhat related to a comment I posted earlier in this file. Perhaps we should just fall back to the full ZIP download immediately when this condition is true.

> Those mass updates are usually non-critical [...]

Is an assumption that might prove to be wrong. It's also possible that such a mass update was done shortly after a new critical vulnerability was added. If we simply cut off here, we'd have severe delays of said critical vuln making it into our database.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+        metricParseTime += end - start;
+        return true;
+    }
 
-            String line = null;
-            StringBuilder out = new StringBuilder();
-            while ((line = reader.readLine()) != null) {
-                out.append(line);
+    private void unzipOsvZipFile(ZipInputStream zipIn, int totalCount) throws IOException {
+        final Pattern jsonPattern = Pattern.compile("\\.json$", Pattern.CASE_INSENSITIVE);
```
nscuro: This should be a class-level constant as it never changes.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+            lastLoggedPercent = currentProcessedPercentage;
+        }
+        return lastLoggedPercent;
+    }
+
+    private void processOsvAdvisoryJsonFromZip(ZipInputStream zipIn, String entryName, OsvAdvisoryParser parser) {
+        try {
+            final BufferedReader reader = new BufferedReader(
+                    new InputStreamReader(zipIn, StandardCharsets.UTF_8), 8192
+            );
```
nscuro, line 567: Can we move the readers to the `try` statement so they're guaranteed to be closed?

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -109,92 +213,462 @@ public OsvDownloadTask() {
...
+
+        if (!fullMirrorFile.exists() || !(fullMirrorFile.length() > 0))
+            // the .zip file does not exist or is empty
+            return false;
+
+        Instant lastFullMirror = readTimestampFile(fullMirrorOsvFileName);
+        Instant lastIncrementalMirror = readTimestampFile(modifiedOsvFileName);
+        if (lastFullMirror == null || lastIncrementalMirror == null)
+            // either one or both timestamp files are not present or timestamp parsing failed
+            return false;
```
nscuro: Please always use curly brackets for `if` blocks, even if it's just a single line.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
```diff
@@ -79,20 +106,97 @@
 
 public class OsvDownloadTask implements LoggableSubscriber {
 
+    public static final Path DEFAULT_OSV_MIRROR_DIR = Config.getInstance().getDataDirectorty().toPath()
+            .resolve("osv").toAbsolutePath();
+    // Max size for ecosystem zip files: 1 GiB, some ecosystem files might reach this size in a few years
+    // (Ubuntu is currently the largest one at ~380 MiB)
+    private static final long MAX_ZIP_BYTES = 1024 * 1024 * 1024;
```
nscuro, line 114: I think this is sensible. I don't envision large files becoming a big issue, but it's always good to have limits in place in case our assumptions are proven wrong.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
jonbally: I'm also not a fan of the nested retries TBH. I have added the "outer" one because I encountered a `SocketTimeoutException: Read timed out` multiple times during testing (specifically when running with the `Jetty` configuration for multiple hours with a OSV cadence of 1 hour, but it's not really reproducible reliably).
This did not affect the initial request, but happened during the download (`Files.copy()`) part of the method, so the request retry would not have caught it.

The issue was usually resolved with the first "outer" retry. I'm not sure what caused this exception, maybe the `HTTPClientPool` is reusing an older connection shortly before the socket times out? This might have to do with my test setup and never actually happen in production use (even though I did not modify anything in the `HTTPClient` handling), but I don't have enough experience to say for certain. 

I can of course remove the `DOWNLOAD_RETRY` and if this issue is observed in the future, it can be analyzed and handled specifically.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
jonbally: Yes, this makes sense. I will think about a way to do this in a good way.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
nscuro: I think your connection reuse idea might be spot on. The problem with retrying requests is that the retry doesn't close responses on its own, so you need to configure it with:

https://github.com/DependencyTrack/dependency-track/blob/2b67f6b552aaaade27b466ed8462288c75fae5c1/src/main/java/org/dependencytrack/tasks/scanners/OssIndexAnalysisTask.java#L107

Otherwise it might be piling up unclosed connections over time.

src/main/java/org/dependencytrack/tasks/OsvDownloadTask.java
jonbally: actually in this case it should not be changed to a try with resources, because then the wrapped `ZipInputStream` would be prematurely closed after processing the first zipentry.

jonbally:
I have implemented the requested changes, the failing tests seem to be unrelated. These `GitHubMetaAnalyzerTest` tests are usually flaky for me...

Issue #3808, opened 2024-06-05
False positives due to erroneous comparison of component versions


lukas-braune:
### Current Behavior

DT attributes CVE-2022-48174 to the component `pkg:apk/alpine/busybox@1.36.1-r18?arch=x86_64&distro=3.19.1` although the vulnerability has already been resolved in the earlier package version `1.36.1-r2`:

![image](https://github.com/DependencyTrack/dependency-track/assets/9955029/2f4eb8eb-c794-4af5-ba0b-867c36f10e9a)
https://osv.dev/vulnerability/CVE-2022-48174

The root cause of this issue appears to be DT's sorting methodology. Instead of a numerical sort, the application is performing a lexical sort on version identifiers, mistakenly interpreting the version `1.36.1-r18` as an earlier, and thus vulnerable, release compared to `1.36.1-r2`.

### Steps to Reproduce

1. Ensure that Google OSV advisory mirroring is enabled.
2. Create a new project.
3. In the newly created project, create a new component with the following purl: `pkg:apk/alpine/busybox@1.36.1-r18?arch=x86_64&distro=3.19.1`.
4. Check whether CVE-2022-48174 has been attributed to the newly created component.

### Expected Behavior

CVE-2022-48174 should not be attributed to the newly created component.

### Dependency-Track Version

4.11.3

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Database Server Version

15.5

### Browser

Microsoft Edge

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

nscuro:
Related: #2826

nscuro:
Support for Alpine versions has been added to versatile: https://github.com/nscuro/versatile/pull/257

The implementation passes [the comparison test suite](https://github.com/package-url/vers-spec/blob/470128dbe7574c360931239f8fbc8a971fa1d6ea/tests/alpine_version_cmp_test.json) of the vers specification.

PR #5559, opened 2025-11-18
Comparing version numbers with a combination of characters and numbers more differentially. 

ElenaStroebele:
### Description
Added a new method within /src/main/java/org/dependencytrack/util/ComponentVersion.java to further split the version number, to be able to compare the combination of characters and numbers more differentially.
Before, version numbers like this: 1.36.1-r2 & 1.36.1-r18 were split and then each part was compared. For the last part: r2 and r18, since it is no int, the method compared lexicographically, making r18 older than r2. That resulted in some false positives. 
Now, the r2 and r18 parts are split between the r and the integer, and then each of the pairs are compared. 

### Addressed Issue
Addressed issue: #3808 ("False positives due to erroneous comparison of component versions")

### Additional Details
Used ChatGPT and Copilot to understand existing structure and pattern matching, as well as sanity checks with example version numbers. 
https://stackoverflow.com/questions/11232801/regex-split-numbers-and-letter-groups-without-spaces

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/util/ComponentVersion.java
+++ b/src/main/java/org/dependencytrack/util/ComponentVersion.java
@@ -19,6 +19,7 @@
 package org.dependencytrack.util;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


ElenaStroebele:
@nscuro Please consider this PR in release 4.13.7 .

nscuro:
@ElenaStroebele I think this is superseded by #5591? The erroneous comparison described in https://github.com/DependencyTrack/dependency-track/issues/3808 seems to be Alpine-specific, so the true fix should be to implement Alpine version support in versatile.

ElenaStroebele:
Yes, I just wasn't sure yet whether I would start to implement the VERS approach, so I started with this little fix. But since you added the alpine support, I will delete this one. Thank you :)!

Issue #5616, opened 2025-12-09
Log-Statements for updating/creating/deleting/cloning projects should contain assigned teams


ElenaStroebele:
### Current Behavior

Currently, the log messages for updating/creating/deleting/cloning projects only contain the project uuid, name and version. 

### Proposed Behavior

The log messages should also contain assigned team(s) if specified.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

nscuro:
See https://github.com/DependencyTrack/dependency-track/pull/5617#issuecomment-3660123044

PR #5617, opened 2025-12-09
Added projectAccessTeams to log statements.

ElenaStroebele:
### Description
Added the projectAccessTeams to the logging statements if projects are created/updated etc. in ProjectResource.java via MDC.

### Addressed Issue
Addressed issue #5616.

### Additional Details
Used ChatGPT and Copilot to understand existing codebase and suggest changes. 

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/common/MdcKeys.java
+++ b/src/main/java/org/dependencytrack/common/MdcKeys.java
@@ -36,6 +36,7 @@ public final class MdcKeys {
     public static final String MDC_PROJECT_NAME = "projectName";
     public static final String MDC_PROJECT_UUID = "projectUuid";
     public static final String MDC_PROJECT_VERSION = "projectVersion";
+    public static final String MDC_PROJECT_TEAMS = "projectTeams";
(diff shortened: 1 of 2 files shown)


ElenaStroebele:
@nscuro Please consider this PR in release 4.13.7 .

nscuro:
@ElenaStroebele What is the underlying requirement for this? I am a bit confused as to what this is supposed to achieve.

ElenaStroebele:
> @ElenaStroebele What is the underlying requirement for this? I am a bit confused as to what this is supposed to achieve.

@nscuro This was a workaround for the following issue: #1409 in the /frontend repository which lead to projects missing an assigned team due to a bug in the code of the create-modal. Since I found the solution to the problem, this fix here is no longer a priority, though I think it would add some information for the future, to be able to identify problems with the assigned teams easier and faster. Let me know what you think! If this is not a beneficial feature to DT at all, I will close this issue.

nscuro:
Then I'd prefer to close this. There could be similar situations for other fields and we can't add log fields for all of them. If there is a need for more visibility then we'll need to think about a more thorough audit logging solution, but that should be generally applicable and not on a field-by-field basis.

PR #5711, opened 2026-01-19
fix: performance issue with PURL lookups #5710

### LINKED ISSUE #5710: Performance: Missing index on VULNERABLESOFTWARE causes high DB load
### Current Behavior

We are running Dependency-Track v4.13.6 with a dataset of **~26 million rows** in the `VULNERABLESOFTWARE` table. We are experiencing severe performance degradation where PostgreSQL CPU and IO usage hit the limit, causing the application to become unresponsive.

Investigation using `pg_stat_activity` showed that `SELECT` queries targeting `VULNERABLESOFTWARE` get stuck in the `active` state with **`LWLock: BufferMapping`** wait events, indicating heavy heap page thrashing.

### The problematic query

The query performs a lookup by PURL components **and** `VERSION` (ensuring no version ranges are involved via `IS NULL` checks):

```sql
SELECT 'org.dependencytrack.model.VulnerableSoftware' AS "DN_TYPE", ... 
FROM "VULNERABLESOFTWARE" "A0" 
WHERE "A0"."PURL_TYPE" = $1 
  AND "A0"."PURL_NAME" = $2 
  AND "A0"."PURL_NAMESPACE" = $3 
  AND "A0"."VERSION" = $4 
  AND "A0"."VERSIONENDEXCLUDING" IS NULL 
  AND "A0"."VERSIONENDINCLUDING" IS NULL 
  AND "A0"."VERSIONSTARTEXCLUDING" IS NULL 
  AND "A0"."VERSIONSTARTINCLUDING" IS NULL
```

### Root Cause

The existing index `VULNERABLESOFTWARE_PURL_TYPE_NS_NAME_IDX` (`PURL_TYPE`, `PURL_NAMESPACE`, `PURL_NAME`) **does not include the `VERSION` column**.
On a large dataset, PostgreSQL uses the existing index to find all rows for a component (which can be thousands) but must visit the **Heap** for every single row to verify the `VERSION` and `IS NULL` constraints. This causes massive random I/O.

### Workaround/Fix

We manually created a composite index including the `VERSION` column:

```sql
CREATE INDEX CONCURRENTLY "idx_vulnsoft_purl_version_lookup"
ON "VULNERABLESOFTWARE" ("PURL_TYPE", "PURL_NAMESPACE", "PURL_NAME", "VERSION");
```

After creating this index and running `ANALYZE`, the database load dropped to normal levels immediately.

### Steps to Reproduce

1. Run Dependency-Track with a large dataset (e.g., `VULNERABLESOFTWARE` table with > 25M rows).
1. Trigger a workload that performs heavy PURL lookups (e.g., BOM analysis).
1. Monitor PostgreSQL `pg_stat_activity`.
1. Observe that queries filtering by PURL + VERSION are slow and causing high I/O wait.

### Expected Behavior

PURL lookups that specify a concrete `VERSION` should utilize a covering index (or a precise index scan) to avoid scanning the table heap and causing IO contention.

### Dependency-Track Version

4.13.6

Container Image

### Database Server

PostgreSQL

### Database Server Version

14.12

### Browser

N/A

### Checklist

WoozyMasta:
### Description

Added a missing database index on the `VULNERABLESOFTWARE` table to include the `VERSION` column. This significantly improves the performance of PURL lookups (specifically when checking for a concrete version) by enabling Index-Only Scans (or precise Index Scans) and eliminating heavy heap page thrashing (`LWLock: BufferMapping`) on large datasets.

### Addressed Issue

Fixes #5710

### Additional Details

Investigation revealed that the existing index `VULNERABLESOFTWARE_PURL_TYPE_NS_NAME_IDX` does not cover the `VERSION` column. Queries filtering by PURL + VERSION were forced to scan the heap for every row matching the component, leading to extreme I/O and CPU usage on tables with millions of records.

This PR:

1. Adds the `@Index` annotation to the `VulnerableSoftware` entity.
1. Implements an upgrade task (`v4136Updater`) to create the index `VULNERABLESOFTWARE_FULL_PURL_IDX` on existing installations.
1. Supports PostgreSQL, MySQL, MSSQL, and H2.

### Checklist


src/main/java/org/dependencytrack/upgrade/v4136/v4136Updater.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Creates an index on VULNERABLESOFTWARE table to optimize PURL lookups with VERSION.
+ * This addresses performance issues when querying by PURL components and VERSION
+ * on large datasets (>25M rows), preventing heap page thrashing.
+ *
+ * @see <a href="https://github.com/DependencyTrack/dependency-track/issues/5710">GitHub Issue</a>
+ * @since 4.13.6
+ */
+public class v4136Updater extends AbstractUpgradeItem {
```
nscuro: Manual migration is not necessary, new indexes are added automatically. Only updating / dropping requires manual migration.

nscuro (changes requested):

src/main/java/org/dependencytrack/upgrade/v4136/v4136Updater.java
WoozyMasta: Done

Issue #5577, opened 2025-11-26
Fuzzy Matching as project setting instead of global


BrightKn1ght:
### Current Behavior

Currently fuzzy matching setting is a global setting which affects all projects. We have a Dependency Track instance running in your company and different departments are using it. Some are now wanting to scan with Fuzzy Matching while the others are complaining about too many False Positives.

### Proposed Behavior

I would suggest the following:

1. Adding a toggle button to the project creation dialog allowing to enable/disable the fuzzying on per project level. 
2. Marking the vulnerabilities found by fuzzying with a flag, which allows filtering them later in the UI. Probably the code section where to flag them could be [here](https://github.com/DependencyTrack/dependency-track/blob/a4ed39ff22d33edeb8fbe15b4cd93d070517cd3d/src/main/java/org/dependencytrack/search/FuzzyVulnerableSoftwareSearchManager.java#L143) and [here](https://github.com/DependencyTrack/dependency-track/blob/a4ed39ff22d33edeb8fbe15b4cd93d070517cd3d/src/main/java/org/dependencytrack/search/FuzzyVulnerableSoftwareSearchManager.java#L129)

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

BrightKn1ght:
I have started working on a solution. PR is coming soon.

PR #5729, opened 2026-01-21
Add fuzzy flag

BrightKn1ght:
### Description

DT supports fuzzy matching just as a global setting which will affect all upcoming projects/scans. Now you will have some users which like to see such matches and some are complaining about too may new matches. When you enable fuzzy matching, a lot of false positives could be generated. With this change findings that are found by fuzzing get annotated with a foundByFuzzing flag. When users can see which findings are caused by fuzzy matching, helps clearing false positives faster. Users which do not like to see such results could filter them out in their DT project => changes in frontend will be also required. 
With this change the flag will be present in the findings endpoint. 

E.g. endpoint **/api/v1/finding/project/your-project-id** returns now the response [findings-response.txt](https://github.com/user-attachments/files/24769496/findings-response.txt) with the BOM [libarchive_purl 1.json](https://github.com/user-attachments/files/24769511/libarchive_purl.1.json)

### Addressed Issue

CLOSES: #5577

### Checklist

- [x ] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)
- [x ] This PR implements an enhancement, and I have provided tests to verify that it works as intended
  // i have introduced a new database field for the fuzzy flag. not sure if this needs an update logic?

### DIFF
--- a/src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
+++ b/src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
@@ -631,16 +631,16 @@ class FindingResourceTest extends ResourceTest {
         Component c1 = createComponent(p1, "Component A", "1.0");
         Component c2 = createComponent(p1, "Component B", "1.0");
         Component c3 = createComponent(p1, "Component C", "1.0");
-        
+
         // Create vulnerabilities with different EPSS scores
         Vulnerability v1 = createVulnerabilityWithEpss("Vuln-1", Severity.CRITICAL, new BigDecimal("0.1"));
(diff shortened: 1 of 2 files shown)


BrightKn1ght:
I understand your concern about linking a specific scanner attribute with a generic attribution concept. That's also one thing which disturbed myself. Could you please elaborate a little more on what you are understanding under confidence score and how one could realize that? Are you meaning something like a score number e.g. from 0 to 100 how certain an issue is identified or do you mean to put the proposed fuzzy flag somewhere where it better belongs to. In case real score number is meant, should it then be supported by all scanners? This might be a real though thingy considering DT supports currently 5 different scanners. 
It's may first PR to DT so my understanding of the code base is limited yet.

PR #5739, opened 2026-01-23
Prevent XSS by removing unsafe innerHTML usage in search results  Signed-off-by:rajesh.sangepu@opsmx.io

Rajesh-Sangepu:
This PR fixes a Semgrep SAST finding where user-controlled search input was rendered using innerHTML, which could lead to a potential XSS issue.

The search rendering logic has been refactored to use safe DOM APIs and avoid generating HTML from user input, while preserving the existing search and highlighting behavior.

### DIFF
--- a/docs/scripts/search.js
+++ b/docs/scripts/search.js
@@ -48,14 +48,51 @@ layout: null
 				preview = preview + "...";
 			}
 
-			// Highlight query parts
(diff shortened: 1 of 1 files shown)


nscuro:
May I ask for some evidence that there is indeed a XSS concern here? How is this supposed to be exploitable?

Issue #3243, opened 2023-11-27
wget  zombie processes


somera:
### Current Behavior

Today I saw this

  => There are 2 zombie processes.

after login to my system. This is new. I don't have any zombie processes on my system.

ps shows

```
$ ps aux | grep 'Z'
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
xxxxxx      6717  0.0  0.0      0     0 ?        Z    20:17   0:00 [wget] <defunct>
xxxxxx      7857  0.0  0.0      0     0 ?        Z    20:18   0:00 [wget] <defunct>
```

next step was

```
           ├─containerd-shim(3136)─┬─java(3940)─┬─wget(6717)
           │                       │            ├─wget(7857)
           │                       │            ├─{java}(4781)
           │                       │            ├─{java}(4812)
           │                       │            ├─{java}(4884)
           │                       │            ├─{java}(4893)
```

And

```
$ ps aux | grep 3940
xxxxxx      3940 19.9  2.6 15333688 849660 ?     Ssl  20:16   1:49 java -XX:+UseParallelGC -XX:MaxRAMPercentage=90.0 --add-opens java.base/java.util.concurrent=ALL-UNNAMED -Dlogback.configurationFile=logback.xml -DdependencyTrack.logging.level=DEBUG -jar dependency-track-apiserver.jar -context /
xxxxxx     16571  0.0  0.0   6608  2432 pts/0    S+   20:26   0:00 grep --color=auto 3940
```

After I stopped the DT container zombie processes are gone. And DT was in idle. After reboot my system I saw the same.

What happens here? Why the zombie processes?

My docker-compose:

```
version: '3.7'

services:
  dtrack-apiserver:
    image: dependencytrack/apiserver
    environment:
      - TZ=Europe/Berlin
      # Database Properties
      - ALPINE_DATABASE_MODE=external
      - ALPINE_DATABASE_URL=jdbc:postgresql://xx.xx.xx.xx:5432/dtrack
      - ALPINE_DATABASE_DRIVER=org.postgresql.Driver
      - ALPINE_DATABASE_USERNAME=dtrack
      - ALPINE_DATABASE_PASSWORD=xxx
      - ALPINE_DATABASE_POOL_ENABLED=true
      - ALPINE_DATABASE_POOL_MAX_SIZE=10
      - ALPINE_DATABASE_POOL_MIN_IDLE=2
[...]

PR #5758, opened 2026-01-29
Ensure container zombie processes are reaped

nscuro:
### Description

Ensures container zombie processes are reaped.

Adds [tini](https://github.com/krallin/tini) as init process to ensure that zombie processes (as possibly caused by health checks) to not pile up over time.

Also extends the health check's timeout by a few seconds such that the curl command (with a lower timeout) is guaranteed to complete.

### Addressed Issue

Fixes https://github.com/DependencyTrack/dependency-track/issues/3243

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -71,7 +71,7 @@ RUN mkdir -p ${APP_DIR} ${DATA_DIR} \
     \
     # Install curl for health check
     && apt-get -yqq update \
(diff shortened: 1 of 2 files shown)


kutzi:
As far as I know, tini has been bundled with docker for a long time and you should use docker run --init to use it.
Or for docker-compose: specify 'init: true' https://docs.docker.com/reference/compose-file/services/#init

nscuro:
@kutzi Thanks, I saw that. But unfortunately we cannot rely on all users setting that flag. I decided to err on the side of caution and just bake it into the image itself.

Issue #5776, opened 2026-02-03
OSV mirroring only parses `affected.versions` when `affected.ranges` is not available, leading to over-reporting


nscuro:
### Current Behavior

The OSV mirroring logic currently only parses `affected.versions` when no `affected.ranges` property is defined for an advisory:

https://github.com/DependencyTrack/dependency-track/blob/050084ce146d5554a7577d4c0814bc1911d223a4/src/main/java/org/dependencytrack/parser/osv/OsvAdvisoryParser.java#L138-L148

This leads to over-reporting for advisories such as https://test.osv.dev/vulnerability/DEBIAN-CVE-2005-1119, which uses both a range, *and* individual versions:

<img width="438" height="590" alt="Image" src="https://github.com/user-attachments/assets/f66a0c3a-2a40-4355-aca2-d2a13804c9a4" />

<img width="1580" height="221" alt="Image" src="https://github.com/user-attachments/assets/4492b029-a099-4dfa-8f34-51da3bd7545c" />

### Steps to Reproduce

1. Enable OSV mirroring for the Debian ecosystem
2. Upload a BOM with vulnerable Debian components

### Expected Behavior

OSV mirroring should recognize this pattern (`range >= 0` + individual versions), and only persist the individual versions.

This **must** be fixed before v4.14.0 is released.

### Dependency-Track Version

4.14.0-SNAPSHOT

### Dependency-Track Distribution

Container Image

### Database Server

N/A

### Browser

N/A

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

nscuro:
Curiously, the OSV spec [says](https://ossf.github.io/osv-schema/#affected-fields):

> A version is considered affected if it lies within any one of the ranges **or** is listed in the versions list.

So it seems the Debian data is simply reported in a wrong way?

nscuro:
Asked a question here: https://github.com/google/osv.dev/discussions/4746

Issue #4725, opened 2025-03-06
DependencyTrack does not honor the ecosystem-version/release-branch/distribution-name


Andre-85:
### Current Behavior

PURLs in SBOMs contain often the information to which distribution they belonging to e.g. pkg:deb/debian/expat@2.5.0-1+deb12u1?arch=amd64&distro=debian-12 means expat in version 2.5.0-1+deb12u1 from distribution debian-12 (which is bookworm).
For expat is a CVE called CVE-2024-4591 with following information which versions are affected:
- debian-11: versions < 2.2.10-2+deb11u6
- debian-12: versions < 2.5.0-1+deb12u1
- debian-13: versions < 2.6.2-2

Without knowing the distribution we can not decide if expat version 2.5.0-1+deb12u1 is affected or not since the vulnerability is not fixed in debian-11, but its fixed in debian-12 and debian-13. So we would need to fallback to the save decision that is vulnerable since it is older than the latest affected version (2.6.2-2, debian-13), meaning that we will create false positives debian-11 and debian-12.

This refers to @mjaix on Jan 22 2025 on Bug #4445:

> We recently stumbled over a bunch of false positives with Debian/purls/OSV, using DT 4.12.2.
N.B. I am by no means a Debian (versioning) expert.
My $0.02 on the problem:

    A solid version comparison functionality is fundamental. However, be aware that ecosystems like Debian or Ubuntu with multiple delivery branches provide semantics (from my limited/simplistic point-of-view, Alpine has a similar but simpler system set up).
    The problem is to a great extent mitigated by the fact that vulnerability sources like OSV usually provide one or more matching rules for each "delivery branch" (that is, a given base version with or without patches or backports). As a consequence, the matching algorithm must look for matching "branches" in the set of matching rules for a vulnerability, and if a match exist, simply the final numbers need to be compared. In case no "branch" matches, simply compare the base numbers. With equal base numbers and no branch specified in the rule, any "branch" or build on top should suffice.

For the test of your solution, to provide some kind of long-term stability, and to avoid creating some "golden code",
I would recommend to create test data and a simple test suite to run against your implementation.
One of the actual false positives we have (and some derived edge cases) could be verified with the YAML test data below.
I am pretty sure that test experts will provide better solutions :-).

---
test_cases:
- vulnerability: CVE-2024-45491
  osv_rules:
  - pkg:deb/debian/expat?arch=source (<2.5.0-1+deb12u1)
[...]

Issue #4445, opened 2024-12-06
SBOMs from Ubuntu with purls seam not to work with Google OSV


Andre-85:
### Current Behavior

Hello together,
I've noticed that in DependencyTrack in version 4.12.1 supports fetching vulnerabilities from Google's OSV service. So wanted to test DependencyTrack with an SBOM containing PURLs i can also use for CVE-fetching via Google API. As example i choose the ubuntu package atftp in version 0.7.git20120829-3.1~0.18.04.1 which has the known CVE'S CVE-2020-6097, CVE-2021-46671 and CVE-2021-41054. But DependencyTrack reports no issues at all.

Since i was a bit unsure about the supported debian/ubuntu PURL format, i made different variants containing arch and/or distro and completely without arch/distro: 
[purl_test_name_version.json](https://github.com/user-attachments/files/18040459/purl_test_name_version.json)
[purl_test_name_version_arch_armhf.json](https://github.com/user-attachments/files/18040464/purl_test_name_version_arch_armhf.json)
[purl_test_name_version_arch_armhf_distro_bionic.json](https://github.com/user-attachments/files/18040467/purl_test_name_version_arch_armhf_distro_bionic.json)
[purl_test_name_version_arch_src.json](https://github.com/user-attachments/files/18040466/purl_test_name_version_arch_src.json)
[purl_test_name_version_arch_src_distro_bionic.json](https://github.com/user-attachments/files/18040468/purl_test_name_version_arch_src_distro_bionic.json)
[purl_test_name_version_distro_bionic.json](https://github.com/user-attachments/files/18040465/purl_test_name_version_distro_bionic.json)

### Steps to Reproduce

Upload the SBOMs which i uploaded to this ticket and you will find no vulnerabilities.

### Expected Behavior

Just last week Google closed issue #2842 (reported by me) on OSV services and made CVE-fetching via purl possible.
So right now it is possible to fetch the CVE's (CVE-2020-6097, CVE-2021-46671 and CVE-2021-41054) belonging to atftp for the given version from Google OSV with:
curl -d '{"package": {"purl": "pkg:deb/ubuntu/atftp@0.7.git20120829-3.1~0.18.04.1"}}' https://api.osv.dev/v1/query

I would be great if DependencyTrack could fetch the CVE's with a SBOM with this PURL in it.

Let me know how I can help to figure this problem out.

### Dependency-Track Version

4.12.1

### Dependency-Track Distribution

Container Image

### Database Server

[...]

Issue #1374, opened 2022-02-07
openssl - ubuntu package - versions not matched correctly


kmoens:
### Current Behavior:
We have a cycloneDX BOM file, as generated by syft for our Docker containers. These Docker containers contain the OpenSSL library. 

This is contained in the BOM file as follows:
```
    <component type="library">
      <publisher>Ubuntu Developers &lt;ubuntu-devel-discuss@lists.ubuntu.com&gt;</publisher>
      <name>openssl</name>
      <version>1.1.1f-1ubuntu2.4</version>
      <cpe>cpe:2.3:a:openssl:openssl:1.1.1f-1ubuntu2.4:*:*:*:*:*:*:*</cpe>
      <purl>pkg:deb/ubuntu/openssl@1.1.1f-1ubuntu2.4?arch=amd64</purl>
      <properties>
        <property name="foundBy">dpkgdb-cataloger</property>
        <property name="type">deb</property>
        <property name="metadataType">DpkgMetadata</property>
        <property name="path">/var/lib/dpkg/status</property>
        <property name="layerID">sha256:5a1aaa0fc304f17de4904b251554ad1d18d210122a495f40a726418f101018ec</property>
        <property name="path">/var/lib/dpkg/info/openssl.md5sums</property>
        <property name="layerID">sha256:d0af6cd0417d041e4be905c5344ff2d88c58294fd7e1eba1fc19d8dba795e1d3</property>
        <property name="path">/var/lib/dpkg/info/openssl.conffiles</property>
        <property name="layerID">sha256:d0af6cd0417d041e4be905c5344ff2d88c58294fd7e1eba1fc19d8dba795e1d3</property>
        <property name="path">/usr/share/doc/libssl1.1/copyright</property>
        <property name="layerID">sha256:d0af6cd0417d041e4be905c5344ff2d88c58294fd7e1eba1fc19d8dba795e1d3</property>
        <property name="installedSize">1257</property>
      </properties>
    </component>
```

When this BOM is uploaded, we get a number of vulnerabilities reported, including CVE-2019-1543. This CVE is stated on NVD as:
"Fixed in OpenSSL 1.1.1c (Affected 1.1.1-1.1.1b). Fixed in OpenSSL 1.1.0k (Affected 1.1.0-1.1.0j).",

Similarly we found this info in the database as follows:
```
[...]

PR #5783, opened 2026-02-04
Consider OS distro during vulnerability matching

nscuro:
### Description

Considers OS distro during vulnerability matching.

* Where possible, enriches an affected package's PURL with `distro` qualifier inferred from the package's `ecosystem`. e.g. `ecosystem=Debian:11` becomes `distro=debian-11`, `ecosystem=Ubuntu:20.04:LTS` becomes `distro=ubuntu-20.04` etc.
* During vulnerability analysis, if both component and matching criteria have a PURL `distro` qualifier, ensures they match. Matching can handle codename <-> version comparisons, e.g. for Ubuntu `focal` would match `20.04` and vice versa.
* Generally improves performance of OSV mirroring by using fewer transactions and disabling ORM features that caused expensive unnecessary queries.

Currently Alpine, Debian, and Ubuntu distribution matching is implemented. These seem to work for SBOMs generated with Trivy and Syft.

The codename <-> version mapping is currently hardcoded for Debian and Ubuntu. There is a fallback mechanism that will handle exact matches, such that when Debian publishes a hypothetical "foo" release, we can still match components with vulnerabilities if both `distro` qualifiers are exactly "foo".

Debian and Ubuntu provide CSV which we could regularly fetch at runtime, but this involves more work to coordinate.

### Addressed Issue

Fixes https://github.com/DependencyTrack/dependency-track/issues/1374 
Fixes https://github.com/DependencyTrack/dependency-track/issues/5776 
Fixes https://github.com/DependencyTrack/dependency-track/issues/4445 
Fixes https://github.com/DependencyTrack/dependency-track/issues/4725

### Additional Details

A container image for this PR is available at `dependencytrack/bundled:feature-distro-matching`. 

OSV is enabled by default and configured to mirror the `Alpine`, `Debian`, and `Ubuntu` ecosystems.

### Checklist

### DIFF
--- a/dev/docker-compose.postgres.yml
+++ b/dev/docker-compose.postgres.yml
@@ -28,6 +28,11 @@ services:
 
   postgres:
     image: postgres:14-alpine
+    command: >-
+      -c 'shared_preload_libraries=pg_stat_statements'
+      -c 'pg_stat_statements.track=all'
+      -c 'pg_stat_statements.max=10000'
+      -c 'track_activity_query_size=2048'
     environment:
       POSTGRES_DB: "dtrack"
       POSTGRES_USER: "dtrack"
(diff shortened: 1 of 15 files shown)


nscuro:
Testing with cdxgen, scalibr, Syft, and Trivy. 

* I can't get cdxgen to generate container image SBOMs that contain OS packages. 
* SBOMs with scalibr and Trivy appear to be effectively free from false positives. 
* SBOMs generated by Syft still lead to false positives due to them including CPEs, which then get matched against the NVD which does not provide distro qualifiers.

Issue #1362, opened 2022-02-01
Dependency-track V4.3.6 is not reporting vulnerability for protobuf-java (3.19.0/3.19.1)


kaushikjeet:
### Current Behavior:
Our software has dependency on protobuf-java component using maven, we are using version 3.19.1 but dependency track unable to detect reported vulnerabilities against given version, maven repositories list reposted vulnerability against this version.

https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java/3.19.1

### Steps to Reproduce:
1. download attached generated sbom for dependency-track ([sbom_arc.zip](https://github.com/DependencyTrack/dependency-track/files/7977064/sbom_arc.zip)
2. import in dependency-track server.
3. No vulnerabilities shown again protobuf-java: 3.19.1 version.

### Expected Behavior:
Dependency-track should display reported (CVE-2021-22569) vulnerability against given component.

### Environment:
  
  - Dependency-Track Version: V4.3.6
  - Distribution: [ Docker | Executable WAR | Traditional WAR ]
  - BOM Format & Version: CycloneDX 1.3
  - Database Server: PostgreSQL
  - Browser: Chrome Version 97.0.4692.99

### Additional Details:

We expect all components used in our product tracked for any reported vulnerabilities which we can trust and improve application security.

stevespringett:
Sonatype OSS Index is not reporting this library to be vulnerable.
https://ossindex.sonatype.org/component/pkg:maven/com.google.protobuf/protobuf-java@3.19.1

Please report false positives and false negatives for OSS Index to the OSS Index project. 
https://ossindex.sonatype.org/doc/report-vulnerability

PR #5827, opened 2026-02-19
projects: show ancestor path

valentijnscholten:
### Description

When searching for projects/components/vulnerabilities there's no visual indication of where in the Project Hierarchy a project is shown. 

### Addressed Issue
(partly) fixes #1353, #1362

### Additional Details
In v4 we need to do perform some additional java code to get the "ancestor path" populated. But in v5 this will be easier due to the materialized view being used there. Happy to forward port the feature to v5.

I haven't created tests yet as I first would like to get confirmation this is something that can be added to v4. I know v5 is around the corner. At the same it might take some time before v5 is released as GA and everybody has moved over. This PR fixes a real need in v4.

FrontEnd companion PR: https://github.com/DependencyTrack/frontend/pull/1448

### Considerations
Having the full hierarchy view when displaying ony a subset/searchresult of projects is non-trivial to implement if we take features such as paging into account. Maybe for v5 someone has a better proposal, for now an additional column seems helpful in v4.

### Checklist

Screenshot 2026-02-01 200245

Screenshot 2026-02-01 200227

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/FindingsSearchQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/FindingsSearchQueryManager.java
@@ -27,6 +27,7 @@ import org.dependencytrack.model.Analysis;
 import org.dependencytrack.model.Component;
 import org.dependencytrack.model.Finding;
 import org.dependencytrack.model.GroupedFinding;
+import org.dependencytrack.model.Project;
 import org.dependencytrack.model.RepositoryMetaComponent;
 import org.dependencytrack.model.RepositoryType;
 import org.dependencytrack.model.Vulnerability;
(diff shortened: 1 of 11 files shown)


valentijnscholten:
@nscuro I have now refactored both PRs to show the ancestor path as tooltip instead of column. Not perfect, but better than nothing and maybe good to be accepted. I've also refactored the api reponse to just return the chain of parents in the parent block instead of a new ancestor path element.

Global Policy Audit doesn't have the full parent chain as that endpoint returned detached JDO Object which only have the direct parent. If we all like the tooltip approach and/or the parent chain response format I could look at refactoring this endpoint to use value objects instead of detached objects.

Plot twist: Should we just render the tooltip on the fly using an API call for that view (or all views)? (To avoid having to build the full parent chain in the backend for every api response containing projects)

Issue #5343, opened 2025-09-29
Enable OSV vulnerability source per default


nscuro:
### Current Behavior

We previously enabled the OSS Index analyzer per default, since it allowed for unauthenticated usage.

The other vulnerability source that is enabled per default is the NVD. It however does not provide PURL matching data and is thus not of great help during vulnerability analysis.

We need another, PURL-based alternative that can be enabled per default.

### Proposed Behavior

Enable the OSV integration per default.

It provides PURL matching data, does not require authentication, and has broad ecosystem coverage.

Consider limiting the default to a small selection of popular ecosystems.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

wkoot:
Enabling by default seems like a terrible idea while the implementation is still in beta. It would however be convenient to configure the source via `application.properties` and environment variables for container-based deployments

nscuro:
I have been pondering whether we should add an OSV *analyzer* instead, which uses [the OSV API](https://osv.dev/#use-the-api) rather than mirroring the data, to bridge this gap.

For the following reasons:

* It would be usable immediately since it doesn't require any configuration or data downloads.
* The component version matching logic is offloaded to the OSV service.
* There is no need (or even possibility) to configure ecosystems. If OSV supports the PURL being sent, it will return vulnerabilities.

wkoot:
That sounds like a better option, as starting a fresh instance of Dependency-Track can take days to fetch all OSV data.
In any case it would be nice to be able to configure this outside of the GUI, especially since changes POSTed don't seem to take effect (or, at least not within a short timespan)

cachescrubber:
Please consider https://github.com/DependencyTrack/dependency-track/discussions/5785 (Discussion) as it seems related to this issue. I enabled OSV on top of NVD (Datasource) + Internal (analyzer) + Sonatype Oss Index (analyzer) and get duplicate findings.

lukas-braune:
> I have been pondering whether we should add an OSV _analyzer_ instead, which uses [the OSV API](https://osv.dev/#use-the-api) rather than mirroring the data, to bridge this gap.

@nscuro It would be helpful to keep the existing mirroring option for DT instances that aren’t allowed to call external APIs with component information.

[...]

PR #5903, opened 2026-03-12
Fix PURL-specific version matching being bypassed for components with CPE

nscuro:
### Description

Fixes PURL-specific version matching being bypassed for components with CPE.

The previous logic did not explicitly "parse" the component's PURL when the component also has a CPE. It would still go on to fetch VulnerableSoftware records for CPE *and* PURL as expected, but would not follow the PURL-specific version comparison algorithms.

A leftover logic that pre-dates the introduction of vers caused version range analysis to be performed against VulnerableSoftware records that had no CPE information (i.e. maybeMatchCpe returns null). VulnerableSoftware records with PURL information were thus still considered, but in a logic branch that doesn't leverage ecosystem-aware or distro-aware matching.

### Addressed Issue

N/A

### Additional Details

Credit to @xavier-calland for identifying this as per https://github.com/DependencyTrack/dependency-track/issues/5343#issuecomment-4046427496

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java
+++ b/src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java
@@ -37,6 +37,7 @@ import us.springett.parsers.cpe.Cpe;
 import us.springett.parsers.cpe.util.Relation;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 
 /**
@@ -87,7 +88,7 @@ public abstract class AbstractVulnerableSoftwareAnalysisTask extends BaseCompone
             Component component,
             VulnerabilityAnalysisLevel vulnerabilityAnalysisLevel) {
         for (final VulnerableSoftware vs : vsList) {
-            if (comparePurlVersions(targetPurl, vs, targetVersion)) {
+            if (matchesPurl(vs, targetPurl) && comparePurlVersions(targetPurl, vs, targetVersion)) {
                 if (vs.getVulnerabilities() != null) {
                     for (final Vulnerability vulnerability : vs.getVulnerabilities()) {
                         NotificationUtil.analyzeNotificationCriteria(qm, vulnerability, component,
@@ -107,8 +108,7 @@ public abstract class AbstractVulnerableSoftwareAnalysisTask extends BaseCompone
             Component component,
             VulnerabilityAnalysisLevel vulnerabilityAnalysisLevel) {
         for (final VulnerableSoftware vs : vsList) {
-            final Boolean isCpeMatch = maybeMatchCpe(vs, targetCpe, targetVersion);
-            if ((isCpeMatch == null || isCpeMatch) && compareCpeVersions(vs, targetVersion, component)) {
+            if (matchesCpe(vs, targetCpe, targetVersion) && compareCpeVersions(vs, targetVersion, component)) {
                 if (vs.getVulnerabilities() != null) {
                     for (final Vulnerability vulnerability : vs.getVulnerabilities()) {
                         NotificationUtil.analyzeNotificationCriteria(qm, vulnerability, component, vulnerabilityAnalysisLevel);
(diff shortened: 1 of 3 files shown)


src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java
```diff
@@ -87,7 +88,7 @@ protected void analyzePurlVersionRange(
             Component component,
             VulnerabilityAnalysisLevel vulnerabilityAnalysisLevel) {
         for (final VulnerableSoftware vs : vsList) {
-            if (comparePurlVersions(targetPurl, vs, targetVersion)) {
+            if (matchesPurl(vs, targetPurl) && comparePurlVersions(targetPurl, vs, targetVersion)) {
```
xavier-calland, line 91: @nscuro 
Is it possible for a vulnerability source to provide both CPE and PURL?

If so, what should the expected behavior be for a component that also has both CPE and PURL when only one of the two matches?

When both CPE and PURL match, would the finding be reported twice (once in `analyzeCpeVersionRange` and once in `analyzePurlVersionRange`)?

xavier-calland:
@nscuro 
I’d like to share another thought about managing vulnerability aliases.
Let me know if it's not the right place.

What would be the "correct" behavior when, for two vulnerability aliases, a component is identified as vulnerable in one but not in the other?

To reduce false positives, could a solution be to allow setting priorities among vulnerability sources and, in case of aliases, prioritize the analysis from the highest-priority source?

For example, if I set the priority as OSV first, then NVD, for analyzing the vulnerability **`NVD CVE-2025-15467`** and its alias **`OSV UBUNTU-CVE-2025-15467`**:

- For a component with both **CPE and PURL**, the analysis of **`NVD CVE-2025-15467`** is skipped, and priority is given to **`OSV UBUNTU-CVE-2025-15467`** because analysis with PURL is possible.
- For a component with only **CPE**, the analysis of **`NVD CVE-2025-15467`** is performed, while the analysis of **`OSV UBUNTU-CVE-2025-15467`** is not possible due to the lack of PURL.

Does this approach seem promising, or does the risk of false negatives appear too high?

src/main/java/org/dependencytrack/tasks/scanners/AbstractVulnerableSoftwareAnalysisTask.java
nscuro: > Is it possible for a vulnerability source to provide both CPE and PURL?

Technically yes, but practically none does today. Vulnerable CPE and PURL details are stored in separate `VulnerableSoftware` records in any case.

> When both CPE and PURL match, would the finding be reported twice (once in analyzeCpeVersionRange and once in analyzePurlVersionRange)?

No, if it's the exact same vulnerability, it will only be added once.

xavier-calland:
> @nscuro I’d like to share another thought about managing vulnerability aliases. Let me know if it's not the right place.
> 
> What would be the "correct" behavior when, for two vulnerability aliases, a component is identified as vulnerable in one but not in the other?
> 
> To reduce false positives, could a solution be to allow setting priorities among vulnerability sources and, in case of aliases, prioritize the analysis from the highest-priority source?
> 
> For example, if I set the priority as OSV first, then NVD, for analyzing the vulnerability **`NVD CVE-2025-15467`** and its alias **`OSV UBUNTU-CVE-2025-15467`**:
> 
>     * For a component with both **CPE and PURL**, the analysis of **`NVD CVE-2025-15467`** is skipped, and priority is given to **`OSV UBUNTU-CVE-2025-15467`** because analysis with PURL is possible.
> 
>     * For a component with only **CPE**, the analysis of **`NVD CVE-2025-15467`** is performed, while the analysis of **`OSV UBUNTU-CVE-2025-15467`** is not possible due to the lack of PURL.
> 
> 
> Does this approach seem promising, or does the risk of false negatives appear too high?

@nscuro 
Were you able to see this proposed feature?
What do you think of it, would this be an acceptable behavior?
I can take a look at it and work on it if you'd like.

PR #5971, opened 2026-03-26
Add DefectDojo auto create context

### LINKED ISSUE #1323: Possibility for DependencyTrack to Create a DefectDoJo Product and Engagement
Would it be possible to add the logic and webhooks for DependencyTrack to call the DefectDoJo API and create a Product with a CI/CD enguagement, then update the DependancyTrack Pjject by adding the integration config.

We have created a Python Flask proxy to glue this workflow together.  

### Current Behavior:
We have many Products and teams working together, and as part of the on-boarding for DependencyTrack we manually need to Create over 2000 Products in DefectDoJo then manually configure the DependencyTrack integration.  

### Proposed Behavior:
When a new BOM is processed, the webhook is fired that connects to the DefectDoJo API creating the new Product (if one doesn't already exist by the same name), and create the CICD engagement.  The response from DefectDoJo includes the engagement ID which can then be processed by DependencyTrack and added to the integration project property for the project that consume the BOM.

mjwrona:
### Description

Adds usage of DefectDojos `auto_create_context` import api flag so that users don't have to build their own glue code to use this integration at scale.

**Global Configuration**
Introduces 4 new global settings
| Property Name | Default Value | Description |
|--------------|---------------|-------------|
| `defectdojo.autocreate.enabled` | `false` | Enable/disable auto context creation |
| `defectdojo.autocreate.engagementName` | `dependencytrack` | Default engagement name for all projects |
| `defectdojo.autocreate.productTypeName` | `Dependency Track` | Default product type name for all projects |
| `defectdojo.autocreate.deduplicationOnEngagement` | `false` | Enable deduplication at engagement level instead of product level |

**Project Properties**
Introduces 4 new project properties that allow to override product name, the global deduplicationOnEngagement, engagement and product type name
| Attribute      | Value                             |
| ---------------| --------------------------------- |
| Group Name     | `integrations`                    |
| Property Name  | `defectdojo.autocreate.productName`         |
| Property Value | Custom product name (defaults to project name if not set) |
| Property Type  | `STRING`                          |

| Attribute      | Value                             |
| ---------------| --------------------------------- |
| Group Name     | `integrations`                    |
| Property Name  | `defectdojo.autocreate.engagementName`      |
| Property Value | Custom engagement name (defaults to global config if not set) |
| Property Type  | `STRING`                          |

| Attribute      | Value                             |
| ---------------| --------------------------------- |
| Group Name     | `integrations`                    |
| Property Name  | `defectdojo.autocreate.productTypeName`     |
| Property Value | Custom product type name (defaults to global config if not set) |
| Property Type  | `STRING`                          |

| Attribute      | Value                             |
| ---------------| --------------------------------- |
| Group Name     | `integrations`                    |
| Property Name  | `defectdojo.autocreate.deduplicationOnEngagement`     |
| Property Value | `true` or `false` (defaults to global config if not set) |
| Property Type  | `BOOLEAN`                         |

**UI Change:**
image

### Addressed Issue

Closes #1323

### Additional Details

See https://docs.defectdojo.com/automation/api/api-v2-docs/#import

### Checklist

<del>- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective </del>
<del>- [ ] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade) </del>


mjwrona:
Frontend PR: https://github.com/DependencyTrack/frontend/pull/1480

valentijnscholten:
Great to see this added as it makes the Defect Dojo integration easier and better. There are some more parameters that can be provided with `auto_create_context`. The following two might be interesting to add:

- `test_title`
- `deduplication_on_engagement`

Do you think this PR would a good opportunity to add these two?

mjwrona:
Sure I just need a bit more context about those settings. 
Would you expect that as global config, project property or both?
What should be the default for those settings to not introduce a breaking change?

valentijnscholten:
The `test_title` should be optional and not sent by default. In that case Defect Dojo will use the name of the parser as title for the test which is "Dependency Track Finding Packaging Format (FPF) Export". By providing a `test_title` value you can choose a more useful value that provides some context, for example the name + version of the project in Depdendency Track. That might even be a good new default, but would break for users using the reimport option.

The `deduplication_on_engagement` value should be empty/left out by default. The default value is `False` in Defect Dojo. By providing `True` from Dependency Track it will instruct Defect Dojo to use a narrow deduplication scope. It goes too far to explain all the intricacies around that here. But allowing users to set it to `True` if needed is helpful.

mjwrona:
@valentijnscholten I added the `deduplication_on_engagement` flag to global and project config.
For `test_title` - this is already configurable on project level using `defectdojo.testTitle`. The switch to a dynamic value like name + version does not only target auto create context but the full integration so i wouldn't implement that in this PR.

@nscuro what are your thoughts on that?

AndreVirtimo:
Thank you for this PR. This has been on my wishlist for a long time.

I propose using only the reimport-scan API. There is no need to use the import-scan API. As far as I know**,** the reimport-scan API is the best option for tool integration and CI/CD pipelines.

This would reduce the complexity of the DD integration.

The reimport-scan API offers additional options. Maybe a more generic mapping between project properties and API options would provide more flexibility. All properties with the prefix `defectdojo.reimport.` could be mapped. So `defectdojo.autocreate.productName` would be `defectdojo.reimport.product_name`. Then I could also use options like `defectdojo.reimport.push_to_jira`.

PR #6085, opened 2026-05-01
Add project filters to component identity search

### LINKED ISSUE #4570: Components Page: Filter by Project Properties
### Current Behavior

Using the Component Page (`/components`) to perform a search will return components from inactive projects....  with not even any indication of which are active and which are inactive. This can result in a lot of noise in the results (noise that increases over time as more projects are "archived").

This is arguably a defect, but thinking more about the problem indicates that having an ability to filter search results by additional project properties (in addition to "inactive/active" would also be very useful.

### Proposed Behavior

The components screen should provide the ability to filter search results by:

* Active projects
* Project is configured as "Latest"
* Project is of type "Collection"  (new functionality to be introduced in DT v4.13.0)
* Project has tag X  (eg, filter by tag `production` or `staging`... each of which might be slightly older than "Latest" )

WRT implementation in Frontend, there is a desire to keep things clean and present a UI that is easy to learn & use.  I believe that the graphs ("Portfolio Vulnerabilities", etc) that are currently displayed at the top of the components screen "bring nothing to the party" and could be dropped.  This would free-up space to offer something else instead...such as a "non-cluttered" UI for specifying filters.

### Checklist

hoobio:
### Description

Adds `excludeInactiveProjects` and `onlyLatestProjectVersions` query parameters to `GET /v1/component/identity` so the Components page can filter out archived projects and older versions during portfolio-wide search.

### Addressed Issue

Closes #4570

### Additional Details

The new parameters combine with the existing `project` UUID filter and with all search modes (name, purl, cpe, swidTagId). Both default to `false`, so existing callers see no change.

Naming follows existing conventions in the codebase:

- `excludeInactiveProjects` mirrors `excludeInactive` from `ProjectResource`, `AccessControlResource`, and `VulnerabilityResource`. The `Projects` suffix disambiguates on this `/component` endpoint, where `excludeInactive` could otherwise read as filtering on component state.
- `onlyLatestProjectVersions` mirrors the existing `Policy.onlyLatestProjectVersion` field used by `PolicyEngine` for the same "scope to latest project versions" semantic, in plural form per PR review feedback.

Frontend toggle UI: DependencyTrack/frontend#1505

### Checklist

- ~~This PR introduces changes to the database model, and I have added corresponding update logic~~
- [x] This PR introduces new or alters existing behavior, and I have updated the documentation accordingly

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
@@ -206,7 +206,7 @@ public class ComponentResource extends AlpineResource {
                                            @Parameter(description = "When true, only return components from active projects")
                                            @QueryParam("excludeInactiveProjects") boolean excludeInactiveProjects,
                                            @Parameter(description = "When true, only return components from projects flagged as the latest version")
-                                           @QueryParam("onlyLatestProjectVersion") boolean onlyLatestProjectVersion) {
+                                           @QueryParam("onlyLatestProjectVersions") boolean onlyLatestProjectVersions) {
         try (QueryManager qm = new QueryManager(getAlpineRequest())) {
             Project project = null;
             if (projectUuid != null) {
@@ -233,7 +233,7 @@ public class ComponentResource extends AlpineResource {
                     && identity.getPurl() == null && identity.getCpe() == null && identity.getSwidTagId() == null) {
                 return Response.ok().header(TOTAL_COUNT_HEADER, 0).build();
             } else {
-                final PaginatedResult result = qm.getComponents(identity, project, true, excludeInactiveProjects, onlyLatestProjectVersion);
+                final PaginatedResult result = qm.getComponents(identity, project, true, excludeInactiveProjects, onlyLatestProjectVersions);
                 return Response.ok(result.getObjects()).header(TOTAL_COUNT_HEADER, result.getTotal()).build();
             }
         }
--- a/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
@@ -202,7 +202,11 @@ public class ComponentResource extends AlpineResource {
                                            @Parameter(description = "The swidTagId of the component")
                                            @QueryParam("swidTagId") String swidTagId,
                                            @Parameter(description = "The project the component belongs to", schema = @Schema(type = "string", format = "uuid"))
-                                           @QueryParam("project") @ValidUuid String projectUuid) {
+                                           @QueryParam("project") @ValidUuid String projectUuid,
+                                           @Parameter(description = "Optionally restrict results to components belonging to active projects only")
+                                           @QueryParam("onlyActive") boolean onlyActive,
+                                           @Parameter(description = "Optionally restrict results to components belonging to projects flagged as latest version only")
+                                           @QueryParam("onlyLatestVersion") boolean onlyLatestVersion) {
         try (QueryManager qm = new QueryManager(getAlpineRequest())) {
             Project project = null;
             if (projectUuid != null) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
```diff
@@ -202,7 +202,11 @@ public Response getComponentByIdentity(@Parameter(description = "The group of th
                                            @Parameter(description = "The swidTagId of the component")
                                            @QueryParam("swidTagId") String swidTagId,
                                            @Parameter(description = "The project the component belongs to", schema = @Schema(type = "string", format = "uuid"))
-                                           @QueryParam("project") @ValidUuid String projectUuid) {
+                                           @QueryParam("project") @ValidUuid String projectUuid,
+                                           @Parameter(description = "When true, only return components from active projects")
+                                           @QueryParam("onlyActive") boolean onlyActive,
+                                           @Parameter(description = "When true, only return components from projects flagged as the latest version")
+                                           @QueryParam("onlyLatestVersion") boolean onlyLatestVersion) {
```
nscuro: The parameter names are ambiguous, given this is an endpoint scoped to *components*. `onlyActive` then reads like "only active components", and `onlyLatestVersion` reads like "only latest component version".

We should also try to align with similar parameters in other endpoints, e.g. `/api/v1/project` uses `excludeInactive`. That could become `excludeInactiveProjects` here.

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
hoobio: Good catch, renamed in [22508cf](https://github.com/DependencyTrack/dependency-track/pull/6085/commits/22508cf28).

- `onlyActive` -> `excludeInactiveProjects` (mirrors `excludeInactive` on `ProjectResource`, `AccessControlResource`, `VulnerabilityResource`, with `Projects` suffix to disambiguate on this endpoint)
- `onlyLatestVersion` -> `onlyLatestProjectVersion` (mirrors the existing `Policy.onlyLatestProjectVersion` field used by `PolicyEngine` for the same "scope to latest project version" semantic)

Pulled the matching frontend across in DependencyTrack/frontend#1505 too. While I was there I also relabelled the two toggles to "Show inactive projects" / "Show non-latest project versions" (both default on), which matches the existing `show_inactive_projects` switch on `ProjectList`, `PolicyViolationAudit`, `VulnerabilityAuditByOccurrence`, `VulnerabilityAuditGroupedByVulnerability`, and `AffectedProjects`.

src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
```diff
@@ -202,7 +202,11 @@ public Response getComponentByIdentity(@Parameter(description = "The group of th
                                            @Parameter(description = "The swidTagId of the component")
                                            @QueryParam("swidTagId") String swidTagId,
                                            @Parameter(description = "The project the component belongs to", schema = @Schema(type = "string", format = "uuid"))
-                                           @QueryParam("project") @ValidUuid String projectUuid) {
+                                           @QueryParam("project") @ValidUuid String projectUuid,
+                                           @Parameter(description = "When true, only return components from active projects")
+                                           @QueryParam("excludeInactiveProjects") boolean excludeInactiveProjects,
+                                           @Parameter(description = "When true, only return components from projects flagged as the latest version")
+                                           @QueryParam("onlyLatestProjectVersion") boolean onlyLatestProjectVersion) {
```
nscuro: I think it should be "onlyLatestProjectVersion**s**". May be a little pedantic but best to get this aligned before we commit to it.

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
hoobio: Renamed in [d32dcdf](https://github.com/DependencyTrack/dependency-track/pull/6085/commits/d32dcdf7a).

hoobio:
Both points addressed in [d32dcdf](https://github.com/DependencyTrack/dependency-track/pull/6085/commits/d32dcdf7a). Frontend follow-up in DependencyTrack/frontend#1505. Ready for another look @nscuro.

nscuro (approved):
Thanks!

PR #6138, opened 2026-05-15
Extending length of PURL and PURLCOORDINATES columns from 786 to 4096

mrgxqq:
### Description

Increased the maximum length of the `PURL` and `PURLCOORDINATES` fields.

In some BOM files generated on Windows environments, the `PURL` value can become excessively long — in certain cases exceeding the current limit of 786 characters. As a result, importing such BOM files into DependencyTrack fails with the following error:

> `in column "PURLCOORDINATES" that has maximum length of 786. Please correct your data!`

This fix increases the allowed field length and resolves the issue when importing such BOM files.

A similar issue has also been discussed previously in `#3560`.

### Addressed Issue

fixes #3560

### Additional Details

The issue was reproducible with BOM files generated on Windows, where the path contained in the `PURL` became significantly longer than expected.

Due to the length limitation of the `PURLCOORDINATES` database field, the database rejected the record during BOM import.

This change is limited to increasing the allowed data size and does not affect the existing PURL processing logic.

### Checklist

### DIFF
--- a/docs/_posts/2026-04-03-v4.14.1.md
+++ b/docs/_posts/2026-04-03-v4.14.1.md
@@ -21,6 +21,15 @@ type: patch
 * Avoid NPE when computing Trivy pkgType - [apiserver/#5987]
 * Remove leading whitespace from vulnerability badge SVG template - [apiserver/#6000]
 * Fix Japanese Trivy analyzer strings - [frontend/#1489]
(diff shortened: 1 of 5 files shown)


nscuro:
Thanks for the PR, but this will break for MSSQL, as mentioned in the PR you linked. The problem is that this column is indexed, and some RDBMSes cannot deal with indexes on columns that large.

mrgxqq:
Thank you for the explanation.

PR #6181, opened 2026-05-28
Include analysis.detail in DefectDojo finding payload

### LINKED ISSUE #6169: Forward `vulnerability.analysis.detail` to DefectDojo
### Current Behavior

Findings forwarded by the DefectDojo integration include the standard fields (title, severity, component name and version, CVE/GHSA identifiers, mitigation, references) plus a single `file_path` that holds the bare artifact filename. The rich audit text stored in `vulnerability.analysis.detail` on the corresponding DT vulnerability is not part of the payload that DefectDojo receives.

### Proposed Behavior

Include `vulnerability.analysis.detail` in the DefectDojo finding the publisher creates. The natural target is either an extension of the `description` field (appended after a separator so existing parsers keep working) or a dedicated field that DefectDojo's parser can render alongside the description.

This would let downstream auditors see the same context DT shows in its audit UI without context switching back to DT for the same vulnerability.

## Why this matters

In setups where DefectDojo is the primary triage surface and DT is the SBOM ingestion engine, the `analysis.detail` text is where the actionable per-finding information lives. Typical content we put there:

- Severity and CVSS at the top for quick scanning
- All `depscan:*` properties (`insights`, `prioritized`, etc.)
- Usage locations from the companion SBOM (`evidence.occurrences[].location` with `#line` anchors for js and ts, consuming projects derived from each component's `SrcFile` properties for dotnet, `ImportedModules` when present)
- Atom call paths from `*reachables.slices.json` when reachability analysis ran

Once a finding is in DefectDojo, none of this is reachable from there today.

## Related

#5843 (closed) requested similar enrichment for CVSS vectors, references, and publish date. This issue is in the same direction (more vulnerability metadata from DT reaching DefectDojo) but covers the audit text specifically.

## Environment

- Dependency-Track version: 4.14
- DefectDojo version: v.2.56.4
- Integration mechanism: built-in DefectDojo publisher driven by project `integrations:defectdojo.engagementId` property

### Checklist

webdevred:
### Description

When DefectDojo is used as the primary triage surface and Dependency-Track as the SBOM ingestion engine, the per-finding audit text stored in `analysis.detail` never made it into the payload sent to DefectDojo. Auditors working in DefectDojo had no way to see that context without switching back to DT for every finding.

This adds `ANALYSIS."DETAILS"` to the SQL queries that back the Finding model, which makes the field flow into the Finding Packaging Format document that the DefectDojo integration uploads. The field appears as `analysis.detail` alongside the existing `analysis.state` and `analysis.isSuppressed` fields.

One option considered was prepending the audit detail directly to `vulnerability.description` in the FPF, which would have made it visible in DefectDojo without any changes on that side. That was rejected because `vulnerability.description` is CVE data and mixing analyst notes into it would affect any other consumer of the FPF. Adding it as a separate field keeps the concerns separated and lets each consumer decide how to present it. A companion PR to DefectDojo appends it to the finding description there.

One pre-existing bug is also fixed here: after the new column was inserted at position 36, the `QUERY_ALL_FINDINGS` call site in `FindingsSearchQueryManager` was reading the project UUID from the wrong index. This would have caused a runtime failure in the global findings view for any project with a finding that has audit detail set.

image
image

Related to https://github.com/DefectDojo/django-DefectDojo/pull/14931

### Addressed issue

Closes #6169

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/Finding.java
+++ b/src/main/java/org/dependencytrack/model/Finding.java
@@ -96,6 +96,7 @@ public class Finding implements Serializable {
                  , "FINDINGATTRIBUTION"."REFERENCE_URL"
                  , "ANALYSIS"."STATE"
                  , "ANALYSIS"."SUPPRESSED"
+                 , "ANALYSIS"."DETAILS"
               FROM "COMPONENT"
              INNER JOIN "COMPONENTS_VULNERABILITIES"
                 ON "COMPONENT"."ID" = "COMPONENTS_VULNERABILITIES"."COMPONENT_ID"
@@ -151,6 +152,7 @@ public class Finding implements Serializable {
                  , "FINDINGATTRIBUTION"."REFERENCE_URL"
                  , "ANALYSIS"."STATE"
                  , "ANALYSIS"."SUPPRESSED"
+                 , "ANALYSIS"."DETAILS"
                  , "PROJECT"."UUID"
                  , "PROJECT"."NAME"
                  , "PROJECT"."VERSION"
@@ -245,10 +247,15 @@ public class Finding implements Serializable {
 
         optValue(analysis, "state", o[34]);
         optValue(analysis, "isSuppressed", o[35], false);
+        if (o[36] instanceof final Clob clob) {
+            optValue(analysis, "detail", toString(clob));
+        } else {
+            optValue(analysis, "detail", o[36]);
+        }
 
-        if (o.length > 36) {
-            optValue(component, "projectName", o[37]);
-            optValue(component, "projectVersion", o[38]);
+        if (o.length > 37) {
+            optValue(component, "projectName", o[38]);
+            optValue(component, "projectVersion", o[39]);
         }
     }
 
--- a/docs/_docs/integrations/defectdojo.md
+++ b/docs/_docs/integrations/defectdojo.md
@@ -14,6 +14,16 @@ Dependency-Track accomplishes this in the following ways:
 * Dependency-Track pushes findings to DefectDojo on a periodic basis (configurable)
 * DefectDojo parses Dependency-Track findings
 
+### Forwarded analysis data
+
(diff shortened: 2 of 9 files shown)


nscuro:
Hi @webdevred, please re-target your PR at either `4.14.x` (for v4) or `main` (for v5) as per https://github.com/DependencyTrack/dependency-track/blob/main/V5_MIGRATION.md#contributors-and-pr-authors.

src/main/java/org/dependencytrack/model/Finding.java
nscuro, line 1: Please bump the FPF version accordingly:

https://github.com/DependencyTrack/dependency-track/blob/03f1364c7a296b133e243bf482553dcc2526e6a7/apiserver/src/main/java/org/dependencytrack/integrations/FindingPackagingFormat.java#L46-L49

And add an entry for the change here:

https://github.com/DependencyTrack/dependency-track/blob/fadd195f16ac0fce30c88e014749ce951d8e402e/docs/_docs/integrations/file-formats.md?plain=1#L29-L40

nscuro (changes requested):

src/main/java/org/dependencytrack/model/Finding.java
webdevred: Hi,
Thanks for the feedback!
What do you think of this entry?

> Finding Packaging Format v1.4 is not yet available in a stable release of Dependency-Track
> It adds an optional `detail` field to `analysis` objects, which contains any analyst notes recorded against the finding.

Best regards,
webdevred

nscuro: Strip the "Finding Packaging Format v1.4 is not yet available in a stable release of Dependency-Track" part, it will go stale and we'll likely forget to remove it once this is released. Otherwise looks good.

nscuro (approved):

Issue #6183, opened 2026-05-15
Add `isDirect` flag in `COMPONENT` for efficient direct/transitive dependency filtering


sahibamittal:
### Current Behavior

Currently, at large scale, determining whether a component is direct or transitive dependency requires evaluating JSONB containment check against `project.direct_dependencies`. It becomes expensive for cross-project reporting and bulk analytics. 

### Proposed Behavior

Add `IS_DIRECT` to `COMPONENT`. This will require setting it during BOM import, can be a bulk-update for all components in a project.

All read paths can be updated by replacing JSONB containment checks with boolean filter.

New field will be nullable to avoid heavy migration for existing components and can be populated upon next bom upload/publish.

Optional: Expose the same in UI.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/hyades/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/hyades/issues) for whether this enhancement was already requested

nscuro:
Adjacent to / duplicate of DependencyTrack/dependency-track#6184?

sahibamittal:
> Adjacent to / duplicate of [#2118](https://github.com/DependencyTrack/dependency-track/issues/6184)?

CEL function will be useful when querying for a specific component and project. In order to get this data for all components/projects (for reporting/analytics), having this flag stored will be more feasible.

sahibamittal:
Upon discussion internally, requirements have changed, this issue can be closed.

PR #6206, opened 2026-05-29
Add `direct_dependency` flag to component for efficient direct/transitive dependency filtering

sahibamittal:
### Description

Add `direct` dependency flag to component for efficient direct/transitive dependency filtering.

### Addressed Issue

Ports PR from Hyades Apiserver

https://github.com/DependencyTrack/dependency-track/issues/6183
https://github.com/DependencyTrack/hyades-apiserver/pull/2168

### Checklist

[ADR]: ../docs/adr/


apiserver/src/main/java/org/dependencytrack/tasks/ImportBomActivity.java
```diff
@@ -698,6 +701,10 @@ private void processDependencyGraph(
         }
 
         qm.getPersistenceManager().flush();
+
+        try (final Handle jdbiHandle = createLocalJdbi(qm).open()) {
+            jdbiHandle.attach(ComponentDao.class).setDirect(project.getId());
+        }
```
nscuro: Whether a component is a direct dependency could be determined here as well: https://github.com/DependencyTrack/dependency-track/blob/85f86f83fb75d47ad18b672fdd49c8735f25dd6d/apiserver/src/main/java/org/dependencytrack/tasks/ImportBomActivity.java#L657-L701

apiserver/src/main/java/org/dependencytrack/model/Component.java
```diff
@@ -426,6 +426,11 @@ public enum FetchGroup {
     @NotNull
     private UUID uuid;
 
+    @Persistent
+    @Column(name = "DIRECT", allowsNull = "true")
+    @JsonProperty("isDirect")
+    private Boolean direct;
```
nscuro: Can we disambiguate the name? e.g. call it `directDependency` / `isDirectDependency`.

apiserver/src/test/java/org/dependencytrack/policy/cel/CelPolicyEngineTest.java
nscuro, line 1: Let's also add the new field to the "all fields" test: https://github.com/DependencyTrack/dependency-track/blob/85f86f83fb75d47ad18b672fdd49c8735f25dd6d/apiserver/src/test/java/org/dependencytrack/policy/cel/CelPolicyEngineTest.java#L72-L396

migration/src/main/resources/org/dependencytrack/migration/V202605291156__add_direct_column_to_component.sql
nscuro, line 1: Might want to think about backfilling the new column. Otherwise existing deployments will be in an awkward state where the new column is `NULL`, but the "is direct dependency" information is *technically* known based on the project's `DIRECT_DEPENDENCIES` column.

That would lead to the new column being unreliable until new BOMs are uploaded to all projects.

nscuro (changes requested):

migration/src/main/resources/org/dependencytrack/migration/V202605291156__add_direct_column_to_component.sql
sahibamittal: Yes plan was to make it nullable to avoid heavy migration for existing components (number can be in millions for deployed components) and can be populated upon next bom upload/publish.

nscuro: Yeah I think that makes sense given your table sizes. But I'd like to have a solution that still backfills data so the feature is "eventually reliable".

Looking into options right now. One might be a scheduled backfill task that runs the backfill at runtime, and chunked, so it doesn't prevent startup and has minimal DB footprint while it's running.

nscuro:
@sahibamittal I looked at this again and have a few concerns:

* The new column duplicates what's already in `PROJECT.DIRECT_DEPENDENCIES`. That's not bad in itself, but it means we have to ensure they're always in sync. Which in turn means `COMPONENT.IS_DIRECT_DEPENDENCY` can't be updated without also updating `PROJECT.DIRECT_DEPENDENCIES` and vice-versa. Currently the there is nothing preventing independent modification.
* The lack of backfill breaks the `onlyDirect` API filters for everyone with existing component data. To make this work w/o backfill we'd need to always check `PROJECT.DIRECT_DEPENDENCIES` **and** `COMPONENT.IS_DIRECT_DEPENDENCY`, at which point the new column only becomes a burden.
* There is currently no query that benefits from the new column that I can see. Issue #6183 says "It becomes expensive for cross-project reporting and bulk analytics", but I don't think we have queries like that at the moment. Virtually all queries on `PROJECT.DIRECT_DEPENDENCIES` are project-scoped. Is there any specific query that does not perform well at the moment?

Regarding backfills, I did look into options to run them separately from schema migrations and asynchronously (like [GitLab's BBM](https://docs.gitlab.com/development/database/batched_background_migrations/)), but even stripped-down versions of that feel like over-engineering at this point. So I guess rather than trying to somehow make the backfill work, I'm trying to probe just about how much this change is actually needed, or whether the desired outcome could be achieved by other means.

sahibamittal:
Upon discussion internally, requirements have changed, this PR can be closed. I'll close the issue.

PR #6210, opened 2026-05-29
Apply OWASP rating from VEX import

fahedouch:
### Description

Re-targeting DependencyTrack/hyades-apiserver#1928 here, since v5 dev moved to this repo (asked by @nscuro: https://github.com/DependencyTrack/hyades-apiserver/pull/1928#issuecomment-4577290072).

Same change as the original PR, just rebased on main. Only thing that changed: renamed the migration so it runs after the latest one on main instead of in the middle.

### Addressed Issue

No separate issue — this is the re-target of DependencyTrack/hyades-apiserver#1928 onto this repo, as requested by @nscuro for the v5 GA move.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/parser/cyclonedx/CycloneDXVexImporter.java
+++ b/apiserver/src/main/java/org/dependencytrack/parser/cyclonedx/CycloneDXVexImporter.java
@@ -34,6 +34,7 @@ import org.dependencytrack.persistence.command.MakeAnalysisCommand;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


nscuro:
@fahedouch Apologies for the delay here, but I'm bringing (I think) good news!

I revisited my earlier comments about the need to track ownership of `ANALYSIS` etc. and ended up deciding that it's not necessary here. The semantics around `ANALYSIS` are already "last writer wins" today, and while improving this would be great, it's a separate feature from your's.

I just merged #6417 which ensures that individual rating overrides will be properly surfaced via UI and the findings API. There was some logic in place that was tailored to how vulnerability policies operate that would have prevented your individual OWASP overrides from ever surfacing.

For your PR, that means:

* You can *delete* `RatingSource`, `RatingSourceTest`, the `Analyis.source` field and associated methods and migrations.
* You can *drop* the entire `canUpdateAnalysis` logic.

One thing I'd like to ask you to *add*, is [test cases](https://github.com/DependencyTrack/dependency-track/blob/main/apiserver/src/test/java/org/dependencytrack/parser/cyclonedx/CycloneDXVexImporterTest.java) for the VEX import that represent your expected usage. Setup a project with findings, apply a realistic VEX, and verify whether the outcome is as expected.

Worth noting is that the severity of findings does *not* change unless it's explicitly overwritten. So if the finding's severity is `HIGH`, and you provide an OWASP vector / score that would evaluate to `LOW`, the finding would remain at `HIGH`. For reference, the corresponding `SELECT` clause is [here](https://github.com/DependencyTrack/dependency-track/blob/9d6913e5840853f81624aa9d3efc68c6df8fe337/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java#L140-L192). You'd need to also explicitly set `Analysis.severity` if that's a desired outcome.

nscuro (approved):
Thanks!

PR #6313, opened 2026-06-08
Feature/policy driven annotation

mehab:
### Description
Vulnerability policies can attach annotations to matching findings. Multiple policies may match a single finding; annotations are merged and persisted on ANALYSIS.POLICY_ANNOTATIONS. Triage (state, suppress, ratings, etc.) is owned by the first non–annotation-only policy in bundle order.

### Addressed Issue
https://github.com/DependencyTrack/dependency-track/issues/6205

### Checklist

[ADR]: ../docs/adr/


apiserver/src/main/java/org/dependencytrack/model/Vulnerability.java
```diff
@@ -406,6 +407,48 @@ public static boolean isKnownSource(String source) {
 
     private transient List<VulnerabilityAlias> aliases;
 
+    private transient AnalysisInfo analysis;
```
sahibamittal: Do we need analysis inside vulnerability model? Any analysis should be at project/component level IMO.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/VulnerabilityDao.java
```diff
@@ -312,7 +318,7 @@ record AffectedProjectListRow(
                 ON "COMPONENTS_VULNERABILITIES"."COMPONENT_ID" = "COMPONENT"."ID"
               LEFT JOIN "ANALYSIS"
             	ON "V"."ID" = "ANALYSIS"."VULNERABILITY_ID"
-               AND "COMPONENT"."PROJECT_ID" = "ANALYSIS"."PROJECT_ID"
+               AND "COMPONENT"."ID" = "ANALYSIS"."COMPONENT_ID"
```
sahibamittal, line 339: This changes the logic of analysis matching. This would result in suppression at each component level instead of current project level.

mehab: this change is only for view of vulnerabilities/ suppressions and policy annotations on per component basis. 
The reconcilliation logic continues to be project level reconcilliation so suppression will continue to happen at project level.
This change will only fetch the project level suppression/ annotation for the specific component where the id match happens.

apiserver/src/main/java/org/dependencytrack/model/Vulnerability.java
mehab: separated this out from the Vulnerability class

Issue #2267, opened 2022-12-10
Add support for the CISA Known Exploited Vulnerabilities (KEV) catalog


troymarshall:
### Current Behavior

The current version of Dependency-Track provides support for several open data sources, including the EPSS. It does not currently ingest data from the CISA KEV. Inclusion in the CISA KEV triggers vulnerability remediation timeline requirements for federal agencies as set forth in CISA Binding Operational Directive (BOD) 22-01. Specifically, the BOD 22-01 specifies:

"Remediate each vulnerability according to the timelines set forth in the CISA-managed vulnerability catalog. The catalog will list exploited vulnerabilities that carry significant risk to the federal enterprise with the requirement to remediate within 6 months for vulnerabilities with a Common Vulnerabilities and Exposures (CVE) ID assigned prior to 2021 and within two weeks for all other vulnerabilities. These default timelines may be adjusted in the case of grave risk to the Federal Enterprise."outlined in

### Proposed Behavior

I am proposing ingesting the CISA KEV database available at https://www.cisa.gov/known-exploited-vulnerabilities-catalog and using this data to enrich the existing exploitability data such as EPSS also to include the date added to KEV and the remediation due date. Once the data is included, I would request a policy to identify vulnerabilities outside of the KEV remediation timeline.

While the vulnerability due date is currently only applicable to US federal agencies, I could foresee a time when the remediation timeline becomes a requirement for federal contractors. 

I had a brief exchange with Steve S. on this in the OWASP Slack channel, and he requested I include a note about building this in a way that isn't specific to the USA. I have spent some time researching international equivalents to the CISA KEV but have not found any. I have found several commercial tools supporting KEV, including:
https://www.tenable.com/sc-dashboards/dhs-cisa-binding-operational-directive-22-01
https://nucleussec.com/cisa-kev/
https://www.paloaltonetworks.com/blog/security-operations/cortex-xpanse-identify-cisa-kev/
https://blog.qualys.com/product-tech/2022/02/22/managing-cisa-known-exploited-vulnerabilities-with-qualys-vmdr

Steve's note from our Slack discussion: "I wonder if other countries have similar things. I'd hate to make it U.S. specific. I'd rather find a non-geographic way to implement this while still being able to use KEV as one of the possible feeds. "

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
[...]

PR #6315, opened 2026-06-08
ADR for feature "Known Exploited Vulnerabilities"

LukasChristel:
### Description

Add an ADR to get the feature of reporting known exploited vulnerabilities started. 

### Addressed Issue

#2267 

### Additional Details

I am unsure about many aspects in this ADR. Especially the data storage proposal.

Also unclear
- [x] Should other sources be mentioned in this ADR? (VulnCheck KEV, inthewild.io, GreyNoise, Metasploit, ExploitDB, Nuclei)
- [x] Remediation due dates are currently not addressed by the ADR

### Checklist

[ADR]: ../docs/adr/


docs/adr/029-actively-exploited-vulnerability-monitoring.md
```diff
@@ -0,0 +1,67 @@
...
+
+The 'Audit Vulnerabilities' panel marks vulnerabilities as actively exploited when any vulnerability in the vulnerability-alias-group ([ADR-014][adr-014]) is actively exploited.
+This allows showing the hint when a component is only identified to be vulnerable to e.g. GHSA-* but the active exploitation is reported for CVE-*. 
+The vulnerability is marked with a red icon in the column 'Severity' (same design as the icon in the column 'Version'). The icon has a tooltip that lists all sources which identified a vulnerability in the vulnerability-alias-group to be actively exploited.
+
+The 'Audit Vulnerabilities' panel gets a toggle to limit the results to actively exploited vulnerabilities only.
+
+The 'Policy Management' panel allows creating a policy that generates a policy violation in case of actively exploited vulnerabilities.
+
+The 'Administration' > 'Notifications' > 'Alerts' panel allows configuring an alert that triggers in case of new actively exploited vulnerabilities.
```
nscuro: Assuming the notification model is updated accordingly, this is covered automatically using [filter expressions](https://dependencytrack.github.io/docs/next/reference/notifications/filter-expressions/).

docs/adr/029-actively-exploited-vulnerability-monitoring.md
```diff
@@ -0,0 +1,67 @@
...
+#### Actively Exploited Vulnerability Assertions
+
+To track provenance of actively exploited vulnerability assertions, a separate `ACTIVELY_EXPLOITED_VULNERABILITY_ASSERTION` table records which entity asserted that a vulnerability is actively exploited:
+
+| Column      | Type           | Constraints             |
+|:------------|:---------------|:------------------------|
+| ASSERTER    | TEXT           | PK                      |
+| VULN_SOURCE | TEXT           | PK                      |
+| VULN_ID     | TEXT           | PK                      |
+| CREATED_AT  | TIMESTAMPTZ(3) | NOT NULL, DEFAULT NOW() |
```
nscuro: I think we need to capture more data than just presence. For example:

* Date added: When the item was added to the KEV list by the asserter
* Due date: Until when this should be fixed, I think CISA provide this

Potentially more. Consider reviewing https://gcve.eu/bcp/gcve-bcp-07/ and seeing what other info might be useful to track. Could even consider tracking the raw source data in a `JSONB` column.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
```diff
@@ -0,0 +1,67 @@
...
+
+Each row records that `ASSERTER` claimed (`VULN_SOURCE`, `VULN_ID`) is actively exploited. 
+
+#### Check for actively exploited vulnerabilities
+
+To check if any vulnerability in the vulnerability-alias-group is actively exploited, other records from `ACTIVELY_EXPLOITED_VULNERABILITY_ASSERTION` need to be joined through the table `VULNERABILITY_ALIAS`. `ACTIVELY_EXPLOITED_VULNERABILITY_ASSERTION` also needs to be queried without joining `VULNERABILITY_ALIAS` in case no alias exists.
```
nscuro: How do we handle multiple assertions for the same vuln? "Has at least one assertion" suffices to check whether a vuln is actively exploited, but auxiliary data like "due date" would need to be disambiguated.

Might be worth synthesizing all assertions into a single "concluded" record whenever an assertion is added or removed. The vuln alias mechanism does something like this, where assertions are used to compute alias groups.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: - I added a `PUBLISHED` column for the date added.
- I am a bit hesitant about the due date because I understand it to be very specific to CISA and the US government. For any project not related to US government it might give the wrong impression that fixing the vulnerability is not necessary until the due date is reached. If you give the go ahead that this is not a concern then I'll add it as well.
- From the sources discussed in #2267 only Metasploit/ExploitDB do not directly offer JSON responses. Raw data as `JSONB` should work well, responses from some future sources might need to be converted to JSON.
- GCVE-BCP-07: I am happy to add status_reason, characteristics, more timestamps, scope, evidence and others but lack the context to judge which of these is useful for the direction Dependency-Track intends to go.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: Multiple assertions for the same vulnerability would each get a row in `KEV_ASSERTION`. This assumes the same source does not repeat the same vulnerability-id multiple times. That would violate the primary key constraint.
Querying would join through the alias group like [ADR-022](https://github.com/DependencyTrack/dependency-track/blob/main/docs/adr/022-read-time-epss-resolution-via-alias.md).

I did not find a good synthesized format. One attempt is listed in the chapter 'Considered alternatives'. Two more attempts here, if you like one of them I'll update the ADR.

## Option 1: Always create alias groups
The alias groups from [ADR-014](https://github.com/DependencyTrack/dependency-track/blob/main/docs/adr/014-new-alias-schema.md) are always created. Even when no alias-identifiers exist. Then it is just a group with only one member.
The table `VULNERABILITY_ALIAS` gets a new boolean column `KNOWN_EXPLOITED`. This column needs to be recomputed after every change to `KEV_ASSERTION` or `VULNERABILITY_ALIAS`. It makes querying whether any vulnerability in the alias-group has a vulnerability a bit easier. Only the `VULNERABILITY_ALIAS` table needs to be queried because it contains entries for ALL vulnerabilities.

## Option 2: Materialize into `KEV` table
This option provides a materialized view like [ADR-014](https://github.com/DependencyTrack/dependency-track/blob/main/docs/adr/014-new-alias-schema.md)
| Column | Type | Constraints |
|:-|:-|:-|
|ASSERTER|TEXT|PK|
|VULN_SOURCE|TEXT|PK|
|VULN_ID|TEXT|PK|
|ALIAS_ID|TEXT|PK|

This table needs to be recomputed after every change to `KEV_ASSERTION` or `VULNERABILITY_ALIAS`.
Recomputing involves that for every entry in `KEV_ASSERTION` an entry in `KEV` is created (`VULN_ID` and `ALIAS_ID` are the same). If aliases exist, then an additional entry in `KEV` is created for each alias.

`ALIAS_ID` is used in queries that want to know if a vulnerability is known to be exploited; `ASSERTER`, `VULN_SOURCE`, `VULN_ID` are the identifiers required to look up more details in table `KEV_ASSERTION`.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: Mentioned the use of filter expressions in the ADR

docs/adr/029-actively-exploited-vulnerability-monitoring.md
nscuro: Upon further research, I'd paddle back on my earlier feedback and would align more with your initial proposal. May be re-stating the obvious but just to have it noted down once:

* We *don't* materialize, we only record the raw assertions.
* The signal that matters in most use-cases is simply the boolean "is this actively exploited?", and that can be answered with a simple `EXISTS(SELECT 1 FROM "ACTIVELY_EXPLOITED_VULNERABILITY_ASSERTION" ...)` subquery, joined through `VULNERABILITY_ALIAS` as you noted.
* For views / endpoints where more detail is desired, we surface *all* assertions, including their data like `published` / `dateAdded` / `dueDate` etc.

Materializing `ACTIVELY_EXPLOITED_VULNERABILITY` and having it depend on alias relationships would involve significant overhead which I'd like to avoid. As you noted, that would require refreshing the materialization when KEV assertions **or** aliases change. The latter is in the hot path during vuln analysis.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+The 'Vulnerability Audit' panel gets a new filter to limit results to actively exploited vulnerabilities.
+
+### Data storage
+
+Information about actively exploited vulnerabilities are saved per source. This way it is easy to know which row to delete when a source no longer lists a vulnerability as actively exploited.
+
+#### Actively Exploited Vulnerability Assertions
+
+To track provenance of actively exploited vulnerability assertions, a separate `ACTIVELY_EXPLOITED_VULNERABILITY_ASSERTION` table records which entity asserted that a vulnerability is actively exploited:
```
nscuro: Nit: Let's call it `KEV_ASSERTION`. KEV is an established name at this point and means practically the same as the proposed table name.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
nscuro: Some more useful fields:

* Description: I think all sources provide a little bit of free-form text that describe the issue.
* Required action. It's free-form text and it seems only CISA and VulnCheck provide it, but it'd be useful to track and surface.
* Known ransomware, e.g.: https://www.cisa.gov/known-exploited-vulnerabilities-catalog?search=CVE-2022-27593&field_date_added_wrapper=all&field_cve=&sort_by=field_date_added&items_per_page=20&url=

So I think we should add these columns:

* `REQUIRED_ACTION` (`TEXT NULLABLE`, `NULL` when not provided)
* `KNOWN_RANSOMWARE` (`BOOL NULLABLE`, `NULL` when not provided)
* `DESCRIPTION` (`TEXT NULLABLE`, `NULL` when not provided)

`RAW JSONB` then covers the rest, including due date.

> responses from some future sources might need to be converted to JSON

Yeah that's acceptable.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+| Column      | Type           | Constraints             |
+|:------------|:---------------|:------------------------|
+| ASSERTER    | TEXT           | PK                      |
+| VULN_SOURCE | TEXT           | PK                      |
+| VULN_ID     | TEXT           | PK                      |
+| CREATED_AT  | TIMESTAMPTZ(3) | NOT NULL, DEFAULT NOW() |
```
nscuro: We should also track `UPDATED_AT` to record when we update assertions.

Also as noted in https://github.com/DependencyTrack/dependency-track/issues/2267#issuecomment-1751725119, KEV entries can be revoked. We should track that as well via a dedicated `REVOKED_AT` column. Rows where `REVOKED_AT IS NULL` should then not be considered when determining the "actively exploited" status.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
nscuro, line 1: In your PR description you asked:

> Should other sources be mentioned in this ADR? (VulnCheck KEV, inthewild.io, GreyNoise, Metasploit, ExploitDB, Nuclei)

My vote would be **no** for most of them (except VulnCheck KEV). 

https://github.com/DependencyTrack/dependency-track/issues/2267#issuecomment-4067393192 is asking for multiple states like `unknown`, `exploited` and `exploit code public`. I think we should focus on the "actively exploited" signal that KEV is about. Mostly to keep things simple, but also because "exploit code public" is a very different risk from KEV, and it shares almost no properties with its model.

Regarding your question, VulnCheck is the exception to the above because it provides KEV data. The others provide "exploit code public" signals.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: I have `KNOWN_RANSOMWARE` as `NOT NULL, DEFAULT FALSE` for now because I couldn't think of a use case where a consumer needs to differentiate between CISA not knowing of any ransomware usage vs a source not providing the value.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: I included your points in the ADR and put the two options from me into 'Considered alternatives'

In making the change I noticed that some sources ([EUVD](https://euvd.enisa.europa.eu/vulnerability/EUVD-2026-34341)) provide multiple IDs (CVE-\*, GHSA-\*, EUVD-\*) for one KEV. 
I guess these sources should then also be an alias asserter and the information should go into the `VULNERABILITY_ALIAS_ASSERTION` table. Another option would be to create a copy of the entry in `KEV_ASSERTION` for each identifier. Any thoughts?

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: Updated wording in ADR and in my comments on this PR to have consistent naming for new reviewers.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: I am not sure if you suggest `UPDATED_AT` as a technical identifier for when the last change to the DB record happened or as data provided by a source ([example](https://euvd.enisa.europa.eu/vulnerability/EUVD-2026-34341)).

Soft deleting with `REVOKED_AT` instead of hard deleting differs from how [ADR-014](https://github.com/DependencyTrack/dependency-track/blob/main/docs/adr/014-new-alias-schema.md) treats assertions and doesn't provide full history. Records in `ASSERTION_ALIAS` are hard deleted and when a KEV is only identified by an alias then the history (that this KEV had been identified in the past) is lost once the alias disappears. I suggest to provide no information about past KEV indicators rather than partial information and therefore not have `REVOKED_AT`.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: ADR updated. I am leaving this conversation unresolved for now in case anyone wants to chime in.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
nscuro: For the purpose of KEV I'd say we should flatten it into multiple records.

But looking at the EUVD page you linked, it says `Added to CISA KEV: 2026-06-09 00:00`. Do you know of the EUVD even publishes its own KEV data, or does it always parrot CISA? In the latter case there'd be zero use in pulling data from the EUVD.

nscuro: Found this post from VulnCheck: https://www.vulncheck.com/blog/enisa-euvd

> Unfortunately, I was disappointed to find that the ENISA EUVD Exploited Vulnerabilities list is merely a subset of CISA’s Known Exploited Vulnerabilities (KEV).

So I guess there is no value in direct EUVD support for KEV.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
nscuro: > I am not sure if you suggest UPDATED_AT as a technical identifier for when the last change to the DB record happened or as data provided by a source ([example](https://euvd.enisa.europa.eu/vulnerability/EUVD-2026-34341)).

As a technical audit field. Every time we update our own record, we update the timestamp. But tbh the suggestion hinges on whether records are updated at all. While I find it *likely* that, say, CISA, updates descriptions or even ransomware flags over time, I don't have hard evidence for that. So it's more speculative.

Agreed on `REVOKED_AT`. Value would be limited if there's a hard dependency on aliases that don't keep track like that.

docs/adr/029-actively-exploited-vulnerability-monitoring.md
LukasChristel: In short: I could find some but not many and not a single official post that claim the EUVD will become relevant for DependencyTrack in the very near future.

Mandatory reporting under the EU Cyber Resilience Act starts September 11th 2026 (https://digital-strategy.ec.europa.eu/en/policies/cra-reporting). I expect that from this date on they will have original data as well.
But posts that share this expectation are rare. There is the last paragraph in https://www.itsecurityguru.org/2025/05/14/enisa-launches-european-vulnerability-database-to-bolster-eu-cyber-resilience/ and the last chapter from https://www.aikido.dev/blog/nist-nvd-changes-2026 that support it.

This article claims that data is exchanged between EUVD and NVD but that they don't necessarily have to agree. https://www.infosecurityeurope.com/en-gb/blog/future-thinking/how-to-use-euvd-enisa.html. It mentions the following authoritative sources available to the EUVD
- Open-source vulnerability databases (such as CVE and NVD)
- National CSIRT advisories and security alerts
- Vendor-provided patching and mitigation recommendations
- Exploitation status indicators showing active abuse in cyber attacks

nscuro: I am asking around if we can get any at least semi-official info from ENISA or ENISA-adjacent folks on this.

nscuro: Ah-ha! There is a dedicated feed with custom data here: https://github.com/enisaeu/CNW/tree/main/advisories/eukev

nscuro: The EUVD has a data dump API at https://euvdservices.enisa.europa.eu/api/kev/dump.

It's a minimal representation of both CISA's and ENISA's KEV data, e.g.:

```json
  {
    "cveId": "CVE-2024-8963",
    "euvdId": "EUVD-2024-49510",
    "vendorProject": "Ivanti",
    "product": "CSA (Cloud Services Appliance)",
    "dateAdded": "2024-09-19",
    "sources": [
      "cisa_kev",
      "eukev_kev"
    ]
  },
  {
    "cveId": "CVE-2021-31166",
    "euvdId": "EUVD-2021-18079",
    "dateAdded": "2022-04-06",
    "sources": [
      "cisa_kev"
    ]
  },
```

So here we'd need to skip `cisa_kev` records to prevent stale data from the EUVD overwriting the authoritative and more up-to-date data from CISA's own feed. With that in mind, I'd argue consuming ENISA's data directly and *not* going through the EUVD is likely the better option.

LukasChristel: - Changed wording so that ENISA KEV is used as a source, not EUVD.
- Mentioned URLs for CISA, ENISA and VulnCheck.
- Specified that KEVs with multiple vulnerability IDs should not be alias asserters but just repeat the KEV database record for each ID

nscuro (approved):
I think this is in a good state and can be worked on now. Is this ready to merge from your POV @LukasChristel?

LukasChristel:
Yes, ready from me as well. Changed the header to 'Accepted'. Thank you for all the great feedback.

nscuro (approved):

PR #6325, opened 2026-06-09
fix: match Red Hat OS distro in OSV vulnerability analysis

### LINKED ISSUE #6156: False positives: RedHat scanner result are compared with wrong version format
### Current Behavior

When producting a SBOM from a Red Hat / Ubi9 minimal container image (in my case registry.access.redhat.com/ubi9/ubi-minimal:9.7-1773939694) with trivy and uploading it to DependencyTrack, false positives are produced because version numbers are not correctly compared.

Image

Image

### Steps to Reproduce

1. Scan your container image which is based on Ubi9 minimal, or Ubi9 itself
```bash
trivy image --format cyclonedx --output results.sbom.json registry.access.redhat.com/ubi9/ubi-minimal:9.7-1773939694
cat results.sbom.json | jq -r '.components[] | select(.name=="libsolv") | .purl'
pkg:rpm/redhat/libsolv@0.7.24-3.el9?arch=x86_64&distro=redhat-9.7
```
2. Make sure the OSV scanner including the Red Hat repo is activated 
3. Upload the SBOM to DependencyTrack
4. Look at the detected CVEs 

### Expected Behavior

My version is "0.7.24-3.el9", the one compared with is "1:0.7.20-6.el8sat" (Red Hat satellite repo).
That difference is not respected but should be.

### Dependency-Track Version

4.14.2

Container Image

### Database Server

N/A

### Database Server Version

_No response_

### Browser

Google Chrome

### Checklist

mvanhorn:
### Description

Adds Red Hat to the set of OS distributions consulted during OSV vulnerability matching, so Red Hat / UBI RPM components are only matched against advisories scoped to the same RHEL major version. This is the Red Hat equivalent of the Alpine/Debian/Ubuntu distro-aware matching introduced in #5783.

SBOMs from Red Hat / UBI9 images (e.g. `pkg:rpm/redhat/libsolv@0.7.24-3.el9?distro=redhat-9.7`) currently produce false positives, because an OSV Red Hat advisory is compared against the component without respecting the Red Hat product-stream / RHEL scope. In #6156 the reporter's RHEL 9 component (`0.7.24-3.el9`) is matched against an unrelated RHEL 8 Satellite advisory (`1:0.7.20-6.el8sat`).

`OsDistribution` already gates version comparison in `InternalVulnAnalyzer.matchesDistro`, but only Alpine/Debian/Ubuntu were modeled, so Red Hat fell through and the false positives stood. As @nscuro noted on the issue, Red Hat needs the same treatment, and OSV encodes the Red Hat product stream in the ecosystem string via a `:<CPE>` suffix.

Changes:

- `RedhatDistribution` (new sealed `OsDistribution` record) modeling the RHEL major version, with `matches` that only matches like RHEL majors.
- `OsDistribution`: added `RedhatDistribution` to the `permits` clause and extended `of(PackageURL)` to resolve `rpm` / `redhat` PURLs carrying a `distro` qualifier.
- `OsvEcosystems.toOsDistribution`: added a `red hat` / `redhat` ecosystem branch that parses the `:<CPE>` suffix.

### Addressed Issue

Fixes #6156

### Additional Details

Red Hat OSV ecosystem CPEs do not all carry the RHEL major in the same field, so the parser handles them explicitly:

- An explicit `elN` OS target is authoritative (e.g. `Red Hat:openshift:4.18::el8` -> RHEL 8, not OpenShift 4), since layered-product versions are not the RHEL major.
- Otherwise the major is taken from a base RHEL product (`rhel`, `rhel_aus`, `rhel_eus`, `rhel_els`, `rhel_tus`, `rhel_e4s`, `enterprise_linux`).
- Layered products without an `elN` target (e.g. `rhel_application_stack:2`, `rhel_atomic:7`, `satellite:6.16`) return null rather than mis-scoping to a wrong RHEL major, which would create new false negatives.

Verified against real OSV feed entries while iterating on the CPE parsing.

Testing: `mvn -pl support/os-distro-metadata test` (115 tests, including 31 new in `RedhatDistributionTest`) and `mvn -pl vuln-data-source/osv test -Dtest=OsvEcosystemsTest` (40 tests) both pass. Tests cover the #6156 regression (RHEL 9 component not matched against a RHEL 8 stream advisory), same-major matches, base-RHEL and layered-product CPE forms, and PURL-qualifier resolution.

AI was used for assistance.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/support/os-distro-metadata/src/main/java/org/dependencytrack/support/distrometadata/RedhatDistribution.java
+++ b/support/os-distro-metadata/src/main/java/org/dependencytrack/support/distrometadata/RedhatDistribution.java
@@ -1,0 +1,130 @@
+/*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.support.distrometadata;
+
+import org.jspecify.annotations.Nullable;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Models a Red Hat product stream as scoped by the RHEL major version.
+ * <p>
+ * Component PURLs carry the distro as a {@code distro} qualifier (e.g.
+ * {@code pkg:rpm/redhat/libsolv@0.7.24-3.el9?distro=redhat-9.7}), whereas OSV
+ * advisories encode the product stream in the ecosystem string via a
+ * {@code :<CPE>} suffix (e.g. {@code Red Hat:rhel_aus:8.4::appstream}). Both forms
+ * carry a RHEL major version, which is the smallest reliably comparable scope:
+ * an advisory for RHEL 8 must not be matched against a RHEL 9 component.
+ *
+ * @since 4.14.0
+ */
+public record RedhatDistribution(String majorVersion) implements OsDistribution {
+
+    // The RHEL OS target embedded in a CPE's version-of-target field, e.g.
+    // "el8" in "openshift:4.18::el8" or "satellite:6.16::el8". This is the
+    // authoritative OS scope and takes precedence over a product version that
+    // happens to differ from the RHEL major (OpenShift 4.18 runs on RHEL 8).
+    private static final Pattern EL_TARGET_PATTERN =
+            Pattern.compile(".*\\bel(\\d+)\\b.*", Pattern.CASE_INSENSITIVE);
+
+    // The major version of a base RHEL product itself, e.g. "8" in
+    // "rhel:8::appstream" or "rhel_aus:8.4::appstream". Used only when the product
+    // is a base RHEL stream whose version IS the RHEL major, and no explicit "elN"
+    // target is present. Deliberately excludes layered products (e.g.
+    // "rhel_application_stack:2", "rhel_atomic:7"), whose version is a product
+    // stream number rather than the RHEL major.
+    private static final Pattern RHEL_PRODUCT_PATTERN =
+            Pattern.compile("^(?:rhel|rhel_aus|rhel_eus|rhel_els|rhel_tus|rhel_e4s|enterprise_linux):(\\d+)(?:[.:].*)?$",
+                    Pattern.CASE_INSENSITIVE);
+
+    // The leading major version of a PURL "distro" qualifier value, e.g. "9" in
+    // "redhat-9.7" or "9.7". PURL qualifiers carry a bare RHEL version, not a CPE.
+    private static final Pattern PURL_VERSION_PATTERN =
+            Pattern.compile("^(\\d+)(?:\\..*)?$");
+
+    public RedhatDistribution {
+        requireNonNull(majorVersion, "majorVersion must not be null");
+    }
+
+    @Override
+    public String purlQualifierValue() {
+        return "redhat-" + majorVersion;
+    }
+
+    @Override
+    public boolean matches(OsDistribution other) {
+        return other instanceof RedhatDistribution(final String otherMajorVersion)
+                && this.majorVersion.equals(otherMajorVersion);
+    }
+
+    /**
+     * Resolves a Red Hat distro from a PURL {@code distro} qualifier value,
+     * e.g. {@code redhat-9.7} or {@code 9.7}.
+     */
+    public static @Nullable RedhatDistribution of(@Nullable String qualifierValue) {
+        if (qualifierValue == null || qualifierValue.isEmpty()) {
+            return null;
+        }
+
+        final String value = qualifierValue.toLowerCase().startsWith("redhat-")
+                ? qualifierValue.substring("redhat-".length())
+                : qualifierValue;
+
+        final Matcher matcher = PURL_VERSION_PATTERN.matcher(value);
+        if (!matcher.matches()) {
+            return null;
+        }
+
+        return new RedhatDistribution(matcher.group(1));
+    }
+
+    /**
+     * Resolves a Red Hat distro from the {@code <CPE>} suffix of an OSV
+     * Red Hat ecosystem string. The string is a CPE with the
+     * {@code cpe:/[oa]:redhat:} prefix removed, e.g. {@code rhel_aus:8.4::appstream},
+     * {@code openshift:4.18::el8}, or {@code satellite:6.16::el8}.
+     * <p>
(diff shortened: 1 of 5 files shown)


support/os-distro-metadata/src/main/java/org/dependencytrack/support/distrometadata/RedhatDistribution.java
```diff
@@ -0,0 +1,130 @@
...
+ * @since 4.14.0
+ */
+public record RedhatDistribution(String majorVersion) implements OsDistribution {
+
+    // The RHEL OS target embedded in a CPE's version-of-target field, e.g.
+    // "el8" in "openshift:4.18::el8" or "satellite:6.16::el8". This is the
+    // authoritative OS scope and takes precedence over a product version that
+    // happens to differ from the RHEL major (OpenShift 4.18 runs on RHEL 8).
+    private static final Pattern EL_TARGET_PATTERN =
+            Pattern.compile(".*\\bel(\\d+)\\b.*", Pattern.CASE_INSENSITIVE);
```
nscuro, line 47: To also handle minor versions:

```regex
.*\\bel(\\d+)(?:\\.(\\d+))?\\b.*
```

support/os-distro-metadata/src/main/java/org/dependencytrack/support/distrometadata/RedhatDistribution.java
```diff
@@ -0,0 +1,130 @@
...
+     * e.g. {@code redhat-9.7} or {@code 9.7}.
+     */
+    public static @Nullable RedhatDistribution of(@Nullable String qualifierValue) {
+        if (qualifierValue == null || qualifierValue.isEmpty()) {
+            return null;
+        }
+
+        final String value = qualifierValue.toLowerCase().startsWith("redhat-")
+                ? qualifierValue.substring("redhat-".length())
+                : qualifierValue;
```
nscuro, line 90: Some SBOM generators like Syft use `rhel-` instead of `redhat-` for the distro qualifier. We should handle both here.

support/os-distro-metadata/src/main/java/org/dependencytrack/support/distrometadata/RedhatDistribution.java
```diff
@@ -0,0 +1,130 @@
...
+ * Models a Red Hat product stream as scoped by the RHEL major version.
+ * <p>
+ * Component PURLs carry the distro as a {@code distro} qualifier (e.g.
+ * {@code pkg:rpm/redhat/libsolv@0.7.24-3.el9?distro=redhat-9.7}), whereas OSV
+ * advisories encode the product stream in the ecosystem string via a
+ * {@code :<CPE>} suffix (e.g. {@code Red Hat:rhel_aus:8.4::appstream}). Both forms
+ * carry a RHEL major version, which is the smallest reliably comparable scope:
+ * an advisory for RHEL 8 must not be matched against a RHEL 9 component.
+ *
+ * @since 4.14.0
```
nscuro, line 38:
```suggestion
 * @since 5.1.0
```

support/os-distro-metadata/src/main/java/org/dependencytrack/support/distrometadata/RedhatDistribution.java
```diff
@@ -0,0 +1,130 @@
...
+    // The major version of a base RHEL product itself, e.g. "8" in
+    // "rhel:8::appstream" or "rhel_aus:8.4::appstream". Used only when the product
+    // is a base RHEL stream whose version IS the RHEL major, and no explicit "elN"
+    // target is present. Deliberately excludes layered products (e.g.
+    // "rhel_application_stack:2", "rhel_atomic:7"), whose version is a product
+    // stream number rather than the RHEL major.
+    private static final Pattern RHEL_PRODUCT_PATTERN =
+            Pattern.compile("^(?:rhel|rhel_aus|rhel_eus|rhel_els|rhel_tus|rhel_e4s|enterprise_linux):(\\d+)(?:[.:].*)?$",
+                    Pattern.CASE_INSENSITIVE);
+
```
nscuro, line 58: Based on the OSV data set, there are a few more product names we should capture:

```suggestion
    private static final Pattern RHEL_PRODUCT_PATTERN = Pattern.compile("""
            ^(?:\
            enterprise_linux\\w*\
            |enterprise_ipa\
            |rhel_application_server\
            |rhel_aus\
            |rhel_cluster\
            |rhel_cluster_storage\
            |rhel_developer_suite\
            |rhel_e4s\
            |rhel_els\
            |rhel_eus\
            |rhel_eus_long_life\
            |rhel_extras\\w*\
            |rhel_global_file_system\
            |rhel_mission_critical\
            |rhel_productivity\
            |rhel_rhn_tools\
            |rhel_sjis\
            |rhel_stronghold\
            |rhel_tus\
            |rhel_virtualization\
            ):(\\d+)(?:\\.(\\d+))?(?:[.:].*)?$\
            """, Pattern.CASE_INSENSITIVE);
```

That would also capture minor versions.

support/os-distro-metadata/src/main/java/org/dependencytrack/support/distrometadata/RedhatDistribution.java
```diff
@@ -0,0 +1,130 @@
...
+ * Component PURLs carry the distro as a {@code distro} qualifier (e.g.
+ * {@code pkg:rpm/redhat/libsolv@0.7.24-3.el9?distro=redhat-9.7}), whereas OSV
+ * advisories encode the product stream in the ecosystem string via a
+ * {@code :<CPE>} suffix (e.g. {@code Red Hat:rhel_aus:8.4::appstream}). Both forms
+ * carry a RHEL major version, which is the smallest reliably comparable scope:
+ * an advisory for RHEL 8 must not be matched against a RHEL 9 component.
+ *
+ * @since 4.14.0
+ */
+public record RedhatDistribution(String majorVersion) implements OsDistribution {
```
nscuro, line 40: We should also track the (optional) minor version. A large chunk of the OSV data provides Red Hat versions in the `major.minor` format. So if OSV claims a vuln to affect `redhat-8.6`, it shouldn't match against a component with `redhat-8.7`. But it should match a component with `redhat-8` and vice-versa. `matches` would need to be updated accordingly.

support/os-distro-metadata/src/main/java/org/dependencytrack/support/distrometadata/RedhatDistribution.java
```diff
@@ -0,0 +1,130 @@
...
+    // "rhel_application_stack:2", "rhel_atomic:7"), whose version is a product
+    // stream number rather than the RHEL major.
+    private static final Pattern RHEL_PRODUCT_PATTERN =
+            Pattern.compile("^(?:rhel|rhel_aus|rhel_eus|rhel_els|rhel_tus|rhel_e4s|enterprise_linux):(\\d+)(?:[.:].*)?$",
+                    Pattern.CASE_INSENSITIVE);
+
+    // The leading major version of a PURL "distro" qualifier value, e.g. "9" in
+    // "redhat-9.7" or "9.7". PURL qualifiers carry a bare RHEL version, not a CPE.
+    private static final Pattern PURL_VERSION_PATTERN =
+            Pattern.compile("^(\\d+)(?:\\..*)?$");
```
nscuro, line 62: To also handle minor versions:

```suggestion
            Pattern.compile("^(\\d+)(?:\\.(\\d+))?(?:\\..*)?$");
```

nscuro (changes requested):
Thanks! A few changes to extend the coverage and accuracy.

nscuro:
Superseded by #6464. Thanks @mvanhorn, some of this work was incorporated in that PR and I added you via `Co-authored-by:`.

Issue #5934, opened 2026-03-18
OSV mirroring hangs with Ubuntu enabled


lukas-braune:
### Current Behavior

With `Ubuntu` enabled, OSV mirroring hangs forever:
<img width="1814" height="219" alt="Image" src="https://github.com/user-attachments/assets/b4f1d723-173d-4553-b5b9-ad9008676fb7" />

This issue showed up in DT v4.14.0.

### Steps to Reproduce

1. Enable OSV mirroring and enable the `Ubuntu` ecosystem
2. OsvDownloadTask will hang at `Processed 5449/54490 advisories (10%)` (or similar)

### Expected Behavior

Working OSV mirroring

### Dependency-Track Version

4.14.0

### Dependency-Track Distribution

Container Image

### Database Server

PostgreSQL

### Database Server Version

18.2

### Browser

Google Chrome

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

nscuro:
So I profiled this, and while it is rather slow to mirror the Ubuntu ecosystem (took ~25min on my laptop, will be slower on a server of course), nothing really stood out as an obvious performance issue in our code.

It might just be that the sheer volume of data is the issue, see https://github.com/DependencyTrack/dependency-track/issues/5674#issuecomment-3737536361.

I will try this again and also check if the database is not using the indexes it should, but tbh I don't expect that to be an issue here.

@lukas-braune how long has it been running before you concluded it being stuck?

lukas-braune:
@nscuro About 1.5 days.

nscuro:
Oof, yeah no that's clearly broken then. I'll try further to reproduce this.

nscuro:
https://github.com/DependencyTrack/dependency-track/pull/6031 should help. Will include that in a 4.14.2 release.

valentijnscholten:
I ran into this is as well, but it did eventually complete. And then the next full mirror was fast again as it did not have to recreate all the records that needed to be recreated due to changes in 4.14.0

nscuro:
#6031 was just released with 4.14.2. Since I was unable to reproduce the issue reported here, I don't know for sure whether that change actually fixes the issue. 

[...]

PR #6345, opened 2026-06-10
Exclude retrieval of components and aliases during OSVDownloadTask

ElenaStroebele:
### Description
**General workflow OSV Download Task**
The OSV download task includes a method called `updateDatasource(final OSVAdvisory advisory)`. Based on profiling results, this method appears to be the main hotspot.

The method has two possible paths: either the vulnerability referenced by the advisory is already persisted and both the source and metadata are unchanged, or the vulnerability has not been persisted yet, has been modified, or is being reported by OSV for the first time.

For a typical full download—where most advisories have already been fetched before—the first case should apply most of the time: the vulnerability already exists and hasn’t changed since the last run. This fix therefore focuses on optimizing that fast path. For an initial full download where no vulnerabilities have been persisted yet, other parts of the code dominate the runtime, and this change won’t affect that.

**Proposed change**
`getVulnerabilityByVulnId()` is called during the fast path and fetches the persisted vulnerability and all associated components. For the call originating from the OSVDownloadTask, components are no longer fetched, as it is unnecessary in this context and takes more time than needed. 

### Addressed Issue
Hopefully this will make the OSV download faster; addresses issue #5934 .

### Additional Details
I’m not sure this PR fully resolves the issue—it’s likely only one piece of a larger solution, but hopefully a good first step.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/QueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/QueryManager.java
@@ -806,19 +806,23 @@ public class QueryManager extends AlpineQueryManager {
     }
 
     public Vulnerability getVulnerabilityByVulnId(String source, String vulnId) {
-        return getVulnerabilityQueryManager().getVulnerabilityByVulnId(source, vulnId, false);
+        return getVulnerabilityQueryManager().getVulnerabilityByVulnId(source, vulnId, false, true);
     }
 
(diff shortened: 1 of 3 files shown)


nscuro (approved):
Awesome work, that might really be the issue! Fingers crossed this solves it.

PR #6415, opened 2026-06-15
Forward analysis.detail to DefectDojo, v5 version

webdevred:
### Description

When DefectDojo is used as the primary triage surface and Dependency-Track as the SBOM ingestion engine, the per-finding audit text stored in `analysis.detail` never made it into the payload sent to DefectDojo. Auditors working in DefectDojo had no way to see that context without switching back to DT for every finding.

This is the v5 companion to #6181, which makes the same change for `4.14.x`. The field is added to `FindingDao.FindingRow` as `analysisDetail`, projected in both `FindingDao` SQL queries as `a."DETAILS" AS "analysisDetail"`, and mapped to `analysis.detail` in the `Finding` constructor.

dtrack-audit-detail
defect-dojo-audit-detail

### Addressed Issue

Related to #6169. v5 companion to #6181 (same change targeting `4.14.x`).

### Additional Details

The v4 fix also corrected an off-by-one in the `QUERY_ALL_FINDINGS` column index after the new column shifted the positions of subsequent fields. That problem does not exist here: `FindingDao` maps query results by column name via JDBI, not by positional index, so adding a column to the SELECT list has no side effects.

### Checklist


apiserver/src/main/java/org/dependencytrack/model/Finding.java
nscuro, line 1: Same as in the v4 PR, please bump the FPF version: https://github.com/DependencyTrack/dependency-track/pull/6181#pullrequestreview-4528101785

nscuro (changes requested):

nscuro (approved):

PR #6130, opened 2026-05-12
Send groupby to defect dojo

webdevred:
### Description

DefectDojo's import-scan and reimport-scan endpoints accept a `group_by` parameter that clusters findings into Finding Groups on import, but Dependency-Track never forwarded it. This adds a new per-project property `defectdojo.groupBy` that, when set, is sent as `group_by` in the multipart form for both import and reimport requests.

Also removes a pre-existing dead `.build()` call at the end of the `reimportDependencyTrackFindings` fluent chain whose return value was discarded before `request.setEntity()` was called.

I have verified that after the findings groups are created in Defect Dojo in my own local test environment.

bild
bild

### Addressed Issue

Closes #6061 

### Additional Details

The implementation follows the same pattern as `defectdojo.testTitle` (#4796): a per-project `ProjectProperty` is read in `DefectDojoUploader` and forwarded as a multipart form field in `DefectDojoClient`. No new abstraction was introduced.

The fix to `reimportDependencyTrackFindings` (removing the dangling `.build()` call at the end of the fluent `MultipartEntityBuilder` chain) is included here because it was discovered while adding `group_by` support to that method. It was pre-existing dead code, the return value was discarded and the actual entity was set correctly via a later `builder.build()` call, so it had no runtime effect.

### Checklist

- [x] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)
- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective
- [x] This PR implements an enhancement, and I have provided tests to verify that it works as intended
- [ ] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)
- [x] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly

Issue #6061, opened 2026-04-23
Add `defectdojo.groupBy` per-project property to control finding grouping on import


august-monitor:
### Current Behavior

Dependency-Track's DefectDojo integration publishes findings via the `import-scan` / `reimport-scan` endpoints but never sets the `group_by` parameter. As a result, findings pushed from DT arrive in DefectDojo as individual ungrouped entries, even though DefectDojo fully supports grouping them into Finding Groups on import.

When the same FPF report is uploaded manually through the DefectDojo UI, the user can select "Component Name" (or `component_name+component_version`, `file_path`, `finding_title`) in the Group By dropdown and the findings are grouped accordingly.

### Why this matters

In our setup, every other scanner we feed into DefectDojo is configured to group findings by component name. That is how our team triages: one row per vulnerable component, not N rows per CVE. It keeps the backlog readable, and it keeps DevOps ticket counts sane (one ticket per component instead of one per CVE, which for projects with many transitive dependencies is the difference between a workable backlog and an unusable one).

Dependency-Track's built-in integration is the only source that breaks this convention. Its findings land ungrouped alongside everyone else's grouped findings, which forces analysts to context-switch between two different triage models in the same product. The only way to restore consistency today is to disable the built-in integration entirely and replace it with a custom pipeline step that calls `/api/v2/reimport-scan/` directly with `group_by` set, which defeats most of the value of having a built-in integration.

### Proposed Behavior

Add a new optional per-project property. When set, its value is forwarded as `group_by` in the multipart form of the DefectDojo upload request:

| Attribute | Value |
| --- | --- |
| Group Name | `integrations` |
| Property Name | `defectdojo.groupBy` |
| Property Value | A valid DefectDojo `group_by` value (e.g. `component_name`, `component_name+component_version`, `file_path`, `finding_title`) |
| Property Type | `STRING` |

When the property is not set, behavior is unchanged (no `group_by` sent). Fully backwards compatible.

### Prior Art

This follows the same pattern as three previously accepted enhancements, all of which exposed a DefectDojo import parameter as a per-project property:

- #1622 : `defectdojo.reimport` (v4.6.0)
- #2424 : `defectdojo.doNotReactivate` (v4.8.0)
- #4796 : `defectdojo.testTitle` (v4.13.0)

The implementation should be mechanically equivalent to #4796: add the property constant, read it in `DefectDojoUploader`, and pass it through `DefectDojoClient` into the multipart form when non-null.

[...]

PR #6416, opened 2026-06-15
Forward group_by parameter to DefectDojo, v5 version

webdevred:
### Description

DefectDojo's import-scan and reimport-scan endpoints accept a `group_by` parameter that clusters findings into Finding Groups on import, but Dependency-Track never forwarded it. Operators who rely on Finding Groups for triage had to set `group_by` outside of DT or give it up entirely.

This adds a per-project property `defectdojo.groupBy` that, when set, is sent as `group_by` in the multipart form body for both import and reimport requests. When not set, behavior is unchanged.

This is the v5 companion to #6130, which makes the same change for `4.14.x`. Adapted for the `MultipartBodyPublisher`-based `DefectDojoClient` in the `apiserver` module.
dtrack-groupby
defectdojo-groupby

### Addressed Issue

Related to #6061. v5 companion to #6130 (same change targeting `4.14.x`).

### Additional Details

The implementation follows the same pattern as `defectdojo.testTitle`: a per-project `ProjectProperty` is read in `DefectDojoUploader` and forwarded as a form field via `MultipartBodyPublisher.addFormField`. No new abstraction was introduced.

The dead `.build()` fix included in the v4 PR does not apply here. The v5 `DefectDojoClient` uses `MultipartBodyPublisher` rather than Apache HttpClient's `MultipartEntityBuilder`, so there is no equivalent dead call to remove.

### Checklist

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/integrations/defectdojo/DefectDojoClient.java
+++ b/apiserver/src/main/java/org/dependencytrack/integrations/defectdojo/DefectDojoClient.java
@@ -60,7 +60,8 @@ public class DefectDojoClient {
             final String engagementId,
             final InputStream findingsJson,
             final Boolean verifyFindings,
-            final @Nullable String testTitle) {
+            final @Nullable String testTitle,
+            final @Nullable String groupBy) {
         LOGGER.debug("Uploading Dependency-Track findings to DefectDojo");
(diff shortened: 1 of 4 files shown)


nscuro (approved):
Thanks!

PR #6426, opened 2026-06-16
Add server release version to the OpenAPI spec info block

### LINKED ISSUE #6414: Restructure OpenAPI Spec Versions to Include Server Release Version (v5 Regression)
### Current Behavior

A regression was introduced in the v5 release regarding how API specification versions are surfaced in the OpenAPI documents.

The documentation is now split into two endpoints, both of which return static semantic versions tied strictly to the API contract rather than the software release:

- `/api/openapi.yaml` returns `version: 1.0.0`
- `/api/v2/openapi.yaml` returns `version: 2.0.0`

### The Problem
These static strings make it impossible to determine the exact server version (e.g., `5.0.1`) from which the spec was pulled. This is particularly problematic for the `/api/v1` endpoint; while the API structure itself might remain v1, the documentation continues to receive improvements, fixes, and updates across different server patch releases.

### Proposed Behavior

Ajust titles and versions as:

```yaml
# Example for /api/openapi.yaml
info:
  title: OWASP Dependency-Track
  version: API v1 / Server v5.0.1

# Example for /api/v2/openapi.yaml
info:
  title: OWASP Dependency-Track
  version: API v2 / Server v5.0.1
```

### Checklist

GabrielBBaldez:
### Description

Surfaces the server release version in the OpenAPI documents. The OpenAPI `version` field is part of the API contract and is intentionally left unchanged; instead, each spec's `info` block is enriched at load time with the server version appended to the `title` (visible in frontends such as Swagger UI / Redocly, which ignore custom fields) and exposed as a machine-readable `x-server-version` extension.

Example for `/api/v2/openapi.yaml`:

```yaml
info:
  title: OWASP Dependency-Track API (Server v5.0.1)
  version: 2.0.0
  x-server-version: 5.0.1
```

### Addressed Issue

Closes #6414

### Additional Details

Following the approach agreed in the issue, the enrichment is applied during the existing one-time spec load in both the v1 and v2 `OpenApiResource`, via a small reusable `OpenApiSpecEnricher`. The server version comes from `new About().getVersion()`; when it is unavailable the spec is returned unchanged. Unit tests cover the title / `x-server-version` enrichment and verify the API `version` is left untouched.

### Checklist

- [ ] ~This PR fixes a defect~
- [ ] ~This PR introduces changes to the database model~
- [ ] ~This PR alters existing behavior and requires documentation updates~ (no user-facing docs cover the OpenAPI `info` block)
- [ ] ~This PR is a substantial change per the ADR criteria~

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/resources/OpenApiSpecEnricher.java
+++ b/apiserver/src/main/java/org/dependencytrack/resources/OpenApiSpecEnricher.java
@@ -1,0 +1,67 @@
+/*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.resources;
+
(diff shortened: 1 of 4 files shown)


nscuro (approved):
Thanks!

PR #6431, opened 2026-06-17
Fix admin re-seeding when USER table is populated

### LINKED ISSUE #6392: On first boot database seeding creates MANAGED admin despite the USER table being already populated
### Current Behavior

Following limitation #6312, we renamed managed `admin` user as `disabled__admin` with v4 to avoid possible conflict with any OIDC user.

### Steps to Reproduce

1. Convert a v4 DB whose MANAGED `admin` user was renamed and contains no other users
2. Start DT v5

-> Seed MANAGED admin user is created again

### Expected Behavior

No seeding of admin user if other users already exists.

### Dependency-Track Version

5.x

### Browser

N/A

### Checklist

Clark1945:
seedDefaultUsers checked only for an existing 'admin' username conflict. After a v4->v5 migration where the admin was renamed (e.g. to 'disabled__admin'), the table is populated but has no user named 'admin', causing a new MANAGED admin with the default password to be created unexpectedly.

Fix by replacing VALUES with SELECT...WHERE NOT EXISTS to skip seeding entirely when any user already exists. Retain ON CONFLICT DO NOTHING as a safety net for concurrent inserts.

Add a regression test covering the renamed-admin scenario.

Closes #6392

"

### Description

Fix admin re-seeding when USER table is already populated.
The seedDefaultUsers method only checked for a username conflict,
causing a new MANAGED admin to be created after v4->v5 migration
when the original admin was renamed (e.g. to 'disabled__admin').

### Addressed Issue

Closes #6392

### Additional Details

Changed the INSERT...VALUES to INSERT...SELECT...WHERE NOT EXISTS
to skip seeding entirely when any user already exists. Retained
ON CONFLICT DO NOTHING as a safety net for concurrent inserts.
Added a regression test covering the renamed-admin scenario.

### Checklist

- [ ] ~This PR implements an enhancement~
- [ ] ~This PR introduces changes to the database model~
- [ ] ~This PR introduces new or alters existing behavior~
- [ ] ~This PR is a substantial change~

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/persistence/DatabaseSeedingInitTask.java
+++ b/apiserver/src/main/java/org/dependencytrack/persistence/DatabaseSeedingInitTask.java
@@ -195,11 +195,17 @@ public final class DatabaseSeedingInitTask implements InitTask {
     }
 
     public static void seedDefaultUsers(final Handle jdbiHandle) {
+        // Only seed the default 'admin' user when no users exist at all.
+        // Checking solely for an existing 'admin' username is not sufficient:
+        // an upgraded v4 instance whose admin user was renamed (e.g. to avoid
+        // collisions with OIDC users) would otherwise get a fresh 'admin' user
+        // with the default password, despite already having users. See #6392.
         final Optional<Long> adminUserId = jdbiHandle.createUpdate("""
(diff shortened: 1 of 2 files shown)


nscuro (approved):
Thanks!

PR #6475, opened 2026-06-19
Add project `isActive` parameter for BOM upload API

### LINKED ISSUE #6285: Support ephemeral CI/CD scan with optional active  parameter and retention-based cleanup
### Proposed Behavior

**Summary**

Introduce an optional active property for project bom (v1/bom) , with a default value of true, allowing users to explicitly mark projects as inactive when they upload bom. This would be particularly beneficial for CI/CD, automation, and other short-lived scanning workflows where projects are temporary and do not require ongoing monitoring.

**Problem**

Dependency-Track v5 significantly improves scalability by enabling dynamic workload scaling and prioritisation of event driven scans over periodic scans. These enhancements make Dependency-Track increasingly attractive for CI/CD pipelines and automated security validation use cases.
However, a potential barrier to adoption remains: many CI/CD-generated projects are ephemeral by nature. They exist only for the duration of a build, pull request, release validation, or other temporary automation workflow. 

These scan/projects typically:
- Do not require long-term portfolio visibility.
- Do not need continuous monitoring for newly disclosed vulnerabilities.
- Can generate substantial portfolio growth and operational overhead when retained indefinitely.
- May require separate cleanup processes to prevent portfolio clutter.

As a result, organisations may hesitate to fully integrate Dependency-Track into high-volume automation workflows because every scan currently results in a standard project lifecycle, regardless of whether the project is intended to be long-lived or temporary.

**Proposed Solution**

Introduce a new optional project attribute:

`active = true (default)`

When set to false, the project would be treated as inactive and intended for temporary or short-lived use cases.

This could be combined with existing Project Retention Policies, allowing organisations to automatically remove inactive projects after a configurable retention period ( https://dependencytrack.github.io/docs/next/guides/administration/configuring-project-retention/#pick-a-policy ).

**Example use cases:**

Pull request validation scans
Feature branch builds
Release candidate verification
Automated security gates
Temporary compliance validation workflows

In these scenarios, users could create projects with active=false, receive all analysis and policy evaluation results, and then rely on retention policies to automatically clean up the projects once they are no longer relevant.

**Benefits**
Improved CI/CD Adoption

Organisations could safely integrate Dependency-Track into high-volume automation workflows without concern about long-term portfolio growth.

Clear Separation of Use Cases

Teams could use the same Dependency-Track instance for both:

Long-lived compliance and governance projects (active=true)
Short-lived CI/CD and automation projects (active=false)
Reduced Operational Overhead

Inactive projects could be automatically removed through existing retention mechanisms, reducing manual cleanup requirements. https://dependencytrack.github.io/docs/next/guides/administration/configuring-project-retention/#pick-a-policy

Better Resource Utilisation

Projects that are no longer relevant would not continue consuming monitoring and portfolio management resources unnecessarily.

### Checklist

sahibamittal:
### Description

Introduce an optional `isActive` property for project bom (v1/bom).
```
isActive = null -> no action
isActive = false -> deactivate project with inactiveSince = now
isActive = true -> activate project
```

### Addressed Issue

Closes https://github.com/DependencyTrack/dependency-track/issues/6285

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
+++ b/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
@@ -188,6 +188,7 @@ final class ProjectQueryManager extends QueryManager {
         project.setParent(parent);
         project.setPurl(purl);
         project.setInactiveSince(inactiveSince);
+        project.setActive(inactiveSince == null);
         project.setIsLatest(isLatest);
         return createProject(project, tags, commitIndex);
     }
--- a/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
+++ b/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
@@ -188,7 +188,6 @@ final class ProjectQueryManager extends QueryManager {
         project.setParent(parent);
         project.setPurl(purl);
         project.setInactiveSince(inactiveSince);
-        project.setActive(inactiveSince == null);
         project.setIsLatest(isLatest);
         return createProject(project, tags, commitIndex);
     }
--- a/apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
+++ b/apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
@@ -334,10 +334,7 @@ public class BomResource extends AbstractApiResource {
                       the response's content type will be <code>application/problem+json</code>.
                     </p>
                     <p>
-                      The upload can target an existing project directly with <code>project</code>, or use
-                      <code>projectName</code> and <code>projectVersion</code> to look up the project and,
-                      with <code>autoCreate</code>, create it when it does not already exist. When creating
-                      projects, <code>parentUUID</code> or <code>parentName</code> and
+                      When creating projects, <code>parentUUID</code> or <code>parentName</code> and
                       <code>parentVersion</code> can place the new project under a parent,
                       <code>projectTags</code> can apply tags, and <code>isLatest</code> can mark it as
                       the latest version. The <code>isActive</code> parameter sets the project's active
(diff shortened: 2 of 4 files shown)


apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
```diff
@@ -188,6 +188,7 @@ public Project createProject(String name, String description, String version, Co
         project.setParent(parent);
         project.setPurl(purl);
         project.setInactiveSince(inactiveSince);
+        project.setActive(inactiveSince == null);
```
nscuro: Nit: Judging by `setActive`'s code, this call is effectively a no-op.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -497,6 +502,15 @@ public Response uploadBom(@Parameter(required = true) BomSubmitRequest request)
         return processUpload(projectInfo, bomBytes);
     }
 
+    private void maybeSetProjectActive(QueryManager qm, Project project, Boolean isActive) {
+        if (isActive == null) {
+            return;
+        }
+
+        project.setActive(isActive);
+        qm.persist(project);
```
nscuro: Nit: Calling `persist` is redundant because `project` is already attached to the persistence context when `maybeSetProjectActive` is called.

You could simply inline this method at the call sites as:

```java
if (isActive != null) {
    project.setActive(isActive);
}
```

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
nscuro, line 1: Could you add a paragraph to both `@Operation`'s `description`s that explains what the parameters do and why one might want to use them? Also the behavior of modifying the active state even for existing projects might be non-obvious to clients and should be mentioned there.

nscuro (changes requested):
Operation docs should be extended so the generated OpenAPI spec will give clients context about the new parameters.

The other comments are just nits, apply or ignore at your own judgement.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -331,6 +333,17 @@ public Response exportComponentAsCycloneDx(
                       a response with problem details in RFC 9457 format will be returned. In this case,
                       the response's content type will be <code>application/problem+json</code>.
                     </p>
+                    <p>
+                      The upload can target an existing project directly with <code>project</code>, or use
+                      <code>projectName</code> and <code>projectVersion</code> to look up the project and,
+                      with <code>autoCreate</code>, create it when it does not already exist. When creating
```
nscuro: This duplicates information from the first paragraph. Same for the other endpoint.

nscuro (changes requested):

nscuro (approved):

Issue #6561, opened 2026-07-01
Updating an analysis fails without `project`, but `project` is never used and should just be ignored


Zusel:
### Current Behavior

Hey,
I found out that I can't add a comment or change the analysis state of a vuln via API. Every time I specify a project, component, and vuln, Dependency-Track adds the comment and changes the state, BUT in a completely different project. Without a project, Dependency-Track throws a 500, even though project isn't required according to Swagger.
It seems to me that the commit "0dbe529ca6f446c42f7ac30327d1e7e9cfdea6b0" by @nscuro broke the endpoint by removing the project filtering (no finger-pointing, just a heads-up — hopefully you're faster with research and fixing the problem :))

After a bit of research, I noticed that project has been deprecated. Now it's not entirely clear to me how I should reference a vuln when a component can be part of multiple projects.

### Steps to Reproduce

Add a comment to a vuln that has more than one affectedProject via /v1/analysis

### Expected Behavior

Create a comment / change state on the specified vuln/project/component

### Dependency-Track Version

4.x

### Browser

Google Chrome

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

nscuro:
> Now it's not entirely clear to me how I should reference a vuln when a component can be part of multiple projects.

Component records are logically bound to a single project, they are **not** shared across projects. That's why specifying the project separately was redundant and thus removed.

Also note how the commit you referenced did not ship in v4, but you specified version 4.x. In 4.x, the project must still be specified.

Zusel:
Hey, thanks for the quick reply. According to the Swagger definition, though, "project" is already no longer required in v4, and it's still throwing the 500 there.
Here's the issue, and also a bit of context on how I got here: We're working with the notifications, which send a notification to one of our web servers when new vulns are found. Now, both the AffectedProjects and the Component are included in the notification. So what should I rely on? Do I get a notification for every AffectedProject, or do I need to look up the other UUIDs for the component in other projects based on the component's purl — or how is this supposed to work?

nscuro:
Hmmm, yeah you'll get a notification for each affected project anyway so I'd recommend ignoring the affectedProjects array normally. It's a historic artifact from v3 where components were indeed global object and not tied to projects.

[...]

PR #6562, opened 2026-07-01
Mark project as not required and only validate when specified.

Zusel:
### Description

"Hey, so after our little chat, I thought I'd just quickly add it myself. My thinking behind it was that I'd simply check whether the component's project and the (optional) project match. That way we can resolve the 500 and also get users used to the idea that the project field will go away entirely. With the conflict, we can then prevent someone (like me) from wondering why what they expected isn't happening :)

### Addressed Issue

https://github.com/DependencyTrack/dependency-track/issues/6561

### Checklist

- [ ] I have read and understand the [contributing guidelines] (No 404 not found)
- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective (no)
- [ ] This PR implements an enhancement, and I have provided tests to verify that it works as intended (no)
- [ ] This PR introduces changes to the database model, and I have updated the [migration changelog] accordingly (no)
- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly (cant find any docs)
- [ ] This PR is a substantial change (per the [ADR criteria]), and I have added an [ADR] under `docs/adr/` (no)

[ADR]: ../docs/adr/

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/AnalysisResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/AnalysisResource.java
@@ -147,7 +147,6 @@ public class AnalysisResource extends AlpineResource {
     public Response updateAnalysis(AnalysisRequest request) {
         final Validator validator = getValidator();
         failOnValidationError(
-                validator.validateProperty(request, "project"),
                 validator.validateProperty(request, "component"),
                 validator.validateProperty(request, "vulnerability"),
                 validator.validateProperty(request, "analysisState"),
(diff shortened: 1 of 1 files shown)


nscuro (changes requested):
Thanks for the PR!

I get the intention behind failing for inconsistent project <-> component parameters, but it will cause all existing clients to break that did not consider this. The previous behavior was mostly harmless because what mattered has always been the `component` parameter. Turning that into an error is quite blunt and not appropriate for a patch version.

I'd suggest we just follow what 0dbe529ca6f446c42f7ac30327d1e7e9cfdea6b0 (v5) did:

* Ignore the `project` parameter entirely
* Deprecate the `project` parameter so it becomes visible in the OpenAPI spec: https://github.com/DependencyTrack/dependency-track/blob/157afd3930ceaa3422684c2d8a2a331647fa17f6/apiserver/src/main/java/org/dependencytrack/resources/v1/vo/AnalysisRequest.java#L40-L43

Zusel:
Heyho,
I get the point that throwing a 409 might be a bit dramatic and that it will/could break some things. Still, I think just deprecating it is the wrong way to go. Nobody notices that this field is deprecated — you'd have to actually read the code for that, and normally you don't. Besides, the 409 only occurs if you specify a projectId and then hit the wrong project. If someone doesn't:

- specify the wrong project
- specify a project at all

then it will never affect them.
So I really do consider it very, very important that this is handled cleanly. Because if you're interacting with the REST API here and your vulnerability collection suddenly gets edited completely wrong, I'd consider that CRITICAL. You could easily destroy your entire vulnerability inventory this way.
Maybe a bit dramatic for a patch update, but I consider it necessary.

PR #6564, opened 2026-07-02
Add GitHub App authentication for the GitHub Advisories mirror

### LINKED ISSUE #6563: Support GitHub App authentication for the GitHub Advisories mirror
### Current Behavior

The GitHub Security Advisories mirror (`vuln-data-source/github`) authenticates only with a
**Personal Access Token**. A PAT is long-lived, tied to an individual user account, inherits that
user's broad scope, doesn't rotate on its own, and breaks when the person leaves. For an unattended
server-side integration that's the wrong credential shape.

### Proposed Behavior

Add **GitHub App authentication as an alternative to PAT** for the advisories mirror. PAT support
stays unchanged and fully backward compatible.

A GitHub App swaps the long-lived, user-owned PAT for a **short-lived (~1h), least-privilege,
app-owned** installation token. (It does *not* remove long-lived secrets — the App's RSA private key
is still long-lived — but the working credential is short-lived, the App scope is minimal, the key can
be vaulted sign-only, and it rotates/revokes per-App instead of per-user.)

Installation tokens read public GHSA data at the full installation rate limit (5,000+/hr), so they
authenticate the GraphQL mirror exactly as a PAT does today (validated against GitHub docs).

## Scope
- GitHub Advisories mirror only.
- GitHub Enterprise Server (derive the token endpoint from the configured `apiUrl`).
- NOT the package-metadata resolver — stays PAT-only (niche `pkg:github` path; App auth doesn't fit it well).
- NO change to private/git-sourced Cargo/Go crate enrichment (those never reach the GitHub resolver).

## Acceptance criteria
- [ ] Admin can configure the advisories source with either a PAT (as today) or a GitHub App
      (App ID + installation ID + private key).
- [ ] Config validation requires exactly one auth method; both/neither is a clear error.
- [ ] Existing PAT configs keep working with zero migration.
- [ ] Installation token is minted, cached, and auto-refreshed so a mirror run exceeding 1h does
      not fail on token expiry.
- [ ] Works against GitHub.com and GitHub Enterprise Server.
- [ ] The App private key is stored as a `SecretManager` secret, like the existing `apiToken`.
- [ ] Tests cover JWT signing, token refresh (fake clock), and the token exchange (WireMock).

## Implementation notes (see ADR-031 for full rationale)
- No new dependencies — RS256 signing via the JDK (`java.security.Signature`), JSON via Jackson,
  HTTP via the existing client. Accept the PKCS#1 key GitHub gives you (wrap to PKCS#8 in ~10 lines);
  do not add BouncyCastle (the build deliberately excludes it).
- Introduce a `GitHubTokenProvider` (interface) with `StaticTokenProvider` (PAT) and
  `GitHubAppTokenProvider` (App) impls; inject via the existing `httpClientSupplier` seam using a
  request interceptor so PAT and App share one auth path.
- Extend the `github-vuln-data-source-config-v1` schema additively with `appId`, `installationId`,
  `appPrivateKey` (secret-ref); infer the mode from which fields are set.

### Checklist

nissessenap:
### Description

Adds GitHub App authentication as an alternative to a Personal Access Token (PAT)
for the **GitHub Security Advisories mirror** (`vuln-data-source/github`). PAT
support is unchanged; the change is additive and backward compatible.

Will need seperate PR for frontend.

### Addressed Issue

Closes #6563

### Additional Details

New, self-contained auth code in `vuln-data-source/github` (no new dependencies —
Jackson is already on the classpath):

- `GitHubTokenProvider` — `currentToken()` seam.
- `StaticTokenProvider` — returns the configured PAT.
- `GitHubAppTokenProvider` — mints & caches an installation token:
  - Builds the App JWT and signs it with the JDK's `Signature` (`SHA256withRSA`) —
    no JWT/crypto library.
  - Reads the App private key with the JDK `KeyFactory`. The key must be in
    **PKCS#8** (`-----BEGIN PRIVATE KEY-----`); GitHub downloads are PKCS#1, so
    operators convert once with `openssl pkcs8 -topk8 -nocrypt`. This keeps key
    handling pure-JDK with no hand-assembled DER and no BouncyCastle (which the
    build deliberately excludes). A PKCS#1 key is rejected with an error carrying
    the exact conversion command.
  - Parses the token-exchange response with Jackson rather than a regex.
  - Exchanges the JWT for an installation token, caches it, and re-mints on
    access within a 5-min skew of expiry (check-on-access; no threads/timers).
    An injected `Clock` makes expiry deterministically testable.
  - Derives the token-exchange endpoint from `apiUrl` so GitHub Enterprise
    Server works.
- `BearerTokenInterceptor` — an Apache HttpClient5 request interceptor that
  removes any existing `Authorization` header and stamps
  `Authorization: Bearer <currentToken()>`. Both PAT and App modes flow through
  it, so there's a single, self-refreshing auth path (a >1h initial sync no
  longer fails on token expiry).

Config (`github-vuln-data-source-config-v1` schema) extended additively with
`appId`, `installationId`, and `appPrivateKey` (`x-secret-ref`). The auth mode is
inferred from presence — `runtimeConfigSpec()` enforces exactly one of: `apiToken`,
or all three App fields. Existing PAT-only configs keep working with zero migration.

### Checklist

- [X] I have read and understand the [contributing guidelines]

[ADR]: ../docs/adr/


vuln-data-source/github/src/main/java/org/dependencytrack/vulndatasource/github/GitHubAppTokenProvider.java
```diff
@@ -0,0 +1,214 @@
...
+    private static final Duration REFRESH_SKEW = Duration.ofMinutes(5);
+
+    /**
+     * Upper bound on the token-exchange request. The interceptor calls
+     * {@link #currentToken()} on the async client's I/O thread, so an unbounded
+     * request could stall it indefinitely if GitHub's token endpoint hangs.
+     */
+    private static final Duration TOKEN_EXCHANGE_TIMEOUT = Duration.ofSeconds(10);
+
+    private static final Pattern TOKEN_FIELD = Pattern.compile("\"token\"\\s*:\\s*\"([^\"]+)\"");
```
samwright: parsing json with regex doesn't sound fun, I would use a library to parse the json instead.

vuln-data-source/github/src/main/java/org/dependencytrack/vulndatasource/github/GitHubAppTokenProvider.java
```diff
@@ -0,0 +1,214 @@
...
+        this.cachedToken = tokenMatcher.group(1);
+        this.cachedTokenExpiresAt = Instant.parse(expiresMatcher.group(1));
+    }
+
+    private static final Pattern PEM_ARMOR = Pattern.compile("-----[^-]+-----|\\s");
+
+    /** DER-encoded AlgorithmIdentifier for rsaEncryption (with NULL parameters). */
+    private static final byte[] RSA_ALGORITHM_ID = {
+            0x30, 0x0D, 0x06, 0x09, 0x2A, (byte) 0x86, 0x48, (byte) 0x86,
+            (byte) 0xF7, 0x0D, 0x01, 0x01, 0x01, 0x05, 0x00
```
samwright: maybe just adding a BouncyCastle would be better than this?

nissessenap: On the BouncyCastle suggestion for the PKCS#1 key parsing: I want to flag a tension before we take it. BC isn't merely absent from this module — it's deliberately excluded in the root pom.xml (lines 468–477), where bcprov-jdk18on is stripped out of the MinIO dependency with the comment "Schlepping a 9MB JAR along for features we don't use is not worth it."

Otion 2 is to make the way of working a bit worse for the admins. 
Instead of copying PEM file directly from GitHub app we ask them to run 
Accept PKCS#8 keys and document the one-time command: 
```
openssl pkcs8 -topk8 -nocrypt conversion
```
then key parsing is 2 lines of pure JDK, no bytes, no dep.

I will go with option 2 and just remove it and document how to do it instead.

nscuro (approved):
Thanks!

Frontend change is not necessary since the config form is rendered from the JSON schema.

But an update to the user docs that explains why / when to use this and how to configure it (i.e. the necessary private key conversion) would be good: https://github.com/DependencyTrack/docs/blob/main/docs/guides/administration/configuring-vulnerability-sources.md#github-advisories

PR #6614, opened 2026-07-07
Fix NPE and redundant queries in VEX/VDR export

### LINKED ISSUE #6613: VEX/VDR export fails with NullPointerException when a finding's component or vulnerability was deleted
### Current Behavior

When exporting a project as VEX or VDR (CycloneDX), ModelConverter.convert() resolves the Component and Vulnerability for each finding by UUID but dereferences them without a null check. If a component or vulnerability referenced by a finding is deleted in the window between the findings query and the export (e.g. a concurrent re-analysis or manual deletion), the lookup returns null and the next line (component.getProject()) throws a NullPointerException, aborting the entire export. The user receives a 500 error instead of a document, even though only one finding out of possibly many is affected.

Additional note: there is a secondary inefficiency in the same method — for VEX/VDR
the analysis lookup re-fetches the Component and Vulnerability by UUID a second time
even though both are already resolved earlier in the method, causing two redundant
database queries per finding. I have a fix ready for both issues, including a
regression test, and would like to submit a PR referencing this issue.

### Steps to Reproduce

1. Create a project with a component that has at least one vulnerability.
2. Delete that component (or trigger a concurrent deletion while an export is in flight).
3. Request a VEX or VDR export: GET /api/v1/bom/cyclonedx/project/{uuid}?variant=vex
4. The export fails with a NullPointerException instead of returning a document that simply omits the deleted item.

### Expected Behavior

The export should skip findings whose component or vulnerability no longer exists and complete successfully with the remaining vulnerabilities included.

### Dependency-Track Version

4.x

### Browser

N/A

### Checklist

heyiamwahab236:
### Description

Fixes a `NullPointerException` that aborts the entire VEX/VDR CycloneDX export when a
finding references a component or vulnerability that was deleted between the findings
query and the export. `ModelConverter.convert()` resolved the `Component` and
`Vulnerability` by UUID but dereferenced them without null checks, so a concurrent
deletion (e.g. re-analysis or manual removal) caused `component.getProject()` to throw
and the whole export to fail with a 500. Such findings are now skipped via a guard in
`convert()` and a `filter` in `generateVulnerabilities()`, so the export completes with
the remaining vulnerabilities.

Also removes two redundant database queries per finding: for the VEX/VDR variant the
analysis lookup re-fetched the `Component` and `Vulnerability` by UUID even though both
were already resolved earlier in the same method. It now reuses the resolved objects.

### Addressed Issue

fixes #6613

### Additional Details

A regression test (`ModelConverterTest`) was added covering both a finding whose
component was deleted (previously reproduced the NPE, now returns an empty result) and
a normal finding that still converts correctly. The test was verified to fail against
the unpatched code and pass with the fix.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
+++ b/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
@@ -837,8 +837,14 @@ public class ModelConverter {
     public static org.cyclonedx.model.vulnerability.Vulnerability convert(final QueryManager qm, final CycloneDXExporter.Variant variant,
                                                                           final Finding finding) {
         final Component component = qm.getObjectByUuid(Component.class, (String)finding.getComponent().get("uuid"));
+        if (component == null) {
+            return null;
+        }
         final Project project = component.getProject();
         final Vulnerability vulnerability = qm.getObjectByUuid(Vulnerability.class, (String)finding.getVulnerability().get("uuid"));
+        if (vulnerability == null) {
+            return null;
+        }
 
         final org.cyclonedx.model.vulnerability.Vulnerability cdxVulnerability = new org.cyclonedx.model.vulnerability.Vulnerability();
         cdxVulnerability.setBomRef(vulnerability.getUuid().toString());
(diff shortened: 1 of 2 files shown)


heyiamwahab236:
@nscuro Can you please have a look at it.

nscuro (approved):
Thanks!

Issue #6383, opened 2026-06-12
NullPointer Exception: [ResolvePackageMetadataActivity] Failed to resolve metadata


mulder999:
### Current Behavior

Following v4->v5 migration, we receive constant: `java.lang.NullPointerException` on some systems.

Logs shows for example:

```text
WARN [ResolvePackageMetadataActivity] Failed to resolve metadata; persisting empty result [pkgMetadataResolverName=nuget, purl=pkg:nuget/Microsoft%C2%AE%20C%20Runtime%20Library@10.0.17134.12]
```

Component name and PURL tends to indicate a flaw in the conversion of the registered trademark sign:

<img width="1051" height="1163" alt="Image" src="https://github.com/user-attachments/assets/3c03a04a-6e30-41a4-bf62-484a721178a3" />

### Steps to Reproduce

1. Convert DT from v4 to v5
2. Launch DTv5 and wait. Errors then is raised every seconds

### Expected Behavior

1. Robust processing in ResolvePackageMetadataActivity
2. Improved conversion logic

### Dependency-Track Version

5.x

### Browser

N/A

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

nscuro:
The NPE is likely fixed via https://github.com/DependencyTrack/dependency-track/pull/6303.

You can apply an ad-hoc fix by running:

```sql
update "REPOSITORY" set "AUTHENTICATIONREQUIRED" = false where "AUTHENTICATIONREQUIRED" is null;
```

mulder999:
Thanks for proposal, unfortunately it does not resolve the issue even after apiserver restart.

nscuro:
Can you post the rest of the warning log please, i.e. the stack trace?

nscuro:
The encoding issue of the PURL caused by multiple bugs in the `packageurl-java` library we use. They're already fixed but not released yet. I asked in https://github.com/package-url/packageurl-java/issues/256#issuecomment-4690217614.

mulder999:
Sorry for the delay; it turns out our log forwarding agent was stuck repeatedly transmitting the same incomplete logs.

[...]

PR #6679, opened 2026-07-13
Issue-6654 : Skip malformed PURLs instead of assigning to empty resolver

### LINKED ISSUE #6654: Missing negative-cache for Malformed PURLs
### Current Behavior

Components with a malformed PURL (e.g. `pkg:nuget/serilog.sinks.file%A%20...5.0.0@5.0.0`) are re=parsed and re-logged as WARN on every package-metadata resolution cycle (daily cron + BOM-triggered runs).

**Root Cause**: the negative-cache path never persists anything for malformed PURLs, so they keep re-qualifying as candidates.

Version tested: 5.0.0-rc.1

### Steps to Reproduce

1. Ingest a BOM containing component with structurally-invalid PURL (bad percent encoding).
2. Trigger package-metadata resolution via bom upload.
3. Same PURL will be re-selected as candidate logging same WARN on repeat every cycle indefinitley.

### Expected Behavior

A malformed PURL should be attempted at most once (or skipped entirely) rather than re-parsed every cycle, and should not spam WARN logs. Valid unresolvable PURLs keep their existing 24h negative-cache behaviour.

Filter un-parseable PURLs out of candidate selection, so malformed PURLs never enter the resolution loop.

### Dependency-Track Version

5.x

### Browser

N/A

### Checklist

sahibamittal:
### Description

When a PURL fails to parse, skip it entirely rather than assigning it to an empty resolver. This prevents unnecessary resolution attempts and reduces log spam from recurring WARN messages on each cycle. Malformed PURLs are a data quality issue and should not be processed.

Added test verifying malformed PURLs are excluded from candidates and not reselected across workflow runs.

### Addressed Issue

Closes https://github.com/DependencyTrack/dependency-track/issues/6654

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/pkgmetadata/FetchPackageMetadataResolutionCandidatesActivity.java
+++ b/apiserver/src/main/java/org/dependencytrack/pkgmetadata/FetchPackageMetadataResolutionCandidatesActivity.java
@@ -73,10 +73,9 @@ public final class FetchPackageMetadataResolutionCandidatesActivity
             try {
                 purl = new PackageURL(purlStr);
             } catch (MalformedPackageURLException e) {
-                LOGGER.warn("Failed to parse PURL '{}'; Assigning to empty resolver", purlStr, e);
-                purlsByResolver
-                        .computeIfAbsent("", k -> new ArrayList<>())
-                        .add(purlStr);
+                // Malformed PURL cannot be resolved and is data quality issue.
+                // Skip to avoid resolution attempts and spamming WARN logs on every cycle.
+                LOGGER.debug("Skipping malformed PURL '{}'", purlStr, e);
                 continue;
             }
 
--- a/apiserver/src/test/java/org/dependencytrack/pkgmetadata/ResolvePackageMetadataWorkflowTest.java
+++ b/apiserver/src/test/java/org/dependencytrack/pkgmetadata/ResolvePackageMetadataWorkflowTest.java
@@ -277,6 +277,65 @@ class ResolvePackageMetadataWorkflowTest extends PersistenceCapableTest {
     }
 
     @Test
+    void shouldSkipMalformedPurlAndNotReselectIt() {
+        final String malformedPurl = "pkg:nuget/serilog.sinks.file%A%20...5.0.0@5.0.0";
+        final Instant resolvedAt = Instant.now();
+        mockResolveFnRef.set(purl -> new PackageMetadata("9.9.9", Instant.now(), resolvedAt, null));
+
+        var project = new Project();
+        project.setName("test-project");
+        project = qm.persist(project);
+
+        final var malformedComponent = new Component();
+        malformedComponent.setProject(project);
+        malformedComponent.setName("serilog.sinks.file");
+        malformedComponent.setVersion("5.0.0");
+        malformedComponent.setPurl(malformedPurl);
+        qm.persist(malformedComponent);
+
+        final var componentFoo = new Component();
+        componentFoo.setProject(project);
+        componentFoo.setGroup("org.acme");
+        componentFoo.setName("foo");
+        componentFoo.setVersion("1.0");
+        componentFoo.setPurl("pkg:maven/org.acme/foo@1.0");
+        qm.persist(componentFoo);
+
+        final var fetchActivity = new FetchPackageMetadataResolutionCandidatesActivity(pluginManager);
+        final List<String> candidatePurls = fetchActivity.execute(null, null)
+                .getCandidateGroupsList().stream()
+                .flatMap(group -> group.getPurlsList().stream())
+                .toList();
+        assertThat(candidatePurls).contains("pkg:maven/org.acme/foo@1.0")
+                .doesNotContain(malformedPurl);
+
+        // Running workflow again should not reselect the malformed PURL
+        for (int i = 0; i < 2; i++) {
+            final UUID runId = workflowTest.getEngine().createRun(
+                    new CreateWorkflowRunRequest<>(ResolvePackageMetadataWorkflow.class));
+            workflowTest.awaitRunStatus(runId, WorkflowRunStatus.COMPLETED);
+        }
+
+        List<String> metadataPurls = withJdbiHandle(handle -> handle
+                .createQuery("""
+                        SELECT "PURL"
+                          FROM "PACKAGE_METADATA"
+                        """)
+                .mapTo(String.class)
+                .list());
+        assertThat(metadataPurls).containsExactly("pkg:maven/org.acme/foo");
+
+        metadataPurls = withJdbiHandle(handle -> handle
+                .createQuery("""
+                        SELECT "PURL"
+                          FROM "PACKAGE_ARTIFACT_METADATA"
+                        """)
+                .mapTo(String.class)
+                .list());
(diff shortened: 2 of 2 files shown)


apiserver/src/main/java/org/dependencytrack/pkgmetadata/FetchPackageMetadataResolutionCandidatesActivity.java
```diff
@@ -73,10 +73,9 @@ public FetchPackageMetadataResolutionCandidatesRes execute(
...
                 purl = new PackageURL(purlStr);
             } catch (MalformedPackageURLException e) {
-                LOGGER.warn("Failed to parse PURL '{}'; Assigning to empty resolver", purlStr, e);
-                purlsByResolver
-                        .computeIfAbsent("", k -> new ArrayList<>())
-                        .add(purlStr);
+                // Malformed PURL cannot be resolved and is data quality issue.
+                // Skip to avoid resolution attempts and spamming WARN logs on every cycle.
+                LOGGER.debug("Skipping malformed PURL '{}'", purlStr, e);
                 continue;
```
nscuro, line 79: Simply skipping these would cause them to always remain eligible for resolution because no result is persisted for them.

The previous behavior assigned the "empty" resolver, which then later trivially persists empty results:

https://github.com/DependencyTrack/dependency-track/blob/13d5b2de05757aff811192ce8bc43c8ec0813828/apiserver/src/main/java/org/dependencytrack/pkgmetadata/ResolvePackageMetadataActivity.java#L97-L108

This prevents the next `FetchPackageMetadataResolutionCandidatesRes` iteration from discovering entries with invalid PURLs again.

By skipping this mechanism, `ResolvePackageMetadataWorkflow` would effectively loop forever and never terminate. It basically breaks the exit condition, i.e. this query returning no results:

https://github.com/DependencyTrack/dependency-track/blob/13d5b2de05757aff811192ce8bc43c8ec0813828/apiserver/src/main/java/org/dependencytrack/pkgmetadata/FetchPackageMetadataResolutionCandidatesActivity.java#L118-L130

nscuro (changes requested):

apiserver/src/main/java/org/dependencytrack/pkgmetadata/FetchPackageMetadataResolutionCandidatesActivity.java
sahibamittal: I'm seeing the same malformed PURL being logged on every resolution cycle with the current implementation. So the empty resolver path is not actually preventing it from being rediscovered. Same malformed PURL should not be processed again after the first successful workflow execution. 
I'll check whether the empty result is being persisted and why `FetchPackageMetadataResolutionCandidatesRes` is still selecting it.

sahibamittal: Okay found it, I noticed that for the malformed PURL, `ResultBuffer.addEmptyResult()` returns immediately without persisting empty results because `PurlUtil.silentPurl(purlStr)` returns null:

```
final PackageURL purl = PurlUtil.silentPurl(purlStr);
if (purl == null) {
    return;
}
```

As a result, no `PackageMetadata` or `PackageArtifactMetadata` records are written, and the same component is selected again by `FetchPackageMetadataResolutionCandidatesActivity` on every cycle. This needs to be fixed.

nscuro: The model currently requires a valid PURL since the identifiers are `PackageURL` objects. We'd need a way to pass empty results for broken PURLs, ideally without loosening the types (i.e. changing from `PackageURL` to `String`) too much.

sahibamittal: OR instead of persisting malformed PURLs (empty results), should it not just reject or sanitize them during BOM ingestion so they never reach the metadata resolver?

Also, currently we decide resolution state based on metadata tables if data exists. How about persisting the outcome of metadata resolution separately from the metadata itself. Say, `PACKAGE_METADATA_RESOLUTION` to record resolution state (e.g. SUCCESS, EMPTY, INVALID_PURL etc) keyed by the raw PURL. This way metadata becomes just the payload.

nscuro: > OR instead of persisting malformed PURLs (empty results), should it not just reject or sanitize them during BOM ingestion so they never reach the metadata resolver?

Hmmm that is already happening:

https://github.com/DependencyTrack/dependency-track/blob/f7e5611ca873eab35991f65947ba1c59bb970d7f/apiserver/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java#L224-L232

Which begs the question how you could still see invalid PURLs reaching the metadata resolver. I wonder if it's related to an encoding bug in the `packageurl-java` library (see https://github.com/DependencyTrack/dependency-track/issues/6383 where it came up recently). Does that error pattern match what you're observing in your instance?

sahibamittal: I checked none of repository records have authenticationRequired set to null. Error is due to malformed escape pair in invalid PURL during metadata resolution:

image

nscuro: Ah, no I meant the PURL encoding issue that's also mentioned there. 

`pkg:nuget/serilog.sinks.file%A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%205.0.0@5.0.0` is likely not how the PURL looked when it entered the system, the `packageurl-java` lib may have borked it when converting it from `PackageURL` to `String`. That would explain why it survived the import:

* First `new PackageURL(cdxComponent.getPurl())` works fine
* Value is persisted to DB using `PackageURL.canonicalize()`, which borks the encoding
* `new PackageURL(<value-from-db>)` now fails

sahibamittal: I checked the BOM before it was ingested into DT by vendor, the component had malformed PURL as it is. It is strange that it did not get filtered out during import bom activity. 

Now if somehow such malformed PURL entered DT, it getting reselected (because empty results are not being persisted in this case) in every resolution cycle needs to be fixed..

nscuro: Yeah. I had a look at this yesterday evening to figure out how #6771 could be addressed. I think both problems require what you had proposed, i.e. some state tracking. There are some nuances there though, mostly around preventing table bloat through excessive `UPDATE`s as much as possible (which can be done with [HOT](https://www.postgresql.org/docs/current/storage-hot.html)).

I've been drafting an ADR and accompanying implementation to address both issues at once, which I hopefully can open a PR for in a few hours. It would be great if you could have a look as to whether it resembles what you had in mind, and whether I missed something.

Issue #6740, opened 2026-07-17
Add heartbeats to policy evaluation with a configurable max timeout to cater to long running policies


mehab:
### Current Behavior

Policy evaluation activities do not heartbeat while processing projects. When an evaluation takes longer than the Dex activity lock timeout (currently five minutes), the activity lock expires even though the original worker is still running.
Another worker or pod can then reclaim the same activity task and begin evaluating the same project concurrently.

## Observed Behaviour
In a multi-pod deployment:
1. Pod A claims a policy evaluation activity.
2. Policy evaluation runs longer than the activity lock timeout.
3. Because no heartbeat is emitted, the lock expires.
4. Pod B reclaims the same activity task.
5. Pod A continues evaluating until it attempts to complete the task and discovers that it no longer owns the lock.
6. This can repeat across workers.
The task attempt remains `1` because this is a lock reclaim, not a normal retry. The lock version continues increasing as workers repeatedly reclaim the task.

## Impact
- Duplicate policy evaluation for the same project
- Increased CPU and database load
- Multiple pods repeatedly processing the same activity
- Long-running workflows may make little or no progress
- Policy violations can be evaluated concurrently
- The issue becomes more likely for projects with many components, vulnerabilities, or expensive CEL policies

## Root cause
`CelPolicyEngine` and the vulnerability policy evaluator perform potentially long-running loops without calling `ActivityContext.maybeHeartbeat()`.
Dex activity locks are leases. If an activity does not heartbeat before its lease expires, Dex assumes the worker is no longer processing the task and allows another worker to reclaim it.
Passing `ActivityContext` directly into policy engines would couple the policy layer to Dex and would also be unsafe as constructor state because an activity context belongs to one specific activity execution.

### Proposed Behavior

Introduce a `Runnable` heartbeat callback into policy evaluation APIs.
Activities pass:
```
PolicyEvaluationDeadline.wrapping(
        activityContext::maybeHeartbeat,
        maxEvaluationDuration);
```
Policy evaluators invoke the callback periodically while iterating over components or vulnerabilities.

This keeps policy engines independent of Dex while allowing the activity to renew its lock.

The callback performs two operations:

Verify that the configured maximum policy evaluation duration has not been exceeded.
Invoke ActivityContext.maybeHeartbeat(), which renews the Dex lock when it is close to expiring.
Maximum evaluation duration
Add a configurable upper bound:

dt.policy.evaluation.max-duration-ms=600000
Default: 10 minutes.

[...]

PR #6741, opened 2026-07-17
dex: enforce execution timeout for activities

mehab:
### Description

Adds an optional per-attempt execution timeout to dex activities, and enables it for policy evaluation. 

Activities may now be registered with an `executionTimeout` in addition to their lock timeout. `ActivityTaskWorker` enforces it: when a single attempt exceeds the timeout, the execution is cancelled (thread interrupt) and the task fails, to be retried according to its retry policy, or terminally once attempts are exhausted.

The timeout is opt-in. Activities registered without one behave exactly as before. It is enabled for `eval-project-policies` with a default of 10 minutes, configurable via `dt.dex.engine.activity.eval-project-policies.execution-timeout-ms`, so that an expensive or pathological policy can no longer occupy a worker indefinitely.

### Addressed Issue
Relates to: https://github.com/DependencyTrack/dependency-track/issues/6740

### Additional Details
Test case provided for replicating the scenario where no heartbeat and lock time out results in another pod picking up the same task.
https://github.com/mehab/dependency-track/blob/feature/policy-eval-heartbeat-max-duration/apiserver/src/test/java/org/dependencytrack/policy/PolicyEvaluationLockThrashTest.java#L81

### Checklist

[ADR]: ../docs/adr/


nscuro (changes requested):
Thanks for the PR!

The heartbeating itself looks good, but I think we can make the deadline portion a bit cleaner, such that it can work for *all* activities.

In `ActivityTaskWorker`, task execution is offloaded to a (virtual) thread:

https://github.com/DependencyTrack/dependency-track/blob/d3429f4b8a9eb48fee27c32f73c5a5612627614e/dex/engine/src/main/java/org/dependencytrack/dex/engine/ActivityTaskWorker.java#L114-L122

The `Future` returned by the above `submit` is then "awaited" via `Future#get`, and the task's outcome is recorded via `engine.onTaskEvent`:

https://github.com/DependencyTrack/dependency-track/blob/d3429f4b8a9eb48fee27c32f73c5a5612627614e/dex/engine/src/main/java/org/dependencytrack/dex/engine/ActivityTaskWorker.java#L124-L167

`Future#get` accepts a timeout via [`Future#get(long, TimeUnit)`](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/concurrent/Future.html#get(long,java.util.concurrent.TimeUnit)). This is where the deadline should go. The invocation will throw a `TimeoutException` if the deadline is exceeded. When this happens, we [cancel](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/concurrent/Future.html#cancel(boolean)) the `Future`, which will interrupt its thread.

Something like this:

```java
final Duration executionTimeout = activityMetadata.executionTimeout();
try {
    final Object activityResult = executionTimeout != null
            ? future.get(executionTimeout.toMillis(), TimeUnit.MILLISECONDS)
            : future.get();
    // ...
} catch (TimeoutException e) {
    future.cancel(/* interruptIfRunning */ true);
    // Report failure, computeRetryAt, etc.
} catch (ExecutionException e) {
    // ...
} catch (InterruptedException e) {
    // ...
}
```

The benefit is that interruption will also work when the activity is blocked on I/O (e.g. DB queries) and activity doesn't have to actively check "did I exceed my deadline?"

apiserver/src/main/java/org/dependencytrack/dex/DexEngineInitializer.java
```diff
@@ -228,6 +228,9 @@ public void contextInitialized(ServletContextEvent event) {
                 voidConverter(),
                 Duration.ofMinutes(1));
 
+        final Duration policyEvaluationMaxDuration = Duration.ofMillis(
+                config.getValue(ConfigKeys.POLICY_EVALUATION_MAX_DURATION_MS, long.class));
```
nscuro: Since timeouts are now a dex engine feature, and we will want to enforce them for other activities as well, let's use a property key structure that supports this. For example:

```
dt.dex.engine.activity.<name>.execution-timeout-ms
```

apiserver/src/main/java/org/dependencytrack/policy/vulnerability/VulnerabilityPolicyEvaluator.java
```diff
@@ -31,9 +31,14 @@ public interface VulnerabilityPolicyEvaluator {
      *
      * @param projectId            ID of the project.
      * @param vulnIdsByComponentId A mapping of component IDs to their associated vulnerability IDs.
+     * @param heartbeat            Callback invoked once per component; used to renew Dex activity locks.
+     *                             Callers outside Dex (e.g. unit tests) may pass a no-op {@code () -> {}}.
```
nscuro: This javadoc leaks implementation details (when to invoke, existence of dex). Can you rephrase to something more generic?

The `Callers outside Dex (e.g. unit tests) may pass a no-op {@code () -> {}}.` section could be replaced with a `default` method.

apiserver/src/main/resources/application.properties
```diff
@@ -933,6 +933,16 @@ dt.vuln-policy-bundle.auth.username=
...
 dt.vuln-policy-bundle.auth.bearer-token=
 
+# Maximum wall-clock duration for a single policy evaluation pass (component policies
+# or vulnerability policies). While evaluation is in progress, the activity lock is
+# renewed via heartbeats. After this limit is reached, evaluation stops, an error is
+# logged, and the rest of the workflow continues without applying incomplete results.
+#
+# @category: General
+# @type:     number
+# @unit:     milliseconds
```
nscuro: These two appear to be an LLM hallucination. Use `integer` for the type. `@unit` is not used anywhere else and is not recognized when building docs from this file.

apiserver/src/test/java/org/dependencytrack/dex/DexEngineInitializerTest.java
```diff
@@ -82,15 +86,24 @@ void afterEach() {
...
 
+    @Test
+    void shouldLoadPolicyEvaluationMaxDurationFromApplicationProperties() {
+        final var config = new SmallRyeConfigBuilder()
+                .addDefaultSources()
+                .build();
+
+        assertThat(config.getValue(ConfigKeys.POLICY_EVALUATION_MAX_DURATION_MS, long.class))
+                .isEqualTo(600_000L);
+    }
```
nscuro: This test basically asserts that SmallRye Config works and adds no value. Let's drop it.

apiserver/src/test/java/org/dependencytrack/policy/PolicyEvaluationLockThrashTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+/**
+ * Multi-worker thrash coverage for the production {@link EvalProjectPoliciesActivity} path.
+ * <p>
+ * Dex lock timeout is shortened at registration time. Per-component cost and whether the
+ * heartbeat {@link Runnable} from the activity is honored are controlled via a
+ * {@link CelPolicyEngine} test double — the real activity always passes
+ * {@code ctx::maybeHeartbeat} and has no switch to disable heartbeats.
+ */
+class PolicyEvaluationLockThrashTest extends PersistenceCapableTest {
```
nscuro: This test seems to duplicate what `DexEngineImplTest.shouldReclaimLongRunningActivityWithoutHeartbeat...` already proves. The `spendTime` pattern and heartbeat debounce math will make it flaky in CI. I think the tests that nail down the reclaim-when-lock-expired behavior belong in the engine tests (where you already added them).

apiserver/src/main/java/org/dependencytrack/policy/cel/CelPolicyEngine.java
```diff
@@ -263,6 +266,7 @@ public void evaluateProject(UUID uuid) {
                 protoVulns = List.of();
             }
 
+            heartbeat.run();
```
nscuro: Consider adding additional invocations before other potentially heavy work. For example, after all components have been evaluated, *before* reconciling violations with the DB.

apiserver/src/main/java/org/dependencytrack/vulnanalysis/ReconcileVulnAnalysisResultsActivity.java
```diff
@@ -195,7 +195,8 @@ public ReconcileVulnAnalysisResultsActivity(
                     projectUuid,
                     reportedFindings,
                     vulnDbIdByVulnKey,
-                    failedAnalyzers);
+                    failedAnalyzers,
+                    ctx);
```
nscuro: Consider only passing the `ctx::maybeHeartbeat` `Runnable` along, since the rest of the context is not needed for downstream callees.

apiserver/src/test/java/org/dependencytrack/policy/EvalProjectPoliciesActivityTest.java
```diff
@@ -0,0 +1,67 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class EvalProjectPoliciesActivityTest {
+
+    @Test
+    void passesActivityHeartbeatToPolicyEngine() {
```
nscuro: I think we can avoid a test with dex machinery by adding a test in `CelPolicyEngineTest` instead. In the end what matters is that `CelPolicyEngine` invokes the heartbeat for the expected N times during its evaluation.

nscuro (changes requested):

nscuro:
@mehab In light of #6795 I've been thinking about solving the heartbeating differently. As it is right now, heartbeating needs to be invoked manually so `ctx.maybeHeartbeat()` "bleeds" everywhere, and everyone needs to remember calling it.

I've been thinking to do heartbeats automatically in the background instead, so the manual invocations would disappear entirely. The work in this PR around execution timeouts already solves one part of your problem (activities running too long), so I think it's worth shipping that in isolation, and handle the heartbeating part separately.

So my suggestion would be:

1. Address the PR comments around execution timeouts (e.g. the property naming)
2. Revert the heartbeat changes
3. Get this merged

The heartbeating would then be an immediate follow-up.

WDYT?

nscuro:
@mehab I added a draft ADR for the heartbeating solution here: https://github.com/DependencyTrack/dependency-track/pull/6807

mehab:
Sounds good @nscuro i will check it on tuesday and complete the changes as you suggest with automated heartbeats. The solution is definitely cleaner than heartbeat tracking by all activities manually.

nscuro (approved):
Thanks!

FYI I updated title and description of the PR to reflect the final state, not what was originally proposed.

mehab:
Thanks niklas 😊

Issue #6194, opened 2023-08-11
Modify frontend to compress BOMs prior to upload, and ensure API server can handle compressed BOMs


nscuro:
Uploading large BOMs can result in extensive network latency, potentially causing clients to time out before a response is received.

Support BOM compression both in frontend and API server.

Nihlus:
I'd be keen on seeing support for this as well. I have some rather large SBOMs (>500MB) that are bogging down both my SBOM generator and DT when uploading them uncompressed.

nscuro:
> rather large SBOMs (>500MB)

Ouch!

gzip compression (with accompanying `Content-Encoding: gzip` request header) should work. Can you give that a try?

Nihlus:
It does! I have the SBOMs in a pre-compressed format, though, and it'd be nice if I could serve them directly to DT as a multipart form parameter. I've been playing around with it locally and should have a PR ready momentarily that adds support for that.

nscuro:
Considering this done with #6769. It's now possible to upload pre-compressed BOMs even via frontend (just select such a file instead of a `bom.json` or whatever).

Looked into having the frontend do transparent compression, but that seems to require buffering the full (compressed) content in browser memory, which for large BOMs that don't compress well will be a huge annoyance.

PR #6769, opened 2026-07-20
Implement support for compressed SBOMs in multipart uploads

Nihlus:
### Description
This PR adds support for pre-compressed SBOMs in the bodies of multipart uploads, enabling faster transfer of already-compressed SBOMs to DependencyTrack.

### Addressed Issue

Partly addresses #6194 by implementing support for compressed SBOMs in the API server.

### Additional Details
Brotli and Zstandard compression have also been added to the overall transparent compression support for the API server for the sake of greater compatibility and API symmetry. gzip, brotli, and zstd together account for >99.5% of all compressed web traffic as of 2026, which should ensure that DT is as compatible as possible with other systems.

There is one caveat - Brotli compression is not tested in the integration test suite due to the lack of an encoding stream in org.brotli. A JNI-based implementation exists, but I was unable to make it work properly during testing and it would add a test-only dependency which may be undesirable. I have tested Brotli-encoded uploads outside of the test suite and they work fine.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
+++ b/apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
@@ -21,6 +21,7 @@ package org.dependencytrack.resources.v1;
 import alpine.model.ConfigProperty;
 import alpine.server.auth.PermissionRequired;
 import com.fasterxml.uuid.Generators;
+import com.github.luben.zstd.ZstdInputStream;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.Content;
@@ -108,6 +109,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import java.util.zip.GZIPInputStream;
 
 import static java.util.function.Predicate.not;
 import static org.dependencytrack.common.MdcKeys.MDC_BOM_UPLOAD_TOKEN;
@@ -529,7 +531,10 @@ public class BomResource extends AbstractApiResource {
                       or <strong>PROJECT_CREATION_UPLOAD</strong> permission.
                     </p>
                     <p>
-                      MediaType supported for BOM artifact is 'application/xml' or 'application/json'.
+                      The BOM artifact may be supplied uncompressed or compressed. If the BOM is uncompressed,
+                      the supported MediaType is 'application/xml' or 'application/json'. If the BOM is compressed,
+                      the supported MediaType is 'application/gzip' or 'application/zstd' and must match the actual 
+                      compression of the data.
                       The BOM will be validated against the CycloneDX schema. If schema validation fails,
                       a response with problem details in RFC 9457 format will be returned. In this case,
                       the response's content type will be <code>application/problem+json</code>.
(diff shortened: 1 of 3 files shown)


apiserver/src/main/java/org/dependencytrack/Application.java
```diff
@@ -266,10 +268,20 @@ public static void main(final String[] args) {
...
         final var gzipCompression = new GzipCompression();
         gzipCompression.setMinCompressSize(1024);
+
+        final var zstdCompression = new ZstandardCompression();
+        zstdCompression.setMinCompressSize(1024);
+
         final var compressionHandler = new CompressionHandler();
+        compressionHandler.putCompression(brotliCompression);
         compressionHandler.putCompression(gzipCompression);
+        compressionHandler.putCompression(zstdCompression);
```
nscuro: These changes are not necessary for your use case. These are about accepting entire compressed requests (`Content-Encoding: gzip`) or transparently compressing entire responses if the client specified `Accept-Encoding: gzip`.

Both zstd and brotli necessitate JNI, so I'd like to defer support for them here until there's an explicit demand for it.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -704,9 +710,22 @@ public Response uploadBom(
...
+             final var encodingStream = (switch (encoding) {
+                 case "application/brotli", "application/x-brotli" -> new BrotliInputStream(inputStream);
+                 case "application/gzip", "application/x-gzip" -> new GZIPInputStream(inputStream);
+                 case "application/zstd", "application/x-zstd" -> new ZstdInputStream(inputStream);
+                 case "*/*", "application/json", "application/xml" -> inputStream;
+                 case null -> inputStream;
+                 default -> throw new WebApplicationException(Response
+                         .status(Response.Status.NOT_ACCEPTABLE)
+                         .entity("BOM has an unsupported Content-Type.")
+                         .build());
```
nscuro: Rejecting other content types is dangerous because the header has historically not been checked at all. So at the moment clients can send all sorts of values (e.g. `application/octet-stream`, or [CycloneDX's own media types](https://cyclonedx.org/specification/overview/)), and we'd break them with this change.

We should check for known-supported compression content types explicitly, and treat everything else as uncompressed like before.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -704,9 +710,22 @@ public Response uploadBom(
...
         final FormDataBodyPart firstPart = artifactParts.getFirst();
+        final var encoding = firstPart.getHeaders().getFirst("Content-Type");
+
         final byte[] bomBytes;
         try (final var inputStream = ((BodyPartEntity) firstPart.getEntity()).getInputStream();
-             final var byteOrderMarkInputStream = BOMInputStream.builder().setInputStream(inputStream).get()) {
+             final var encodingStream = (switch (encoding) {
+                 case "application/brotli", "application/x-brotli" -> new BrotliInputStream(inputStream);
+                 case "application/gzip", "application/x-gzip" -> new GZIPInputStream(inputStream);
+                 case "application/zstd", "application/x-zstd" -> new ZstdInputStream(inputStream);
```
nscuro: Do you have an immediate need for brotli and zstd on top of gzip? I'm particularly torn about brotli because that library [hasn't seen a new version since 2017](https://mvnrepository.com/artifact/org.brotli/dec). It's currently used for the Nix integration but I'd like to avoid its usage spreading if we can avoid it.

nscuro (changes requested):
Thanks! A few questions and suggestions. Also, please sign-off your commits so the DCO check passes.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
Nihlus: I'm fine skipping Brotli, personally. zstd is becoming more and more common these days, especially with the improved compression ratios it brings to the table so I would like to keep support for that.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
Nihlus: Sounds good to me.

apiserver/src/main/java/org/dependencytrack/Application.java
Nihlus: Fair enough! I mainly added it for the sake of feature symmetry.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
nscuro: Tend to agree on zstd. We already use it to compress files before storing them on disk or in S3 so it's established in the code base.

Nihlus:
All comments should be resolved now.

nscuro (approved):

Issue #5907, opened 2026-03-13
Display Fixed version if available as a separate field on its own


prabhushan:
### Current Behavior

vulnerabilities are shown, but the fixed versions are usually hidden in the description of the fixes. Calling out could help the team or automate the remediation in the future

### Proposed Behavior

Is there a way to display Fixed versions for vulnerabilities atleast for the ones for which fixes are available.. 

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

wadimj:
It would be great to have this, since the Trivy analyzer already provides the “Fixed Version” information, but it is currently being silently dropped in Dependency-Track.
In my other workflows, I rely on that field to produce compact reports for developers, including a list of affected packages and the minimum required version to address all findings.

webdevred:
Hello @nscuro 

I'd like to implement this. Here is what I found on `main` and what I would change, split by component.

**What already works.** The GHSA mirror (`vuln-data-source/github` module, `ModelConverter`) parses the advisory's affected version range, and the apiserver exposes it on `GET /v1/vulnerability/source/{source}/vuln/{vulnId}` as `affectedComponents`, each with its version bounds. So the affected range is already in the API response.

**The gap.** The frontend never renders it. In the `frontend` repo, `src/views/portfolio/vulnerabilities/Vulnerability.vue` fetches `affectedComponents` but does not display it. The "Affected Projects" tab only lists the user's own impacted projects, not the advisory's version ranges. So on a vulnerability page you cannot see which versions are affected or fixed.

I would like to make two separate changes:

**1. Display the affected ranges (frontend repo, `Vulnerability.vue`).**
Render `affectedComponents` on the vulnerability page: the affected version range per package, with the `versionEndExcluding` bound shown as the fixed version. This alone resolves the original request and needs no backend change.

**2. Capture GHSA `firstPatchedVersion` (dependency-track repo).**
[...]

PR #6772, opened 2026-07-21
Add affected version ranges to the Finding Packaging Format export

webdevred:
### Description

Fixed versions are only visible inside a vulnerability's free text description today, and the Finding Packaging Format (FPF) export drops the structured version ranges entirely. Tools that consume FPF, in our case DefectDojo, therefore cannot tell which release resolves a finding even though Dependency-Track already holds the affected ranges. Auditors end up opening each advisory by hand, and automated remediation has nothing to key off.

This adds an `affectedVersions` array to each finding's `vulnerability` object in the FPF export. Each entry carries a component's identity (PURL or CPE) and its version bounds, in the same shape already exposed as `affectedComponents` on the vulnerability API. A consumer reads the upper bound (`versionEndExcluding`) as the first patched version and derives the release to upgrade to. `FPF_VERSION` is bumped from 1.5 to 1.6.

The ranges are attached in `FindingPackagingFormat`, not in `FindingDao`, so the shared findings query used by the REST API, the CycloneDX export and the Kenna integration stays untouched. Vulnerable software for all findings in the export is fetched in a single batched query and grouped per vulnerability.

### Addressed Issue

Related to #5907. Companion to the DefectDojo issue, DefectDojo/django-DefectDojo#15239, which suggests reading these ranges and set the finding mitigation to upgrade the version with a fix.

### Additional Details

The original request asked for a single "fixed version" field. Following maintainer feedback on #5907, this exposes the full ranges instead. A vulnerability can affect several release trains of a component and several components at once, so there is no single fixed version to emit without it being misleading. Handing consumers the complete ranges lets them derive the fix themselves and sidesteps that ambiguity. Reusing the `AffectedComponent` shape keeps the payload consistent with the vulnerability API, so a client that already parses `affectedComponents` needs no new shape.

Because the change never touches the `FindingDao` SQL, and `FindingDao` maps results by column name via JDBI rather than by positional index, existing FPF consumers see no difference beyond the new `affectedVersions` key.

### Checklist


webdevred:
Caught this up with `main` and resolved the conflicts, so it's mergeable again.

@nscuro would love a look whenever you have a moment.

Issue #6771, opened 2026-07-21
apiserver: completed `resolve-package-metadata` task seems to get rescheduled indefinitely


migus-rs:
### Current Behavior

Checking logs I noticed a specific warning message occurring repeatedly (every ~50-90s).
```
A successfully completed activity task with ID ActivityTaskId[queueName=default, workflowRunId=019edf6f-b706-766e-8743-90e97c654731, createdEventId=0] could not be deleted, likely because its lock expired, causing another worker to pick it up. Expect duplicate task executions. To prevent this from happening, consider increasing the activity's lock timeout, and sending heartbeats more frequently.
```
This seems to have been ongoing for a month now, as can also be seen in the UI when checking the UUID:

<img width="823" height="523" alt="Image" src="https://github.com/user-attachments/assets/5627c015-dae8-4bf5-88b5-73259fbb5ea3" />

I was unable to find any notes in documentation ([here](https://dependencytrack.github.io/docs/next/reference/configuration/dex-engine/) and [here](https://dependencytrack.github.io/docs/next/reference/configuration/task-scheduler/#scheduled-tasks)) about how to tweak the lock timeout as the message states (or if it is even possible).

There also seems no way to manually cancel this (seemingly stuck) operation.

Seeing #6740 reminded me of this behavior. That issue only mentions policy evaluation, though.
Feel free to mark this a duplicate if this refers to the same problem.

### Steps to Reproduce

1. Run DTv5 with some projects.
2. Observe some tasks getting rescheduled repeatedly.

### Expected Behavior

All workflow tasks should be able to progress or fail within a reasonable time frame.

### Dependency-Track Version

5.x

### Browser

N/A

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

nscuro:
It looks like the candidate query (i.e. "which components need metadata resolution?") takes too long in your case. Working on a fix that should also cover #6654. #6740 may sound related but fundamentally the candidate query here should never take even remotely as long as the default lock timeout.

migus-rs:
Interestingly enough, this behavior seems to persist in 5.0.4, but more subtle.

1. There is no more log messages like the one described before anymore.
2. There is consistent spikes in database load in a fixed interval.
[...]

PR #6794, opened 2026-07-23
Introduce package metadata resolution state tracking

### LINKED ISSUE #6654: Missing negative-cache for Malformed PURLs

nscuro:
### Description

Introduces package metadata resolution state tracking:

* Adds a new `PACKAGE_METADATA_RESOLUTION` table that tracks when a given PURL was last resolved, and with what outcome.
* Drives the resolution candidate query from the new table instead of the full `COMPONENT` table, significantly reducing query cost. This also allows permanently skipping resolution for PURLs that are unresolvable, e.g. because they're malformed.
* Uses a cursor within a single `resolve-package-metadata` workflow run to avoid having to re-scan the full `PACKAGE_METADATA_RESOLUTION` table from the beginning for every batch.
* Enables HOT updates on the `PACKAGE_METADATA` and `PACKAGE_ARTIFACT_METADATA` tables by dropping the now-unused indexes on `RESOLVED_AT` columns.

### Addressed Issue

Fixes #6654
Closes #6679
Relates to #6771

### Additional Details

See the included ADR for more details, and alternative options that were considered.

### Checklist

- ~This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly~

[ADR]: ../docs/adr/


nscuro:
@sahibamittal FYI. Please have a look when you get the time.

sahibamittal:
@nscuro Thanks a lot for the fix, apologies I missed the notification. This looks good and covers the case I raised. Once I sync this, I'll also test it on my instance. 👍🏼

PR #6804, opened 2026-07-24
Add Checkmarx SCA vulnerability analyser

sahibamittal:
### Description

Add an optional Checkmarx SCA analyzer, enabled only when configured (credentials, feature flag, org/project identifiers as required by Checkmarx APIs), analogous to Snyk.

https://checkmarx.stoplight.io/docs/checkmarx-one-api-reference-guide 
**Batch PURL API**: https://checkmarx.stoplight.io/docs/checkmarx-one-api-reference-guide/drwbfz2dzursa-retrieve-package-risks-data 

### Addressed Issue

Closes https://github.com/DependencyTrack/dependency-track/issues/6203

### Additional Details

Frontend changes not required, part of existing extension system.

Screenshot 2026-08-05 at 11 37 46

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxApiClient.java
+++ b/vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxApiClient.java
@@ -72,12 +72,15 @@ final class CheckmarxApiClient {
      * @param purls the package URLs to analyze
      * @return vulnerabilities response from Checkmarx API
      */
-    CheckmarxVulnerabilitiesResponse fetchVulnerabilities(Collection<String> purls) throws IOException, InterruptedException {
+    CheckmarxApiResponse fetchVulnerabilities(Collection<String> purls) throws IOException, InterruptedException {
         if (purls.isEmpty()) {
-            return new CheckmarxVulnerabilitiesResponse(List.of());
+            return new CheckmarxApiResponse(List.of());
         }
 
         LOGGER.debug("Fetching Checkmarx vulnerabilities for {} PURLs", purls.size());
+
+        final URI requestUrl = apiBaseUrl.resolve("/api/sca/packages/vulnerabilities"
+                                + "?IncludeRiskDetails=true" + "&IncludeVersionDetails=true" + "&IncludeVersionRemediation=true");
 
         // Ensure valid access token (will be cached if still valid)
         final String accessToken = tokenManager.getAccessToken(authApiBaseUrl, orgId, refreshToken);
@@ -85,7 +88,7 @@ final class CheckmarxApiClient {
         final String requestBody = objectMapper.writeValueAsString(new VulnerabilityRequest(new ArrayList<>(purls)));
 
         final var request = HttpRequest.newBuilder()
-                .uri(apiBaseUrl.resolve("/api/sca/packages/vulnerabilities"))
+                .uri(requestUrl)
                 .header("Authorization", "Bearer " + accessToken)
                 .header("Content-Type", "application/json")
                 .header("Accept", "application/json")
@@ -97,14 +100,14 @@ final class CheckmarxApiClient {
 
         try (final InputStream body = response.body()) {
             if (response.statusCode() == 200) {
-                return objectMapper.readValue(body, CheckmarxVulnerabilitiesResponse.class);
+                return objectMapper.readValue(body, CheckmarxApiResponse.class);
             }
 
             body.transferTo(OutputStream.nullOutputStream());
 
             if (response.statusCode() == 404) {
                 LOGGER.debug("No vulnerabilities found for provided PURLs");
-                return new CheckmarxVulnerabilitiesResponse(List.of());
+                return new CheckmarxApiResponse(List.of());
             }
 
             throw new IOException("Checkmarx API request failed with status " + response.statusCode());
--- a/vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxApiClient.java
+++ b/vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxApiClient.java
@@ -1,0 +1,118 @@
+/*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.vulnanalysis.checkmarx;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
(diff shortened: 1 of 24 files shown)


vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxModelConverter.java
```diff
@@ -0,0 +1,175 @@
...
+        if (vulnDetails.references() != null) {
+            for (final var ref : vulnDetails.references()) {
+                vulnBuilder.addAdvisories(Advisory.newBuilder().setUrl(ref.url()));
+            }
+        }
+
+        if (remediation != null) {
+            final var recommendations = new ArrayList<String>();
+            recommendations.add("Smallest package upgrade that resolves the identified risks in the current package version: " + remediation.nearest().version());
+            recommendations.add("Latest version of the package: " + remediation.latest().version());
```
nscuro, line 98: Is it possible that `nearest()` or `latest()` are `null`, or are both always set when `remediation` is present?

vuln-analysis/checkmarx/src/main/resources/org/dependencytrack/vulnanalysis/checkmarx/checkmarx-vuln-analyzer-config-v1.schema.json
```diff
@@ -0,0 +1,52 @@
...
+      "description": "Organization ID to authenticate with the Checkmarx REST API.",
+      "minLength": 1
+    },
+    "refreshToken": {
+      "type": "string",
+      "title": "Refresh Token",
+      "description": "Refresh token (API Key) to authenticate with the Checkmarx REST API.",
+      "minLength": 1,
+      "x-secret-ref": true
+    }
```
nscuro: Based on https://checkmarx.stoplight.io/docs/checkmarx-one-api-reference-guide/ybti0d5u0auto-authentication#prerequisites, it seems like API keys are bound to a user, whereas OAuth2 client credentials can be used to authenticate as client (i.e., non-human account). Given the use-case, I think the latter would be preferable here. Is there a specific reason you chose API key auth instead?

vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxApiClient.java
```diff
@@ -0,0 +1,121 @@
...
+
+            body.transferTo(OutputStream.nullOutputStream());
+
+            if (response.statusCode() == 404) {
+                LOGGER.debug("No vulnerabilities found for provided PURLs");
+                return new CheckmarxApiResponse(List.of());
+            }
+
+            throw new IOException("Checkmarx API request failed with status " + response.statusCode());
+        }
```
nscuro, line 123: Use [`RetryableVulnAnalysisException`](https://github.com/DependencyTrack/dependency-track/blob/main/vuln-analysis/api/src/main/java/org/dependencytrack/vulnanalysis/api/RetryableVulnAnalysisException.java) to ensure retriable exceptions are classified correctly and lead to retries. Example:

https://github.com/DependencyTrack/dependency-track/blob/29e2f50b816591fe1b923e59826d4a6ec383e573/vuln-analysis/oss-index/src/main/java/org/dependencytrack/vulnanalysis/ossindex/OssIndexVulnAnalyzer.java#L262-L277

Add this to all HTTP requests you're making, including the one that fetches access tokens.

vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxVulnAnalyzer.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        final var vulnerabilitiesByPurl = new HashMap<String, List<CheckmarxDataObject>>();
+        final var entriesToCache = new HashMap<String, byte @Nullable []>(purlBatch.size());
+
+        if (!response.packageRisks().isEmpty()) {
+            for (final CheckmarxDataObject cxDataObject : response.packageRisks()) {
+                final var status = cxDataObject.pkg().status();
+                if (!status.equalsIgnoreCase("Ok")) {
+                    LOGGER.warn("Received non-OK status '{}' for package '{}:{}'; Skipping", status, cxDataObject.pkg().name(), cxDataObject.pkg().version());
```
nscuro, line 164: It seems CX tells us when a package can't ever be successfully analyzed (e.g. invalid PURL, but I think there are more statuses like that). Would it make sense to cache `null` for them too? Otherwise we'll be re-hitting the CX API at least daily with PURLs that'll always fail.

nscuro (changes requested):
Retryability classification is needed, everything else is just suggestions.

vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxModelConverter.java
sahibamittal: As per their doc, `remediation` itself can be `null` but `latest` and `nearest` are non-null objects with `required` `version` field.

vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxVulnAnalyzer.java
sahibamittal: Status can be one of `[Ok, NotFound, UnsupportedPackageType, MissingVersion, InvalidPurl]`. Yes it would be good to cache null for such cases.

sahibamittal: Okay below POC already caches `null` for every submitted PURL that did not produce accepted vulnerability data including skipped cases because status was not `Ok`, because those entries never get added to `vulnerabilitiesByPurl`.

```
for (final String purl : purlBatch) {
    if (!vulnerabilitiesByPurl.containsKey(purl)) {
        entriesToCache.put(purl, null);
    }
}
```

vuln-analysis/checkmarx/src/main/resources/org/dependencytrack/vulnanalysis/checkmarx/checkmarx-vuln-analyzer-config-v1.schema.json
sahibamittal: CHX provided us accounts with permissions to only create API keys. I'll align with them for OAuth2 client instead, for more ideal integration.

vuln-analysis/checkmarx/src/main/java/org/dependencytrack/vulnanalysis/checkmarx/CheckmarxVulnAnalyzer.java
nscuro: Ah, good catch! Nevermind then.

vuln-analysis/checkmarx/src/main/resources/org/dependencytrack/vulnanalysis/checkmarx/checkmarx-vuln-analyzer-config-v1.schema.json
sahibamittal: I confirmed this internally, OAuth2 client credentials are not currently available. So the proposed authentication mechanism is API key-based authentication only. 

If support for OAuth2 client credentials becomes available in the future, I'd be happy to revisit the authentication mechanism and migrate to that approach.

Currently API Key will be FID bound at org level.

nscuro: OK, makes sense. Should we rename the parameter to `apiKey` then so it's more consistent with CX's own wording? Judging by the docs, their API keys just happen to be implemented as OAuth refresh tokens, but the user-facing name is still "API key".

nscuro (approved):

PR #6824, opened 2026-07-27
Exclude project from components loaded during ImportBomActivity

### LINKED ISSUE #6810: ImportBomActivity::processComponents can cause memory runaway with large BOMs
### Current Behavior

Currently, ImportBomActivity::processComponents loads all components of a project into memory in one go for processing. While this is fine for small-to-mid-size BOMs, larger BOMs can use a significant amount of memory far exceeding the size of the raw BOM itself. In extreme cases, the application runs out of memory when trying to import BOMs and worker threads crash causing a full-application degradation that can't be gracefully recovered from.

Large BOMs are common when including the full inventory of source files in build-time BOMs, for example, and can have a component count in excess of 300 000 unique components. In our usage of DependencyTrack, there are about 30-40 projects like this with the component count likely to grow once we start including symbol information in the BOMs. Our instance has 64G of RAM allocated to it, of which DependencyTrack is allowed to use 60G with a soft cap of 48G; however, the application still routinely runs out of memory and crashes.

We have tried to tune down the artifact import concurrency, but even with `dt.dex-engine.activity-worker.artifact-import.max-concurrency` set to `1` we're not moving past the worst offenders in the task backlog.

There are likely other parts of the application with assumptions around the sizes of result sets, but this activity is the one that's causing most issues at the moment.

### Steps to Reproduce

1. Upload a BOM with a very large component count to an instance with insufficient working memory
2. Observe OutOfMemory errors

### Expected Behavior

A common solution to large result sets like this is to employ a streaming approach, where parts of the result set are retrieved, processed, and deallocated before the next batch. A cursory glance at `ImportBomActivity::processComponents` suggests that it should be possible to work iteratively over the components rather than loading all of the into memory in one go.

### Dependency-Track Version

5.x

### Browser

N/A

### Checklist

Nihlus:
### Description
This PR alters ImportBomActivity to exclude the project from components loaded during an import.

### Addressed Issue
Closes #6810.

### Additional Details
The project a component belongs to contains a full JSON string of all direct dependencies by default. For projects with large component counts, this results in a massive memory runaway as the entire project object is loaded for every component. Potential load sizes of 10TB have been observed in production (50MB per directDependencies field,
\>200k components), which far exceeds the resources of most computers on the planet.

By excluding the project from the default load set during BOM imports, we drastically reduce memory usage without affecting other parts of the application. Ideally, we'd remove the project or the direct dependency field from the default load sets of `Project` and `Component`, but there are many parts of the code that depend on that behavior today.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/tasks/ImportBomActivity.java
+++ b/apiserver/src/main/java/org/dependencytrack/tasks/ImportBomActivity.java
@@ -55,6 +55,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
+import javax.jdo.FetchGroup;
 import javax.jdo.PersistenceManager;
 import javax.jdo.Query;
 import java.io.ByteArrayInputStream;
@@ -862,6 +863,25 @@ public final class ImportBomActivity implements Activity<ImportBomArg, Void> {
 
     private static List<Component> getAllComponents(final QueryManager qm, final Project project) {
         final Query<Component> query = qm.getPersistenceManager().newQuery(Component.class);
+
+        // Every component contains a reference to its parent project, which also contains the set of direct references
+        // it has. For large BOM uploads, this results in massively duplicated data being fetched that we don't
+        // actually need. Unfortunately, a lot of other code relies on the project being part of the default fetch
+        // group, so it can't just be removed everywhere just yet. Instead, we'll selectively remove it here.
+        final var trimmedFetchGroup = qm.getPersistenceManager().getFetchGroup(Component.class, "DEFAULT_TRIMMED");
+        if (!trimmedFetchGroup.isUnmodifiable()) {
+            final var defaultFetchGroup = qm.getPersistenceManager().getFetchGroup(Component.class, FetchGroup.DEFAULT);
+            for (final var member : defaultFetchGroup.getMembers()) {
+                trimmedFetchGroup.addMembers(member.toString());
+            }
+
+            trimmedFetchGroup.removeMember("project");
+            trimmedFetchGroup.setUnmodifiable();
+        }
+
+        query.getFetchPlan().removeGroup(FetchGroup.DEFAULT);
+        query.getFetchPlan().addGroup("DEFAULT_TRIMMED");
+
         query.getFetchPlan().addGroup(Component.FetchGroup.BOM_UPLOAD_PROCESSING.name());
         query.getFetchPlan().setFetchSize(FETCH_SIZE_GREEDY);
         query.setFilter("project.id == :projectId");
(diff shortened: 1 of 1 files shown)


apiserver/src/main/java/org/dependencytrack/tasks/ImportBomActivity.java
```diff
@@ -862,6 +863,25 @@ private static void resolveAndApplyLicense(
...
+            for (final var member : defaultFetchGroup.getMembers()) {
+                trimmedFetchGroup.addMembers(member.toString());
+            }
+
+            trimmedFetchGroup.removeMember("project");
+            trimmedFetchGroup.setUnmodifiable();
+        }
+
+        query.getFetchPlan().removeGroup(FetchGroup.DEFAULT);
+        query.getFetchPlan().addGroup("DEFAULT_TRIMMED");
```
nscuro, line 883: Should also do this for `getAllServices` to avoid the same issue there:

https://github.com/DependencyTrack/dependency-track/blob/29e2f50b816591fe1b923e59826d4a6ec383e573/apiserver/src/main/java/org/dependencytrack/tasks/ImportBomActivity.java#L877-L881

apiserver/src/main/java/org/dependencytrack/tasks/ImportBomActivity.java
```diff
@@ -862,6 +863,25 @@ private static void resolveAndApplyLicense(
...
+        // group, so it can't just be removed everywhere just yet. Instead, we'll selectively remove it here.
+        final var trimmedFetchGroup = qm.getPersistenceManager().getFetchGroup(Component.class, "DEFAULT_TRIMMED");
+        if (!trimmedFetchGroup.isUnmodifiable()) {
+            final var defaultFetchGroup = qm.getPersistenceManager().getFetchGroup(Component.class, FetchGroup.DEFAULT);
+            for (final var member : defaultFetchGroup.getMembers()) {
+                trimmedFetchGroup.addMembers(member.toString());
+            }
+
+            trimmedFetchGroup.removeMember("project");
+            trimmedFetchGroup.setUnmodifiable();
```
nscuro, line 879: Creating the fetch group dynamically is fine here, just noting that there already exists a `BOM_UPLOAD_PROCESSING` fetch group:

https://github.com/DependencyTrack/dependency-track/blob/29e2f50b816591fe1b923e59826d4a6ec383e573/apiserver/src/main/java/org/dependencytrack/model/Component.java#L85-L88

... which currently only lists fields that are *not* included in the default FG. So an alternative solution could be to explicitly list *all* required fields there, while still removing the default fetch group. But that is more likely to drift over time because it's easy to forget adding fields there.

Nihlus: Yeah, I considered doing that but opted not to for that exact reason. The long-term fix is to get away from JDO and the whole auto-loading of stuff as a full object graph, really.

apiserver/src/main/java/org/dependencytrack/tasks/ImportBomActivity.java
Nihlus: Done!

nscuro (approved):

PR #6835, opened 2026-07-28
Add checksum matching support for Snyk vuln analyzer

### LINKED ISSUE #6204: Enhance Snyk vuln analyzer for checksum-qualified Maven PURLs + surface Snyk match confidence
### Summary
Hyades processes uploaded SBOMs. Some SBOM generators (e.g., vendor-distributed Maven artifacts such as Cloudera) produce Maven PURLs with a checksum qualifier.

Snyk’s packages/issues API can successfully evaluate those checksum-qualified PURLs and returns:

- vulnerability data, and
- authoritative per-submitted-package match metadata under meta.packages[<purl>].match indicating full, partial, or none.

We want to update only the Snyk vulnerability analyzer so it can:

- correctly handle checksum qualifiers in PURLs (for Maven only, where this is required), and
- record/surface Snyk’s match confidence to users without gating (i.e., vulnerabilities returned by Snyk are still shown; they’re just flagged as uncertain where appropriate),
- do so with minimal caching/performance impact and correct cache isolation.

### Current behavior
**Qualifier handling**
- The Snyk analyzer correlates PURLs largely using coordinates-only keys (PackageURL.getCoordinates()), which drops qualifiers like checksum.
- This can prevent correct correlation of the Snyk response to the exact uploaded SBOM identity when Snyk relies on checksum-qualified matching.

**Match metadata visibility**
- Snyk returns meta.packages[*].match (starting api version 2025-11-05, but the current analyzer does not use this to annotate results to users.

**Cache-key considerations**

- Snyk and OSS Index both use the cache namespace name "results" and cache keys are strings derived from PURLs.
- If Snyk changes keying strategy (e.g., to include checksum), we must avoid cache collisions—especially negative-cache (null) behavior that can suppress remote fetches.

### Part 1: Preserve and handle checksum qualifier PURLs (Snyk analyzer only)

**Update hyades-apiserver’s Snyk vuln analyzer (vuln-analysis/snyk) so that for Maven PURLs that include a checksum qualifier:**

- The analyzer must treat the full canonical PURL (including qualifiers) as the correlation/request identity for that subset.
- Correlation must align with Snyk’s response mapping:
  - Snyk response meta.packages entries are keyed by the submitted PURL, so we must correlate using that same identity form.
 - For Maven PURLs without checksum, keep existing coordinates-only behavior to minimize changes.
(Implementation-wise, this means updating the logic in SnykVulnAnalyzer where it collects “analyzable PURLs”, where it keys cache entries, and where it matches Snyk response PURLs back to submitted components.)

### Part 2: Surface meta.packages[*].match as a user-visible flag

**Parse and store the match metadata from meta.packages[submittedPurl].match:**

- full → confidence 100% (applicable)
- partial → show vuln but mark as partial match / verify
- none → show vuln but mark as not matched / verify (since the vulnerability data may still be present; user must decide)

**How to expose the flag in Hyades:**

Can use existing attribution metadata in Hyades:

- FindingAttribution.matchingPercentage (confidence)
- FindingAttribution.alternateIdentifier (can store match type + match.description)

This avoids introducing new DB schema or changing vulnerability visibility logic.

### Part 3: Cache update with minimal performance impact and correct isolation
Because Snyk and OSS Index share the same cache namespace name "results", we must ensure Snyk’s new checksum-qualified keying cannot corrupt/mask OSS Index or vice versa.

**Proposed cache handling:**
- Keep cache-key changes localized to Snyk:
 - For Snyk, cache using the chosen identity key form:
  - coordinates-only for PURLs without checksum
  - full canonical PURL (including checksum) for Maven PURLs with checksum
- Ensure cache isolation to prevent collisions:
 - Option A (preferred): use a Snyk-specific cache namespace name, e.g. cacheManager.getCache("snyk-results")
 - Option B: prefix cache keys with snyk: to prevent cross-analyzer collisions
This also protects against the “negative cache” problem (null cached entries) where one analyzer could otherwise suppress another analyzer’s remote fetch for a colliding key.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/hyades/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/hyades/issues) for whether this enhancement was already requested

mehab:
### Description

Adds opt-in Snyk checksum-qualified PURL matching for Maven components. Match quality from `meta.packages` (full / partial / none) decides whether findings are attached.

When `checksumMatchingEnabled` is true, Maven PURLs with a checksum qualifier are sent to Snyk as full canonical PURLs. Coordinates-only and flag-off behavior is unchanged.

### Addressed Issue

Closes #6204

#### Why

Snyk’s batch packages/issues API (`2025-11-05`) accepts checksum-qualified PURLs (e.g. from `snyk sbom --include-provenance`) and returns match quality in `meta.packages`. Previously Dependency-Track stripped PURL qualifiers, ignored meta, and could not distinguish a trustworthy empty result from a none match.

#### Runtime behaviour

**Config:** `checksumMatchingEnabled` (default `false`) on the Snyk vuln-analyzer extension.

**Coordinates-only** (flag off, Maven PURL without a checksum, or non-Maven even if a checksum qualifier is present):

- Request / cache key = lowercase coordinates
- Findings come from `data[]` only; `meta.packages` is not used for branching
- Exact request-key match only (issue PURLs are not stripped to force a coordinates match)

**Checksum-qualified** (flag on + Maven + checksum qualifier):

- Request / cache key = full canonical lowercase PURL (checksum preserved)
- **full** (name/version and checksum matched) → attach findings; cache
- **partial** (name/version matched, checksum did not) → attach findings; cache
- **none** (neither matched) → skip findings; cache as empty
- Missing usable `meta.packages` **and** non-empty `meta.errors` → skip findings; negative-cache (permanent failure, e.g. unsupported ecosystem)
- Missing usable `meta.packages` **without** `meta.errors` → skip findings; do not cache (transient)
- PURL keys are correlated via `PackageURL.canonicalize().toLowerCase()` so `sha1:` and `sha1%3A` match
- Issue PURLs may be coordinates-only while the request was checksum-qualified; lookup uses the coordinates-only key only, not other checksum-qualified keys that share coordinates

No apiserver / BOM prep changes — checksums must already be present on component PURLs.

Partial matches are reported like full matches. There is no UI or API signal yet that a finding came from a partial match; that needs a per-component (not per-vulnerability) attribution field and frontend work.

#### Cache

Keys are shared by request PURL across projects (scales with unique PURLs, not project count). This cache is only used to skip repeat Snyk HTTP calls. It is not the findings database.

- Findings present: JSON `SnykCachedPurlResult`
- No findings (full-empty, partial-empty, or none): `null` negative entry
- Legacy cache values remain readable (`null`, `SnykIssue[]` JSON, structured `SnykCachedPurlResult` JSON)

### Checklist

[ADR]: ../docs/adr/


vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykPurlUtil.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        return purl.getCoordinates().toLowerCase();
+    }
+
+    /**
+     * Whether this PURL should use checksum matching and {@code meta.packages} semantics.
+     */
+    static boolean requiresChecksumMeta(PackageURL purl, boolean checksumMatchingEnabled) {
+        return checksumMatchingEnabled
+                && MAVEN.equals(purl.getType())
```
nscuro: Is the limitation to Maven a vendor restriction or an active choice by you?

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykCacheCodec.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Compact cache encoding for Snyk analyzer results.
+ * <p>
+ * Empty or untrusted outcomes use single-byte sentinels or {@code null} so portfolio-scale
+ * cache reads avoid JSON deserialization for the common negative-cache case.
+ *
+ * @since 5.1.0
+ */
+final class SnykCacheCodec {
```
nscuro: I don't quite follow why this was needed. The old code already cached `null` (i.e. no serialization involved) for negative hits:

https://github.com/DependencyTrack/dependency-track/blob/5a323d4292f3d93dfa1531b84a36f032e56fb2d8/vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java#L247-L251

And `null` values were simply skipped, there's no deserialization:

https://github.com/DependencyTrack/dependency-track/blob/5a323d4292f3d93dfa1531b84a36f032e56fb2d8/vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java#L118-L126

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
```diff
@@ -224,34 +232,257 @@ private Map<String, List<SnykIssue>> analyzePurlBatch(
...
+
+    private void attachCachedIssuesIfPresent(
+            String requestPurl,
+            SnykCachedPurlResult cached,
+            Map<String, List<SnykIssue>> issuesByPurl) {
+        if (cached.issues() != null && !cached.issues().isEmpty()) {
+            issuesByPurl.put(requestPurl, cached.issues());
+        }
+        // Empty issues is a cached negative result (shared across projects until TTL).
+    }
```
nscuro: Handling for match types `PARTIAL`, `NONE`, and `FULL`-with-empty-issues is behaviorally the same. I think for all those cases we should just cache `null`, which removes the `SnykCacheCodec` overhead, slightly reduces storage cost, and makes the code simpler.

vuln-analysis/snyk/src/test/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzerChecksumMatchingTest.java
```diff
@@ -0,0 +1,451 @@
...
+                        .setPurl(CHECKSUM_PURL)
+                        .build())
+                .build();
+
+        assertThat(analyzer.analyze(bom).getVulnerabilitiesList()).isEmpty();
+        // No usable meta.packages → not cached → second call hits HTTP again.
+        assertThat(analyzer.analyze(bom).getVulnerabilitiesList()).isEmpty();
+
+        verify(2, postRequestedFor(anyUrl()));
+    }
```
nscuro, line 288: Should this not cause a negative cache? If a PURL can't be analyzed for a permanent reason (e.g. due to `Unsupported Ecosystem` as in the test file), it won't be analyzable on the next run.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
```diff
@@ -224,34 +232,257 @@ private Map<String, List<SnykIssue>> analyzePurlBatch(
...
+                }
+            }
+        } else {
+            // Issue PURL may include qualifiers; match on coordinates.
+            for (final var entry : issuesByIssuePurl.entrySet()) {
+                final String coords = coordinatesLower(entry.getKey());
+                if (requestPurl.equals(coords) && bomRefsByPurl.containsKey(requestPurl)) {
+                    issues.addAll(entry.getValue());
+                }
             }
```
nscuro: This could be a source of false positives, and also somewhat defeats the purpose of the checksum matching, since the checksums here are also just PURL qualifiers which would be discarded.

vuln-analysis/snyk/src/main/resources/org/dependencytrack/vulnanalysis/snyk/snyk-vuln-analyzer-config-v1.schema.json
```diff
@@ -18,6 +18,13 @@
       "description": "Whether to include alias information in vulnerability data.",
       "existingJavaType": "boolean"
     },
+    "checksumMatchingEnabled": {
+      "type": "boolean",
+      "title": "Checksum Matching Enabled",
+      "description": "When enabled, send checksum qualifiers from Maven component PURLs to Snyk and interpret meta.packages match results. Only pkg:maven PURLs with a checksum qualifier use checksum matching; all others use coordinates-only matching.",
```
nscuro: There are a few preconditions to make this feature work to begin with, which should be [documented](https://github.com/DependencyTrack/docs/blob/main/docs/reference/analyzers.md#snyk). For example, from what I understand, this practically only works when you generate your SBOMs with Snyk's tooling, as no other generator emits `checksum` PURL qualifiers.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykPurlUtil.java
nscuro:
```suggestion
                && PackageURL.StandardTypes.MAVEN.equals(purl.getType())
```

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
```diff
@@ -224,34 +232,257 @@ private Map<String, List<SnykIssue>> analyzePurlBatch(
...
+    }
+
+    private boolean isChecksumQualifiedRequestPurl(String requestPurl) {
+        if (!checksumMatchingEnabled) {
+            return false;
+        }
+        try {
+            return SnykPurlUtil.requiresChecksumMeta(new PackageURL(requestPurl), true);
+        } catch (MalformedPackageURLException e) {
+            return requestPurl.contains("checksum=");
```
nscuro: `requestPurl` always comes from `SnykPurlUtil.toSnykRequestPurl` and should never be invalid. The fallback is dangerous because it could match on arbitrary prefixes of `checksum` like `foo_checksum=`. Either return `false` here or propagate the exception.

nscuro (changes requested):

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykPurlUtil.java
mehab: it is a limitation from vendor. They are gradually increasing support. Recently talked about gradle I think. But that needs to be tested first

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
nscuro: This is still unfixed: https://github.com/mehab/dependency-track/blob/a05c994b089080538d37fc7de815c4bb5a7c6bad/vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java#L410-L423

Issues keyed under `...@1.0?checksum=A` are returned for the request `...@1.0?checksum=B`.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
```diff
@@ -319,6 +580,8 @@ private Bom assembleVdr(Map<String, List<SnykIssue>> issuesByPurl, Map<String, S
                 final Vulnerability.Builder vulnBuilder = vulnBuilderByVulnId.computeIfAbsent(
                         issue.id(), _ -> SnykModelConverter.convert(issue, aliasSyncEnabled));
 
+                SnykModelConverter.applyMatchingPercentage(vulnBuilder, matchingPercentage);
```
nscuro: This applies matching percentages per vuln, but what Snyk provides is a match type per *component*. What happens if component A and B are both affected by vuln V, and component A is a partial match and B a full match?

apiserver/src/main/java/org/dependencytrack/vulnanalysis/ReconcileVulnAnalysisResultsActivity.java
```diff
@@ -298,6 +302,20 @@ vulnKey, new ReportedVulnerability(vdrVuln, internalVulnId), (existing, incoming
...
+                try {
+                    return Short.parseShort(prop.getValue());
+                } catch (NumberFormatException e) {
+                    LOGGER.warn("Invalid matching percentage: {}", prop.getValue());
+                }
+            }
+        }
+
+        return null;
+    }
```
nscuro: `MATCHING_PERCENTAGE` is a leftover from v5 work that got reverted before the v5 GA release. It doesn't currently do anything and nothing reads it. Even if we surface it through the UI, it's not actionable.

So I'd prefer we skip the matching percentage in this PR entirely, and if it's desired, we figure out how it'll actually be used.

The core value this PR provides is filtering out true false positives when we know *for sure* that component checksums don't match (i.e. match type `none`). To that end I'd recommend:

* `full`, `partial` -> Report the finding
* `none` -> Discard the finding

Alternatively, make it configurable whether `partial` is considered a match or a no-match. But only if that's really a distinction users would want.

That would remove all changes in the API server, which are currently also entirely untested.

nscuro (changes requested):

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
mehab: Fixed. Checksum-qualified issue correlation now uses meta package URL, exact/normalized request keys, and a coordinates-only key lookup only. It no longer matches other checksum-qualified keys that share coordinates, so issues for `...@1.0?checksum=A` are not attached to `...@1.0?checksum=B`.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
mehab: Agreed — match type is per component, so a vuln-level matching percentage cannot represent mixed full/partial affects correctly. Matching percentage is removed from this PR.

apiserver/src/main/java/org/dependencytrack/vulnanalysis/ReconcileVulnAnalysisResultsActivity.java
mehab: Done. Matching percentage and the apiserver changes are removed. Behaviour is now: `full` and `partial` report findings, `none` discards them.

vuln-analysis/snyk/src/test/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzerChecksumMatchingTest.java
mehab: Yes — for a permanent `meta.errors` failure (e.g. Unsupported Ecosystem) we now negative-cache. `shouldNegativeCacheWhenMetaErrorsPresentForChecksumPurl` asserts a second analyze() does not hit Snyk again (`verify(1, postRequestedFor(anyUrl()))`). Missing meta *without* errors is still not cached, so a transient gap can be retried.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykCachedPurlResult.java
```diff
@@ -0,0 +1,62 @@
...
+ * {@link SnykMatchType#FULL}, and empty {@link SnykMatchType#PARTIAL}) are cached as
+ * {@code null} and do not use this type.
+ * <p>
+ * {@link SnykMatchType#PARTIAL} may still include issues (name/version matched, checksum
+ * did not); those are cached here with {@code matchType = PARTIAL} so cache hits retain
+ * findings without re-fetching.
+ *
+ * @since 5.1.0
+ */
+record SnykCachedPurlResult(
```
nscuro: I believe this is no longer needed, and you can return to just caching `SnykIssue[]` like it is on the `main` branch.

nscuro (changes requested):
Please look if you can remove `SnykCachedPurlResult` and rebase on `main` to resolve the merge conflicts.

On `main` there is now a toggle to allow for non-batched requests. It would be good to verify whether the non-batched endpoint also supports `meta.packages` or whether it's a batch-only feature.

nscuro (approved):

Issue #2778, opened 2023-05-19
Extend project filtering by tags and versions


FabioRighe:
### Current Behavior

The filtering of the projects is done by a input text field, where is possible to filter exclusively by project name.

![image](https://github.com/DependencyTrack/dependency-track/assets/37578397/fbb66647-e188-41a2-b721-33794901b9e0)

### Proposed Behavior

It would be nice if the filter was improved, adding the possibility of doing mixed search of different columns: Project Name + Tags (1 or more) + Version, in a single search.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

Issue #926, opened 2021-02-10
Support project search/filtering by tags


ScottChapman:
Tags are nice, but there is no way to search/filter for projects by tags in the UI.

stevespringett:
Yes there is. click a tag of a project and it will show you only the projects with that tag. This works very similar to how GitHub itself works. If you're looking for something different, please be very specific about what you want.

ScottChapman:
ah, I see. that does work! I think what I was looking for was something on the search page, but it's less important if the mechanism exists.
A solution would be to have a `tags` field in search. Then one could simply select the tag there.

stevespringett:
Yes, I think there's room for improvement. I think if you want to turn this enhancement request into a search by tag title, then that's something that would be specific and actionable.

ScottChapman:
Sorry, what I meant above was to have a `tags` column on the project search table that you could make visible. And it would have the list of tags that one could simply click like one would do on the page for a specific project.

msymons:
See also ##238 and 586 (tag management).  This latter covers some of the problems which could cause problems for searching...  ie, dont want to have tags "scott" and "SCOTT" and "Scott", want to control who can create tags,  want to be able to delete tags (tidy up).

ckazimie:
@stevespringett you wrote it is possible to click on a single tag to view only projects with that tag. Is there an url where you can combine 2 tags so that you can view the projects that include both tags?

jhrv:
We are using tags extensively, and would be very nice to have a way of combining multiple tags when searching projects.

Issue #144, opened 2018-05-04
Filter to show only projects with vulnerabilities


StephenTrombetti:
This is an enhancement idea/suggestion.

Under the projects section add a filter (check box, toggle button, something similar...) to display only the projects containing vulnerabilities in the table.

This would be useful when there are a large number of projects and keeps the user from clicking through multiple pages to look for vulnerable projects.

Perhaps clicking the "Projects at Risk" display box could trigger this or take the user to a separate page displaying only the projects at risk.

stevespringett:
Nice suggestion. Agreed, this would be very useful.

oxdef:
+1 Current badges with numbers, e.g. "Projects at Risk" may mislead user who think that it's links to filter projects.
 
![2020-01-14_15-55](https://user-images.githubusercontent.com/3337518/72346318-ca24aa80-36e6-11ea-969e-ac559347d819.png)

PR #6873, opened 2026-07-31
Implemented project filtering.

ElenaStroebele:
### Description
Implemented project filtering for GET /api/v2/projects.

Adds a paginated list endpoint with query-based filters, sorting, optional metrics expansion, and ACL-aware results. Text filters are case-insensitive; 
Frontend implementation: [https://github.com/DependencyTrack/frontend/pull/1721](https://github.com/DependencyTrack/frontend/pull/1721)

### Addressed Issue
#926 
#2778 
(#144)

### Additional Details

**Filter specification**
- Name (name_contains) — substring match on project name; empty/whitespace values are ignored
- Version (version_contains) — substring match on version; empty/whitespace values are ignored
- Tags (tags) — exact match on tag names; multiple values require all tags (AND)
- Teams (teams) — exact match on team names; multiple values match any team (OR)
- Classifier (classifier) — exact match; multiple values match any classifier (OR); unknown values → 400
- Severity (severity) — projects with ≥1 vulnerability of the given severities (from metrics); multiple values match any (OR); unknown values → 400
- Parent (parent_uuid) — only direct children of that parent
- Ancestor (ancestor_uuid) — all descendants at any depth (excluding the ancestor)
- Only root (only_root=true) — only projects without a parent
- Has children (has_children=true) — only projects with ≥1 direct child visible to the caller
- Active (is_active) — true = active only; false = inactive only; omit = both
- Latest (is_latest) — true = latest only; false = non-latest only; omit = both
- Last BOM import (last_bom_import_since / last_bom_import_before) — inclusive lower / exclusive upper bound; projects that never imported a BOM are excluded when either is set

Screenshot 2026-07-31 163750

### Checklist


apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -710,6 +726,403 @@ OR EXISTS (
...
+            <#-- @ftlvariable name="includeMetrics" type="boolean" -->
+            <#-- @ftlvariable name="whereConditions" type="java.util.Collection<String>" -->
+            <#-- @ftlvariable name="collectionMetricsSubquery" type="String" -->
+            <#-- @ftlvariable name="leafMetricsSubquery" type="String" -->
+            <#-- @ftlvariable name="sortByColumn" type="org.dependencytrack.persistence.jdbi.query.ListAllProjectsQuery.SortBy" -->
+            <#-- @ftlvariable name="sortDirection" type="org.dependencytrack.common.pagination.SortDirection" -->
+            <#-- @ftlvariable name="hasCursor" type="boolean" -->
+            <#-- @ftlvariable name="filterHasChildren" type="boolean" -->
+            <#-- @ftlvariable name="apiProjectAclCondition" type="String" -->
+            <#assign childProjectAclCondition = apiProjectAclCondition?replace('"PROJECT"."ID"', '"CHILD_PROJECT"."ID"')>
```
nscuro: Access checks on child projects are redundant. ACL is inherited, so having access to the parent implies access to the children:

* https://github.com/DependencyTrack/dependency-track/blob/main/docs/PORTFOLIO_ACCESS_CONTROL.md#data-model
* https://dependencytrack.github.io/docs/next/concepts/access-control/#further-reading

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/PaginationSupport.java
```diff
@@ -97,13 +118,22 @@ default TotalCount getBoundedTotalCount(
...
                       ${fromWhereClause}
                 <#if includeAcl>
                        AND ${apiProjectAclCondition}
+                </#if>
+                <#if filterHasChildren>
+                       AND EXISTS (
+                         SELECT 1
+                           FROM "PROJECT" AS "CHILD_PROJECT"
+                          WHERE "CHILD_PROJECT"."PARENT_PROJECT_ID" = "PROJECT"."ID"
+                            AND (${childProjectAclCondition}))
```
nscuro: Same as in `ProjectDao`, access checks on child projects are redundant. 

`filterHasChildren` is also a filter specific to a single query, but `PaginationSupport` is shared code used by multiple queries. Domain-specific filters don't belong here.

api/src/main/openapi/resources/projects/schemas/list-projects-response-item.yaml
```diff
@@ -0,0 +1,93 @@
...
+  teams:
+    type: array
+    description: Teams with access to the project. Empty when no teams are assigned.
+    items:
+      type: object
+      properties:
+        name:
+          type: string
+      required:
+        - name
```
nscuro, line 70: Shouldn't be included in the response by default. 10s to 100s of teams can have access to a project. Always including it makes queries unnecessarily expensive and results bloated. API v1 had no `expand` concept, but v2 does, and we should make use of it here.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -710,6 +726,403 @@ OR EXISTS (
...
+                   ) AS "teamNames"
+                 , (
+                     SELECT JSONB_STRIP_NULLS(JSONB_BUILD_OBJECT(
+                              'supplier', "SUPPLIER"::JSONB,
+                              'authors', "AUTHORS"::JSONB,
+                              'tools', "TOOLS"::JSONB
+                            ))
+                       FROM "PROJECT_METADATA"
+                      WHERE "PROJECT_METADATA"."PROJECT_ID" = "PROJECT"."ID"
+                   ) AS "metadataJson"
```
nscuro: Selected but never used and not returned by the API.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -710,6 +726,403 @@ OR EXISTS (
...
+            <#-- @ftlvariable name="hasCursor" type="boolean" -->
+            <#-- @ftlvariable name="filterHasChildren" type="boolean" -->
+            <#-- @ftlvariable name="apiProjectAclCondition" type="String" -->
+            <#assign childProjectAclCondition = apiProjectAclCondition?replace('"PROJECT"."ID"', '"CHILD_PROJECT"."ID"')>
+            SELECT "PROJECT"."ID"
+                 , "PROJECT"."CLASSIFIER" AS "classifier"
+                 , "PROJECT"."CPE"
+                 , "PROJECT"."DESCRIPTION"
+                 , "PROJECT"."DIRECT_DEPENDENCIES"
+                 , "PROJECT"."EXTERNAL_REFERENCES"
```
nscuro: Selected but never used and not returned by the API.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -710,6 +726,403 @@ OR EXISTS (
...
+                     SELECT EXISTS(
+                       SELECT 1
+                         FROM "PROJECT" AS "CHILD_PROJECT"
+                        WHERE "CHILD_PROJECT"."PARENT_PROJECT_ID" = "PROJECT"."ID"
+                          AND (${childProjectAclCondition}))
+                   ) AS "hasChildren"
+                 , "PROJECT"."NAME" AS "name"
+                 , "PROJECT"."PUBLISHER"
+                 , "PROJECT"."PURL"
+                 , "PROJECT"."SWIDTAGID"
```
nscuro: Selected but never used and not returned by the API.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -710,6 +726,403 @@ OR EXISTS (
...
+                 , "PROJECT"."UUID"
+                 , "PROJECT"."VERSION" AS "version"
+                 , "PROJECT"."SUPPLIER"
+                 , "PROJECT"."MANUFACTURER"
```
nscuro: Selected but never used and not returned by the API.

api/src/main/openapi/resources/projects/schemas/list-projects-response-item.yaml
```diff
@@ -0,0 +1,93 @@
...
+    maxLength: 255
+  version:
+    type: string
+    maxLength: 255
+  group:
+    type: string
+    maxLength: 255
+  classifier:
+    type: string
+    maxLength: 255
```
nscuro: Use [shared classifier](https://github.com/DependencyTrack/dependency-track/blob/7264057017b778f0b754a121753d014f0b72d2fb/api/src/main/openapi/shared/schemas/classifier.yaml).

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+            - DEVICE
+            - FIRMWARE
+            - FILE
+            - PLATFORM
+            - DEVICE_DRIVER
+            - MACHINE_LEARNING_MODEL
+            - DATA
+            - CRYPTOGRAPHIC_ASSET
+      style: form
+      explode: true
```
nscuro, line 207: Use [shared classifier](https://github.com/DependencyTrack/dependency-track/blob/7264057017b778f0b754a121753d014f0b72d2fb/api/src/main/openapi/shared/schemas/classifier.yaml).

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+      schema:
+        type: boolean
+    - name: is_active
+      in: query
+      description: |-
+        When `true`, return only active projects.
+        When `false`, return only inactive projects.
+        Omit to include projects regardless of active state.
+      schema:
+        type: boolean
```
nscuro: Should use [shared project state](https://github.com/DependencyTrack/dependency-track/blob/7264057017b778f0b754a121753d014f0b72d2fb/api/src/main/openapi/shared/schemas/project-state.yaml).

api/src/main/openapi/resources/projects/schemas/list-projects-response-item.yaml
```diff
@@ -0,0 +1,93 @@
...
+  tags:
+    type: array
+    description: Tag names assigned to the project. Empty when no tags are assigned.
+    items:
+      type: object
+      properties:
+        name:
+          type: string
+      required:
+        - name
```
nscuro: Should just be an array of `string`. There won't ever be more fields than `name`.

api/src/main/openapi/resources/projects/schemas/list-projects-response-item.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - AGGREGATE_DIRECT_CHILDREN
+      - AGGREGATE_DIRECT_CHILDREN_WITH_TAG
+      - AGGREGATE_LATEST_VERSION_CHILDREN
+  collection_tag:
+    type: object
+    properties:
+      name:
+        type: string
+    required:
+      - name
```
nscuro: Should just be a `string`. There won't ever be more fields than `name`.

api/src/main/openapi/resources/projects/schemas/list-projects-response-item.yaml
```diff
@@ -0,0 +1,93 @@
...
+    description: Teams with access to the project. Empty when no teams are assigned.
+    items:
+      type: object
+      properties:
+        name:
+          type: string
+      required:
+        - name
+  parent:
+    $ref: "../../components/schemas/component-project.yaml"
```
nscuro: https://github.com/DependencyTrack/dependency-track/tree/main/api#conventions

> **Cross-resource `$ref`s only into `shared/`.** A path or schema under
  `resources/<a>/` may not reference files under `resources/<b>/`. Promote to
  `shared/` when reuse arises.

api/src/main/openapi/resources/projects/schemas/list-projects-response-item.yaml
```diff
@@ -0,0 +1,93 @@
...
+  has_children:
+    type: boolean
+    description: >-
+      Whether the project has at least one direct child that is accessible to the
+      caller. When portfolio access control is disabled, this reflects all direct
+      children in the database.
+  metrics:
+    $ref: "../../components/schemas/dependency-metrics.yaml"
```
nscuro: See note about cross-resource reuse above, but also this should be *project* metrics, not dependency metrics.

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+    * `inactive_since`
+    * `is_latest`
+    * `last_bom_import`
+    * `last_inherited_risk_score`
+
+    ### Expandable fields
+
+    The following fields can be included via `expand`:
+
+    * `metrics`
```
nscuro, line 47: Should also be expandable, not included by default:

* `parent`
* `teams`

apiserver/src/main/java/org/dependencytrack/resources/v2/mapping/ModelMapper.java
```diff
@@ -104,6 +108,85 @@ public static DependencyMetrics mapDependencyMetrics(org.dependencytrack.model.D
                 .build();
     }
 
+    public static DependencyMetrics mapProjectMetrics(@Nullable ProjectMetrics metrics) {
```
nscuro: Should map to `ProjectMetrics`.

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+        items:
+          type: string
+          enum:
+            - critical
+            - high
+            - medium
+            - low
+            - unassigned
+      style: form
+      explode: true
```
nscuro, line 194: This resource is the wrong place to filter by severity. There are already finding endpoints that should be used, or could be extended, for this use case.

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+      description: |-
+        Filter by tag names (exact match, case-insensitive).
+        When multiple values are provided, only projects matching **all** tags are returned.
+      schema:
+        type: array
+        items:
+          type: string
+      style: form
+      explode: true
+    - name: teams
```
nscuro: To communicate how the provided values are matched:

```suggestion
    - name: teams_any
```

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+    - name: version_contains
+      in: query
+      description: |-
+        Filter by version (substring match).
+
+        Leading and trailing whitespace is ignored. Values that are empty or
+        whitespace-only are treated as if the parameter was omitted.
+      schema:
+        type: string
+    - name: tags
```
nscuro: To communicate how the provided values are matched:

```suggestion
    - name: tags_all
```

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+      schema:
+        type: array
+        items:
+          type: string
+          enum:
+            - critical
+            - high
+            - medium
+            - low
+            - unassigned
```
nscuro: [Enum values should be upper snake case](https://github.com/DependencyTrack/dependency-track/blob/352e26f10579907fcb57ba697e296bd3ce4bc010/api/src/main/spectral/zalando.yaml#L206-L215).

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+            - PLATFORM
+            - DEVICE_DRIVER
+            - MACHINE_LEARNING_MODEL
+            - DATA
+            - CRYPTOGRAPHIC_ASSET
+      style: form
+      explode: true
+    - name: expand
+      in: query
+      description: Optional fields to include in each project response item. Unknown values are silently ignored.
```
nscuro, line 210: > Unknown values are silently ignored.

Is not true when the parameter is an enum. Use a simple `string` here, which aligns with how other endpoints do it.

api/src/main/openapi/resources/projects/projects.yaml
```diff
@@ -0,0 +1,259 @@
...
+      in: query
+      description: |-
+        Filter by parent project UUID.
+
+        Returns only direct children of the given parent.
+
+        When portfolio access control is enabled and the caller does not have
+        access to the parent project, an empty result set is returned.
+
+        Used for tree-view navigation (lazy loading of one hierarchy level).
```
nscuro: Nit: the API has no concept of tree-view navigation.

api/src/main/openapi/resources/projects/schemas/list-projects-response-item.yaml
```diff
@@ -0,0 +1,93 @@
...
+  group:
+    type: string
+    maxLength: 255
+  classifier:
+    type: string
+    maxLength: 255
+  is_latest:
+    type: boolean
+  is_active:
+    type: boolean
```
nscuro: Should use [shared project state](https://github.com/DependencyTrack/dependency-track/blob/7264057017b778f0b754a121753d014f0b72d2fb/api/src/main/openapi/shared/schemas/project-state.yaml).

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -530,6 +539,13 @@ SELECT ARRAY_AGG("TAG"."NAME")
                          ON "PROJECTS_TAGS"."PROJECT_ID" = "PROJECT"."ID"
                       WHERE "TAG"."ID" = "PROJECTS_TAGS"."TAG_ID"
                    ) AS "tagNames"
+                 , (
+                     SELECT ARRAY_AGG("TEAM"."NAME")
+                       FROM "TEAM"
+                      INNER JOIN "PROJECT_ACCESS_TEAMS"
+                         ON "PROJECT_ACCESS_TEAMS"."TEAM_ID" = "TEAM"."ID"
+                      WHERE "PROJECT_ACCESS_TEAMS"."PROJECT_ID" = "PROJECT"."ID"
+                   ) AS "teamNames"
```
nscuro: This query is for the API v1 endpoint, but that one doesn't return `teamNames`.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -710,6 +726,403 @@ OR EXISTS (
...
+                        OR (${castedLastSortValue} IS NOT NULL AND "PROJECT"."${sortByColumn}" IS NOT NULL
+                            AND ("PROJECT"."${sortByColumn}" > ${castedLastSortValue}
+                                 OR ("PROJECT"."${sortByColumn}" = ${castedLastSortValue} AND "PROJECT"."ID" > :lastId)))
+                        OR (${castedLastSortValue} IS NOT NULL AND "PROJECT"."${sortByColumn}" IS NULL))
+               </#if>
+            <#elseif hasCursor>
+               AND "PROJECT"."ID" > :lastId
+            </#if>
+            <#if sortByColumn?has_content>
+             ORDER BY "PROJECT"."${sortByColumn}" ${sortDirection!"ASC"}, "PROJECT"."ID" ASC
```
nscuro: This breaks pagination when `sortByColumn=LAST_RISKSCORE`, because it doesn't work for collection projects. Earlier in the query there is a `CASE` statement that computes `lastInheritedRiskScore` in a way that takes collections into consideration. We thus need to sort by `lastInheritedRiskScore`.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -710,6 +726,403 @@ OR EXISTS (
...
+                query.includeMetrics(),
+                decodedPageToken != null
+                        ? decodedPageToken.lastId()
+                        : null,
+                effectiveSortBy,
+                effectiveSortDirection,
+                decodedPageToken != null,
+                filterHasChildren,
+                COLLECTION_METRICS_SUBQUERY,
+                LEAF_METRICS_SUBQUERY);
```
nscuro: Use `withJitDisabled` like the other project listing queries:

https://github.com/DependencyTrack/dependency-track/blob/352e26f10579907fcb57ba697e296bd3ce4bc010/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java#L688-L710

nscuro (changes requested):

api/src/main/openapi/resources/projects/projects.yaml
ElenaStroebele: Hey @nscuro , 
I wasn't sure what you had in mind here. How would you implement the severity filtering, instead of the current approach?

nscuro: Argh, apologies @ElenaStroebele I didn't see this comment earlier because GitHub had hidden the conversation behind a "Show more" button. I assumed you were still going to address some other comments and didn't follow up. Please feel free to click the "Resolve conversation" button on comments you addressed so only the truly pending items remain visible.

On this one, let's split it out into a separate PR, everything else is pretty much good for merge so dropping this particular item would unblock things for now.

To rephrase my reservation: We already have `/api/v1/finding/*` endpoints that filter on severity and more. My worry is that adding a severity filter to the project list invites later requests for "also filter by CVSS score / EPSS / KEV" etc. and the endpoint becomes a kitchen sink. That said I'm open to discuss the use case behind this and adding it if it makes sense. But as stated above, for the sake of getting this PR in I'd decouple that.

WDYT?

api/src/main/openapi/resources/projects/schemas/list-projects-response-item.yaml
```diff
@@ -0,0 +1,84 @@
...
+    description: >-
+      Teams with access to the project. Only present when `expand=teams` is requested.
+      Empty array when expanded and no teams are assigned.
+    items:
+      type: object
+      properties:
+        name:
+          type: string
+      required:
+        - name
```
nscuro, line 70: This could also be a list of strings instead.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -713,6 +774,404 @@ OR EXISTS (
...
+                  THEN cm."inheritedRiskScore"
+                  ELSE "PROJECT"."LAST_RISKSCORE"
+                END
+            </#assign>
+            SELECT "PROJECT"."ID"
+                 , "PROJECT"."CLASSIFIER" AS "classifier"
+                 , "PROJECT"."GROUP" AS "group"
+                 , "PROJECT"."LAST_BOM_IMPORTED" AS "lastBomImport"
+                 , "PROJECT"."LAST_BOM_IMPORTED_FORMAT" AS "lastBomImportFormat"
+                 , "PROJECT"."LAST_VULNERABILITY_ANALYSIS"
```
nscuro, line 1063: Also selected but never used and not returned by the API.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -511,7 +571,8 @@ record ListProjectsRow(
                      SELECT EXISTS(
                        SELECT 1
                          FROM "PROJECT" AS "CHILD_PROJECT"
-                        WHERE "CHILD_PROJECT"."PARENT_PROJECT_ID" = "PROJECT"."ID")
+                        WHERE "CHILD_PROJECT"."PARENT_PROJECT_ID" = "PROJECT"."ID"
+                          AND (${childProjectAclCondition}))
```
nscuro, line 575: Also no longer necessary.

PR #6874, opened 2026-08-02
Support AWS-managed credentials for S3 storage

### LINKED ISSUE #6851: S3 file storage: support ambient AWS credentials (ECS task roles, EKS IRSA) instead of static keys only
### Current Behavior

  The `s3` file storage provider only supports **static credentials**. In `S3FileStorageProvider.create()` the MinIO client is built like this:

```java
final var accessKey = config.getOptionalValue("dt.file-storage.s3.access-key", String.class).orElse(null);
final var secretKey = config.getOptionalValue("dt.file-storage.s3.secret-key", String.class).orElse(null);
if (accessKey != null && secretKey != null) {
    clientBuilder.credentials(accessKey, secretKey);
}
```

When `access-key` / `secret-key` are not configured, no credentials provider is installed at all, so the client sends **unsigned (anonymous) requests** (`BaseS3Client.executeAsync`: `Credentials credentials = (provider == null) ? null : provider.fetch();`). There is no fallback to ambient AWS credential resolution — environment variables, `~/.aws` config, **ECS task roles** (`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`), **EKS IRSA** (web identity), EKS Pod Identity, and EC2 instance profiles are all ignored.

On AWS this makes the `s3` provider effectively unusable in well-configured environments: on ECS and Kubernetes/EKS the standard (and often org-mandated) way to grant S3 access is the task role / IRSA, not long-lived IAM user keys. Minting a static key pair just for file storage is an anti-pattern (no rotation, secret sprawl) and is frequently prohibited outright.

This matters more than it first appears, because the default `local` provider silently breaks horizontally scaled deployments: with more than one apiserver replica, a BOM accepted by one replica may have its `import-bom` workflow picked up by another (the workflow engine coordinates through PostgreSQL), where `FileStorage#get` throws `NoSuchFileException`, the workflow fails terminally, and the project is left empty — while `/api/v1/event/token/{uuid}` reports processing complete. We hit exactly this running 2 replicas on ECS (verified on 5.0.3). Shared storage is the fix, and S3 is the natural choice — but only static keys are supported today. (We worked around it with an EFS volume mounted over the data directory.)

Observed on 5.0.3; the code on `main` is unchanged.

### Proposed Behavior

When `dt.file-storage.s3.access-key` / `secret-key` are **not** configured, fall back to ambient credential resolution instead of anonymous access. The MinIO client already bundled with Dependency-Track ships everything needed in `io.minio.credentials` — no new dependencies:

```java
if (accessKey != null && secretKey != null) {
    clientBuilder.credentials(accessKey, secretKey);
} else {
    clientBuilder.credentialsProvider(new ChainedProvider(
            new AwsEnvironmentProvider(),
            new AwsConfigProvider(),
            new IamAwsProvider(null, httpClient)));
}
```

`IamAwsProvider` already handles the ECS container credentials endpoint (`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` / `_FULL_URI`), EKS IRSA (`AWS_WEB_IDENTITY_TOKEN_FILE`), and EC2 IMDS, so this single change makes the provider work with ECS task roles and Kubernetes service-account-based auth out of the box, while keeping current static-key behavior fully backward compatible. If truly anonymous access should remain reachable, it could be kept behind an explicit opt-in flag rather than being the silent default.

Happy to submit a PR for this.

### Checklist

fouad-j:
### Description

The S3 file storage provider only authenticated when static access and secret keys were configured, and otherwise issued anonymous requests. This prevented the use of AWS-managed credentials (IRSA, ECS task roles, EC2 instance profiles), which are the recommended way to grant access on AWS and avoid long-lived static keys.

This PR adds a `dt.file-storage.s3.credentials-source` option, defaulting to `static` to preserve existing behavior. When set to `aws`, credentials are resolved from the workload's AWS identity, so no static keys need to be configured.

### Addressed Issue

Closes #6851

### Additional Details

When `credentials-source=aws`, credentials are resolved via MinIO's `ChainedProvider`, composed of `AwsEnvironmentProvider` (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`), `AwsConfigProvider` (`~/.aws/credentials`), and `IamAwsProvider`, which itself supports web identity tokens (IRSA), ECS container credentials (`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` `AWS_CONTAINER_CREDENTIALS_FULL_URI`) and the EC2 instance metadata service, in that order of precedence.

Note that this is close to, but not exactly, the AWS default credential provider chain: the AWS SDK consults web identity tokens before `~/.aws/credentials`, whereas the chain above is evaluated in the order listed.

Behavior change worth highlighting: configuring only one of the two static keys now fails at startup instead of silently falling back to anonymous requests. Such a configuration could never authenticate, since SigV4 requires both an access key and a secret key, so this only surfaces already-broken setups. Setting static keys together with `credentials-source=aws` fails at startup as well, so a leftover key cannot silently shadow the intended source.

The credential resolution logic is covered by `S3FileStorageProviderTest`, which deliberately does not use Testcontainers so it runs without Docker.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/resources/application.properties
+++ b/apiserver/src/main/resources/application.properties
@@ -1138,10 +1138,15 @@ dt.file-storage.local.directory=${dt.data-directory}/storage
 # `dt.file-storage.s3.secret-key` are used. When both are missing, requests are performed
 # anonymously. When only one of the two is set, startup fails.
 # <br/><br/>
-# When set to `aws`, credentials are resolved from the AWS credential chain. This supports
-# ECS task roles, IRSA / web identity tokens, and the EC2 instance metadata service, and
-# does not require static keys to be configured. `dt.file-storage.s3.access-key` and
-# `dt.file-storage.s3.secret-key` are ignored in this mode.
+# When set to `aws`, credentials are resolved from the environment, mirroring the credential
+# provider chain of the AWS SDK: `AWS_*` environment variables, the shared AWS config file
+# (~/.aws/credentials), ECS task roles, EKS IRSA / web identity tokens, and the EC2 instance
+# metadata service, in that order. Static keys are not required and are ignored in this mode.
+# Startup fails if no credentials can be resolved.
+# <br/><br/>
+# EKS Pod Identity is not supported: the S3 client does not read
+# AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE and rejects the non-loopback Pod Identity Agent
+# endpoint. Use IRSA instead.
 #
 # @category:     Storage
 # @default:      static
--- a/apiserver/src/main/resources/application.properties
+++ b/apiserver/src/main/resources/application.properties
@@ -1132,13 +1132,32 @@ dt.file-storage.local.directory=${dt.data-directory}/storage
 # @type:     string
 # dt.file-storage.s3.bucket=
 
+# Defines the source of the credentials used to authenticate against the S3 endpoint.
+# <br/><br/>
+# When set to `static`, the statically configured `dt.file-storage.s3.access-key` and
+# `dt.file-storage.s3.secret-key` are used. When both are missing, requests are performed
+# anonymously. When only one of the two is set, startup fails.
+# <br/><br/>
+# When set to `aws`, credentials are resolved from the AWS credential chain. This supports
+# ECS task roles, IRSA / web identity tokens, and the EC2 instance metadata service, and
+# does not require static keys to be configured. `dt.file-storage.s3.access-key` and
+# `dt.file-storage.s3.secret-key` are ignored in this mode.
+#
+# @category:     Storage
+# @default:      static
+# @type:         enum
+# @valid-values: [static, aws]
+# dt.file-storage.s3.credentials-source=static
+
 # Defines the S3 access key / username.
+# Has no effect when dt.file-storage.s3.credentials-source is `aws`.
 #
 # @category: Storage
 # @type:     string
 # dt.file-storage.s3.access-key=
 
 # Defines the S3 secret key / password.
+# Has no effect when dt.file-storage.s3.credentials-source is `aws`.
 #
 # @category: Storage
 # @type:     string
--- a/file-storage/provider-s3/src/main/java/org/dependencytrack/filestorage/s3/S3FileStorageProvider.java
+++ b/file-storage/provider-s3/src/main/java/org/dependencytrack/filestorage/s3/S3FileStorageProvider.java
@@ -21,6 +21,9 @@ package org.dependencytrack.filestorage.s3;
 import com.github.luben.zstd.Zstd;
 import io.minio.BucketExistsArgs;
 import io.minio.MinioClient;
+import io.minio.credentials.AwsConfigProvider;
+import io.minio.credentials.AwsEnvironmentProvider;
+import io.minio.credentials.ChainedProvider;
 import io.minio.credentials.IamAwsProvider;
 import io.minio.credentials.Provider;
 import io.minio.credentials.StaticProvider;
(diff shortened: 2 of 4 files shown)


file-storage/provider-s3/src/main/java/org/dependencytrack/filestorage/s3/S3FileStorageProvider.java
```diff
@@ -38,6 +44,8 @@ public final class S3FileStorageProvider implements FileStorageProvider {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(S3FileStorageProvider.class);
     static final String NAME = "s3";
+    static final String CREDENTIALS_SOURCE_STATIC = "static";
+    static final String CREDENTIALS_SOURCE_AWS = "aws";
```
nscuro: Use an enum instead. You can use enums in `config.getOptionalValue` etc. and the config provider will handle conversion and validation automatically.

file-storage/provider-s3/src/test/java/org/dependencytrack/filestorage/s3/S3FileStorageTest.java
```diff
@@ -58,26 +61,112 @@ void shouldHaveNameS3() {
...
+    }
+
+    @Test
+    void resolveCredentialsProviderShouldIgnoreCaseAndWhitespaceForSource() {
+        final Config config = configOf(Map.ofEntries(
+                Map.entry("dt.file-storage.s3.credentials-source", "  AWS  ")));
+
+        assertThat(S3FileStorageProvider.resolveCredentialsProvider(config, new OkHttpClient()))
+                .hasValueSatisfying(provider -> assertThat(provider).isInstanceOf(IamAwsProvider.class));
+    }
```
nscuro: I don't think we should be lenient with config values. If the expected values are `static` and `aws`, then that is what users should provide. Everything else is not valid and should not be accepted.

file-storage/provider-s3/src/main/java/org/dependencytrack/filestorage/s3/S3FileStorageProvider.java
```diff
@@ -90,6 +94,56 @@ public FileStorage create(Config config, ProxySelector proxySelector) {
...
+     *     ({@code AWS_WEB_IDENTITY_TOKEN_FILE}), ECS task roles
+     *     ({@code AWS_CONTAINER_CREDENTIALS_RELATIVE_URI} / {@code AWS_CONTAINER_CREDENTIALS_FULL_URI}),
+     *     and the EC2 instance metadata service, in that order of precedence. No static keys are
+     *     required. Note that this is not the full AWS default credential provider chain:
+     *     {@code AWS_ACCESS_KEY_ID} / {@code AWS_SECRET_ACCESS_KEY} environment variables and
+     *     {@code ~/.aws/credentials} profiles are not consulted; use {@code static} for those.</li>
+     * </ul>
+     *
+     * @return The resolved {@link Provider}, if one should be configured.
+     */
```
nscuro: Nit: since this is a purely internal method and it's only a few lines of easy-to-read code, restating what it does in javadoc is redundant.

nscuro (changes requested):
Thanks. Please use an enum for the credential source option, the other comments are suggestions.

Please also update the user-facing documentation here: https://github.com/DependencyTrack/docs/blob/main/docs/reference/configuration/file-storage.md#s3 so it explains how to configure the provider for AWS, and what sources are checked.

apiserver/src/main/resources/application.properties
```diff
@@ -1132,13 +1132,37 @@ dt.file-storage.local.directory=${dt.data-directory}/storage
...
+# Defines the source of the credentials used to authenticate against the S3 endpoint.
+# <br/><br/>
+# When set to `static`, the statically configured `dt.file-storage.s3.access-key` and
+# `dt.file-storage.s3.secret-key` are used. When both are missing, requests are performed
+# anonymously. When only one of the two is set, startup fails.
+# <br/><br/>
+# When set to `aws`, credentials are resolved from the environment, mirroring the credential
+# provider chain of the AWS SDK: `AWS_*` environment variables, the shared AWS config file
+# (~/.aws/credentials), ECS task roles, EKS IRSA / web identity tokens, and the EC2 instance
+# metadata service, in that order. Static keys are not required and are ignored in this mode.
```
nscuro: > Static keys are not required and are ignored in this mode.

Is no longer true since the validation has been added.

nscuro (approved):
Thanks!

FYI I updated the PR description so it accurately states what's being shipped now.

The docs in `application.properties` are also outdated from an earlier iteration, i.e. "Static keys are not required and are ignored in this mode." is no longer the case. I'll fix that myself in a follow-up PR.

PR #6889, opened 2026-08-03
S3 ambient aws credentials

### LINKED ISSUE #6851: S3 file storage: support ambient AWS credentials (ECS task roles, EKS IRSA) instead of static keys only

fffinkel:
### Description

The `s3` file storage provider only supported static credentials. When `dt.file-storage.s3.access-key` and `secret-key` were both absent, no credentials provider was installed at all, so the client sent unsigned (anonymous) requests with no fallback to ambient credential resolution. Task-based roles and IRSA are a common (and often org-mandated) way to grant S3 access.

When no static credentials are configured, the provider now resolves them from the environment instead, covering `AWS_*` environment variables, the shared AWS config file, ECS task roles, EKS IRSA, and EC2 instance profiles. Static credentials continue to work exactly as before, so this is backwards compatible for every configuration that sets them.

A second commit closes a related gap: configuring only one of `access-key` / `secret-key` used to be silently ignored, and now fails at startup naming both properties.

### Addressed Issue

Closes #6851

### Additional Details

**No new dependencies.** This uses the credential providers already bundled in `minio-java` (`io.minio.credentials`), chained in AWS-SDK-like order:

```java
clientBuilder.credentialsProvider(new ChainedProvider(
        new AwsEnvironmentProvider(),
        new AwsConfigProvider(null, null),
        new IamAwsProvider(null, httpClient)));
```

Switching to the AWS SDK v2 was considered and rejected, per @nscuro's own evaluation in DependencyTrack/helm-charts#384: roughly 4 MB of additional dependencies, and the synchronous S3 client lacks the streaming uploads `minio-java` provides.

**On the open question in the issue thread** (@nscuro: *"we'd need to verify how well Minio handles the case where no static creds are provided and it cannot resolve AWS creds dynamically. IIRC it fails loudly and the AWS S3 SDK doesn't"*). Verified against `minio-java` 9.0.3: `ChainedProvider.fetch()` throws `java.security.ProviderException("All providers fail to fetch credentials")` when every provider fails. So it does fail loudly. Resolution is lazy, consulted on the first request, and the existing bucket-existence check at startup is that first consumer, so unresolvable credentials fail startup rather than surfacing later at runtime.

**Behavior change worth calling out.** Truly anonymous (unsigned) access is no longer reachable via "configure no credentials". Anyone relying on that today would now fail at startup. I judged this acceptable because it requires a publicly writable bucket, and the Helm chart already refuses to render without credentials, so no chart user can be in that state. The issue floated keeping anonymous access behind an explicit opt-in flag. Happy to add one if you'd prefer that over the loud failure.

**EKS Pod Identity is not covered**, despite being requested in DependencyTrack/helm-charts#384. Two concrete blockers in `minio-java` 9.0.3's `IamAwsProvider`: it reads `AWS_CONTAINER_AUTHORIZATION_TOKEN` but not the `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` variant that the Pod Identity Agent injects, and `checkLoopbackHost()` rejects the agent's non-loopback endpoint (`169.254.170.23`). IRSA, ECS task roles, and IMDS all work. This limitation is documented in `application.properties`.

**Documentation.** `apiserver/src/main/resources/application.properties` is updated for both properties, including the Pod Identity caveat, so the generated property reference will pick it up. The hand-written file-storage page in DependencyTrack/docs is not generated, so it needs its own change: a companion PR there adds an "Authentication" section covering both credential modes, the resolution order, and the Pod Identity exclusion.

**ADR.** No new ADR. Per the criteria in `CONTRIBUTING.md`, this introduces no module or extension point, no schema or migration change, no REST contract change, no new runtime dependency or external integration, and no change to concurrency or consistency characteristics. It alters credential resolution inside an existing provider. It does adjust a default of an accepted ADR, though, so per the convention in `docs/adr/README.md` I added a `## Follow-up (2026-07-31)` section to [ADR-004](docs/adr/004-file-storage-plugin.md) instead. Happy to promote that to a full ADR if you read the criteria differently.

**Companion changes:**

* DependencyTrack/helm-charts#384 adds a `fileStorage.s3.authentication.type: auto` mode that stops injecting credentials so this resolution path can take over.
* A PR against DependencyTrack/docs documents the new behavior on the file-storage reference page.

**Testing notes.** Three tests added (ambient resolution via the environment; both half-configured-credential cases). Verified locally with JDK 25 + Docker: `file-storage/provider-s3` goes from 11 to 14 tests with my three passing. Two pre-existing failures in the `WhenHostIsUnavailable` nested class fail identically on an unmodified `main` in my environment. They assert `ConnectException` but a VM-based Docker runtime (colima) returns `SocketException: Connection reset`. They are unrelated to this change and should pass on CI's native Docker.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/file-storage/provider-s3/src/main/java/org/dependencytrack/filestorage/s3/S3FileStorageProvider.java
+++ b/file-storage/provider-s3/src/main/java/org/dependencytrack/filestorage/s3/S3FileStorageProvider.java
@@ -69,24 +69,43 @@ public final class S3FileStorageProvider implements FileStorageProvider {
 
         final var accessKey = config.getOptionalValue("dt.file-storage.s3.access-key", String.class).orElse(null);
         final var secretKey = config.getOptionalValue("dt.file-storage.s3.secret-key", String.class).orElse(null);
-        if (accessKey != null && secretKey != null) {
-            clientBuilder.credentials(accessKey, secretKey);
-        } else if (accessKey != null || secretKey != null) {
-            throw new IllegalStateException(
-                    "Incomplete static credentials: dt.file-storage.s3.access-key and dt.file-storage.s3.secret-key "
-                    + "must either both be configured, or both be omitted to resolve credentials from the environment");
-        } else {
-            // No static credentials. Resolve them from the environment instead, mirroring the
-            // credential provider chain of the AWS SDK. This covers AWS_* environment variables,
-            // the shared AWS config file, ECS task roles, EKS IRSA, and EC2 instance profiles.
-            // Resolution is lazy: the chain is consulted on the first request, and throws
-            // java.security.ProviderException when no provider yields credentials. The bucket
-            // existence check below is that first request, so unresolvable credentials fail startup.
-            LOGGER.debug("No static credentials configured; resolving credentials from the environment");
-            clientBuilder.credentialsProvider(new ChainedProvider(
-                    new AwsEnvironmentProvider(),
-                    new AwsConfigProvider(/* filename */ null, /* profile */ null),
-                    new IamAwsProvider(/* customEndpoint */ null, httpClient)));
+        final var credentialsSource = config
+                .getOptionalValue("dt.file-storage.s3.credentials-source", String.class)
+                .orElse("static");
+        switch (credentialsSource) {
+            case "aws" -> {
+                if (accessKey != null || secretKey != null) {
+                    throw new IllegalStateException(
+                            "Conflicting credentials configuration: dt.file-storage.s3.credentials-source is aws, "
+                            + "but dt.file-storage.s3.access-key or dt.file-storage.s3.secret-key is also configured");
+                }
+                // Resolve credentials from the environment, mirroring the credential provider
+                // chain of the AWS SDK. This covers AWS_* environment variables, the shared AWS
+                // config file, ECS task roles, EKS IRSA, and EC2 instance profiles.
+                // Resolution is lazy: the chain is consulted on the first request, and throws
+                // java.security.ProviderException when no provider yields credentials. The bucket
+                // existence check below is that first request, so unresolvable credentials fail startup.
+                LOGGER.debug("Resolving credentials from the environment");
+                clientBuilder.credentialsProvider(new ChainedProvider(
+                        new AwsEnvironmentProvider(),
+                        new AwsConfigProvider(/* filename */ null, /* profile */ null),
+                        new IamAwsProvider(/* customEndpoint */ null, httpClient)));
+            }
+            case "static" -> {
+                if (accessKey != null && secretKey != null) {
+                    clientBuilder.credentials(accessKey, secretKey);
+                } else if (accessKey != null || secretKey != null) {
+                    throw new IllegalStateException(
+                            "Incomplete static credentials: dt.file-storage.s3.access-key and "
+                            + "dt.file-storage.s3.secret-key must either both be configured, or both be omitted");
+                } else {
+                    // Installing no credentials provider makes the client send unsigned requests.
+                    // Intended for S3-compatible endpoints that allow anonymous access.
+                    LOGGER.debug("No static credentials configured; requests will be unsigned");
+                }
+            }
+            default -> throw new IllegalStateException(
+                    "Invalid credentials source: must be static or aws, but is " + credentialsSource);
         }
 
         config.getOptionalValue("dt.file-storage.s3.region", String.class).ifPresent(clientBuilder::region);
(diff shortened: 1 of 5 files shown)


file-storage/provider-s3/src/main/java/org/dependencytrack/filestorage/s3/S3FileStorageProvider.java
```diff
@@ -67,6 +71,22 @@ public FileStorage create(Config config, ProxySelector proxySelector) {
...
+            // credential provider chain of the AWS SDK. This covers AWS_* environment variables,
+            // the shared AWS config file, ECS task roles, EKS IRSA, and EC2 instance profiles.
+            // Resolution is lazy: the chain is consulted on the first request, and throws
+            // java.security.ProviderException when no provider yields credentials. The bucket
+            // existence check below is that first request, so unresolvable credentials fail startup.
+            LOGGER.debug("No static credentials configured; resolving credentials from the environment");
+            clientBuilder.credentialsProvider(new ChainedProvider(
+                    new AwsEnvironmentProvider(),
+                    new AwsConfigProvider(/* filename */ null, /* profile */ null),
+                    new IamAwsProvider(/* customEndpoint */ null, httpClient)));
```
fouad-j: Anonymous access stops being reachable. You call this out, and I agree it's a poor default on AWS — but it's currently the only way to run against an S3-compatible endpoint with no credentials, and that isn't limited to Helm users

fffinkel: Good catch, and agreed. I've reworked it so the default matches today's behavior, and ambient resolution is now opt-in via `dt.file-storage.s3.credentials-source=aws`. I believe that matches your proposal from the issue thread, name and semantics, which also avoids the IMDS-probe-on-non-AWS concern you raised there.

One thing my LLM turned up: anonymous *writes* are already broken on `main` independent of this PR. The minio 8.6.0 to 9.0.0 bump dropped the automatic Content-MD5 computation, and 9.x rejects unsigned request bodies without one, so `store()` throws on any credential-less config. I added a commit that restores them by spooling anonymous uploads to a temp file and sending a single part with an explicit `Content-MD5` (streaming multipart can't provide per-part hashes). That caps anonymous uploads at 5GiB; credentialed uploads stream as before. Arguably a minio-java regression worth filing upstream.

I know that's a bit out of scope for this issue, and it's also worth mentioning that I'm a horrible Java programmer, so please let me know what you think.

I've updated all three repos (dependency-track, helm-charts, docs) with this latest change.

fffinkel:
Closing this issue in favor of #6874, see discussion in #6851

PR #6914, opened 2026-08-07
Fix notification rule test to work regardless of rule level

valentijnscholten:
Today I lost a lot time debugging email alerts not working. The "Perform Test" button confirmed to me that a test email was submitted to the queue succesfully. But it didn't arrive. Turns out no e-mail was queued after all. The reason is that my alert has level "Error" and the test button only has an "Informational" template to use. So it didn't find a suitable template and didn't send a test e-mail. This PR simplifies the template/supplier matrix to always generate an event regardless of the chosen level.

## Summary
- Fixes the notification rule test endpoint silently succeeding without dispatching when a rule is configured at Error level but the selected event group only has Informational test templates.
- Looks up test notifications by scope and group only, then applies the rule's level to the dummy notification so publisher delivery can be verified independently of production routing semantics.
- Adds a regression test for Error-level rules with `NEW_VULNERABILITY`.

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/resources/v1/NotificationPublisherResource.java
+++ b/apiserver/src/main/java/org/dependencytrack/resources/v1/NotificationPublisherResource.java
@@ -429,23 +429,15 @@ public class NotificationPublisherResource extends AbstractApiResource {
     }
 
     private @Nullable Notification buildTestNotification(Scope scope, Group group, Level ruleLevel) {
-        final List<Level> levelsToTry = switch (ruleLevel) {
-            case LEVEL_INFORMATIONAL -> List.of(Level.LEVEL_INFORMATIONAL, Level.LEVEL_WARNING, Level.LEVEL_ERROR);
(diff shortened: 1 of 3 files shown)


nscuro:
> This PR simplifies the template/supplier matrix to always generate an event regardless of the chosen level.

I don't believe this is a good idea as it creates a false sense of assurance. The system would never emit such a notification beyond this test. So the manual test could "pass", but in production you'd never see a notification from the rule.

Better approaches would be to:

* Adjust the notification groups shown in the UI based on the configured level, so e.g. `NEW_VULNERABILITY` is not even shown for levels `WARNING` and `ERROR` (it's only emitted for `INFORMATIONAL`)
* Have the test endpoint return an error if no test notification was emitted (this hints at the alert configuration being ineffective)

valentijnscholten:
It sounds a bit overly complex to me, but maybe there just needs to be a button on the SMTP config screen to send a test e-mail because that was my main goal that I wanted to make sure emails are/can be delivered.

Issue #6852, opened 2026-07-30
Support CISA 2026 SBOM Minimum Elements in CycloneDX exports


mihaimacarie98:
## Current Behavior

On July 29, 2026, CISA and 17 partner agencies (incl. BSI, ANSSI, ACN, METI) published the
**2026 Minimum Elements for a Software Bill of Materials**, replacing the 2021 NTIA elements:
https://www.cisa.gov/resources-tools/resources/2026-minimum-elements-software-bill-materials-sbom

v5 already covers most of the 17 data fields (and the `?version=` export parameter resolves
the old v4 spec-version limitation). Remaining gaps in `GET /api/v1/bom/cyclonedx/project/{uuid}`:

| 2026 Element | Gap | Code reference |
|---|---|---|
| SBOM Generation Context (new) | `metadata.lifecycles` (CDX ≥1.5) is never populated. | `ModelConverter#createMetadata` |
| SBOM Version (new) | `serialNumber` is a fresh `UUID.randomUUID()` per export and `version` is hardcoded to `1`, so no export indicates a relationship to earlier iterations. | `CycloneDXExporter#create` |
| SBOM Author (major update) | Export uses `Metadata#setManufacture` — deprecated since CDX 1.6 in favor of `manufacturer`. There is also no instance-level "SBOM authoring organization" setting; author identity must be maintained per project. | `ModelConverter#createMetadata` |
| SBOM Author Signature (new) | No signing support in the export path. | — |
| Explicitly Identifying Unknown Information (major update) | Components without `supplier`, `hashes`, or `licenses` have those fields silently omitted. The 2026 practice requires explicitly distinguishing unknown from withheld (e.g. unknown-provenance marker for supplier, `NOASSERTION` for licenses). | `ModelConverter#convert(Component)` |
| Coverage (major update) | Ingestion de-duplicates by `ComponentIdentity` (purl/cpe/swid/group/name/version), collapsing multiple instances of a component that differ only in other metadata (hashes, licenses, scope). The 2026 Coverage element requires listing such instances separately with their own dependency relationships. Data is lost at ingest and cannot be reconstructed on export. | `ImportBomActivity` ("One identity can point to multiple BOM refs, due to component and service de-duplication.") |

## Proposed Behavior

- [ ] Populate `metadata.lifecycles` on export (configurable; sensible default `post-build`/`operations` for an inventory export).
[...]

PR #6935, opened 2026-08-10
Add SBOM manufacturer during export

### LINKED ISSUE #3935: Populate Manufacturer field in downloaded SBOM's
### Current Behavior

Property is not included

### Proposed Behavior

When a SBOM is generated by Dependency track the manufacturer property as defined in (https://github.com/CycloneDX/specification/blob/62a669075f1897193a14060e0784e6a7576b693d/schema/bom-1.6.schema.json#L677) should be populated with information about OWASP etc.

### Checklist

Zuplyx:
### Description

This adds support for the SBOM manufacturer field (author of the SBOM).
If an imported SBOM contains the manufacturer field, it is persisted.
Additionally a new instance setting can be used to provide a default manufacturer, which is used if the imported SBOMs do not contain a manufacturer.

### Addressed Issue

fixes #3935
related to #6852 

### Additional Details
 
Note that the CycloneDX manufacturer field is only available from v1.6 onward, therefore the manufacturer is not included in the SBOM downloaded via the UI (hardcoded to v1.5).
If requested it would also be possible to enhance this feature with a per-project manufacturer setting, so projects could deviate from the instance default or override the imported SBOM's manufacturer.

Related PRs: DependencyTrack/frontend#1737 DependencyTrack/docs#215

### Checklist

- [X] I have read and understand the [contributing guidelines]
- [X] This PR introduces changes to the database model, and I have updated the [migration changelog] accordingly
- [X] This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly
~- [ ] This PR is a substantial change (per the [ADR criteria]), and I have added an [ADR] under `docs/adr/`~

[ADR]: ../docs/adr/


apiserver/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
```diff
@@ -867,9 +864,57 @@ public static org.cyclonedx.model.Metadata createMetadata(final Project project)
...
+     * <p>
+     * A manufacturer captured from a previously imported BOM for this project
+     * takes precedence over the system-wide default configured in Admin > General.
+     */
+    private static OrganizationalEntity resolveBomManufacturer(final Project project, final QueryManager qm) {
+        if (project != null && project.getMetadata() != null && project.getMetadata().getManufacturer() != null) {
+            return project.getMetadata().getManufacturer();
+        }
+        return manufacturerFromConfig(qm);
+    }
```
nscuro, line 887: Always falling back to a global default is not safe.

This assumes all projects in the portfolio are your own, which in many cases they aren't. For example if you consume BOMs from your vendors, or you monitor other OSS projects, and so on.

Even if all projects were strictly internal, you could still have multiple sub-organizations / departments with different logical business entities, also making a single global default impractical.

Regardless of the global default, import and export of manufacturers data is something we should support. So perhaps it's worth decoupling the two things.

migration/src/main/resources/org/dependencytrack/migration/V202608101126__add_project_metadata_manufacturer.sql
nscuro, line 1: Should use `JSONB` instead of `TEXT`.

nscuro (changes requested):

apiserver/src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
Zuplyx: But does it really matter who owns the project? The manufacturer in question here is the manufacturer of the SBOM itself and not the project/component described by the SBOM. In my opinion a global default makes sense in this case, since whoever operates the DependencyTrack instance is de-facto the manufacturer of all SBOMs exported from it.


Alternatively, I could also make the SBOM manufacturer configurable at a project level, maybe in the Identiy tab. This way individual projects could overwrite the global default, without having to alter the SBOMs they import.

Zuplyx: @nscuro Did you have time to consider my suggestions? How should I proceed with this PR?

PR #6976, opened 2026-08-13
Log child projects deleted with their parent

### LINKED ISSUE #6975: Log child projects that are deleted together with their parent
### Current Behavior

When a project is deleted, its child projects are deleted as well. Only the requested project is recorded in the audit log.

### Proposed Behavior

The audit log should also record child projects that are removed as a result of deleting a parent, so it is clear which projects were deleted and why.

### Checklist

ElenaStroebele:
### Description

When a project is deleted, its children are now recorded in the security audit log as well, not only the project that was explicitly requested.

### Addressed Issue

Closes #6975

### Additional Details
For batch delete, projects that were explicitly requested keep the existing “Deleted project” log and are not also logged as cascade children. When several requested projects sit on the same path, each descendant is attributed to the closest requested ancestor.

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
+++ b/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
@@ -760,10 +760,10 @@ public interface ProjectDao extends SqlObject, PaginationSupport {
     List<DeletedProjectRow> deleteProjects(@Bind Collection<UUID> projectUuids);
 
     record DeletedProjectRow(
-            @ColumnName("UUID") UUID uuid,
-            @ColumnName("NAME") String name,
-            @ColumnName("VERSION") String version,
-            @ColumnName("ANCESTOR_UUID") @Nullable UUID ancestorUuid) {
+            UUID uuid,
+            String name,
+            @Nullable String version,
+            @Nullable UUID ancestorUuid) {
     }
 
     @SqlQuery("""
--- a/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
+++ b/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
@@ -737,6 +737,44 @@ public interface ProjectDao extends SqlObject, PaginationSupport {
     @GetGeneratedKeys
     Set<UUID> deleteProjects(@Bind Collection<UUID> projectUuids);
 
+    /**
+     * Projects that will be removed by {@code ON DELETE CASCADE} when the given
+     * ancestors are deleted. When multiple requested ancestors are on the same
+     * path, each descendant is attributed to the closest one ({@code DEPTH} ascending).
+     */
+    @SqlQuery("""
+            SELECT DISTINCT ON (descendant."ID")
+                   descendant."UUID" AS "UUID"
+                 , descendant."NAME" AS "NAME"
+                 , descendant."VERSION" AS "VERSION"
+                 , ph."DEPTH" AS "DEPTH"
+                 , ancestor."UUID" AS "ANCESTOR_UUID"
+                 , ancestor."NAME" AS "ANCESTOR_NAME"
+                 , ancestor."VERSION" AS "ANCESTOR_VERSION"
+              FROM "PROJECT_HIERARCHY" AS ph
+             INNER JOIN "PROJECT" AS ancestor
+                ON ancestor."ID" = ph."PARENT_PROJECT_ID"
+             INNER JOIN "PROJECT" AS descendant
+                ON descendant."ID" = ph."CHILD_PROJECT_ID"
+             WHERE ancestor."UUID" = ANY(:ancestorUuids)
+               AND ph."DEPTH" > 0
+             ORDER BY descendant."ID"
+                    , ph."DEPTH"
+                    , ancestor."ID"
+            """)
+    @RegisterConstructorMapper(ProjectDeletedAsDescendant.class)
+    List<ProjectDeletedAsDescendant> getProjectsDeletedAsDescendants(@Bind Collection<UUID> ancestorUuids);
+
+    record ProjectDeletedAsDescendant(
+            @ColumnName("UUID") UUID uuid,
+            @ColumnName("NAME") String name,
+            @ColumnName("VERSION") String version,
+            @ColumnName("DEPTH") int depth,
+            @ColumnName("ANCESTOR_UUID") UUID ancestorUuid,
+            @ColumnName("ANCESTOR_NAME") String ancestorName,
+            @ColumnName("ANCESTOR_VERSION") String ancestorVersion) {
+    }
+
     @SqlQuery("""
              WITH "CTE" AS (
                SELECT "ID"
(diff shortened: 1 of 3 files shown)


apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -737,6 +737,44 @@ WITH cte_locked AS (
...
+             INNER JOIN "PROJECT" AS descendant
+                ON descendant."ID" = ph."CHILD_PROJECT_ID"
+             WHERE ancestor."UUID" = ANY(:ancestorUuids)
+               AND ph."DEPTH" > 0
+             ORDER BY descendant."ID"
+                    , ph."DEPTH"
+                    , ancestor."ID"
+            """)
+    @RegisterConstructorMapper(ProjectDeletedAsDescendant.class)
+    List<ProjectDeletedAsDescendant> getProjectsDeletedAsDescendants(@Bind Collection<UUID> ancestorUuids);
```
nscuro: It would make sense to combine this select with the delete itself, something along the lines of:

```sql
WITH
cte_locked AS (
  SELECT "ID"
    FROM "PROJECT"
   WHERE ${apiProjectAclCondition}
     AND "UUID" = ANY(:projectUuids)
   ORDER BY "ID"
     FOR UPDATE
),
cte_deleted AS (
  DELETE
    FROM "PROJECT"
   WHERE "ID" IN (SELECT "ID" FROM cte_locked)
  RETURNING "ID"
)
SELECT DISTINCT ON (project."ID")
       project."UUID" AS uuid
     , project."NAME" AS name
     , project."VERSION" AS version
     , CASE WHEN ph."DEPTH" = 0 THEN NULL ELSE ancestor."UUID" END AS ancestor_uuid
  FROM "PROJECT_HIERARCHY" AS ph
 INNER JOIN "PROJECT" AS ancestor
    ON ancestor."ID" = ph."PARENT_PROJECT_ID"
 INNER JOIN "PROJECT" AS project
    ON project."ID" = ph."CHILD_PROJECT_ID"
 WHERE ph."PARENT_PROJECT_ID" IN (SELECT "ID" FROM cte_deleted)
 ORDER BY project."ID"
        , ph."DEPTH"
```

(`cte_locked`'s `ORDER BY "ID"` ensures that locks are acquired in deterministic order so concurrent queries won't deadlock)

This gives you:

* deletion of multiple projects like before,
* ACL enforcement, and
* a row for each deleted project with `uuid`, `name`, `version`, and `ancestor_uuid` (`null` when root project).

You also no longer need `getAccessibleProjectUuids` to determine accessible projects before calling this query. `cte_locked` already does that filtering.

apiserver/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -1040,11 +1040,18 @@ public Response deleteProject(
                 }
                 requireAccess(qm, project);
 
+                final List<ProjectDao.ProjectDeletedAsDescendant> deletedAsDescendants = withJdbiHandle(
+                        getAlpineRequest(),
+                        handle -> handle.attach(ProjectDao.class)
+                                .getProjectsDeletedAsDescendants(Set.of(project.getUuid())));
```
nscuro: Don't call JDBI in a context that already has opened a DB transaction using the JDO API (`QueryManager#runInTransaction` etc.). Starting a transaction keeps the backing DB connection open, and JDBI doesn't share that connection, so you'd have two connections open at the same time.

If you need to call JDBI methods here, convert the full method to no longer use `QueryManager`. You can see how that may look in other methods that exclusively use the JDBI API for reference.

nscuro (changes requested):

apiserver/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -1022,34 +1022,35 @@ private <T> boolean setIfDifferent(final Project source, final Project target, f
...
+            if (deletedProject.ancestorUuid() != null) {
+                continue;
+            }
+            try (var _ = MDC.putCloseable(MDC_PROJECT_UUID, deletedProject.uuid().toString());
+                 var _ = MDC.putCloseable(MDC_PROJECT_NAME, deletedProject.name());
+                 var _ = MDC.putCloseable(MDC_PROJECT_VERSION, String.valueOf(deletedProject.version()))) {
+                LOGGER.info(SecurityMarkers.SECURITY_AUDIT,
+                        "Project {} deletion request by {}",
+                        formatDeletedProject(deletedProject),
+                        super.getPrincipal().getName());
```
nscuro: The message is now emitted *after* the project got deleted (which is correct), but the message still says "deletion request". In the bulk-delete endpoint it correctly says "deleted project". Let's get them aligned.

apiserver/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -1070,15 +1071,41 @@ public Response deleteProject(
...
-                handle -> handle.attach(ProjectDao.class).deleteProjects(uuids));
-        for (final UUID uuid : deletedProjectUuids) {
-            LOGGER.info(SecurityMarkers.SECURITY_AUDIT, "Deleted project {}", uuid);
+        final List<ProjectDao.DeletedProjectRow> deletedProjects = inJdbiTransaction(getAlpineRequest(), handle ->
+                handle.attach(ProjectDao.class).deleteProjects(uuids));
+        for (final ProjectDao.DeletedProjectRow deletedProject : deletedProjects) {
+            if (deletedProject.ancestorUuid() != null) {
+                continue;
+            }
+            LOGGER.info(SecurityMarkers.SECURITY_AUDIT, "Deleted project {}", deletedProject.uuid());
```
nscuro: MDC is not populated here, but is for the descendants. Should be uniform for all.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -720,22 +719,52 @@ OR EXISTS (
...
             """)
-    @GetGeneratedKeys
-    Set<UUID> deleteProjects(@Bind Collection<UUID> projectUuids);
+    @RegisterConstructorMapper(DeletedProjectRow.class)
+    List<DeletedProjectRow> deleteProjects(@Bind Collection<UUID> projectUuids);
+
+    record DeletedProjectRow(
+            @ColumnName("UUID") UUID uuid,
+            @ColumnName("NAME") String name,
+            @ColumnName("VERSION") String version,
```
nscuro: Nit: `version` is also nullable.

```suggestion
            @Nullable String version,
```

The `@ColumnName` annotations are also not necessary, JDBI can map `ANCESTOR_UUID` to `ancestorUuid` etc.

apiserver/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -1070,15 +1071,41 @@ public Response deleteProject(
...
+            try (var _ = MDC.putCloseable(MDC_PROJECT_UUID, deletedProject.uuid().toString());
+                 var _ = MDC.putCloseable(MDC_PROJECT_NAME, deletedProject.name());
+                 var _ = MDC.putCloseable(MDC_PROJECT_VERSION, String.valueOf(deletedProject.version()))) {
+                LOGGER.info(SecurityMarkers.SECURITY_AUDIT,
+                        "Project {} deleted as child of {}",
+                        formatDeletedProject(deletedProject),
+                        deletedProject.ancestorUuid());
+            }
+        }
+    }
```
nscuro: Consider adjusting this so it works for both ancestors and descendants:

```
for deletedProject in deletedProjects:
  # ... populate MDC ...
  if deletedProject.ancestorUuid == null:
    log("Deleted project {}", ...)
  else:
    log("Deleted project {} as descendant of {}", ...)
```

You can then also reuse it for both endpoints and can directly pass `ProjectDao#deleteProjects` output to it.

nscuro (changes requested):

nscuro (approved):

Issue #6991, opened 2026-08-14
Project create team picker only shows the first 100 teams


ElenaStroebele:
### Current Behavior

When creating a project in admin mode, team picker only shows the first 100 teams and not all. Hidden projects can further not be assigned when typing them in.

### Steps to Reproduce

1. Enable portfolio access control.
2. Create more than 100 teams.
3. Sign in as a user who is allowed to manage access and create projects.
4. Open Create Project.
5. Open the team field and look for a team whose name sorts after the first 100 team names (for example a name starting with Z).

### Expected Behavior

The team field lists every team the user is allowed to assign. Any of those teams can be selected and is applied to the new project.

### Dependency-Track Version

5.x

### Browser

N/A

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this defect was already reported

PR #6992, opened 2026-08-14
Fix project-create team picker stopping at 100 teams.

ElenaStroebele:
### Description
The API that lists teams for project creation now supports search and paging, so the new team picker can find teams without loading the full list.

### Addressed Issue
#6991 

### Additional Details
Frontend PR: [https://github.com/DependencyTrack/frontend/pull/1755](https://github.com/DependencyTrack/frontend/pull/1755)

### Checklist

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
+++ b/apiserver/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
@@ -27,6 +27,7 @@ import alpine.security.InvalidApiKeyFormatException;
 import alpine.server.auth.PermissionRequired;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.enums.ParameterIn;
 import io.swagger.v3.oas.annotations.headers.Header;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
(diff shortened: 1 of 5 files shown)


apiserver/src/main/java/org/dependencytrack/resources/v1/TeamResource.java
```diff
@@ -260,8 +260,7 @@ public Response availableTeams() {
                             || super.hasPermission(Permissions.Constants.ACCESS_MANAGEMENT_READ);
             List<Team> teams = new ArrayList<>();
             if (isAllTeams) {
-                var paginatedResult = qm.getTeams();
-                teams = paginatedResult.getList(Team.class);
+                teams = qm.getAllTeams();
```
nscuro: I think the more efficient and ergonomic fix would be to handle the `searchText` query parameter, such that clients can filter the result set by name. This is how the project parent selection also works: https://github.com/DependencyTrack/frontend/blob/2678e31e6b68c8710db4e542ec508a9a284e7a35/src/views/portfolio/projects/ProjectCreateProjectModal.vue#L447-L457

The API shouldn't return unbounded sets of data. Or at least it shouldn't _introduce_ this behavior anymore, since some legacy endpoints don't handle pagination consistently.

nscuro (changes requested):

nscuro (approved):
Thanks!

PR #6995, opened 2026-08-14
Fix POLICY_VIOLATION notification templates referencing nonexistent p…

### LINKED ISSUE #6994: POLICY_VIOLATION notifications fail with HTTP 400: default templates reference policyCondition, but v5 subject field is condition
### Current Behavior

On v5.0.x, notifications for the `POLICY_VIOLATION` group are never delivered via the Slack publisher — both real violations and the rule "Test" button fail. The API server logs a terminal publish failure:

```
WARN [ActivityTaskWorker] Activity failed terminally after 1 attempt(s) [activityName=publish-notification, ...]
org.dependencytrack.dex.api.failure.TerminalApplicationFailureException: Failed to publish notification with non-retryable cause
Caused by: java.lang.IllegalStateException: Request failed with unexpected response code: 400
	at org.dependencytrack.notification.publishing.http.AbstractHttpNotificationPublisher.publish(AbstractHttpNotificationPublisher.java:68)
```

The same webhook URL accepts a minimal `{"text":"hi"}` payload via curl (HTTP 200), so the webhook itself is fine.

### Steps to Reproduce

1. Create a notification rule: scope `PORTFOLIO`, group `POLICY_VIOLATION`, publisher Slack, destination a valid `https://hooks.slack.com/services/...` webhook URL
2. Click **Test** on the rule (or trigger a real policy violation via BOM upload)
3. Nothing arrives in Slack; the API server logs the 400 failure above
4. `curl -X POST -H 'Content-type: application/json' --data '{"text":"hi"}' <same webhook URL>` returns 200

### Expected Behavior

The templates reference `subject.policyViolation.condition.*` so the rendered payload is valid and POLICY_VIOLATION notifications are delivered.

### Dependency-Track Version

5.x

### Browser

N/A

### Checklist

fffinkel:
### Description

Notifications for the `POLICY_VIOLATION` group are never delivered by the Slack publisher (and render incorrectly in the MS Teams, Mattermost, and email publishers). The v5 notification proto names the `PolicyViolation` condition field `condition`, but the default templates still reference `policyCondition` (the v4 name). Pebble resolves the missing attribute to null and renders empty strings; Slack rejects payloads containing empty `plain_text` blocks with HTTP 400 (`invalid_blocks`), the publisher treats 400 as non-retryable, and the notification is dropped. This affects both real policy violations and the rule "Test" button.

This PR replaces `subject.policyViolation.policyCondition` with `subject.policyViolation.condition` in the Slack, MS Teams, Mattermost, and email default templates, and adds `POLICY_VIOLATION` payload coverage to `SlackNotificationPublisherTest` so a regression would be caught.

### Addressed Issue

Fixes #6994

### Additional Details

- The email template's other two `policyCondition` references (inside the `NEW_POLICY_VIOLATIONS_SUMMARY` loop over `projectEntry.violationsList`) are intentionally left unchanged: the summary subject's `Violation` message does define a `policy_condition` field, so those resolve correctly.
- The new test pins the full rendered Block Kit payload for the `POLICY_VIOLATION` test notification, including the `View Project` / `View Component` action buttons.
- One pre-existing quirk was noticed but deliberately not addressed here to keep the change focused: the Slack template's `POLICY_VIOLATION` branch renders the project as `subject.project.toString` (a protobuf `TextFormat` dump) rather than `subject.project | summarize` as the other branches do. The output is valid, just verbose. Happy to fix that in a follow-up if desired.

### Checklist

[ADR]: ../docs/adr/


nscuro (approved):
Thanks!

Issue #1665, opened 2022-05-27
BOM upload fails without feedback due to field max length


ecaisse:
A BOM file that contains a component with a "publisher" field with more than 255 character fails due to the constraints of the field. However, there is no feedback or way to know that the BOM upload failed other than figuring it out from the log file.

### Current Behavior:
When uploading a broken BOM file (see bom-broken.xml in [bom-broken.zip](https://github.com/DependencyTrack/dependency-track/files/8788432/bom-broken.zip)), no components will be loaded, and there is no way to see that the BOM processing failed outside the log file. See "Additional Details" section for the stacktrace.

I manually created the BOM file to only include the broken component, however, this is how https://github.com/CycloneDX/cyclonedx-dotnet would generate the component for Hangfire.PostgreSql@1.9.6.

### Steps to Reproduce:
- Create a BOM file with a component that has a publisher value of more than 255 characters
-- Or use the bom-broken.xml in [bom-broken.zip](https://github.com/DependencyTrack/dependency-track/files/8788432/bom-broken.zip)
- Upload the BOM file to Dependency Track manually in a test project
- Note that no errors are raised, and no components are loaded

To test that the issue is with the publisher field, simply truncate the field in the XML file and reupload it to Dependency Track. The component should appear correctly in the Components tab.

### Expected Behavior:
I expect one of 2 behaviors:
1. The BOM processing should succeed and limits put on fields should be removed unless explicitly stated by the CycloneDX specification
2. There should be a way to see that the BOM processing failed in the frontend, especially when uploaded manually

I think option 1 is better than option 2. The restriction makes it a problem for using automation as there is no way to predict if a component will break Dependency Track. Since the specification does not contain restrictions on field lengths, Dependency Track should not enforce arbitrary ones.

### Environment:
  
  - Dependency-Track Version: 4.4.2 and 4.5.0
  - Distribution: Docker
  - BOM Format & Version: XML 1.3
  - Database Server: PostgreSQL (but should apply to all)
  - Browser: Chrome 100.0.4896.127

### Additional Details:
DT stacktrace
```
2022-05-27 16:37:25,861 [] ERROR [org.dependencytrack.tasks.BomUploadProcessingTask] Error while processing bom
[...]

PR #7127, opened 2026-08-26
Expose workflow run status on BOM token polling endpoint

Suhas2109:
### Description

Fixes the BOM upload token polling endpoints (`GET /v1/bom/token/{uuid}` and `GET /v1/event/token/{uuid}`) so they now expose the actual processing status of the associated workflow run instead of silently reporting only a boolean `processing` flag. The response now includes a `status` field using a dedicated, consumer-facing `IsTokenBeingProcessedResponse.Status` enum (`PENDING`, `RUNNING`, `COMPLETED`, `FAILED`), decoupled from the internal `WorkflowRunStatus` used by the workflow engine. Additionally, the underlying `listRuns` lookups by BOM upload token now specify an explicit sort order (`CREATED_AT` descending) to ensure deterministic results when multiple runs share the same label.

### Addressed Issue

Relates to #1665

### Additional Details

- Introduced `IsTokenBeingProcessedResponse.Status`, a dedicated enum mirroring the pattern used in `VulnDataSourceMirrorService.MirrorStatus.Status`, to avoid leaking internal `WorkflowRunStatus` semantics (and the word "workflow") into the public API surface.
- `Status.of(WorkflowRunStatus)` maps: `CREATED` / `SUSPENDED` -> `PENDING`, `RUNNING` -> `RUNNING`, `COMPLETED` -> `COMPLETED`, `CANCELLED` / `FAILED` -> `FAILED`.
- Added explicit `.withSortBy(CREATED_AT)` + `.withSortDirection(DESC)` to the `listRuns` calls in `BomResource` and `EventResource` for deterministic run resolution.
- Verified manually: a corrupted BOM upload yields `{"processing": false, "status": "FAILED"}`, and a valid BOM upload yields `{"processing": false, "status": "COMPLETED"}` once processing completes.
- Rebasing onto latest upstream `main` to resolve a merge conflict caused by a concurrent unrelated refactor of the same endpoint.

### Checklist

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
+++ b/apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
@@ -38,10 +38,11 @@ import org.cyclonedx.Version;
 import org.cyclonedx.exception.GeneratorException;
 import org.dependencytrack.auth.Permissions;
 import org.dependencytrack.auth.ProjectAccess;
+import org.dependencytrack.common.pagination.SortDirection;
 import org.dependencytrack.dex.engine.api.DexEngine;
-import org.dependencytrack.dex.engine.api.WorkflowRunStatus;
+import org.dependencytrack.dex.engine.api.WorkflowRunMetadata;
 import org.dependencytrack.dex.engine.api.request.CreateWorkflowRunRequest;
-import org.dependencytrack.dex.engine.api.request.ExistsWorkflowRunRequest;
+import org.dependencytrack.dex.engine.api.request.ListWorkflowRunsRequest;
 import org.dependencytrack.filestorage.api.FileStorage;
 import org.dependencytrack.filestorage.proto.v1.FileMetadata;
 import org.dependencytrack.model.BomValidationMode;
@@ -975,17 +976,21 @@ public class BomResource extends AbstractApiResource {
                     String uuid) {
         final UUID token = UUID.fromString(uuid);
 
-        final boolean isProcessing;
-        if (dexEngine.existsRun(new ExistsWorkflowRunRequest(
-                WorkflowRunStatus.NON_TERMINAL_STATUSES, Map.of(WF_LABEL_BOM_UPLOAD_TOKEN, token.toString())))) {
-            isProcessing = true;
-        } else {
-            final var runMetadata = dexEngine.getRunMetadataById(token);
-            isProcessing = runMetadata != null && !runMetadata.status().isTerminal();
-        }
+        final var page = dexEngine.listRuns(new ListWorkflowRunsRequest()
+                .withLabels(Map.of(WF_LABEL_BOM_UPLOAD_TOKEN, token.toString()))
+                .withSortBy(ListWorkflowRunsRequest.SortBy.CREATED_AT)
+                .withSortDirection(SortDirection.DESC)
+                .withLimit(1));
+        final WorkflowRunMetadata runMetadata =
+                !page.items().isEmpty() ? page.items().getFirst() : dexEngine.getRunMetadataById(token);
 
         final var response = new IsTokenBeingProcessedResponse();
-        response.setProcessing(isProcessing);
+        if (runMetadata != null) {
+            response.setProcessing(!runMetadata.status().isTerminal());
+            response.setStatus(IsTokenBeingProcessedResponse.Status.of(runMetadata.status()));
+        } else {
+            response.setProcessing(false);
+        }
         return Response.ok(response).build();
     }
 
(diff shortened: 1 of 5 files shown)


apiserver/src/main/java/org/dependencytrack/resources/v1/vo/IsTokenBeingProcessedResponse.java
```diff
@@ -29,11 +30,26 @@ public class IsTokenBeingProcessedResponse implements Serializable {
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
     private Boolean processing;
 
+    @Schema(nullable = true,
+            description = "The workflow run status associated with the token. "
+                    + "Null when no workflow run is associated with the token.")
```
nscuro: The fact that workflows are used internally is an implementation detail that shouldn't bleed into consumer APIs. Please remove the "workflow" wording from the description.

apiserver/src/main/java/org/dependencytrack/resources/v1/vo/IsTokenBeingProcessedResponse.java
```diff
@@ -29,11 +30,26 @@ public class IsTokenBeingProcessedResponse implements Serializable {
...
+    @Nullable
+    private String status;
```
nscuro: Should be an enum so consumers know what values to expect. Should **not** reuse `WorkflowRunStatus`, as that again would leak internals. Define a dedicated, simplified `enum` instead, like it's done here:

https://github.com/DependencyTrack/dependency-track/blob/50234414b980dc89719de43bbc4301718209444f/apiserver/src/main/java/org/dependencytrack/vulndatasource/VulnDataSourceMirrorService.java#L112-L128

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -881,19 +881,21 @@ public Response isTokenBeingProcessed(
...
-                        Map.of(WF_LABEL_BOM_UPLOAD_TOKEN, token.toString())))) {
-            isProcessing = true;
-        } else {
-            final var runMetadata = dexEngine.getRunMetadataById(token);
-            isProcessing = runMetadata != null && !runMetadata.status().isTerminal();
-        }
+        final var page = dexEngine.listRuns(
+                new ListWorkflowRunsRequest()
+                        .withLabels(Map.of(WF_LABEL_BOM_UPLOAD_TOKEN, token.toString()))
+                        .withLimit(1));
```
nscuro: Add explicit `withSortBy` and `withSortDirection` here so the result stays deterministic. Similar to:

https://github.com/DependencyTrack/dependency-track/blob/50234414b980dc89719de43bbc4301718209444f/apiserver/src/main/java/org/dependencytrack/vulndatasource/VulnDataSourceMirrorService.java#L139-L144

apiserver/src/main/java/org/dependencytrack/resources/v1/EventResource.java
```diff
@@ -97,30 +96,27 @@ public Response isTokenBeingProcessed(
...
-                dexEngine.existsRun(new ExistsWorkflowRunRequest(
-                        WorkflowRunStatus.NON_TERMINAL_STATUSES,
-                        Map.of(WF_LABEL_BOM_UPLOAD_TOKEN, token.toString())));
-        if (hasNonTerminalByLabel) {
-            return true;
+    private WorkflowRunMetadata resolveRunByToken(final UUID token) {
+        final var page = dexEngine.listRuns(
+                new ListWorkflowRunsRequest()
+                        .withLabels(Map.of(WF_LABEL_BOM_UPLOAD_TOKEN, token.toString()))
+                        .withLimit(1));
```
nscuro: Same as in `BomResource`, add explicit sorting details for determinism.

nscuro (changes requested):
Thanks for the PR! A few changes please.

Also, make sure to follow the [PR template](https://raw.githubusercontent.com/DependencyTrack/dependency-track/refs/heads/main/.github/PULL_REQUEST_TEMPLATE.md). The corresponding build check is currently failing.

Suhas2109:
Hi @nscuro ,
Thanks for the detailed review! Will address all points and will push updates shortly.

Suhas2109:
Hi @nscuro !
I've pushed the requested changes.
Let me know if anything else needs adjusting.

apiserver/src/main/java/org/dependencytrack/resources/v1/vo/IsTokenBeingProcessedResponse.java
nscuro, line 1: Add `@JsonInclude(JsonInclude.Include.NON_NULL)` to the class to resolve the remaining test failures:

```
[ERROR]   BomResourceTest.shouldReportTokenNotBeingProcessed:2710 JSON documents are different:
Different keys found in node "", extra: "status", expected: <{"processing":false}> but was: <{"processing":false,"status":null}>

[ERROR]   EventResourceTest.isTokenBeingProcessedNotExistsTest:181 JSON documents are different:
Different keys found in node "", extra: "status", expected: <{"processing":false}> but was: <{"processing":false,"status":null}>
```

nscuro (approved):
Thanks!

FYI I replaced the "Fixes" with "Relates to" for #1665 because this change is API-only, whereas the issue was asking for making upload status observable by users:

> There should be a way to see that the BOM processing failed in the frontend, especially when uploaded manually

Issue #7141, opened 2026-08-27
Support for oauth2-proxy


phmcder:
### Current Behavior

For our system we authenticate all connections via oauth2-proxy before they reach the backed. This give us a simple pane of glass when it comes to IT health checks. All authentication for apps that support proxy authentication are done in the same way and we can apply rules to not allow any API traffic without authentication

When is comes to application such a dependency track which support the own implementation of OIDC we need to do much more due dilligance to make sure the service complies with our security standards. We have to allow unauthenticated access to the dependency track apis which make our security team nervous.

### Proposed Behavior

Allow Dependency track to support oauth2-proxy so that a user, once authenticate with the gateway, does not have to do any more actions to access the dashboard.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

nscuro:
Anyone interested in this, can you kindly have a look at the proposed ADR so we can flesh out how you expect an implementation to work: #7152

Issue #7056, opened 2026-08-23
Support workload identity federation as alternative to API keys


nscuro:
### Current Behavior

Currently, non-human clients must authenticate via long-lived API keys. This is both risky (leaked keys stay valid until manually revoked), and cumbersome (manual minting, distribution, rotation).

The industry is converging on tackling this problem using [workload identity federation](https://nhimg.org/glossary/workload-identity-federation/) (WIF):

> Workload identity federation is the practice of letting a workload prove who it is in one trust domain and receive a short-lived credential in another, without storing a long-lived password, API key, or shared secret. In NHI operations, that makes it a control pattern for reducing credential sprawl across cloud, Kubernetes, CI/CD, and hybrid environments.

Some high-assurance environments are actively banning use of long-lived credentials and require tedious exception processes if WIF is not available.

### Proposed Behavior

We should support WIF to make it easier and more secure to integrate with us.

> [!WARNING]
> Requires service accounts (#7055) to be implemented first.

References:

* https://nhimg.org/glossary/workload-identity-federation/
* https://developers.openai.com/api/docs/guides/workload-identity-federation
* https://platform.claude.com/docs/en/manage-claude/workload-identity-federation
* https://docs.snowflake.com/en/user-guide/workload-identity-federation

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

Issue #7055, opened 2026-08-23
Support service / robot accounts


nscuro:
### Current Behavior

Currently, non-human clients must authenticate via API keys. API keys are bound to teams, which makes it hard to reason about and trace who exactly performed an action when said teams have multiple members.

### Proposed Behavior

Introduce a new principal type for non-human clients, i.e. service / robot accounts (final wording to be determined, I'll use the abbreviation SA below).

> [!NOTE]
> SAs are a precondition for implementing [workload identity federation](https://nhimg.org/glossary/workload-identity-federation/) support.

SAs should be assignable to teams like normal users. This is required to ensure compatibility with [PAC](https://dependencytrack.github.io/docs/next/concepts/access-control/#portfolio-access-control).

It must be possible to create API keys for SAs, not just for teams. This will allow API keys to be traced back to specific SAs instead of just vaguely "Team A".

References:

* https://docs.quay.io/glossary/robot-accounts.html
* https://goharbor.io/docs/2.15.0/working-with-projects/project-configuration/create-robot-accounts/

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/dependency-track/issues) for whether this enhancement was already requested

VinodAnandan:
Thanks for this, Niklas. I completely agree that service accounts are a much better fit than team-based concepts. Most organisations typically have service accounts or Non-Human Identity (NHI) accounts, and for backend integrations, it is particularly important to bind the identity to a specific service account/NHI. That said, I’m not sure whether service accounts should necessarily be modelled exactly the same way as human identity, authentication, access.

A few things I would kindly suggest considering:

- Service account principles: Provide the ability to either map or define service account principles according to an organisation’s internal naming standards and conventions.
- Stronger authentication methods: Support more secure service-account authentication mechanisms, particularly OAuth 2.0 client_credentials grant with private_key_jwt for first-party authentication ( https://medium.com/@phat.tan.nguyen/oauth-2-0-the-client-credentials-grant-type-with-keycloak-2debb88a1c70 )
- Federated client authentication: For federated client authentication, consider supporting mechanisms such as SPIFFE or OIDC federated/brokered authentication. There is some additional information on this approach here: [Keycloak – Federated Client Authentication](https://www.keycloak.org/2026/01/federated-client-authentication)

[...]

PR #7152, opened 2026-08-28
Propose OAuth2 Proxy authentication

jjj-n:
### Description

Proposes optional support for deployments where an OIDC-backed OAuth2 Proxy authenticates
incoming browser requests before they reach Dependency-Track.

The revised design treats the OAuth2 Proxy ID token as the credential. OAuth2 Proxy forwards it
in a separate `X-Forwarded-ID-Token` header, and Dependency-Track validates it through the
existing OIDC ID token path. Validation includes the issuer, signature, expiration, and audience.
A new `dt.auth-proxy.client-id` setting declares the expected audience and must match the
OAuth2 Proxy client identifier, so a token issued to another client of the same identity provider
cannot be exchanged for a Dependency-Track session.

Native OIDC continues to use `dt.oidc.client-id`. The native and proxy login paths may coexist
with different client registrations, but both must use the configured `dt.oidc.issuer` in the
first version.

The integration does not trust forwarded username, email, or group headers. It also does not
authenticate from a forwarded access token or rely on UserInfo as proof that a token was issued
for Dependency-Track. Users remain normal OIDC users and reuse existing provisioning, team
synchronization, permissions, and session behavior. The configured username claim, and the teams
claim when team synchronization is enabled, must be present in the ID token. A missing required
claim fails the exchange rather than synchronizing an incomplete profile.

Native Dependency-Track OIDC remains the default. This integration is limited to interactive
users in deployments that require the OAuth2 Proxy session to initiate the application session.
Service accounts and workload identity federation remain separate work in #7055 and #7056.

### Addressed Issue

Related to #7141

### Additional Details

The ADR compares five approaches:

- Use native Dependency-Track OIDC behind OAuth2 Proxy.
- Exchange trusted proxy identity headers for a Dependency-Track session.
- Validate an access token forwarded by OAuth2 Proxy.
- Validate an ID token forwarded by OAuth2 Proxy.
- Authenticate every API request directly from proxy headers.

It proposes the ID token exchange because it reuses Dependency-Track's existing OIDC model while
retaining cryptographic issuer, signature, expiration, and audience validation. The ID token uses
a separate header so `Authorization` remains available for the Dependency-Track session.

Implementation pull requests will follow after the ADR is accepted.

### Checklist

- [x] ~~This PR fixes a defect, and I have provided tests~~
- [x] ~~This PR implements an enhancement, and I have provided tests~~
- [x] ~~This PR introduces changes to the database model~~
- [x] ~~This PR introduces new or alters existing behavior~~
- [x] This PR is a substantial change, and I have added an ADR under `docs/adr/`

### DIFF
--- a/docs/adr/040-oauth2-proxy-authentication.md
+++ b/docs/adr/040-oauth2-proxy-authentication.md
@@ -129,9 +129,15 @@ same short-lived bearer session used by other login methods.
 
 The ID token must pass the standard checks defined by [OpenID Connect Core] and already performed
 by Dependency-Track. These include issuer, signature, expiration, and audience validation. A new
-`dt.oauth2-proxy.client-id` setting must match the client identifier used by OAuth2 Proxy. It is
-the expected audience for the forwarded ID token. A token issued to a different client of the
-same identity provider will be rejected.
+`dt.auth-proxy.client-id` setting must match the client identifier used by the reverse proxy that
+forwards the ID token. It is the expected audience for the forwarded ID token. A token issued to
+a different client of the same identity provider will be rejected.
+
+The setting is named generically because the exchange validates a forwarded ID token and does not
+depend on which reverse proxy provides it. OAuth2 Proxy is the reference implementation for the
+first version: other reverse proxies that can inject a bearer ID token into
+`X-Forwarded-ID-Token` should work with the same validation path, but only the OAuth2 Proxy
+configuration is documented and validated.
 
 Native Dependency-Track OIDC continues to use `dt.oidc.client-id`. The two login paths can be
 enabled together and can use different client registrations, but both registrations must use the
--- a/docs/adr/040-oauth2-proxy-authentication.md
+++ b/docs/adr/040-oauth2-proxy-authentication.md
@@ -10,24 +10,51 @@ server validates the tokens, creates or finds an OIDC user, and returns a Depend
 session.
 
 Some deployments put [OAuth2 Proxy] in front of every application. OAuth2 Proxy authenticates
-the browser before a request reaches the application. When it uses an OIDC provider, it can pass
-the access token or ID token to the upstream application. Starting another OIDC flow in the
-Dependency-Track frontend gives the user a second login step.
+the browser before a request reaches the application. These deployments use the proxy to apply
+the same access policy to every application. Starting another OIDC flow in the Dependency-Track
+frontend adds another redirect and makes the proxy session separate from the application login.
 
-OAuth2 Proxy can also pass the authenticated username, email, and groups in HTTP headers. These
+The existing Dependency-Track OIDC flow remains the default. It gives Dependency-Track its own
+OIDC client and does not depend on a proxy credential. The integration in this decision is only
+for deployments that require an OAuth2 Proxy session to initiate the application session.
+
+OAuth2 Proxy can pass the authenticated username, email, and groups in HTTP headers. These
 identity headers are not signed. A client can forge them if it can bypass the proxy, or if the
 proxy preserves values supplied by the client. Accepting them as credentials would make the
-deployment network part of the authentication mechanism.
+deployment network the only authentication boundary.
 
-The frontend and REST API use a Dependency-Track bearer session after login. Passing an OIDC ID
-token to the upstream in the `Authorization` header conflicts with this session because both use
-the same header. The browser also cannot read a header that the proxy adds to its request. A
-server-side exchange is therefore required even when the existing OIDC validation is reused.
+OAuth2 Proxy can also pass the provider access token or ID token to the upstream application.
+Calling the OIDC UserInfo endpoint proves that an access token is accepted by that endpoint. It
+does not prove that the token was issued for Dependency-Track. An access token issued to another
+client of the same identity provider could therefore be accepted by the existing UserInfo path.
+Using an access token as a Dependency-Track credential would require resource server validation,
+including a Dependency-Track audience. That is broader than reusing the existing OIDC login.
+
+An OIDC ID token has the client identifier in its audience. Dependency-Track already validates
+the ID token issuer, signature, expiration, and audience. The OAuth2 Proxy client identifier can
+therefore be configured as the Dependency-Track OIDC client identifier and used as a
+cryptographic boundary between clients of the same identity provider.
+
+The frontend and REST API use a Dependency-Track bearer session after login. OAuth2 Proxy's
+stable direct-upstream option puts the ID token in the `Authorization` header. This conflicts
+with the Dependency-Track session because both use the same header. The browser also cannot read
+a header that the proxy adds to its request. A separate header and a server-side exchange are
+therefore required.
 
 This decision addresses interactive users opening the Dependency-Track dashboard. Authentication
 for service and robot accounts is tracked separately in [issue 7055] and [issue 7056].
 
 ### Possible Solutions
+
+#### Use native Dependency-Track OIDC behind OAuth2 Proxy
+
+OAuth2 Proxy can authenticate the gateway request, then the frontend can start the existing
+Dependency-Track OIDC authorization flow. An existing identity provider session can complete the
+second flow without another credential prompt.
+
+This keeps an independent application authentication boundary and uses only standard OIDC
+behavior. It still requires a Dependency-Track browser flow, redirect URI, and client
+configuration. It also does not use the OAuth2 Proxy session as the application login event.
 
 #### Exchange trusted identity headers for a Dependency-Track session
 
@@ -41,23 +68,26 @@ user provisioning and team synchronization behavior.
 #### Validate an access token forwarded by OAuth2 Proxy
 
 OAuth2 Proxy can pass the provider access token in the `X-Forwarded-Access-Token` header. The API
-server can use the existing OIDC UserInfo validation, user provisioning, team synchronization,
-and session creation behavior.
+server can call the existing OIDC UserInfo endpoint and reuse the returned user claims.
 
-This does not trust plain identity headers and does not require another user type. OAuth2 Proxy's
-existing client registration can be reused. However, the provider must support OIDC, and the API
-server must be able to access OIDC discovery and the UserInfo endpoint.
+This does not prove that the access token is intended for Dependency-Track. Correct resource
+server support must validate a Dependency-Track audience, signature, issuer, token type, and
+expiration as described by the [OAuth 2.0 JWT access token profile]. It must also define scopes
+for API authorization. These requirements apply to both human and machine clients and are
+tracked separately.
 
 #### Validate an ID token forwarded by OAuth2 Proxy
 
-The API server can validate a forwarded ID token with the existing OIDC issuer, audience, and
-signature checks. This avoids a UserInfo request when the token contains all required claims.
+The API server can validate a forwarded ID token with the existing OIDC issuer, audience,
+signature, and expiration checks. The audience binds the token to the configured OAuth2 Proxy
+client instead of only trusting the network path. The resulting identity can reuse the normal
+OIDC user, provisioning, team synchronization, and session behavior.
 
-OAuth2 Proxy's stable direct-upstream option sends the ID token in the `Authorization` header.
-This would overwrite the Dependency-Track session on later requests. A separate header requires
-structured OAuth2 Proxy configuration or an additional reverse proxy. Access token refresh also
-does not guarantee that the stored ID token is replaced. This makes the option less portable for
-the first version.
+The ID token must use a separate header so it does not replace the Dependency-Track session in
+`Authorization`. Direct-upstream deployments need OAuth2 Proxy's structured header injection.
+NGINX `auth_request` deployments can remap the ID token from the authentication response. The ID
+token must contain every claim required by the configured username and team synchronization
+settings.
 
 #### Authenticate every API request from proxy headers
 
@@ -70,7 +100,7 @@ harder to define.
 
 ## Decision
 
-We will validate an access token forwarded by OAuth2 Proxy and exchange it for a normal
+We will validate an ID token forwarded by OAuth2 Proxy and exchange it for a normal
 Dependency-Track session. The integration will only support OAuth2 Proxy deployments backed by
 an OIDC provider. It will only authenticate interactive users.
 
@@ -79,13 +109,13 @@ sequenceDiagram
     participant Browser
     participant Proxy as OAuth2 Proxy
     participant API as Dependency-Track API
-    participant IdP as OIDC UserInfo
+    participant IdP as OIDC Provider
 
     Browser->>Proxy: Request dashboard
     Proxy->>Proxy: Authenticate browser
-    Proxy->>API: Session exchange with forwarded access token
-    API->>IdP: Validate token and load claims
-    IdP-->>API: OIDC user claims
+    Proxy->>API: Session exchange with forwarded ID token
+    API->>IdP: Load OIDC discovery and signing keys
+    API->>API: Validate issuer, signature, expiry, and audience
     API-->>Proxy: Dependency-Track session
     Proxy-->>Browser: Dependency-Track session
     Browser->>Proxy: API request with Dependency-Track session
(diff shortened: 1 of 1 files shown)


docs/adr/040-oauth2-proxy-authentication.md
```diff
@@ -0,0 +1,107 @@
...
+The API server will extend the existing version 1 login API with a session exchange endpoint.
+This is an extension of the existing login contract, not a new authentication method for every
+REST endpoint. The endpoint will have these properties:
+
+* It will only accept identity when proxy authentication is enabled.
+* It will read the user and email values from configurable header names. The defaults will
+  match the `X-Forwarded-*` request headers emitted by OAuth2 Proxy.
+* It will reject a request when the user header is missing or empty.
+* It will create or find a persistent proxy user. A username that already belongs to another
+  user type will be rejected.
```
nscuro: How will we make sure these headers are not forged, e.g. by clients bypassing oauth2-proxy? [Grafana supports whitelisting IPs / IP ranges to trust](https://grafana.com/docs/grafana/latest/setup-grafana/configure-access/configure-authentication/auth-proxy/), but maybe there are more options we should consider? What do the other products you integrate oauth2-proxy with do?

docs/adr/040-oauth2-proxy-authentication.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### Validate a token forwarded by OAuth2 Proxy
+
+OAuth2 Proxy can forward an access token or OIDC ID token. Dependency-Track can validate that
+token with its existing OIDC support before it creates a session.
+
+This option does not trust plain identity headers. However, it only works for providers and
+tokens that Dependency-Track can validate through OIDC. It may require another client
+registration, access to the identity provider, and provider-specific claim configuration. It
+does not solve the general proxy authentication use case.
```
nscuro: The existing OIDC support already works by accepting ID/access tokens from the IdP, and validating them either via JWK keyset (ID token) or via the IdP's `/userinfo` endpoint (access token).

If oauth2-proxy were to forward either of these tokens, we could just use the existing OIDC infrastructure. Is there anything else that makes this option impractical that is not currently listed here?

docs/adr/040-oauth2-proxy-authentication.md
```diff
@@ -0,0 +1,107 @@
...
+the selected identity headers, not preserve values sent by clients. They must also prevent
+clients from reaching the API server without the gateway. The configuration documentation will
+include examples for direct upstream mode and NGINX `auth_request` mode.
+
+## Consequences
+
+* Users authenticated by OAuth2 Proxy can open the dashboard without a second login action.
+* OAuth2 Proxy can use OIDC or another provider. Dependency-Track does not need to validate or
+  store the provider's token.
+* Existing API key clients and Dependency-Track sessions keep their current behavior.
```
nscuro: @phmcder, does excluding non-human clients from this functionality fulfill your requirement? From your issue I'm not 100% sure whether this needs to also work for machine clients that are currently using API keys.

If that is a requirement, please have a look at #7055 and #7056, since they target a similar problem.

docs/adr/040-oauth2-proxy-authentication.md
```diff
@@ -0,0 +1,107 @@
...
+  match the `X-Forwarded-*` request headers emitted by OAuth2 Proxy.
+* It will reject a request when the user header is missing or empty.
+* It will create or find a persistent proxy user. A username that already belongs to another
+  user type will be rejected.
+* It will issue the same short-lived bearer session used by other login methods. Later API
+  requests will use that session and the existing authorization filters.
+* It will not accept passwords, access tokens, or ID tokens from the browser.
+
+Proxy users will be a distinct user type in the shared user model. This keeps their origin clear
+and avoids treating an unverified proxy assertion as an OIDC identity. Proxy users will use the
```
nscuro: Is there any scenario in which you'd have oauth2-proxy *and* OIDC enabled, assuming option 2 really is not viable? In that case, why differentiate the two? A new user type adds considerable maintenance burden.

docs/adr/040-oauth2-proxy-authentication.md
jjj-n: Thanks, this is the main weakness of the plain-header option. I checked the official integration
guidance: Grafana protects plain auth-proxy headers with a trusted IP/CIDR allowlist, while the
OAuth2 Proxy integrations for Headlamp and Kubernetes Dashboard forward an OIDC token for the
downstream to validate.

I revised the ADR to use the latter pattern. Dependency-Track will validate a forwarded access
token through its existing OIDC UserInfo path and will not accept forwarded username or email
headers as credentials. The proxy must still replace client-supplied token headers and the API
server must not be reachable around the proxy, but forging an identity header alone will no
longer authenticate a user. A generic plain-header mode is now outside the proposed scope.

docs/adr/040-oauth2-proxy-authentication.md
jjj-n: You're right. After checking both the existing OIDC implementation and OAuth2 Proxy's forwarding
behavior, this option is viable for OIDC-backed deployments and is now the selected design.

The revised ADR uses `X-Forwarded-Access-Token` by default and reuses the existing UserInfo
validation, OIDC user provisioning, team synchronization, and session creation. It does not need
a second client registration. Dependency-Track still needs access to OIDC discovery and UserInfo,
which is now listed as a limitation.

A small server-side exchange endpoint is still needed because browser JavaScript cannot read a
request header injected by the proxy. I did not select the ID token for the first version because
OAuth2 Proxy's stable direct-upstream option puts it in `Authorization`, which would overwrite the
Dependency-Track session bearer token on later requests.

docs/adr/040-oauth2-proxy-authentication.md
jjj-n: I have scoped the proposed decision to interactive dashboard users and made the limitation
explicit. It does not add a new authentication mechanism for machine clients.

Dependency-Track's API key validation remains unchanged, but OAuth2 Proxy may reject a machine
request before it reaches the API server. If the requirement is that machine traffic must also
authenticate at the gateway without an API-key bypass, I agree that #7055 and #7056 are the
appropriate foundation. This point still needs confirmation from @phmcder.

docs/adr/040-oauth2-proxy-authentication.md
jjj-n: Agreed. With the forwarded-token design there is no reason to introduce another user type. I
removed the proxy-specific user, discriminator, migration, and management API from the decision.

The exchange will reuse the normal OIDC user, including existing provisioning and team
synchronization. The API server can have OIDC validation enabled for the forwarded token while
the frontend OIDC client settings remain empty, so the frontend does not offer or start a second
OIDC login flow.

docs/adr/040-oauth2-proxy-authentication.md
phmcder: I agree with @jjj-n that header spoofing is an issue but is one that has exisiting mitigation patterns.

At the Network Gateway and the IDP itself rules should be in place reject any incomming headers that match those that they output for user identification. These are normally applied by default.

Withing the a kubernetes cluster you are more open to unauthenticated access by hosted servcies. This is mitigated by using a networking service which provide an authentication layer such as Istio. With those you can restrict access to know services.

Interms of what headers to look for most services I have used have allow deployers to configure these using arguments or environmental varibales. This is because there are no standards around what they should be. For example, when playing with IDP/Proxy combinations I have seen the following formats of username header:
- X-Forwarded-User
- X-Auth-Request-User
- X-authentik-username

I'm don't have much of an issue with requiring as OIDC token. This is not much of an issue but does restict the options availiable to deployers. If you go ahead an fully trust the headers issued to you then the deployer can choose what protections to put in place. OIDC, certificates, API keys, service account of MFA. Resticting to having to provide and OIDC token limits these choices

I suppose it depends how your much your team is taking on the risk of users. Can you put in your documents, "the deployer takes responsibility for mitigating any missuse when opting to use Header authentication"?

docs/adr/040-oauth2-proxy-authentication.md
```diff
@@ -0,0 +1,154 @@
...
+
+OAuth2 Proxy can pass the provider access token in the `X-Forwarded-Access-Token` header. The API
+server can use the existing OIDC UserInfo validation, user provisioning, team synchronization,
+and session creation behavior.
+
+This does not trust plain identity headers and does not require another user type. OAuth2 Proxy's
+existing client registration can be reused. However, the provider must support OIDC, and the API
+server must be able to access OIDC discovery and the UserInfo endpoint.
+
+#### Validate an ID token forwarded by OAuth2 Proxy
```
arjavdongaonkar, line 79: One thing I’m not sure about here. The UserInfo path verifies that the token is valid with the IdP, not that it was issued for Dependency-Track.

`OidcIdTokenAuthenticator` checks the audience against the configured ClientID, but `OidcUserInfoAuthenticator` doesn’t. So an access token issued to another client of the same IdP could also pass and get a DT session.

Is the intention to rely only on the network path for this? If the `Authorization` header clash is the main blocker for the ID token, could oauth2-proxy pass it through a different header instead?

VinodAnandan:
@jjj-n  Have you had a chance to compare this proxy approach versus with native client/resource server support for OAuth 2.0 (for non-human identities) and OIDC (for human identities)?

Personally, I have some reservations about the proxy approach, particularly from a defence-in-depth, zero trust principles perspective. In many cases, placing a proxy in front of a system can leave additional security gaps or weaknesses within the underlying application that the proxy alone may not address.

In my view, the proxy approach would be most appropriate for legacy systems that cannot natively implement or support OAuth 2.0 or OIDC. For systems where native support is feasible, there are now many mature SDKs ( https://connect2id.com/products/nimbus-oauth-openid-connect-sdk#specs ) and libraries available that can significantly simplify the implementation of both OAuth 2.0 client/resource server capabilities and OIDC.

It would be interesting to understand whether these options were considered as part of the design and what the key trade-offs were.

docs/adr/040-oauth2-proxy-authentication.md
jjj-n: Good catch. You are right that the existing UserInfo path proves that the token is accepted by
the IdP, but does not prove that it was issued for Dependency-Track.

I revised the ADR to make the ID token the only credential accepted by the proxy session
exchange. Dependency-Track will validate it through the existing ID token path, including issuer,
signature, expiration, and audience checks. The configured `dt.oidc.client-id` must match the
OAuth2 Proxy client identifier, so a token issued to another client of the same IdP will be
rejected.

The ID token will be forwarded in `X-Forwarded-ID-Token` instead of `Authorization`. Direct
upstream deployments can inject the session `id_token` with OAuth2 Proxy's structured header
configuration. NGINX `auth_request` deployments can remap the ID token returned by
`--set-authorization-header`. In both cases, the proxy or NGINX must replace any client-supplied
value.

The first version will not support an access-token-only exchange or call UserInfo during the
exchange. This means the required username and team claims must be present in the ID token, which
is now documented as a limitation.

jjj-n:
Yes. I have revised the ADR to compare the proxy integration with both existing native OIDC and
native OAuth 2.0 resource server support.

The existing Dependency-Track OIDC flow remains the default and the more portable option. It
keeps an independent application authentication flow and does not depend on credentials
forwarded by a proxy. The proposed integration is narrower: it is for deployments that already
require OAuth2 Proxy in front of every browser request and want that proxy session to initiate the
Dependency-Track session without a second OIDC flow.

To preserve defense in depth, the revised design no longer trusts plain identity headers and no
longer accepts an access token merely because UserInfo accepts it. It forwards an ID token in a
separate header and validates the issuer, signature, expiration, and audience against the
configured OAuth2 Proxy client identifier. Network isolation is still required to enforce the
gateway policy, but it is not the only protection for identity integrity.

A native OAuth 2.0 resource server is a possible authentication layer for non-human identities.
It needs a Dependency-Track audience, token type and signature validation, scopes, and
authorization rules. It should be designed with the service-account and workload identity use
cases tracked in #7055 and #7056, rather than as a side effect of this interactive proxy login.

docs/adr/040-oauth2-proxy-authentication.md
arjavdongaonkar: Thanks, the ID token with an audience check covers what I was thinking

Two follow-ups from dropping UserInfo.

1. Username and teams now have to be in the ID token, since `dt.oidc.username-claim` and `dt.oidc.teams-claim` are read from it. Groups often are not there by default, Keycloak needs a mapper with "add to ID token" and Entra only emits them when configured. So `team-synchronization` could end up syncing nothing in deployments where it works fine today over UserInfo.

2. Audience is checked against `dt.oidc.client-id`, which is a single value, and the forwarded token is issued to the OAuth2 Proxy client. If a deployment also uses normal OIDC login, that token has a different audience. Do both need to be the same client, or are the two mutually exclusive?

jjj-n: Thanks, both points are valid. I checked the current profile-completeness behavior and revised the
ADR to make the outcomes explicit.

1. The proxy exchange will require the configured username claim and, when team synchronization
   is enabled, the configured teams claim to be present in the ID token. One clarification: the
   current OIDC authentication service does not treat a missing teams claim as an empty list. It
   rejects the incomplete profile, so existing memberships are not silently removed. An
   explicitly present empty teams claim is valid and will synchronize the user's mapped team
   memberships to empty. Providers such as Keycloak or Entra therefore need to be configured to
   include the required claims in the ID token; there is no UserInfo fallback for this exchange.

2. Requiring the browser client and OAuth2 Proxy to share a registration would be unnecessarily
   restrictive. I revised the design to add a separate `dt.oauth2-proxy.client-id` setting as the
   expected audience for forwarded ID tokens. Native OIDC continues to use
   `dt.oidc.client-id`. The two login paths may coexist with different client registrations, but
   both must use the provider configured by `dt.oidc.issuer` in the first version.

phmcder: > Thanks, both points are valid. I checked the current profile-completeness behavior and revised the ADR to make the outcomes explicit.
> 
> 1. The proxy exchange will require the configured username claim and, when team synchronization
>    is enabled, the configured teams claim to be present in the ID token. One clarification: the
>    current OIDC authentication service does not treat a missing teams claim as an empty list. It
>    rejects the incomplete profile, so existing memberships are not silently removed. An
>    explicitly present empty teams claim is valid and will synchronize the user's mapped team
>    memberships to empty. Providers such as Keycloak or Entra therefore need to be configured to
>    include the required claims in the ID token; there is no UserInfo fallback for this exchange.
> 2. Requiring the browser client and OAuth2 Proxy to share a registration would be unnecessarily
>    restrictive. I revised the design to add a separate `dt.oauth2-proxy.client-id` setting as the
>    expected audience for forwarded ID tokens. Native OIDC continues to use
>    `dt.oidc.client-id`. The two login paths may coexist with different client registrations, but
>    both must use the provider configured by `dt.oidc.issuer` in the first version.

Hi @jjj-n I know I asked for oauth2-proxy support specifically, but the passing of authentications headers is pretty generic and there are multiple solutions out there. I suggest being more generic with setting names such as changing `dt.oauth2-proxy.client-id` to `dt.auth-proxy.client-id`

The rest looks great so far. Looking forward to testing it out

phmcder:
> @jjj-n Have you had a chance to compare this proxy approach versus with native client/resource server support for OAuth 2.0 (for non-human identities) and OIDC (for human identities)?
> 
> Personally, I have some reservations about the proxy approach, particularly from a defence-in-depth, zero trust principles perspective. In many cases, placing a proxy in front of a system can leave additional security gaps or weaknesses within the underlying application that the proxy alone may not address.
> 
> In my view, the proxy approach would be most appropriate for legacy systems that cannot natively implement or support OAuth 2.0 or OIDC. For systems where native support is feasible, there are now many mature SDKs ( https://connect2id.com/products/nimbus-oauth-openid-connect-sdk#specs ) and libraries available that can significantly simplify the implementation of both OAuth 2.0 client/resource server capabilities and OIDC.
> 
> It would be interesting to understand whether these options were considered as part of the design and what the key trade-offs were.

@VinodAnandan you are correct. Making these changes pushes the ownness of security onto the deployers and not the developers of this application. Deployers can make mistakes and provide unauthenticated access to dependency-track. However, well executed deployments with well managed systems can be impacted by services that only support their own workflows. In our case, NOT having proxy support has raised the same concerns about defence-in-depth and zero-trust. 

To support dependency-track authentication we have had to reduce the protection provided at our ingress gateway. Normally we would intercept any unauthenticated traffic and block it from getting access to our cluster. We also block any authorisation headers not issued by us from reaching that backend. Both of these protections prevent dependency track from working so need disabling. 

By disabling some of our protections we will have to fully trust that the developers of dependency-track have done a good job of implementing security.  Each authentication flow that is not handled by our proxy is another point highlighted in our internal security review that needs to managing. We need to be more diligent monitoring the vulnerabilities reported in the packages dependency track uses and implement playbooks to manage responses. If vulnerabilities are reported in any authentication packages we would be exposed until a new version is released.

Implementing this feature brings Dependecy Tack in line with our other applications. This simplifies the effort involved in supporting Dependency Track for our users.

docs/adr/040-oauth2-proxy-authentication.md
jjj-n: Agreed — the exchange validates a forwarded ID token and does not depend on which reverse proxy provides it, and the existing `dt.oidc.*` settings are already vendor-neutral, so a generic name is more consistent. I renamed the setting to `dt.auth-proxy.client-id` in the ADR.

To avoid overpromising, the ADR now states that OAuth2 Proxy is the reference implementation for the first version: other reverse proxies that can inject a bearer ID token into `X-Forwarded-ID-Token` should work with the same validation path, but only the OAuth2 Proxy configuration is documented and validated.

VinodAnandan:
Apologies, everyone. I don’t think I articulated my point very clearly earlier.

The current OIDC integration that Dependency-Track (DT) provides is not a fully compliant, end-to-end OIDC implementation. What I meant by full OIDC support is a model where the DT Web Application (SPA) acts as the OIDC Relying Party (Client), while the DT Backend REST API acts as an OIDC/OAuth 2.0 Resource Server. DT would then trust a central Authorisation Server, such as Keycloak.

In this model:

- The DT SPA/Relying Party (Client) authenticates with the central Authorisation Server and obtains an ID token and access token (JWT).
- The SPA uses the access token to call the DT Backend REST API/ Resource Server..
- The DT API validates the JWT access token and enforces the appropriate authorisation policies.
- The access token would typically be short-lived and signed by central Authorisation Server 

I am strongly supportive of native, standards-compliant OIDC/OAuth 2.0 support. This approach would allow the DT API to be securely protected using a central OIDC/Authorisation Server without introducing an additional proxy layer.
In my view, this also avoids the need for a proxy to act as a pseudo Client/Relying Party and Resource Server. The SPA and API can instead participate directly in the standard OIDC/OAuth 2.0 flows, which provides a cleaner architecture and better separation of responsibilities.

jjj-n:
Thanks for clarifying. I understand the distinction now: you are proposing that the SPA call the DT API with an access token issued by the central authorization server, with the API acting as an OAuth 2.0 resource server, rather than exchanging the login credential for a DT session.

I agree that native resource server support would benefit both interactive and machine clients. The current ADR acknowledges the validation and authorization work this requires, but does not implement that model.

The narrower requirement here is the deployment described by @phmcder, where authentication at the ingress gateway is already mandatory. Native resource server support would not by itself remove that gateway requirement. This proposal lets the gateway-authenticated identity initiate a DT session while retaining token validation within DT.

Would the maintainers prefer to pursue native resource server support first, or consider this optional proxy integration separately? I can clarify that distinction further in the ADR.

VinodAnandan:
Just to clarify my thoughts on gateway validation, I think it is perfectly reasonable for an API gateway to perform validation and reject invalid requests early, similar to a security guard checking people before they enter a building. However, I don’t think the gateway’s validation or rejection should be relied upon as the sole or primary mechanism for ensuring security. The downstream resource server must still perform its own validation and security checks.

The analogy isn’t perfect, but hopefully it helps explain my point: imagine a customer being allowed to withdraw money from a bank cashier or ATM simply because a security guard checked them before they entered the building. The security guard provides an important first layer of protection, but the cashier or ATM still needs to perform its own authentication and authorisation checks before allowing the transaction.

phmcder:
> Just to clarify my thoughts on gateway validation, I think it is perfectly reasonable for an API gateway to perform validation and reject invalid requests early, similar to a security guard checking people before they enter a building. However, I don’t think the gateway’s validation or rejection should be relied upon as the sole or primary mechanism for ensuring security. The downstream resource server must still perform its own validation and security checks.
> 
> The analogy isn’t perfect, but hopefully it helps explain my point: imagine a customer being allowed to withdraw money from a bank cashier or ATM simply because a security guard checked them before they entered the building. The security guard provides an important first layer of protection, but the cashier or ATM still needs to perform its own authentication and authorisation checks before allowing the transaction.

@VinodAnandan I do see your point about checking the user's authorisation before allowing them access to your services. This can still be an option by validating the user token signature presented to you api boundary. 

Along with the Authentication headers passed by the proxy you could also require the passing of an oauth token with which you can verify the token issuer. This has been mentioned before but I apologies that I could find who said it.
https://github.com/oauth2-proxy/oauth2-proxy/issues/843 

To map with your analogy, users will authenticate with the security at the front door and be issued with a pass. When they get to your service boundary you have another guard who checks their issued pass and confirm it was issued by a trusted source. Only then are they allowed access.

PR #7174, opened 2026-08-30
Support expiry of vulnerability suppressions

### LINKED ISSUE #3627: Expiration support for vulnerability suppressions
### Current Behavior

Currently vulnerabilities can be suppressed but no expiration can be set like in OWASP Dependency-Check.

### Proposed Behavior

Setting a suppression date would be helpful in cases where:
- A patch for a vulnerability doesn't exist yet
- The vulnerability might not be currently be relevant but worth re-checking in the future

### Checklist

arjavdongaonkar:
### Description

Suppressing a finding is permanent today until someone remembers to undo it. That does not fit the two cases in the issue, suppressing a finding that has no patch yet, and suppressing one that is not relevant right now but is worth re-checking later.

This allows an expiry to be recorded alongside a suppression. A maintenance task un-suppresses findings whose expiry has elapsed and records it in the analysis audit trail, so it is not mistaken for a manual decision.

### Addressed Issue

Fixes https://github.com/DependencyTrack/dependency-track/issues/3627

### Additional Details

On the open question from the issue about whether expiry should reset the analysis as a whole, this only clears the suppression and leaves the state, justification and response untouched.

The reason is consistency. makeAnalysis already treats state and suppression as independent, a manual unsuppression does not reset the state either. Expiry is just a deferred unsuppression, so it behaving differently from the manual one seemed wrong. It also avoids having to store previous values somewhere in order to restore them. ANALYSISCOMMENT records changes as display text like "Field: old → new", which is fine for humans but not something to parse and restore from, so a real reset would need new storage. Happy to revisit if you would rather it reset the analysis.

Manually unsuppressing clears any expiry that was set, since an expiry has no meaning for a finding that is no longer suppressed.

Expiry is applied by a scheduled task rather than being derived when findings are read. SUPPRESSED is consumed in a lot of places, findings, metrics, notifications, policy evaluation, CEL, VEX export, and deriving it at read time would mean touching all of them and getting inconsistent behaviour if one was missed. Flipping the stored flag keeps them all consistent.

The task follows the existing maintenance task pattern, batched through AbstractBatchingMaintenanceTask, with dt.task.suppression-expiry.cron defaulting to hourly.

One thing worth flagging, suppressionExpiresAt is the first Instant in the v1 API, everything else there is a Date. It serialises as an epoch number like the existing Date fields, and accepts an ISO-8601 string on input. Let me know if you would prefer it to be a Date for consistency.

### Checklist

- [X] This PR introduces changes to the database model, and I have updated the migration changelog accordingly

The documentation box is unchecked, the suppression docs live in the docs repo. Will raise a PR there once the shape here is agreed.


arjavdongaonkar:
Wrong branch, this picked up an old branch of the same name on my fork. Reopening from a clean branch.

Issue #6572, opened 2026-07-03
Vulnerability analysis fails permanently with NPE in PrepareVulnAnalysisActivity when a component property was uploaded with an empty value


flobrunner:
### Current Behavior

When a project's BOM contains a component property with an **empty string value** (which is valid per the CycloneDX schema, and is what Syft emits for `syft:package:metadataType` on some packages), the vulnerability analysis workflow for that project fails terminally and the project never receives any analysis results (no metrics, no vulnerabilities, risk score stays empty).

The `prepare-vuln-analysis` activity throws a `NullPointerException` and gives up after 6 attempts:

```
WARN [ActivityTaskWorker] Activity failed terminally after 6 attempt(s) [activityName=prepare-vuln-analysis, workflowRunId=..., activityTaskAttempt=6]
java.lang.NullPointerException
    at org.cyclonedx.proto.v1_7.Property$Builder.setValue(Property.java:654)
    at org.dependencytrack.vulnanalysis.PrepareVulnAnalysisActivity.lambda$assembleBom$1(PrepareVulnAnalysisActivity.java:167)
    at org.jdbi.v3.core.result.ResultBearing.lambda$reduceRows$1(ResultBearing.java:276)
    at org.jdbi.v3.core.result.ResultBearing$1.scanResultSet(ResultBearing.java:59)
    at org.jdbi.v3.core.statement.Query.scanResultSet(Query.java:61)
    at org.jdbi.v3.core.result.ResultBearing.reduceRows(ResultBearing.java:269)
    at org.dependencytrack.vulnanalysis.PrepareVulnAnalysisActivity.lambda$assembleBom$0(PrepareVulnAnalysisActivity.java:154)
    at org.dependencytrack.persistence.jdbi.JdbiFactory.lambda$withJdbiHandle$0(JdbiFactory.java:80)
    ...
    at org.dependencytrack.vulnanalysis.PrepareVulnAnalysisActivity.assembleBom(PrepareVulnAnalysisActivity.java:142)
    at org.dependencytrack.vulnanalysis.PrepareVulnAnalysisActivity.execute(PrepareVulnAnalysisActivity.java:95)
```

### Steps to Reproduce

1. Run Dependency-Track v5.0.2 with the Trivy analyzer enabled (any other analyzer configuration is irrelevant; the bug only triggers via the `COMPONENT_PROPERTIES` requirement).
2. Upload the following minimal BOM to a new project:
   ```json
   {
[...]

PR #7178, opened 2026-08-30
Make the Trivy API token optional

### LINKED ISSUE #4843: Trivy API Token
### Current Behavior

Hi,

I have a Trivy scanner in the same kubernetus network where Dependency Track is running.

Currently, I cannot configure it as it requires an from me API token. My trivy does not require any.

Can this setting be changed so that it allows only BaseURL?

### Proposed Behavior

Allow save Trivy settings without providing API token

### Checklist

arjavdongaonkar:
### Description

Trivy servers only require a token when one was configured with --token, so a deployment can legitimately have none to provide. The analyzer rejected any configuration without one, which makes a stock Trivy server unusable with Dependency-Track.

### Addressed Issue

Fixes https://github.com/DependencyTrack/dependency-track/issues/4843

### Additional Details

On v5 the effect is worse than not being able to save the settings. The validation throws while the analyzer is being enabled, and getApplicableAnalyzers calls isEnabled on every analyzer during prepare-vuln-analysis, so the activity fails terminally after its retries and the project gets no results from any analyzer, not just Trivy.

I hit this by accident while setting up a local Trivy server. With Trivy enabled and no token, prepare-vuln-analysis failed 6 times and the project ended up with zero findings even though the internal analyzer was enabled:

```
org.dependencytrack.plugin.api.config.InvalidRuntimeConfigException: No API token provided
    at TrivyVulnAnalyzerFactory.lambda$runtimeConfigSpec$0(TrivyVulnAnalyzerFactory.java:111)
    at TrivyVulnAnalyzerFactory.isEnabled(TrivyVulnAnalyzerFactory.java:84)
    at PrepareVulnAnalysisActivity.getApplicableAnalyzers(PrepareVulnAnalysisActivity.java:122)
...
Activity failed; Failing task terminally after 6 attempt(s) [activityName=prepare-vuln-analysis]
```

v4 handled this by skipping the analyzer, TrivyAnalysisTask logged "No API token provided; Skipping" and carried on.

Verified against a stock `trivy server` with no token. Before the change the workflow failed terminally with no findings. After it, the same setup returns findings attributed to trivy, including CVE-2021-44228 on log4j-core 2.14.1, and prepare-vuln-analysis no longer fails.

That one analyzer's invalid configuration can take down analysis for every analyzer seems worth looking at separately, this PR only removes the most likely trigger. #6572 was a similar case of prepare-vuln-analysis failing terminally and leaving projects without any results. Happy to raise that as its own issue if useful.

### Checklist

The documentation box is unchecked, the Trivy configuration docs live in the docs repo. Will raise a PR there if the token should be documented as optional.

### DIFF
--- a/vuln-analysis/trivy/src/main/java/org/dependencytrack/vulnanalysis/trivy/TrivyVulnAnalyzer.java
+++ b/vuln-analysis/trivy/src/main/java/org/dependencytrack/vulnanalysis/trivy/TrivyVulnAnalyzer.java
@@ -329,14 +329,17 @@ final class TrivyVulnAnalyzer implements VulnAnalyzer {
     }
 
     private byte[] sendProtobufRequest(String url, byte[] body) throws InterruptedException {
-        final var request = HttpRequest.newBuilder()
+        final HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
                 .uri(java.net.URI.create(url))
                 .header("Accept", "application/protobuf")
                 .header("Content-Type", "application/protobuf")
-                .header(TOKEN_HEADER, apiToken)
                 .timeout(Duration.ofSeconds(30))
-                .POST(HttpRequest.BodyPublishers.ofByteArray(body))
-                .build();
+                .POST(HttpRequest.BodyPublishers.ofByteArray(body));
+        // Trivy servers only require a token when one was configured with --token.
+        if (apiToken != null) {
+            requestBuilder.header(TOKEN_HEADER, apiToken);
+        }
+        final HttpRequest request = requestBuilder.build();
 
         final HttpResponse<byte[]> response;
         try {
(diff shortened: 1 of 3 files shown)


nscuro:
> With Trivy enabled and no token [...]

Did you modify the config directly in the database? Because the same validation you removed in this PR is what prevents such a config from being submitted via API in the first place.

arjavdongaonkar:
> > With Trivy enabled and no token [...]
> 
> Did you modify the config directly in the database? Because the same validation you removed in this PR is what prevents such a config from being submitted via API in the first place.

Yes, directly in the database, my mistake. updateExtensionConfig runs the same validator so that config can't be submitted normally, so scratch that part.

The ask in #4843 still stands though, trivy server only needs a token when it is started with --token. Tested against a token-less server and it works fine.

nscuro (approved):

PR #7179, opened 2026-08-30
Expose patched versions to CEL policy expressions

### LINKED ISSUE #5731: Add Policy Evaluator for Patched Version Availability (Patchable vs Non-Patchable)

arjavdongaonkar:
### Description

Policies cannot tell apart vulnerabilities that can be remediated by upgrading from those with no fix available, so SLA style policies end up treating both the same even when there is nothing to upgrade to.

The data is already there, VULNERABILITY.PATCHEDVERSIONS is populated by the data sources, it just is not reachable from an expression. This exposes it:

```
vulns.exists(v, v.patched_versions != "")
```

### Addressed Issue

Fixes https://github.com/DependencyTrack/dependency-track/issues/5731

### Additional Details

This is a plain vulnerability level field, so it needs nothing beyond a proto field, a field mapping and a row mapper line. It is only selected when an expression uses it, like the other fields.

I used field number 40 because 38 and 39 are taken by #7173 and #7177, so the three do not clash whichever order they land in.

This replaces #5202, which was raised against master before the v5 restructure and added a dedicated PolicyEvaluator class, an extension point that no longer exists.

One thing worth noting for whoever picks this up, PATCHEDVERSIONS is a free-form string from the source, not a parsed version range, so an expression can test whether a fix exists but cannot reliably compare it against the component version. That felt like the right scope for this issue, which asks for patchable versus non-patchable.

### Checklist

The documentation box is unchecked, the expression reference lives in the docs repo. Will raise a PR there once agreed.

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/policy/cel/persistence/CelPolicyFieldMappingRegistry.java
+++ b/apiserver/src/main/java/org/dependencytrack/policy/cel/persistence/CelPolicyFieldMappingRegistry.java
@@ -92,6 +92,7 @@ public final class CelPolicyFieldMappingRegistry {
             new FieldMapping("owasp_rr_vector", "v.\"OWASPRRVECTOR\""),
             new FieldMapping("cwes", "STRING_TO_ARRAY(v.\"CWES\", ',')"),
             new FieldMapping("aliases", "CAST(JSONB_VULN_ALIASES(v.\"SOURCE\", v.\"VULNID\") AS TEXT)"),
+            new FieldMapping("patched_versions", "v.\"PATCHEDVERSIONS\""),
             new FieldMapping("epss_score", "ep.\"SCORE\""),
             new FieldMapping("epss_percentile", "ep.\"PERCENTILE\""));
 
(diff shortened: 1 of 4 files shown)


arjavdongaonkar:
Closing this.

@nscuro pointed out on #5202 that nothing populates PATCHEDVERSIONS, and that still holds on v5. On an instance with a full NVD and OSV mirror plus Trivy findings, 0 of 384715 vulnerabilities have a value in that column.

So this would expose a field that is always empty, and a policy using it would silently never match. My test only passed because it set the field by hand, which no ingestion path does.

The gap is upstream of the policy engine, so #5731 needs a source to populate patched versions before exposing it to expressions makes sense.

PR #7195, opened 2026-08-31
Add option to use Snyk's per-package endpoint when batching is not available

### LINKED ISSUE #7137: Regression (5.x): Snyk analyzer fails with 403 — no fallback to 4.x per-PURL GET
### Current Behavior

After upgrading to DT 5.0.4, the Snyk analyzer fails for every analysis. Our Snyk token is valid, but DT 5.x calls POST /rest/orgs/{org_id}/packages/issues (batch), which returns HTTP 403 / SNYK-OSSI-1040 for our org. The per-PURL GET endpoint used by DT 4.x still returns 200 with the same token. There is no fallback; the vuln-analyzer task fails terminally.

According to Snyk support, this is not a misconfiguration: the batch endpoint POST /rest/orgs/{org_id}/packages/issues is intentionally restricted at platform level and is not publicly available for integration with all customers (documented as "not available to all customers"). Access cannot be enabled on request via support; it requires a controlled internal/commercial process. Snyk confirmed DT 5.x switched from the 4.x per-PURL GET to this restricted batch API with no fallback, and recommended raising this with DT maintainers.

Api-server log (2026-08-27):
Caused by: java.io.IOException: Snyk API request failed with status 403
  at org.dependencytrack.vulnanalysis.snyk.SnykVulnAnalyzer.fetchIssues(SnykVulnAnalyzer.java:289)
  at org.dependencytrack.vulnanalysis.snyk.SnykVulnAnalyzer.analyzePurlBatch(SnykVulnAnalyzer.java:209)

### Steps to Reproduce

1. Configure Snyk analyzer in DT 5.0.4 (base URL, org ID, API token).
2. Use a Snyk org entitled for per-PURL GET but not for batch POST /packages/issues.
3. Trigger vulnerability analysis on a project with supported PURLs.

Or reproduce with curl (same token):
- POST .../rest/orgs/{org_id}/packages/issues → 403 SNYK-OSSI-1040
- GET .../rest/orgs/{org_id}/packages/{purl}/issues → 200

### Expected Behavior

Snyk analyzer should work on 5.x for orgs that worked on 4.x. When batch POST returns 403 / SNYK-OSSI-1040, DT should automatically fall back to the 4.x per-PURL GET endpoint instead of failing entirely.

### Dependency-Track Version

5.x

### Browser

Google Chrome

### Checklist

arjavdongaonkar:
### Description

The batch endpoint `POST /rest/orgs/{org}/packages/issues` is not available to all Snyk organizations and answers 403 for those it is not available to. v5 uses it exclusively and 403 is not retryable, so the analyzer fails outright for those organizations. The per-package endpoint that v4 used still works with the same token.

This falls back to the per-package endpoint when the batch endpoint answers 403.

### Addressed Issue

Fixes https://github.com/DependencyTrack/dependency-track/issues/7137

### Additional Details

The fallback sticks for the remainder of the analysis, so a single 403 does not repeat once per batch. It is logged at warn level, because the per-package endpoint means one request per package rather than one per 100, and on a large portfolio that difference is worth knowing about.

The URL is the same one v4 used:

```
GET %s/rest/orgs/%s/packages/%s/issues?version=%s
```

Only 403 triggers the fallback. Every other status behaves as before, so retryable errors still retry and anything else still fails.

Two things I was unsure about and am happy to change.

I made this automatic rather than a configuration option, on the grounds that the affected organizations are currently broken and would otherwise have to discover a setting. A config flag would make the request volume change explicit instead, if you prefer that.

The other is more general and I did not touch it here. getApplicableAnalyzers calls into every analyzer during prepare-vuln-analysis, so an analyzer failing terminally means the project gets no results from any analyzer, not just the one that failed. The reporter of #7137 saw that, and #6572 was a similar case. Worth a separate issue.

Two things worth calling out about correlating results.

The per-package response does not have to repeat the package coordinates, since the package is implied by the request URL. Correlating through SnykModelConverter.getIssuePurl, which reads a fixed representation index of the batch shape, would therefore drop every issue and then cache the PURL as having no vulnerabilities, which is a silent false negative. The per-package path keys results by the PURL it requested instead, so it does not parse coordinates at all.

A 404 for a package Snyk does not know is treated as having no issues. The batch endpoint omits such packages, and v4 logged and carried on, so a single unknown package should not fail the analysis.

I do not have access to a Snyk org where the batch endpoint is forbidden, so the per-package fixture is constructed rather than captured from a real response. It deliberately omits the package coordinates so the test fails if correlation regresses to reading them, but someone with a real entitlement confirming the response shape, and that the configured API version still serves this endpoint, would be worth having before merge.

Tested with WireMock, no Snyk account needed. One test asserts the fallback returns issues after a 403, another asserts the per-package endpoint is not touched when the batch endpoint works.

### Checklist

- ~[ ] This PR introduces new or alters existing behavior, and I have updated the documentation accordingly~

### DIFF
--- a/vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
+++ b/vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
@@ -37,11 +37,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.net.URI;
+import java.net.URLEncoder;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.BodyPublishers;
 import java.net.http.HttpResponse;
 import java.net.http.HttpResponse.BodyHandlers;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -61,6 +63,14 @@ final class SnykVulnAnalyzer implements VulnAnalyzer {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SnykVulnAnalyzer.class);
     private static final int REQUEST_BATCH_SIZE = 100;
+
+    /**
+     * The batch endpoint is not available to all Snyk organizations, and answers 403 for
+     * those it is not available to. Such organizations can still use the per-package
+     * endpoint, so fall back to it for the remainder of the analysis once that happens.
+     */
+    private boolean usePerPackageEndpoint = false;
+
     private static final int CACHE_BATCH_SIZE = 500;
     private static final Set<String> SUPPORTED_PURL_TYPES = Set.of(
             "cargo",
@@ -213,7 +223,7 @@ final class SnykVulnAnalyzer implements VulnAnalyzer {
 
         final SnykIssuesResponse response;
         try {
-            response = fetchIssues(purlBatch);
+            response = fetchIssuesWithFallback(purlBatch);
         } catch (IOException e) {
             final var message = "Failed to fetch Snyk issues";
             RetryableVulnAnalysisException.throwIfRetryableNetworkError(e, message);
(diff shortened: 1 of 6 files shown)


vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
```diff
@@ -264,6 +250,133 @@ private Map<String, List<SnykIssue>> analyzePurlBatch(
...
+
+    private SnykIssuesResponse fetchIssuesForPackage(String purl) throws InterruptedException, IOException {
+        final String encodedPurl = URLEncoder.encode(purl, StandardCharsets.UTF_8);
+
+        final var request = HttpRequest.newBuilder()
+                .uri(URI.create("%s/rest/orgs/%s/packages/%s/issues?version=%s"
+                        .formatted(apiBaseUrl, orgId, encodedPurl, apiVersion)))
+                .header("Authorization", "token " + apiToken)
+                .header("Accept", "application/vnd.api+json")
+                .timeout(Duration.ofSeconds(30))
```
nscuro: Timeout for individual package requests should be lower, ~10s.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
```diff
@@ -296,6 +409,9 @@ private SnykIssuesResponse fetchIssues(Collection<String> purls) throws Interrup
             }
 
             RetryableVulnAnalysisException.throwIfRetryableHttpError(response);
+            if (response.statusCode() == 403) {
+                throw new BatchEndpointUnavailableException();
+            }
```
nscuro: This feels dangerous, since Snyk might respond with 403 for other reasons than batching not being allowed. I think the better option would be an explicit toggle in the settings, defaulting to batching enabled to stay backward compatible. That's both easier to reason about *and* less wasteful, since the initial request currently always fails for users without batch API access.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
```diff
@@ -264,6 +250,133 @@ private Map<String, List<SnykIssue>> analyzePurlBatch(
...
+            return fetchIssuesPerPackage(purlBatch);
+        }
+
+        try {
+            return correlateIssues(fetchIssues(purlBatch), bomRefsByPurl);
+        } catch (BatchEndpointUnavailableException e) {
+            LOGGER.warn(
+                    "The Snyk batch endpoint is not available for this organization; "
+                            + "Falling back to the per-package endpoint for the remainder of this analysis. "
+                            + "Note that this issues one request per package, instead of one per {} packages.",
```
nscuro: Nit: use a text block (`"""`) here.

nscuro (changes requested):

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
arjavdongaonkar: Lowered to 10s.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
arjavdongaonkar: Agreed, switched to a `batchRequestsEnabled` setting defaulting to true. Hadn't considered that the 403 also wastes a request every analysis for orgs without batch access.

vuln-analysis/snyk/src/main/java/org/dependencytrack/vulnanalysis/snyk/SnykVulnAnalyzer.java
arjavdongaonkar: That log line went away with the auto-detection, so nothing left to reformat.

vuln-analysis/snyk/src/main/resources/org/dependencytrack/vulnanalysis/snyk/snyk-vuln-analyzer-config-v1.schema.json
```diff
@@ -18,6 +18,12 @@
       "description": "Whether to include alias information in vulnerability data.",
       "existingJavaType": "boolean"
     },
+    "batchRequestsEnabled": {
+      "type": "boolean",
+      "title": "Batch Requests Enabled",
+      "description": "Whether to request issues for multiple packages at once. The batch endpoint is not available to all Snyk organizations; disable this to request issues one package at a time instead.",
+      "existingJavaType": "boolean"
+    },
```
nscuro, line 27: Set a [default](https://json-schema.org/draft/2020-12/json-schema-validation#name-default) of `true` here so existing configs in user databases where the new field doesn't exist still deserialize to `true`.

nscuro (changes requested):

vuln-analysis/snyk/src/main/resources/org/dependencytrack/vulnanalysis/snyk/snyk-vuln-analyzer-config-v1.schema.json
arjavdongaonkar: added `"default": true` to the schema. The factory default only covered new configs, existing ones without the field would have deserialized to false and silently stopped batching. Added a test for it.

nscuro (approved):

PR #7172, opened 2026-08-30
v4-migrator: Reconcile PROJECT collection logic and tag

arjavdongaonkar:
### Description

Fixed a check constraint violation on PROJECT_COLLECTION_TAG_REQUIRED_check during the load phase. v5 requires a COLLECTION_TAG_ID exactly when COLLECTION_LOGIC is AGGREGATE_DIRECT_CHILDREN_WITH_TAG. v4 has no constraint tying the two columns together, so a v4 instance can hold combinations that v5 rejects.

### Addressed Issue

Fixes https://github.com/DependencyTrack/dependency-track/issues/6870

### Additional Details

There are two combinations possible in v4 and both fail the check:

AGGREGATE_DIRECT_CHILDREN_WITH_TAG with no tag, which is the case reported in the issue. The logic is dropped, since it cannot aggregate anything without a tag. That ends up in the same state as fixing the project by hand in v4, which is what the reporter said they were going to do.

A collection tag set alongside a logic that does not use one, for example AGGREGATE_DIRECT_CHILDREN. Not part of the report, but v4 permits it and it fails the same check, so the tag is dropped. Let me know if you would rather I leave this case out and keep the PR to the reported one.

Well formed collection projects are untouched. This is in the same spirit as the sanitisation already done for CLASSIFIER and PROPERTYTYPE values outside the v5 enums.

ProjectCollectionLogicIT covers all three cases. Also ran ProjectDedupIT, ProjectJoinTablesIT, PropertyTablesIT, ProjectAccessUsersIT, DryRunAndSampleIT and PreflightIT.

### Checklist

- [X] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)
- [X] This PR fixes a defect, and I have provided tests to verify that the fix is effective

PR #7205, opened 2026-09-01
Treat connection resets as transient regardless of exception type

arjavdongaonkar:
### Description

A connection reset does not always surface as a `SocketException`. Depending on where in the exchange it lands, the JDK HTTP client reports a plain `IOException` with "Connection reset by peer" instead, and `TransientNetworkErrors.isTransient` only matched on exception types, so it missed that form.

`isTransient` is shared by notification publishing, package metadata resolution and vulnerability analysis, so in all three a reset taking that path was treated as permanent: no retry, and no stale cache fallback in the places that have one.

### Addressed Issue

N/A

### Additional Details

Found while investigating a recurring CI failure rather than from a filed issue.

It also explains why `CachingHttpClientTest.shouldThrowRetryableExceptionOnConnectionResetWhenNoCachedEntry` is flaky. WireMock's `CONNECTION_RESET_BY_PEER` fault surfaces as either form depending on timing, so the test passes or fails accordingly. It failed on #7172 on 30 Aug, and on #7195 on 31 Aug and again on 1 Sep:

```
Expecting actual throwable to be an instance of:
  org.dependencytrack.pkgmetadata.resolution.api.RetryableResolutionException
but was:
  java.io.UncheckedIOException: java.io.IOException: Connection reset by peer
    at CachingHttpClient.lambda$sendWithStaleFallback$2(CachingHttpClient.java:206)
```

The message check is deliberately narrow, it only applies to `IOException`, so an unrelated `IOException` such as "malformed response" is still classified as permanent. There is an existing test asserting that, and it still passes.

I ran `CachingHttpClientTest` three times locally after the change, 53 tests green each time, though I cannot prove absence of flakiness from that alone. The unit test on the classifier itself is the deterministic part.

### Checklist

- ~[ ] This PR introduces new or alters existing behavior, and I have updated the documentation accordingly~

### DIFF
--- a/support/net/src/main/java/org/dependencytrack/support/net/TransientNetworkErrors.java
+++ b/support/net/src/main/java/org/dependencytrack/support/net/TransientNetworkErrors.java
@@ -46,10 +46,13 @@ public final class TransientNetworkErrors {
                 return true;
             }
 
-            // A connection reset does not always surface as a SocketException. Depending
-            // on where in the exchange it lands, the JDK HTTP client reports a plain
-            // IOException instead, which is just as retryable.
-            if (cause instanceof IOException && isConnectionReset(cause.getMessage())) {
+            // The JDK HTTP client does not always report transport failures as a socket
+            // or channel exception. For the same peer reset it may throw a plain
+            // IOException whose cause chain carries no more specific type, in which case
+            // the message is the only thing left to go on. Matching on messages is not
+            // something to be happy about, but the alternative is treating a retryable
+            // transport failure as permanent.
+            if (cause instanceof IOException && isTransportFailureMessage(cause.getMessage())) {
                 return true;
             }
         }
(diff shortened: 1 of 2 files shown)


support/net/src/main/java/org/dependencytrack/support/net/TransientNetworkErrors.java
```diff
@@ -42,8 +45,19 @@ public static boolean isTransient(Throwable throwable) {
...
+            if (cause instanceof IOException && isConnectionReset(cause.getMessage())) {
+                return true;
+            }
         }
 
         return false;
     }
+
+    private static boolean isConnectionReset(@Nullable String message) {
+        return message != null && message.toLowerCase().contains("connection reset");
```
nscuro: Is there a way to avoid checking the exception message? This has been super brittle in my past experience and I would like to avoid having that in production code. IIRC some messages of IO failures depend on the platform the code is running on.

PR #7206, opened 2026-09-01
Add `analyzer_identity` in `NEW_VULNERABILITY` notifications

### LINKED ISSUE #6701: Add `analyzer_identity` in `NEW_VULNERABILITY ` notifications
### Current Behavior

`NEW_VULNERABILITIES_SUMMARY` already exposes `analyzer_identity` for each finding, but `NEW_VULNERABILITY` notifications do not. Consumers of event-based notifications currently cannot tell which analyzer attributed the vulnerability unless they switch to scheduled summary notifications, which changes delivery semantics.

### Proposed Behavior

`NEW_VULNERABILITY` notification subject includes `analyzer_identity` when available.

### Checklist

sahibamittal:
### Description

Expose the `analyzer_identity` for vulnerabilities in `NEW_VULNERABILITY` notifications. This lets notification consumers see which analyzer produced the finding.

### Addressed Issue

Closes https://github.com/DependencyTrack/dependency-track/issues/6701

### Checklist

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/NotificationSubjectDao.java
+++ b/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/NotificationSubjectDao.java
@@ -155,6 +155,7 @@ public interface NotificationSubjectDao extends SqlObject {
                  , STRING_TO_ARRAY(v."CWES", ',') AS "vulnCwes"
                  , JSONB_VULN_ALIASES(v."SOURCE", v."VULNID") AS "vulnAliasesJson"
                  , <@sql.isKevColumn vulnSource='v."SOURCE"' vulnId='v."VULNID"'/> AS "vulnIsKev"
+                 , fa."ANALYZERIDENTITY" AS "vulnAnalyzerIdentity"
               FROM UNNEST(:componentIds, :vulnerabilityIds)
                 AS req(component_id, vulnerability_id)
              INNER JOIN "COMPONENTS_VULNERABILITIES" AS cv
@@ -166,6 +167,9 @@ public interface NotificationSubjectDao extends SqlObject {
                 ON p."ID" = c."PROJECT_ID"
              INNER JOIN "VULNERABILITY" AS v
                 ON v."ID" = req.vulnerability_id
+             INNER JOIN "FINDINGATTRIBUTION" AS fa
+                ON fa."COMPONENT_ID" = req.component_id
+               AND fa."VULNERABILITY_ID" = req.vulnerability_id
               LEFT JOIN "ANALYSIS" AS a
                 ON a."COMPONENT_ID" = req.component_id
                AND a."VULNERABILITY_ID" = req.vulnerability_id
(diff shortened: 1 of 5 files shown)


apiserver/src/main/java/org/dependencytrack/persistence/jdbi/NotificationSubjectDao.java
```diff
@@ -166,6 +167,9 @@ AS req(component_id, vulnerability_id)
                 ON p."ID" = c."PROJECT_ID"
              INNER JOIN "VULNERABILITY" AS v
                 ON v."ID" = req.vulnerability_id
+             INNER JOIN "FINDINGATTRIBUTION" AS fa
+                ON fa."COMPONENT_ID" = req.component_id
+               AND fa."VULNERABILITY_ID" = req.vulnerability_id
```
nscuro: A finding can have multiple attributions, and attributions can be soft-deleted. Doing an `INNER JOIN` can thus cause to more than one row being returned.

Make it a `LATERAL` and align with how the attribution is selected in findings views: https://github.com/DependencyTrack/dependency-track/blob/8c37ab99d1b9977f99ff6b4bbee72bda2fbb4d75/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java#L990-L1001

notification/api/src/main/proto/org/dependencytrack/notification/v1/notification.proto
```diff
@@ -226,6 +226,7 @@ message Vulnerability {
 
   // Whether the vulnerability is known to be exploited.
   optional bool is_kev = 21 [json_name = "isKev"];
+  optional string analyzer_identity = 22;
```
nscuro, line 231: While you're here, consider adding a short comment as to what this field is (i.e. the name of the analyzer that identified the vulnerability first).

nscuro (changes requested):

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/NotificationSubjectDao.java
```diff
@@ -166,6 +167,14 @@ AS req(component_id, vulnerability_id)
                 ON p."ID" = c."PROJECT_ID"
              INNER JOIN "VULNERABILITY" AS v
                 ON v."ID" = req.vulnerability_id
+             INNER JOIN LATERAL (
+               SELECT *
+                 FROM "FINDINGATTRIBUTION" AS fa
+                WHERE c."ID" = fa."COMPONENT_ID"
+                  AND v."ID" = fa."VULNERABILITY_ID"
+                ORDER BY fa."ID"
```
nscuro, line 175: I see you removed the `DELETED_AT` part in 1b5e667, can you share why? As it stands it could report incorrect analyzer identities for the case where a finding was previously inactive with >=2 attributions, and now gets re-activated because one of them reports the finding again. 

That is, the *oldest* attribution (lowest `ID`) may still have a `DELETED_AT` value, but would now get reported as identifying analyzer.

sahibamittal: `getForNewVulnerabilities` query is used for both `GROUP_NEW_VULNERABILITY` and `GROUP_VULNERABILITY_RETRACTED` notification paths.

After changing the attribution join to only select active attributions: `AND fa."DELETED_AT" IS NULL`, the query no longer returns rows for retracted findings.

This can be fixed by separating the attribution-selection semantics by notification type. ex:

// New or reactivated finding: must use active attribution only.
`getForNewVulnerabilities(componentIds, vulnDbIds, false)`

// Retracted finding
`getForNewVulnerabilities(componentIds, vulnDbIds, true)`

nscuro (approved):

PR #7262, opened 2026-09-08
Resolve project parent before other fields in updateProject

### LINKED ISSUE #7241: HTTP 500 when converting a collection project back to a regular project, if the request body carries parent
### Current Behavior

On 5.1.0 (PostgreSQL 18), `POST /api/v1/project` returns `500 Uncaught internal server error` when a
request clears `collectionLogic`, sets a `classifier`, and carries a `parent`. The same request
without `parent` returns `200` — but the project then loses its parent, `POST` replacing the whole
resource. So a collection project cannot be turned back into a regular project while keeping its
parent.

```
ERROR [logger] Update of object with id "org.dependencytrack.model.Project:4734" using statement
"UPDATE "PROJECT" SET "CLASSIFIER"=? WHERE "ID"=?" failed :
org.postgresql.util.PSQLException: ERROR: new row for relation "PROJECT"
violates check constraint "PROJECT_COLLECTION_CLASSIFIER_check"
```

That constraint — `CHECK (COLLECTION_LOGIC IS NULL OR CLASSIFIER IS NULL)` — is intentional ([`V202605022031__init.sql#L886`](https://github.com/DependencyTrack/dependency-track/blob/5.1.0/migration/src/main/resources/org/dependencytrack/migration/V202605022031__init.sql#L886)).
The problem is that `ProjectQueryManager.updateProject` lets the row transit through a state it
forbids:

1. [L262](https://github.com/DependencyTrack/dependency-track/blob/5.1.0/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L262)    sets the classifier on the persistent object, marking the field dirty.
2. [L291-L292](https://github.com/DependencyTrack/dependency-track/blob/5.1.0/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L291-L292)    resolves the parent with a query. DataNucleus flushes dirty state before any query, emitting
   `UPDATE "PROJECT" SET "CLASSIFIER"=?` while `COLLECTION_LOGIC` still holds `AGGREGATE_*` — both
   columns non-null, constraint violated.
3. [L330](https://github.com/DependencyTrack/dependency-track/blob/5.1.0/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L330), which clears `COLLECTION_LOGIC`, is never reached.

With no `parent` in the body, step 2 is replaced by `setParent(null)`, which runs no query, so
nothing is flushed early and the single flush in `persist(project)` writes a consistent row.

The same hazard is already documented and guarded a few lines below for `collectionTag` ([L315-L318](https://github.com/DependencyTrack/dependency-track/blob/5.1.0/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L315-L318)); the parent lookup is not.

Clients cannot avoid this: omitting `collectionLogic`, sending it as `null`, and sending
`classifier: null` all fail identically, because the resource layer forces a non-null classifier
whenever `collectionLogic` is null ([`ProjectResource.java#L792-L796`](https://github.com/DependencyTrack/dependency-track/blob/5.1.0/apiserver/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L792-L796)).

### Steps to Reproduce

1. Create a project `parent-demo`.
2. Create a project `child-demo` version `1.0.0`, with `parent-demo` as its parent.
3. Edit `child-demo`: tick **Collection project**, choose **Aggregate direct children**, save.
   This succeeds — the UI clears the classifier by itself, as the constraint requires.
4. Edit `child-demo` again: untick **Collection project**, pick any classifier, save.
   → **500**, nothing is saved.

Step 4 fails because the form posts `classifier`, `collectionLogic: null` and `parent` in the same
payload ([`ProjectDetailsModal.vue`](https://github.com/DependencyTrack/frontend/blob/5.1.0/src/views/portfolio/projects/ProjectDetailsModal.vue#L822-L845)).
It fails whether or not a classifier is picked, since the API substitutes `APPLICATION` for a null
one.

The same happens on a direct `POST /api/v1/project` with that payload. Dropping `parent` from it is
what isolates the cause — the request then returns `200`, but the project loses its parent:

```jsonc
// POST /api/v1/project → 500
{"uuid":"<child>","name":"child-demo","version":"1.0.0","active":true,
 "classifier":"LIBRARY","parent":{"uuid":"<parent>"}}

// same, without "parent" → 200, and then: collectionLogic=null, classifier=LIBRARY, parent=null
{"uuid":"<child>","name":"child-demo","version":"1.0.0","active":true,
 "classifier":"LIBRARY"}
```

### Expected Behavior

Step 4 should succeed, leaving `child-demo` with `collectionLogic = null`, the chosen classifier,
and its parent unchanged.

`collectionLogic` should be applied before any query that can flush dirty state. Moving the parent
resolution block ([L286-L304](https://github.com/DependencyTrack/dependency-track/blob/5.1.0/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L286-L304)) above the `setClassifier` call, or setting `collectionLogic` before the classifier, should suffice —
the same shape as the existing `collectionTag` guard. If the combination is meant to be rejected,
a `400`/`409` with a message would be preferable to an uncaught `500`.

### Dependency-Track Version

5.x

### Browser

N/A

### Checklist

adityaanikam:
### Description

`ProjectQueryManager.updateProject` sets the new classifier on the persistent project, then resolves the new parent (if any) via a `getObjectByUuid` call, and only clears `collectionLogic` afterward. `getObjectByUuid` runs a query, and DataNucleus flushes pending dirty state before any query executes. If a request converts a collection project back to a regular one (clearing `collectionLogic`, setting a `classifier`) while also carrying a `parent`, that flush writes a row with both `classifier` and `collectionLogic` non-null, violating the `PROJECT_COLLECTION_CLASSIFIER_check` database constraint and surfacing as an uncaught 500 instead of the expected 200.

This is the same class of hazard already documented and guarded a few lines below for `collectionTag` (`resolveTags` is deliberately called before `collectionLogic` is set, for the same reason) — parent resolution just wasn't given the same treatment.

This PR moves parent resolution to the very start of `updateProject`, before any field is set on the persistent object, so the query runs while it's still clean and nothing flushes prematurely. No other ordering changes.

**Scope note:** `hasActiveChild` (raw SQL) and `getLatestProjectVersion` (JDOQL), used by the `isActive`/`isLatest` branches, run queries in the same position in the method and have the same latent hazard. Neither is exercised by the reported issue or the added test, so I left them alone to keep this change scoped to what's reported and verified — flagging here in case a maintainer wants a follow-up.

### Addressed Issue

fixes #7241

### Additional Details

Added `shouldConvertCollectionProjectBackToRegularWithParent` to `ProjectResourceTest`, reproducing the exact reported request shape: a collection project's classifier is set, `collectionLogic` is omitted (cleared), and a `parent` is included in the same `POST /api/v1/project` body. Verified it fails with the original 500 when the fix is reverted (with the test kept), and passes with the fix applied.

### Checklist

- ~This PR introduces changes to the database model, and I have updated the [migration changelog] accordingly~
- ~This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly~
- ~This PR is a substantial change (per the [ADR criteria]), and I have added an [ADR] under `docs/adr/`~

[ADR]: ../docs/adr/

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
+++ b/apiserver/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java
@@ -251,6 +251,32 @@ final class ProjectQueryManager extends QueryManager {
     public Project updateProject(Project transientProject, boolean commitIndex) {
         return callInTransaction(() -> {
             final Project project = getObjectByUuid(Project.class, transientProject.getUuid());
+
+            // NB: Resolve the parent BEFORE setting any field below, for the same reason the
+            // collection tag is resolved before collectionLogic further down: getObjectByUuid
+            // triggers a query, which flushes dirty state. Doing that after e.g. setClassifier
+            // has run, but before setCollectionLogic clears it, can flush a row that violates
+            // PROJECT_COLLECTION_CLASSIFIER_check (see #7241).
+            if (transientProject.getParent() != null
+                    && transientProject.getParent().getUuid() != null) {
+                if (project.getUuid().equals(transientProject.getParent().getUuid())) {
+                    throw new IllegalArgumentException("A project cannot select itself as a parent");
+                }
+                Project parent = getObjectByUuid(
+                        Project.class, transientProject.getParent().getUuid());
+                if (parent.getInactiveSince() != null) {
+                    throw new IllegalArgumentException("An inactive project cannot be selected as a parent");
+                } else if (isChildOf(parent, transientProject.getUuid())) {
+                    throw new IllegalArgumentException(
+                            "The new parent project cannot be a child of the current project.");
+                } else {
+                    project.setParent(parent);
+                }
+                project.setParent(parent);
+            } else {
+                project.setParent(null);
+            }
+
             project.setAuthors(transientProject.getAuthors());
             project.setPublisher(transientProject.getPublisher());
             project.setManufacturer(transientProject.getManufacturer());
@@ -282,26 +308,6 @@ final class ProjectQueryManager extends QueryManager {
                 }
             }
             project.setIsLatest(transientProject.isLatest());
-
-            if (transientProject.getParent() != null
-                    && transientProject.getParent().getUuid() != null) {
-                if (project.getUuid().equals(transientProject.getParent().getUuid())) {
-                    throw new IllegalArgumentException("A project cannot select itself as a parent");
-                }
-                Project parent = getObjectByUuid(
-                        Project.class, transientProject.getParent().getUuid());
-                if (parent.getInactiveSince() != null) {
-                    throw new IllegalArgumentException("An inactive project cannot be selected as a parent");
-                } else if (isChildOf(parent, transientProject.getUuid())) {
-                    throw new IllegalArgumentException(
-                            "The new parent project cannot be a child of the current project.");
-                } else {
-                    project.setParent(parent);
-                }
-                project.setParent(parent);
-            } else {
-                project.setParent(null);
-            }
 
             // Prevent illegal states of collection projects (must not contain components or services).
             final ProjectCollectionLogic newCollectionLogic = transientProject.getCollectionLogic();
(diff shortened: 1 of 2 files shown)


nscuro (changes requested):
Thanks, looks good. Please resolve the lint failure, then this is good to merge.

nscuro (approved):

Issue #6553, opened 2026-06-30
Expose matched subject data in policy violation notifications


ben-edna:
### Current Behavior

When a policy condition triggers a violation (for example EPSS ≥ 0.5 or a specific severity) a `PolicyViolationSubject` notification is sent. The notification carries the condition configuration (subject, operator, configured threshold value) and the affected component and project, but never the actual data that caused the match.

A recipient cannot tell from the notification alone which CVE, license, or other subject entity triggered the violation. For set-based subjects such as EPSS or SEVERITY, a follow-up API call may not even be deterministic: a component can have multiple vulnerabilities above the threshold, and there is no way to know which one was responsible.

### Proposed Behavior

At notification dispatch time, resolve the matched subject data and include it in both `PolicyViolationSubject` and `NewPolicyViolationsSummarySubject` protos so it is available in notification templates. For example:

- EPSS / SEVERITY / VULNERABILITY_ID / CWE: vulnerability UUID, CVE ID, and the score or severity that satisfied the condition
- LICENSE / LICENSE_GROUP: license name and SPDX ID that matched
- PACKAGE_URL / CPE / COORDINATES / VERSION: the component-level value that matched

The EXPRESSION subject is excluded since there is no single matched entity to resolve.

For set-based subjects like SEVERITY, EPSS, or LICENSE a component can have multiple entities satisfying the condition at the same time (e.g. several vulnerabilities above the EPSS threshold). All matching entities should be included.

It seems like the `NewPolicyViolationsSummary` already does some lookups and at some points all the required info.

This makes notifications self-contained. Recipients can triage a violation without a follow-up API call and downstream integrations (email digests, dashboards, ticketing) do not need N additional calls per violation.

For instance when using `subject.violationsByProjectList` in the summary notification:
```pebble
Condition: {{ violation.policyCondition.subject }} {{ violation.policyCondition.operator }} {{ violation.policyCondition.value }}
{% if violation.matchedVulnerabilitiesList is not empty %}
Matched vulnerabilities:
{% for vuln in violation.matchedVulnerabilitiesList %}
  - {{ vuln.vulnId }} (EPSS: {{ vuln.epssScore }}, Severity: {{ vuln.severity }})
{% endfor %}
{% endif %}
```

This would produce something like:

```plain
Condition: EPSS >= 0.5
Matched vulnerabilities:
  - CVE-2024-12345 (EPSS: 0.823, Severity: CRITICAL)
[...]

PR #7322, opened 2026-09-15
Add ADR for policy condition message expressions

fffinkel:
Proposes an optional CEL message expression per policy condition, evaluated when the condition is violated and stored on the violation, so notifications can say what caused the match.

### Description

Adds ADR 041, proposing an optional CEL message expression on policy conditions. When a condition is violated, the engine evaluates the expression with the same variables as the condition and stores the resulting string on the violation. `POLICY_VIOLATION` and `NEW_POLICY_VIOLATIONS_SUMMARY` notifications include it, so recipients can see which vulnerability, license, or component value caused the match without a follow-up API call.

This follows the `messageExpression` direction suggested in the issue, rather than structured matched-entity fields. The ADR compares both options.

### Addressed Issue

Refs #6553

### Additional Details

* Reuses the unused `POLICYVIOLATION."TEXT"` column (widened to `text`) instead of adding a new one.
* Fail-open: a runtime error in the message expression logs a warning and records the violation without a message.
* Stored message is updated on re-evaluation when it changes; the violation row and its analysis are kept.
* Default messages for legacy subjects (EPSS, SEVERITY, LICENSE) are listed as a possible follow-up, not in scope.

### Checklist

- ~~This PR fixes a defect, and I have provided tests to verify that the fix is effective~~
- ~~This PR implements an enhancement, and I have provided tests to verify that it works as intended~~
- ~~This PR introduces changes to the database model, and I have updated the [migration changelog] accordingly~~
- ~~This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly~~

[ADR]: ../docs/adr/


docs/adr/041-policy-condition-message-expression.md
```diff
@@ -0,0 +1,179 @@
...
+
+The expression is validated when a condition is created or updated. It is compiled against the same
+CEL environment as the condition, and its result type must be `string`. An invalid expression is
+rejected with an [RFC 9457] problem details response, in the same way invalid `EXPRESSION` conditions
+are rejected today.
+
+The policy engine evaluates the message expression only for conditions that were violated. It uses
+the same variables as the condition: `component`, `project`, `vulns`, and `now`. The fields the
+message expression reads are added to the engine's data requirements, so the data is loaded together
+with the data the conditions need and no extra database queries are issued per component. The
```
nscuro: There is a performance risk here, where the message expression contains more fields than the policy expression itself. The data for the message expression is only needed for real matches, but by merging its requirements into the reqs of the policy expression, the fetch cost is always paid upfront.

There's also a risk of users utilizing functions that trigger ad-hoc SQL queries, such as dependency graph traversal, in the message expression, making them more expensive.

It's a bit hard to quantify just about how bad that would be, but I think we should at least consider making the message expression CEL environment more restricted, e.g. by not allowing policy functions and not allowing more data than the policy expression itself.

docs/adr/041-policy-condition-message-expression.md
```diff
@@ -0,0 +1,179 @@
...
+`POLICY_VIOLATION` notifications, and on the per-violation entry used by
+`NEW_POLICY_VIOLATIONS_SUMMARY` notifications. Both are read from the stored violation. The default
+notification templates for email, Slack, Microsoft Teams, and Mattermost print the message when it is
+present.
+
+Out of scope for this decision:
+
+* Default message expressions for legacy subjects. The engine could generate a sensible message for
+  `EPSS`, `SEVERITY`, or `LICENSE` conditions when the user did not provide one. This is a possible
+  follow-up and does not change the mechanism decided here.
```
nscuro: I'd propose we move this in scope, as the mechanism working for legacy subjects would provide immediate value for users and simultaneously proof the mechanism works as intended.

docs/adr/041-policy-condition-message-expression.md
```diff
@@ -0,0 +1,179 @@
...
+* Structured matched-entity data in notifications.
+* The frontend. The policy condition form needs an input for the message expression and the
+  violation views should show the message. This work happens in the frontend repository.
```
nscuro: The frontend has no separate ADR process. This feature should be delivered with the frontend changes as otherwise its use is quite limited. This is implicitly assumed and you don't need to spell out how the frontend changes will look here, so you can just remove this statement.

docs/adr/041-policy-condition-message-expression.md
```diff
@@ -0,0 +1,179 @@
...
+trade-off for a single mechanism that covers all subjects.
+
+The policy condition table and the violation table each change by one column. The v1 REST API for
+policy conditions gains an optional field. Both changes are additive and backward compatible. The
+`text` field on violations, which was always present but never populated, now carries data.
+
+Each violated condition with a message expression costs one additional CEL evaluation per component.
+Conditions that are not violated and conditions without a message expression cost nothing extra. The
+data requirements of message expressions are merged into the existing preload, so the number of
+database queries per evaluation does not change.
```
nscuro: "cost nothing extra" is not entirely true, as reqs being merged can cause additional data to be loaded that may never be used, see comment above.

docs/adr/041-policy-condition-message-expression.md
```diff
@@ -0,0 +1,179 @@
...
+
+* Covers all use cases.
+
+Cons:
+
+* Two mechanisms for one problem, with all the costs of the first option.
+
+## Decision
+
+We will add an optional message expression to policy conditions, the second option above. We will
```
nscuro: nit:

> the second option above

Consider labeling the options (e.g. A, B, C) so you can refer to them by that label, and readers don't have to scroll around to remember what you're referring to.

docs/adr/041-policy-condition-message-expression.md
```diff
@@ -0,0 +1,179 @@
...
+condition applies to every component, message expressions have to use CEL to select what is relevant.
+For example, a condition `vulns.exists(vuln, vuln.epss_score >= 0.5)` may use the message expression:
+
+```
+"EPSS >= 0.5 matched by: " + vulns.filter(vuln, vuln.epss_score >= 0.5)
+  .map(vuln, vuln.id + " (EPSS " + string(vuln.epss_score) + ")").join(", ")
+```
+
+Evaluation of a message expression never affects whether a violation is recorded. If the expression
+fails at runtime, the engine logs a warning, records the violation without a message, and continues.
```
nscuro: Could consider adding a placeholder text that hints at evaluation having failed, without disclosing *why*. That way, users know immediately that something is broken.

nscuro (changes requested):
Good ADR. A few comments but overall looks good. Please also bump the ADR number since [041 has been taken in the meantime](https://github.com/DependencyTrack/dependency-track/blob/e75bb5150521284e28904a1be868932150809bfc/docs/adr/042-workload-identity-federation.md).

fffinkel:
ADR renumbered to 043.

docs/adr/041-policy-condition-message-expression.md
fffinkel: Removed.

docs/adr/041-policy-condition-message-expression.md
fffinkel: Reworded to say conditions without a message expression cost no extra evaluation, and explain the bounded data cost from the restricted environment.

docs/adr/041-policy-condition-message-expression.md
fffinkel: Labeled them A, B, C and updated the references.

docs/adr/041-policy-condition-message-expression.md
fffinkel: Good idea, on runtime failure the engine now stores a fixed placeholder, "Message expression failed to evaluate. See server logs.", and logs the cause.

docs/adr/041-policy-condition-message-expression.md
fffinkel: Moved into scope. The same builder that generates the condition script generates the default message expression, so they cannot drift, and the user-provided expression overrides it.

docs/adr/041-policy-condition-message-expression.md
fffinkel: Gotcha. I restricted the environment in two ways: no policy functions (depends_on, matches_range, etc.), and the message expression may only read fields the condition itself reads, plus a fixed allowlist of identity fields like component.name, component.version, vuln.id, and vuln.source.

I did not go as far as "no more data than the condition" because the engine only selects required columns, so vuln.id is not loaded for an EPSS condition, and naming the matching vulnerability is the whole point. Identity fields are plain columns on tables already queried, so they add a column but never a query or join.

I'm open to trimming the allowlist if you think it is still too wide.

nscuro: Thanks! I've been mulling this over and I think there's another solution to this: We allow policy expressions to return strings instead of just booleans.

The idea being that:

* if a boolean value is returned, we only record the violation without text as we do today
* if a string is returned, we record the violation with the returned value as text
* if an empty or blank string is returned, we record no violation (i.e. it's equivalent to `false`)

So the expression generating the message is the same that determines whether a condition matched. No separate evaluation, no data loading restrictions, no allowlisting. By still allowing boolean results, all existing conditions continue to work.

For EPSS, such a condition could look like this:

```js
cel.bind(
  vulnIds,
  vulns.filter(v, has(v.epss_score) && v.epss_score > 0.5).map(v, v.id),
  size(vulnIds) > 0
    ? "Vulnerabilities with EPSS score exceeding 0.5: " + vulnIds.join(", ")
    : ""
)
```

*(Note: variable binding via [`cel.bind`](https://github.com/cel-expr/cel-java/blob/main/extensions/src/main/java/dev/cel/extensions/README.md#celbind) is an [extension](https://github.com/cel-expr/cel-java/tree/main#extensions) we must enable)*

Granted the expressions you're writing here are a bit more complex, but I feel like it's still manageable (assuming proper documentation and examples). A separate message expression would have similar complexity **and** largely duplicate the condition expression.

WDYT?

PR #7335, opened 2026-09-18
Make the monitoring dashboards usable outside the dev stack

nissessenap:
### Description

Minor tweaks to the dashboard, I know they aren't supported but hopefully this make it easier for future users to just copy them and get a more general base.

### Addressed Issue

https://github.com/DependencyTrack/dependency-track/issues/7336

### Additional Details

**Why the `default` pseudo-option and not an empty `current`.** I first tried `"current": {}`, expecting Grafana to fall back to the org default. It does not. Verified on Grafana 12.1.1 (the version this compose stack pins) by provisioning a second, non-default Prometheus named `AAA-Decoy` alongside the real one: with `"current": {}` the dashboard loaded with `var-datasource=decoy-first-alphabetically` and rendered nothing, while `"current": {"text": "default", "value": "default"}` loaded with `var-datasource=default` and rendered normally. So the explicit `default` option is what makes this deterministic.

### Checklist

[ADR]: ../docs/adr/

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dev/monitoring/prometheus.yml
```diff
@@ -1,7 +1,7 @@
 ---
 scrape_configs:
 - job_name: apiserver
-  scrape_interval: 15s
+  scrape_interval: 1m
```
nscuro: Increasing to 1m is a regression for dev setups where faster feedback / more datapoints are desired. Since you added `$__rate_interval` the interval configured here no longer matters for correctness AFAICT.

nscuro (changes requested):

dev/monitoring/prometheus.yml
nissessenap: That is a fair point, I will fix it

nissessenap: Fixed on my phone, let's hope it worked :)

nscuro (approved):

